PR #289, opened 2026-06-16
fix(suref-react): defer CardImage cached-load check off the hydration commit

alxjrvs:
## What

`CardImage` detected images that finished loading before React attached its `onLoad` handler (cached / preloaded) via a `useRef` callback that called `setLoaded(true)`. Ref callbacks fire during the commit phase, so this mutated state mid-hydration and desynced the server-rendered markup from the client, forcing a tree regeneration (React #418).

## Change

- Move the cached/preloaded detection into a `useEffect` keyed on the display URL, which runs after the hydration commit instead of during it.
- Defer the `setLoaded(true)` to a `queueMicrotask` so there's no synchronous `setState` in the effect body (satisfies the `react-hooks/set-state-in-effect` lint rule).
- The live-load path (`onLoad` on the ``) is unchanged.

## Notes

- Scoped to a single file; no API or behavior change for the common load path.
- Pre-commit lint/format/test are green.

### DIFF
--- a/packages/suref-react/src/components/shared/CardImage.tsx
+++ b/packages/suref-react/src/components/shared/CardImage.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useRef, useState } from 'react'
+import { useEffect, useRef, useState } from 'react'
 import { cn } from '../../utils/cn'
 
 type CardImageEditable = {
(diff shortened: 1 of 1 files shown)


alxjrvs:
## Code review — ✅ Approve

Low-risk, correctly targets the bug. A ref callback fires during the commit phase, so `setLoaded` there during hydration is what produced React #418. Moving to `useEffect` (which runs after the commit) resolves it.

### Correctness
- ✅ Core fix is sound. The live-load path (`onLoad`) is untouched; only the cached/preloaded detection moved.

### Notes (non-blocking)
- **Rationale nuance:** the comment says the microtask makes the update "land after the hydration commit instead of mutating state during it." That's not quite the reason — `useEffect` *already* runs after hydration completes, so moving to an effect is what fixes #418 on its own. The `queueMicrotask` is really there to satisfy the `react-hooks/set-state-in-effect` lint rule (no synchronous `setState` in an effect body). Code is correct; just the stated "why" conflates the two.
- **Deferred setState after unmount:** if the component unmounts before the microtask runs, `setLoaded` fires on an unmounted component — a silent no-op in React 18+, so harmless.

### Follow-ups (pre-existing, out of scope)
- `loaded` is never reset to `false` when `displayUrl` changes. Swapping to a new image shows it at `opacity: 1` immediately (no fade, brief flash of a not-yet-decoded image) until `onLoad` re-fires. Candidate fix: reset `loaded=false` on `displayUrl` change to restore the fade on image swaps.
- No test covers this hydration-timing regression (awkward to unit-test). Acceptable, but noting the coverage gap.

alxjrvs:
## Follow-up addressed — `loaded` reset on URL change (d2cdd883)

The image-swap fade is now handled. `CardImage` tracks the URL its current `loaded` state belongs to and resets `loaded=false` during render when `displayUrl` changes, so a swapped image fades in (and no longer briefly shows a not-yet-decoded frame at full opacity).

Implementation uses React's render-phase "store previous prop" pattern rather than an effect — it re-renders before commit, so there's no extra paint and no synchronous `setState` in an effect body. Lint / typecheck / 303 package tests green.

One related item left out of scope (separate, pre-existing): `showImage` also isn't reset on a `displayUrl` change, so an image that previously errored (`onError` → `showImage=false`) won't reappear if the URL later changes to a valid one. Happy to fold that in here or track it separately — let me know.

PR #292, opened 2026-06-17
fix(suref-react): drive Legal Starting Pattern badge from data tag

alxjrvs:
## Problem

A lot of patterns were badged as **Legal Starting Pattern** that shouldn't be — e.g. every Gopher pattern (a T2 chassis). Two distinct issues:

1. The badge was **computed** at display time instead of reading the data tag.
2. The underlying `legalStarting` data tags didn't match the source books.

## "Legal Starting" is a source-book data tag

The designation comes verbatim from the Workshop Manual (Core Book): *"This Pattern is a legal Tech 1 build for a starting Mech if you are new to the game."* It is an explicit per-pattern tag, set only where the source book prints that text — never derived from a tech-level/SV rule.

Verified across every source book the chassis data draws from (Core Book / Workshop Manual, Starter Set + all its booklets, False Flag, Rainmaker, We Were Here First, asset-pack booklets): the text appears **only** in the Core Book, marking exactly 5 patterns — Hauler (Mule), Buzzard (Mazona), Leaky (Scrapper), Settler (Spectrum), Shepherd (Thresher).

## Changes

**1. Drive the badge from the data tag** (`suref-react`)
- `isLegalStartingMech` now reads `legalStarting` off the resolved pattern record in `useChassisPatternConfig`.
- Removed the computed `checkLegalStartingMech` helper (wrong mechanism).
- Regression test `legalStartingBadge.test.tsx`: badge shows for a tagged pattern, hidden for an untagged one.

**2. Correct the data** (`salvageunion-reference`)
- Removed 7 `legalStarting` tags with no "Legal Starting Mech" text in their source book: Warrior (Bobcat), Moth (Gatecrasher), 10 Finger (Kelpie), Deerstalker (Pioneer), DronTek (Trooper), Stefanus (Parasite) — all `true` — and Hunchback (Ravager) `false`.
- The 5 Workshop Manual tags remain. Text-level edit; prettier confirms `chassis.json` formatting unchanged.

## Checks

`bun --filter suref-react test` (305 pass), typecheck, lint, knip, and `validate:all` (27 files) all clean.

### DIFF
--- a/packages/salvageunion-reference/data/chassis.json
+++ b/packages/salvageunion-reference/data/chassis.json
@@ -6178,7 +6178,6 @@
     "patterns": [
       {
         "name": "10 Finger Pattern",
-        "legalStarting": true,
         "systems": [
           {
             "name": "Rigging Arm"
@@ -6271,7 +6270,6 @@
     "patterns": [
       {
         "name": "DronTek Pattern",
-        "legalStarting": true,
         "systems": [
           {
             "name": "K4 Rifle"
@@ -6348,7 +6346,6 @@
     "patterns": [
       {
         "name": "Deerstalker Pattern",
-        "legalStarting": true,
         "systems": [
           {
             "name": "Mech Melee Armament"
@@ -6420,7 +6417,6 @@
     "patterns": [
       {
         "name": "Stefanus Pattern",
-        "legalStarting": true,
         "systems": [
           {
             "name": "Needle Missile Pod",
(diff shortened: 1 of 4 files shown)


alxjrvs (commented):
✅ Approving — badge now driven purely by the explicit legalStarting data tag (computed SV-budget check removed); test covers tagged/untagged cases. CI green.

PR #294, opened 2026-06-17
fix(suref-react): surface matching-action stats in resolved data row

alxjrvs:
## Symptom

On the Brawler "Gladiator Pattern" chassis, the **Mech Melee Armament** System rendered an **empty data block** — no Range, Damage, or Traits.

## Root cause

That System carries its stats (Range/Damage/Traits) and its freeform "describe the weapon" choice on a **same-named action** (resolved via the action map), not as a root `datavalues` content block or root `choices`. The freeform choice routes the display through `ReferenceEntityResolvedDataRow` → `resolveChoiceView`, which only reads the **root** datavalues/traits/choices shape (the granted-equipment shape, e.g. the Custom Sniper Rifle). For matching-action systems it therefore resolved an empty row and dropped the action-sourced stats. The data and ORM are correct — this is purely a display-layer bug.

## Fix

In `ReferenceEntityResolvedDataRow`, when the entity has **no root `datavalues` block and no root `choices`**, seed the resolved data row's base stats from the matching action's resolved Range/Damage/Traits (via `getDamage`/`getRange`/`getTraits`) before handing off to `resolveChoiceView`. Granted equipment (own `datavalues` block or root `choices`, e.g. the Custom Sniper Rifle) passes through unchanged.

## Testing

- New regression test `mechMeleeArmamentStats.test.tsx`: asserts Mech Melee Armament renders Damage (SP), Range (Close), and the Melee/Wield traits; and a sibling stat-less matching-action System (Industrial Body Kit) shows **no** spurious Damage/Range.
- `bun --filter suref-react test` — **308 pass, 0 fail** (new test passes; existing `grantedEquipmentDisplay.test.tsx` Custom Sniper Rifle still passes — no regression).
- `bun run typecheck` — pass (all packages, 0 errors).
- `bun run lint` — pass (exit 0).
- `bun run format` — pass (formatting applied).

### DIFF
--- a/packages/suref-react/src/components/referenceEntity/ReferenceEntityDisplay/ReferenceEntityResolvedDataRow.tsx
+++ b/packages/suref-react/src/components/referenceEntity/ReferenceEntityDisplay/ReferenceEntityResolvedDataRow.tsx
@@ -1,5 +1,15 @@
-import type { SURefEntity, SURefObjectDataValue } from 'salvageunion-reference'
-import { getChoices, resolveChoiceView } from 'salvageunion-reference'
+import type {
+  SURefEntity,
+  SURefObjectContentBlock,
+  SURefObjectDataValue,
+} from 'salvageunion-reference'
+import {
+  getChoices,
+  getDamage,
+  getRange,
+  getTraits,
+  resolveChoiceView,
+} from 'salvageunion-reference'
 import { DataValueDisplayView } from '../DataValueDisplayView'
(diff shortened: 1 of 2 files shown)


alxjrvs (commented):
✅ Approving — seeds synthetic datavalues from the matching action's resolved Range/Damage/Traits so matching-action-choice systems (Mech Melee Armament) surface their stats; pass-through for granted equipment is preserved and the negative case is covered by the Industrial Body Kit test. CI green.

PR #295, opened 2026-06-17
refactor(suref-react): keep editable inputs out of the SRD reference

alxjrvs:
## Problem

The static SRD reference site (`apps/suref-web`) was rendering live, editable text inputs that belong only in the character builder (ITUN). The most visible case: free-text choice fields like the **Sestra Drone's "A.I. Personality"** rendered as an editable `<input>`/`<textarea>` on the reference site, where there's no persistence — keystrokes were ephemeral and lost on refresh.

## Changes

**1. Free-text choice cards render as static text on the SRD (still editable in ITUN).**
`FreeTextChoiceCard` (Name / Appearance / A.I. Personality) now takes a `readOnly` flag, threaded through `ReferenceEntityResolvedChoices → ChoiceGroups → ChoiceGroup`. The flag is derived from the absence of a real consumer change handler (`onSelectionChange === undefined`):
- **SRD reference** → renders the prompt (or saved value) as plain text, no input.
- **ITUN live builder** → unchanged, fully editable.
- **Bonus:** ITUN read-only snapshots also go static — they previously showed inputs that silently dropped keystrokes.

**2. Removed the dead shared NPC-name input.**
`NpcNameInput` plus its `onNpcNameChange` / `onNpcNameBlur` props (and the `NpcConfig.onNameChange` / `onNameBlur` fields) were unused: ITUN edits NPC names through its own `InlineEditField` / `NpcInset`, and no consumer passed `npcConfig`'s name-edit callbacks. NPC names now render as static text.

## Out of scope (intentionally left alone)

- Choice **option toggle** cards (buttons, not text inputs).
- `GuideStepsDisplay`'s interactive text inputs — gated behind an `interactive` config that no app ever passes, so they never appear in the SRD. Removing that whole config surface is a separate dead-code cleanup.

## Validation

- `bun run typecheck` — clean across all packages
- `bun --filter suref-react test` — 303 pass
- `bun --filter in-the-union-now test` — 888 pass
- `bun run lint` — clean

### DIFF
--- a/packages/suref-react/src/components/referenceEntity/ReferenceEntityDisplay/ReferenceEntityNpcDisplay.tsx
+++ b/packages/suref-react/src/components/referenceEntity/ReferenceEntityDisplay/ReferenceEntityNpcDisplay.tsx
@@ -1,4 +1,3 @@
-import { useState, useEffect, useRef } from 'react'
 import type { ReactNode } from 'react'
 import { getNpc } from 'salvageunion-reference'
 import type { SURefMetaEntity } from 'salvageunion-reference'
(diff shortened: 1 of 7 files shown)


alxjrvs (commented):
✅ Approving — clean refactor: removes inline NPC-name + free-text choice inputs, renders them as static text via readOnly threaded through ChoiceGroups/ChoiceGroup/FreeTextChoiceCard; hook usage and prop removal are consistent. CI green.

PR #296, opened 2026-06-21
feat(suref-web): branded per-entity OG preview cards

alxjrvs:
## What

Every entity page now gets a **branded 1200×630 Open Graph card**, generated at build time. Sharing a link to any chassis / system / ability / creature (etc.) on Discord or social shows a Salvage Union card *representing* the entity — name, tech-level badge, key stats, traits, source — instead of the generic site banner.

## How

- **Pipeline:** Satori (element tree → SVG) + resvg (SVG → PNG). No headless browser, no runtime backend — fits the static-site architecture.
- **Data reuse:** feeds `extractStaticEntitySummary` (name, tech level, stats, traits, source/page) into the template — no new data plumbing.
- **Type/brand match:** loads the self-hosted Barlow / Barlow Semi Condensed WOFF files and the SU palette so cards match the site.
- **Endpoint:** `[itemId].og.png.ts` static endpoint emits one PNG per entity via `getStaticPaths` (mirrors the existing `[itemId].json.ts` pattern).

## Changes

- New `src/lib/og/renderOgCard.ts` + `[itemId].og.png.ts` endpoint
- Item page uses the generated card as `og:image` (real artwork still preloads for the on-page view)
- `BaseLayout` now emits `og:image:width/height` for the cards (previously only for the default banner)
- Sitemap excludes the `.og.png` routes
- Deps: `satori`, `@resvg/resvg-js`
- Changelog entry

## Verification

- **Build:** exit 0 — 1590 pages in 3m35s; **1560 OG cards** emitted, ~40 KB avg (15–60 KB), ≈62 MB total. Spot-checked a dist PNG (Iron Wyrm) — renders correctly.
- **Visual:** verified across chassis / equipment / abilities / creatures — handles long names, sparse stats, and missing tech level.
- typecheck (0 errors), lint, knip all clean; 953 suref-web tests pass.

## Decisions (easy to flip)

1. **Generated card is the OG image for every entity** — uniform/on-brand. Artwork (`assetUrl`) still preloads on-page but no longer drives the share preview. Flipping to artwork-first-with-card-fallback is a one-line change.
2. **Build cost:** the ~1560 renders add roughly 60–90s to the build (est. ~50ms/card). Total build 3m35s.

### DIFF
--- a/apps/suref-web/astro.config.mjs
+++ b/apps/suref-web/astro.config.mjs
@@ -11,7 +11,8 @@ export default defineConfig({
   integrations: [
     react(),
     sitemap({
-      filter: (page) => !page.includes('/image') && !page.includes('/greembeem'),
+      filter: (page) =>
+        !page.includes('/image') && !page.includes('/greembeem') && !page.includes('.og.png'),
     }),
     AstroPWA({
       registerType: 'autoUpdate',
--- a/apps/suref-web/package.json
+++ b/apps/suref-web/package.json
@@ -26,10 +26,12 @@
     "@base-ui/react": "1.5.0",
     "@fontsource/barlow": "5.2.8",
     "@fontsource/barlow-semi-condensed": "5.2.7",
+    "@resvg/resvg-js": "2.6.2",
     "astro": "^6.4.6",
     "react": "^19.2.7",
     "react-dom": "^19.2.7",
     "salvageunion-reference": "workspace:*",
+    "satori": "0.26.0",
     "suref-react": "workspace:*"
   },
   "devDependencies": {
--- a/apps/suref-web/src/layouts/BaseLayout.astro
+++ b/apps/suref-web/src/layouts/BaseLayout.astro
@@ -52,7 +52,6 @@ const {
 const pathname = Astro.url.pathname.endsWith('/') ? Astro.url.pathname : Astro.url.pathname + '/'
 const canonicalUrl = canonical || new URL(pathname, SITE_URL).href
 const ogImageUrl = ogImage.startsWith('http') ? ogImage : new URL(ogImage, SITE_URL).href
-const isDefaultOgImage = ogImage === DEFAULT_OG_IMAGE
 ---
 
 <!doctype html>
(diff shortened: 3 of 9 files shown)


alxjrvs:
## 🔍 Self-review

Ran a high-effort review (correctness + cleanup/robustness finder angles, then verification). **No correctness bugs** — and the build empirically rendered all 1560 cards without error. A few quality/robustness suggestions, none blocking.

### Refuted (checked, not real)
A finder flagged possible undefined-deref on `summary.stats` / `.traits` / `.contentParagraphs` and on `props.item` in the endpoint. Verified against the types: in `StaticEntitySummary` those three are non-optional `[]` (always initialized in `extractStaticEntitySummary`), and `getItemStaticPaths` only pushes paths when `'id' in item && item.id`, so `props.item` is always present. No guard needed.

### Worth addressing (non-blocking)

1. **Duplicated `techLevelLabel`** — `renderOgCard.ts` re-implements `techLevelLabel`, but it's already exported from the `suref-react` barrel (`src/index.ts:62`). Reusing it honors the repo's "reuse existing shared components" rule and keeps the BIO/NANITE labels in one place. *(I inlined it to avoid a UI-package import in a build script, but it's a pure function and the package is already a dep — happy to swap.)*

2. **Long stat values can overflow a chip** — stat chips render `value` at `fontSize: 44` with no `whiteSpace: 'nowrap'`/`maxWidth`. Most values are short numbers, but `Range` (`"Close, Medium, Far"`) and `Damage` (`"2 Damage"`) are strings that can be long enough to collide/clip in the 4-up flex row. Low-frequency, visual-only (Satori clips silently). Easy hardening: cap the chips to numeric/short stats, or add `whiteSpace: 'nowrap'` + truncate the value.

3. **`Cache-Control` header is a no-op** — the endpoint sets `Cache-Control` on the `Response`, but under `output: 'static'` Astro writes the PNG to disk and the host (Netlify) serves it with its own headers; the endpoint header is dropped. Harmless but misleading — could remove it (Netlify already long-caches fingerprinted-path assets).

4. **Traits line** (minor, same class as #2) — truncated to 78 chars, which bounds it, but there's no explicit wrap/clip guard. Low risk given the cap.

### Verified good
typecheck (0 errors), lint, knip clean; 953 tests pass; build exit 0 (1590 pages / 3m35s); og:image URL ↔ endpoint route match confirmed; sitemap excludes the `.og.png` routes; `og:image:width/height` correct for all pages (every OG image the site emits is 1200×630).

I'd lean toward doing #1 and #2 before merge — want me to apply them?

alxjrvs:
## ✅ Follow-ups addressed (67b1fe17)

**#1 — Reuse `techLevelLabel`:** now imported from the `suref-react` barrel instead of the inlined copy. Confirmed the barrel imports cleanly in the build-time Node context (75ms, no browser-global eval issues) and the full build is unaffected.

**#2 — Long stat-value overflow:** stat-chip values now shrink the font for longer strings (44 → 32 → 24 by length), stay on one line (`whiteSpace: nowrap`), and truncate to 16 chars as a hard backstop, with the chip column capped at `maxWidth: 240` + `overflow: hidden`. Verified on **Torpedo Tubes** (`Range: "Close, Long"`) — the value shrinks to fit its column and aligns with the numeric chips, no collision.

Rebuild green: 1590 pages / **1560 cards** / 3m33s, exit 0. typecheck/lint/knip/validate/test all pass.

Leaving **#3** (no-op `Cache-Control` header) and **#4** (traits-line wrap guard) as-is — both are harmless and lower-value than the noise of touching them again; happy to drop #3 if you'd prefer the dead header gone.

PR #305, opened 2026-06-23
test(itun): add pilotSnapshot coverage + rule-util edge cases (M4)

alxjrvs:
## Feedback

**M4: Unit tests for rule-enforcement utilities — add pilotSnapshot coverage + documented edge cases**

REQ-NF-21 (M4) asks for comprehensive unit tests across all rule-enforcement utilities beyond the M1 core (capacity/scrap/cargo/softWarnings), with edge cases documented in test names. Auditing `apps/in-the-union-now/src/lib/rules/`, every utility had a test file **except `pilotSnapshot.ts`** (`enrichPilotSnapshot` — resolves stored ability slug refs into tiered `AbilityInput[]` and classifies the pilot's class tier / Salvager flag to feed `evaluateSoftWarnings`). The remaining work was filling documented edge-case gaps in the existing rule-util tests.

## UX area

`apps/in-the-union-now/src/lib/rules/` — the pure rule-enforcement utility layer (specifically the previously-untested `pilotSnapshot.ts` and `__tests__/` edge-case coverage). These utilities feed the soft-warning system surfaced in the pilot/mech wizards and sheet, but the deliverable here is **unit tests against the pure functions** — no route/component/store touched. This is a test-only change: no source, schema, UI, or store edits, no backend.

## SURules references

- **Salvage Union Core Book Digital Edition 2.0a, p.23 ("Gaining Abilities")** — pins the tier structure the enrichment classifies: Core Abilities live in a class's Core Ability trees (taken in order); Advanced/Hybrid Abilities require 6 Core Abilities and 3 in the linked Core tree; Legendary Abilities require 6 Core + 3 Advanced/Hybrid and *"You may only have one Legendary Ability."* `enrichPilotSnapshot` treats level `'L'` as Legendary, matching this.
- **Salvage Union Core Book Digital Edition 2.0a, p.44 (Salvager class entry)** — the Salvager draws from Core trees of any class, has a 12-ability cap, and has no Advanced/Legendary abilities. The enrichment's `isSalvager` flag (which raises the soft cap to 12) and `classTier` are pinned to this; the dataset's Salvager carries `coreTrees`, so it classifies as `'base'` with `isSalvager: true`.

## Fix

- **`pilotSnapshot.test.ts` (new)** — covers `enrichPilotSnapshot`: (a) tier resolution to core/advanced/legendary via the class-tree index; (b) level-`'L'` → legendary regardless of tree; (c) unclassified tree (Generic, level `'G'`) → `tier` undefined; (d) unknown refs survive as bare `{ ref }` with no fabricated tier; (e) ability resolution by id and by name; (f) class resolution by id / exact name / case-insensitive name; (g) `classTier` base vs advanced-hybrid vs undefined; (h) `isSalvager` only for Salvager; (i) empty abilities + missing classRef. Runs against **real** salvageunion-reference data (preloads `'abilities'` + `'classes'`, mirroring `derivedStats.test.ts`).
- **`capacity.test.ts`** — one-under / exactly-one-over system-cap boundaries; modules exactly at the module cap.
- **`cargo.test.ts`** — exactly-at vs one-over capacity boundaries; zero-cap parent with empty item list.
- **`crawlerCapacity.test.ts`** — exactly-at vs one-over bay boundaries; min (TL1) and max (TL6) tech-level extremes; above-max (TL7) + negative TL treated as unknown.
- **`scrap.test.ts`** — min/max TL salvage value; TL1→TL6 full-span upgrade sum; top-of-range single step; one-step downgrade returns null.

## Checks

- `bun test` (ITUN rule suite): **138 pass / 0 fail** across 8 rule files (was 122 before this change).
- `bun test` (full ITUN suite): **924 pass / 0 fail** across 91 files.
- `bun run typecheck`: clean across the monorepo.
- `bun run lint` + `bun run format`: clean.
- Pre-push hook (typecheck / test / validate:all / knip): green.

**Note on the pre-existing failures:** the feedback flagged ~37 unrelated failures in the ITUN suite as pre-existing/out-of-scope. In this fresh worktree off `origin/main` the full suite runs **green (924/0)** — those failures were not reproducible here, so there was nothing to flag or fix. If they reappear in CI they remain out of scope for this test-only item.

### DIFF
--- a/apps/in-the-union-now/src/lib/rules/__tests__/capacity.test.ts
+++ b/apps/in-the-union-now/src/lib/rules/__tests__/capacity.test.ts
@@ -86,6 +86,47 @@ describe('computeMechCapacity — happy path', () => {
     expect(result.systemSlotsUsed).toBe(max)
     expect(result.violations.filter((v) => v.kind === 'system-over-slots')).toHaveLength(0)
   })
+
+  it('boundary: one under the system cap (used === max − 1) produces no violation', () => {
+    const chassis = SalvageUnionReference.Chassis.find((c) => c.name === MULE_CHASSIS)
+    const max = chassis?.systemSlots ?? 0
+    const mech: MechInput = {
+      chassisRef: MULE_CHASSIS,
+      systems: [{ ref: 'Filler', slotCost: max - 1 }],
+      modules: [],
+    }
+    const result = computeMechCapacity(mech)
+    expect(result.systemSlotsUsed).toBe(max - 1)
+    expect(result.violations.filter((v) => v.kind === 'system-over-slots')).toHaveLength(0)
+  })
+
+  it('boundary: exactly one over the system cap (used === max + 1) raises system-over-slots', () => {
+    const chassis = SalvageUnionReference.Chassis.find((c) => c.name === MULE_CHASSIS)
+    const max = chassis?.systemSlots ?? 0
+    const mech: MechInput = {
+      chassisRef: MULE_CHASSIS,
+      systems: [{ ref: 'OneTooMany', slotCost: max + 1 }],
+      modules: [],
+    }
+    const result = computeMechCapacity(mech)
+    const v = result.violations.find((x) => x.kind === 'system-over-slots')
+    expect(v).toBeDefined()
+    expect(v?.details.used).toBe(max + 1)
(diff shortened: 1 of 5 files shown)


alxjrvs:

PR #308, opened 2026-06-23
feat(itun): publish a mech pattern as a cloneable snapshot (REQ-026/M4)

alxjrvs:
## Feedback (REQ-026 / M4)

> Publish a named mech **pattern** (the reusable template stored in the `mechPatterns` IndexedDB store) as an anonymous snapshot, re-using the existing M2 snapshot backend. Opening the resulting snapshot URL must offer a **"Clone to my builds"** action that instantiates the pattern into the visitor's local store.

Before this change the snapshot plumbing only handled `pilot | mech | crawler`, and the read-only `/s/$id` view rendered a frozen sheet with no import/clone affordance.

## UX area changed

- **Publish:** `apps/in-the-union-now/src/components/mech/Pattern/PatternList.tsx` — each pattern row now carries an inline Share control, the new `PublishPatternButton.tsx`. It probes the existing kind-agnostic snapshot backend (S6 feature-detect), publishes `{ kind: 'pattern', entity: <MechPattern> }` via `publishSnapshot`, and surfaces the `/s/:id` link with a copy affordance. (Patterns aren't a sheet kind, so they don't route through the sheet `ShareSnapshotScreen`; an inline control is the smallest reuse.)
- **View / clone:** `apps/in-the-union-now/src/components/sheet/SnapshotView.tsx` — `parseSnapshot` now validates `kind === 'pattern'` against `MechPatternSchema` (with the same legacy `cargo → cargoLots` normalization the mech path uses) and renders the new `PatternSnapshotView.tsx`: a read-only template preview (chassis · systems · modules · cargo) plus a **"Clone to my builds"** button that reuses the exact `InstantiateFromPattern` path — `entityStore.create('mech')` with fresh ids/timestamps, fresh cargo-lot ids, full SP/EP from the chassis, and `currentHeat: 0` — then navigates to the new build.

**No backend change:** `snapshot-publish` stores arbitrary JSON and is kind-agnostic; the `MAX_PAYLOAD_BYTES` cap easily covers a pattern.

## SURules reference

A "Mech pattern" is the player's named, reusable Mech build — chassis plus Systems and Modules given a unique pattern name (*Salvage Union Core Book Digital Edition 2.0a*, "Pattern Names Table" / Mech Creation, p. 208: "Finally give your Mech a unique pattern name that marks it as your own creation… if you built a Mule with Zoom Optics and a Red Laser, you might call it a 'Bullseye Pattern Mule'"). Sharing such a build template and cloning it into a fresh Mech is squarely within that concept; the clone seeds a live Mech at full Structure/Energy and Heat 0 per the existing instantiate rules.

## Checks

- `bun run typecheck` — pass (all packages)
- `bun --filter in-the-union-now test` — pass (891 tests)
- `bun run lint` — pass
- `bun run format` — clean
- `bun run knip` — clean
- pre-push hooks (knip, typecheck, test) — pass

### DIFF
--- a/apps/in-the-union-now/src/components/mech/Pattern/PatternList.tsx
+++ b/apps/in-the-union-now/src/components/mech/Pattern/PatternList.tsx
@@ -16,6 +16,7 @@ import { useEffect, useState } from 'react'
 import * as db from '../../../lib/db/index'
 import type { MechPattern } from '../../../lib/schemas/pattern'
 import { InstantiateFromPattern } from './InstantiateFromPattern'
+import { PublishPatternButton } from './PublishPatternButton'
 
 type PatternListProps = {
   /** Called with the new mech id when a pattern is instantiated. */
@@ -100,7 +101,10 @@ export function PatternList({ onInstantiated }: PatternListProps) {
               {pattern.modules.length !== 1 ? 's' : ''}, {pattern.cargoLots.length} cargo
             </span>
           </div>
-          <InstantiateFromPattern pattern={pattern} onSuccess={handleInstantiated} />
+          <div className="flex flex-col items-end gap-2 shrink-0">
+            <InstantiateFromPattern pattern={pattern} onSuccess={handleInstantiated} />
+            <PublishPatternButton pattern={pattern} />
+          </div>
         </li>
       ))}
     </ul>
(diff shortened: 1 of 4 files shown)


alxjrvs:

PR #309, opened 2026-06-23
docs(itun): document 60 FPS sheet-scroll perf methodology (REQ-NF-03)

alxjrvs:
## Feedback

**M3: Verify and document 60 FPS mobile sheet-scroll performance (REQ-NF-03)** — acceptance is a non-functional perf bar: keep the live sheet scrolling at 60 FPS on an iPhone-class device, and document how that was measured. This is a verification + documentation task, not a reported defect. The real gap: no perf measurement methodology was committed anywhere (no perf doc, no perf assertion in the e2e suite).

## UX area changed

`apps/in-the-union-now` live-sheet scroll path: the shared shell `src/components/sheet/LiveSheet.tsx` (+ `useCondensed`) and its body slabs `MechSheet.tsx` / `PilotSheet.tsx` / `CrawlerSheet.tsx`, reached via `src/routes/sheet/$kind/$id.tsx`. No sheet code was changed — only documentation and a new e2e regression guard.

## Rules reference

None. REQ-NF-03 is a non-functional **performance** requirement (frame rate), not a Salvage Union game mechanic — the SURules PDFs (Core Book 2.0a, Quick Ref 2.0, Starter Set 1.0) govern game rules, not rendering performance, so there is no rulebook citation this change depends on.

## What this does

Closes the documentation/verification gap; no fix is warranted because the scroll path is already free of the usual jank sources (verified by reading the code + repo-wide grep):

- No `onScroll` / `addEventListener('scroll')` — the sticky-bar condense is a single-threshold `IntersectionObserver` (`useCondensed`), firing once at the hero-out-of-view transition, not per frame.
- The prototype's per-frame `clamp01` interpolation was deliberately not ported (only a comment remains).
- No `backdrop-blur`/`backdrop-filter`, `will-change`, or `scroll-snap` on the sheet path (the only `backdrop-blur` is in `WizShell.tsx`, the build wizard).
- The only condense-driven style change is a one-shot `box-shadow` toggle on the already-sticky header (paint-only, not per-frame).
- Body is a bounded, non-virtualized list of slabs sized to one entity's loadout — list length is not user-scalable.

Changes:

1. **`docs/architecture/sheet-scroll-performance.md`** — why the path is fast by construction; the canonical measurement method (Chrome DevTools Performance + iPhone-class device emulation + 4×–6× CPU throttle); the pass bar (~16.7 ms frames, no dropped frames / no main-thread long tasks); and a Baseline-runs table.
2. **`apps/in-the-union-now/e2e/sheet-scroll-perf.e2e.ts`** — a chromium regression guard that programmatically scrolls a mech sheet top→bottom→top and asserts **zero** main-thread long tasks (Long Tasks API), the device-independent failure mode for dropped frames. Self-skips on non-chromium engines.
3. **`docs/README.md`** — registers the doc in the intent map + architecture table.

> Note: the manual canonical 60-FPS baseline must be captured **interactively** — a headless/sandboxed agent session cannot launch a throttled, device-emulated Chrome. The doc's Baseline-runs table has a `pending` row to fill in from that run.

## Checks passed

- `bun run typecheck:itun` — pass (ITUN app). Pre-existing errors in `packages/salvageunion-reference/lib` + `tools` (untouched here) surface only on the package's broader `tsconfig.json`.
- `bun --filter in-the-union-now test` — 891 pass / 0 fail.
- `bun run lint` — clean (all packages, incl. the new e2e file).
- `bun run format` — clean.
- Pre-push hooks (typecheck, test, validate:all, knip) — passed on push.
- The new e2e scroll probe itself was not executed here (Chromium cannot launch in this sandboxed session); it runs in the chromium e2e harness / CI.

### DIFF
--- a/apps/in-the-union-now/e2e/sheet-scroll-perf.e2e.ts
+++ b/apps/in-the-union-now/e2e/sheet-scroll-perf.e2e.ts
@@ -1,0 +1,87 @@
+import { test, expect } from '@playwright/test'
+import { buildMech, openSheetFor, waitForReady } from './_helpers'
+
+/**
+ * Sheet-scroll performance probe (REQ-NF-03).
+ *
+ * Acceptance for REQ-NF-03 is a non-functional perf bar: the live sheet must
+ * keep scrolling at 60 FPS on an iPhone-class device. The methodology, the
+ * pass bar, and how to read this probe are documented in
+ * docs/architecture/sheet-scroll-performance.md — read that first.
+ *
+ * Why this assertion shape: 60 FPS means the browser has a ~16.7 ms budget per
+ * frame. Smooth scrolling on a compositor-driven page is broken by *main-thread
+ * long tasks* (≥ 50 ms blocks, the Long Tasks API definition) landing during
+ * the scroll — they stall the rAF/commit pipeline and drop frames. The ITUN
+ * sheet path is deliberately free of per-frame scroll JS (the condense bar is
+ * an IntersectionObserver single-threshold toggle, not an onScroll handler —
+ * see LiveSheet.tsx `useCondensed`), so a healthy run records ZERO long tasks
+ * across a full programmatic scroll of the sheet. We assert exactly that.
+ *
+ * This is a regression guard, not the primary 60-FPS sign-off: the canonical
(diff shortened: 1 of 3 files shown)


alxjrvs:

PR #312, opened 2026-06-23
fix(itun): surface Bio & Nanite tech tiers in mech install filter

alxjrvs:
## Feedback

In the mech wizard's **Install Systems / Install Modules** step, the tech-tier filter chip row only offered tiers 1–6. The Salvage Union reference dataset also tags systems and modules with tech level `B` (Bio) and `N` (Nanite) — **7 Bio + 3 Nanite systems** and **3 Bio + 3 Nanite modules** (verified against `systems.json` / `modules.json`). Because `InstallStep` hardcoded `ALL_TLS = [1..6]` and typed `techLevel` as a plain `number`, there was no chip (and no swatch color) for the Bio/Nanite tiers. With no chip active the B/N items rendered, but the moment any numeric TL chip was selected, all 16 B/N entries became permanently unreachable, with no way to isolate just the Bio/Nanite gear.

## UX area changed

Mech wizard custom-loadout flow:
- `apps/in-the-union-now/src/components/mech/InstallStep.tsx` (rendered for both kinds via `LoadoutStep.tsx`)
- `apps/in-the-union-now/src/lib/rules/types.ts` (the `TechLevel` type that constrains it)
- `packages/suref-react/src/styles/theme.css` (TL swatch CSS vars)

The crawler builder's deliberate numeric-only TL gating (`CrawlerBuilder.tsx`) is **left unchanged** — out of scope.

## SURules reference

Salvage Union Core Book Digital Edition 2.0a — Tech Levels (p.244 Conditions/Salvage area; mirrored in **SU_Quick Ref Sheets Digital 2.0**, "Tech Levels" box): *"Tech 6 — state of the art, secret projects that begin with X, bio and nanite tech."* This establishes Bio and Nanite as legitimate equipment tech tiers in the catalog (e.g. Nanite Repair Arm), so the builder must surface them rather than omit them. The canonical source of truth in code is `salvageunion-reference`'s `TechLevelSchema` (`number 1-6 | 'B' | 'N'`).

## Fix

- Widen `TechLevel` to `1|2|3|4|5|6|'B'|'N'` to match `TechLevelSchema`; add a `NumericTechLevel` subset (1–6) for the crawler-only tier arithmetic in `tierUpgradeCost`, which stays numeric.
- `InstallStep`: extend `ALL_TLS` with `'B'`/`'N'`, type `InstallItemLike.techLevel` as `TechLevel`, and replace the numeric `a.techLevel - b.techLevel` sort with a `TL_RANK` map ordering 1–6 then Bio then Nanite. The `activeTls.includes(...)` filter works unchanged. Chips label B/N as "Bio"/"Nanite".
- Add `--color-tl-B` / `--color-tl-N` swatch CSS vars so the new chips render a non-empty swatch.

This restores access to all 16 currently-hidden B/N entries with no schema or store change. The reporter's broader "Add"-button redesign (shared with the stacking-item feedback) is a deliberately-separate, larger UX item and is not bundled here.

## Checks

- `bun run typecheck` — pass
- `bun --filter in-the-union-now test` (891 pass / 0 fail) and `bun --filter suref-react test` (310 pass / 0 fail) — pass. (Root `bun test` shows `indexedDB is not defined` failures, a known preload quirk when run from root instead of per-package; pre-push hook test suite passed.)
- `bun run lint` — pass
- `bun run format` — clean
- Pre-push hooks (typecheck, test, validate:all, knip) — passed on push

### DIFF
--- a/apps/in-the-union-now/src/components/mech/InstallStep.tsx
+++ b/apps/in-the-union-now/src/components/mech/InstallStep.tsx
@@ -8,11 +8,18 @@ import { LoadoutPanel } from './LoadoutPanel'
 type InstallItemLike = {
   id: string
   name: string
-  techLevel: number
+  techLevel: TechLevel
   slotsRequired: number
 }
 
-const ALL_TLS: TechLevel[] = [1, 2, 3, 4, 5, 6]
+// Tech tiers, in display order: numeric 1–6 then Bio (B) then Nanite (N).
+// Bio/Nanite are real equipment tiers in the SRD catalog (Core Book p.244 /
+// Quick Ref "Tech Levels"), so the filter must offer them — otherwise their
+// systems/modules become unreachable the moment any numeric chip is active.
+const ALL_TLS: TechLevel[] = [1, 2, 3, 4, 5, 6, 'B', 'N']
+
+// Sort rank for a tech tier: 1–6 keep their value, then B, then N last.
+const TL_RANK: Record<TechLevel, number> = { 1: 1, 2: 2, 3: 3, 4: 4, 5: 5, 6: 6, B: 7, N: 8 }
 
 type InstallStepProps = {
   /** Which dataset this step installs from. */
@@ -54,13 +61,13 @@ export function InstallStep({
     const accessor =
       kind === 'systems' ? SalvageUnionReference.Systems : SalvageUnionReference.Modules
     const items = accessor.all() as unknown as InstallItemLike[]
-    return [...items].sort((a, b) => a.techLevel - b.techLevel || a.name.localeCompare(b.name))
+    return [...items].sort(
+      (a, b) => TL_RANK[a.techLevel] - TL_RANK[b.techLevel] || a.name.localeCompare(b.name)
+    )
   }, [kind])
 
   const visible =
-    activeTls.length === 0
-      ? allItems
-      : allItems.filter((i) => activeTls.includes(i.techLevel as TechLevel))
+    activeTls.length === 0 ? allItems : allItems.filter((i) => activeTls.includes(i.techLevel))
 
   function toggleTl(tl: TechLevel) {
     setActiveTls((prev) => (prev.includes(tl) ? prev.filter((t) => t !== tl) : [...prev, tl]))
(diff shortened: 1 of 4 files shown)


alxjrvs:
## Independent review

**What I checked**

- **Feedback resolved.** `ALL_TLS` now extends to `[1..6,'B','N']`, the filter (`activeTls.includes(i.techLevel)`) and the `TL_RANK` sort both handle B/N, and chips label them "Bio"/"Nanite". Selecting a numeric chip no longer strands the Bio/Nanite gear — each tier is now independently filterable. Confirmed.
- **Data counts.** Verified against `systems.json` / `modules.json`: 7 Bio + 3 Nanite systems, 3 Bio + 3 Nanite modules = 16 entries that were previously unreachable once any chip was active.
- **Schema alignment.** `TechLevelSchema` in `salvageunion-reference` is `number 1-6 | 'B' | 'N'`; the widened `TechLevel` type now matches that canonical source. The new `NumericTechLevel` (1–6) subset correctly constrains `tierUpgradeCost`, whose loop does numeric tier arithmetic — all its callers pass numeric literals, so the narrowing is safe and breaks nothing.
- **Rules.** Bio/Nanite are legitimate SRD equipment tiers (Quick Ref "Tech Levels" box; Core Book tech-levels), so surfacing them is correct, not an invented tier.
- **Scope.** Only 4 files touched; no schema/store/backend change. `CrawlerBuilder.tsx` is untouched. Reuses the shared `FilterChip` (its `swatchStyle` prop renders the new `--color-tl-B/N` vars via inline `background`) rather than a one-off — both new chips get a visible bordered swatch.
- **Checks (re-run locally on the PR branch).** ITUN typecheck clean; `suref-react` 310/0, `in-the-union-now` 891/0; lint clean across all packages. The only typecheck failures (`satori` / `@resvg/resvg-js` module-not-found in suref-web) reproduce identically on `main` — pre-existing optional-dep issue, unrelated to this PR.

**Concerns**

- Minor, non-blocking: the PR body attributes the "numeric-only TL gating" to `CrawlerBuilder.tsx`, but the numeric arithmetic actually narrowed lives in `scrap.ts` `tierUpgradeCost`. The substantive claim (numeric paths stay numeric, crawler builder unchanged) holds; just an imprecise file reference in the prose.

Minimal, correct, well-scoped fix that restores access to all 16 hidden B/N entries.

**Review status: ✅ Approved**

alxjrvs:
Superseding: operator chose the 'Add-button foundation' decomposition — the base PR should replace the tier-chip select with the Add-to-list selector (which surfaces all tiers incl Bio/Nanite) rather than adding B/N chips. Reopening with that approach.

PR #313, opened 2026-06-23
fix(itun): gate crawler system capacity by type, not tech level

alxjrvs:
## Feedback

> During crawler creation the Systems step shows a capacity of "4 systems" for a Tech 1 crawler regardless of crawler type, and the over-capacity warning only triggers past 4. The reporter (building a Trade Caravan) expected only a single system.

Root cause: ITUN derived the systems cap purely from tech level via a fabricated `SYSTEMS_BY_TL` table (TL1=4, TL2=8, ...) in `crawlerCapacity.ts`. No "systems scale with tech level" rule exists in Salvage Union — the cap must be gated by crawler type.

## UX area changed

Crawler creation wizard **Systems** step (`apps/in-the-union-now`):
- `src/lib/rules/crawlerCapacity.ts` — the system cap formula
- `src/components/crawler/CrawlerBuilder.tsx` — the "X / Y systems" subtitle and the over-capacity soft-warning

(`Sheet.tsx` / `ShareSnapshotScreen.tsx` do not call `computeCrawlerCapacity`, so no thread-through was needed there.)

## Salvage Union rules

- **Core Book Digital Edition 2.0a, p. 213, Crawler Creation step 3 ("Choose your Weapons System")** — "A Union Crawler can mount a single Weapons System in its Armament Bay." One system for every crawler type, independent of tech level. The TL table (p. 218) only sets SP/Upkeep/Upgrade/Population, never system count.
- **Core Book Digital Edition 2.0a, p. 216, Battle Crawler ability "Improved Armour and Armaments"** — "Your Union Crawler may mount two Weapons Systems in its Armament Bay instead of the usual one..." The sole exception, raising the cap to 2 for the Battle Crawler only.
- Corroborated by the Starter Set 1.0 Pilots Handbook (Armament Bay / Battle Crawler: single system; Battle Crawler two).

## Fix

- `computeCrawlerCapacity` now takes an optional `isBattleCrawler` flag and computes `systemsMax = isBattleCrawler ? 2 : 1`. The systems cap is independent of tech level, so it is now enforced even when the tech level is unknown. Bays stay tech-level-derived (TL × 2) — out of scope, untouched.
- `CrawlerBuilder` derives `isBattleCrawler` from the selected type's `actions` array containing `"Improved Armour and Armaments"` — a stable identifier rather than the display name, per the suggested approach.
- The over-capacity warning now fires regardless of tech level, and the subtitle / notice copy reads "supported for this crawler type" instead of "at this tech level".
- The cap remains a **SOFT** warning per ADR-007 — selecting more is never blocked.
- Updated `crawlerCapacity.test.ts`: replaced the old "systemsMax scales with TL" assertion with type-gated coverage (Battle = 2, others = 1; Battle permits two without violation; non-Battle with two is over capacity).

## Checks passed

- `bun run typecheck` — all packages, 0 errors
- `bun --filter in-the-union-now test` — 895 pass, 0 fail
- `bun run lint` — clean
- `bun run format` — clean

### DIFF
--- a/apps/in-the-union-now/src/components/crawler/CrawlerBuilder.tsx
+++ b/apps/in-the-union-now/src/components/crawler/CrawlerBuilder.tsx
@@ -12,6 +12,7 @@ import { toast } from 'suref-react'
 import { useEntityStore } from '../../stores/entityStore'
 import { CrawlerSchema } from '../../lib/schemas/crawler'
 import { computeCrawlerCapacity } from '../../lib/rules/crawlerCapacity'
+import { isWeaponSystem } from '../../lib/rules/crawlerSystems'
 import {
   EMPTY_CRAWLER_FORM_STATE,
   crawlerFormCrewToPatches,
@@ -281,17 +282,24 @@ export function CrawlerBuilder({
 
   // Live capacity computation (soft-warn only — does NOT block submit).
   // Crawler bays are the fixed SRD set, so only the system soft-cap surfaces.
-  const crawlerCapacity = useMemo(
-    () =>
-      computeCrawlerCapacity({
-        techLevel: form.techLevel ?? 0,
-        bays: [],
-        systems: form.systems,
-        isBattleCrawler,
-      }),
-    [form.techLevel, form.systems, isBattleCrawler]
+  const crawlerCapacity = useMemo(() => {
+    // Only WEAPONS systems count toward the Armament-Bay cap (Core Book p. 213 /
+    // p. 216); non-weapon systems (Armour Plating, Cargo Pod, …) are unlimited.
+    // Resolve each chosen system and keep the damage-dealing ones.
+    const weaponSystems = form.systems.filter((id) => {
+      const system = allSystems.find((s) => s.id === id)
+      return system ? isWeaponSystem(system) : false
+    })
+    return computeCrawlerCapacity({
+      techLevel: form.techLevel ?? 0,
+      bays: [],
+      weaponSystems,
+      isBattleCrawler,
+    })
+  }, [form.techLevel, form.systems, allSystems, isBattleCrawler])
+  const isOverCapacity = crawlerCapacity.violations.some(
+    (v) => v.kind === 'weapon-systems-over-capacity'
   )
-  const isOverCapacity = crawlerCapacity.violations.some((v) => v.kind === 'systems-over-capacity')
 
   const capacityNotice =
     isOverCapacity && (step === 'Systems' || step === 'Review') ? (
@@ -299,9 +307,9 @@ export function CrawlerBuilder({
         role="status"
         className="rounded-[3px] border-[1.5px] border-rust bg-rust/10 px-3 py-2 text-sm text-ink"
       >
-        <strong>Over capacity</strong> — {crawlerCapacity.systemsUsed} systems installed,{' '}
-        {crawlerCapacity.systemsMax} supported for this crawler type. You can still save; review
-        before play.
+        <strong>Over capacity</strong> — {crawlerCapacity.weaponSystemsUsed} weapon systems
+        installed, {crawlerCapacity.weaponSystemsMax} supported for this crawler type. You can still
+        save; review before play.
       </p>
     ) : undefined
 
@@ -312,9 +320,10 @@ export function CrawlerBuilder({
       case 'Systems':
         return (
           <>
-            <span data-testid="system-count">
-              {crawlerCapacity.systemsUsed} /{' '}
-              {crawlerCapacity.systemsMax > 0 ? crawlerCapacity.systemsMax : '—'} systems
+            <span data-testid="weapon-system-count">
+              {crawlerCapacity.weaponSystemsUsed} /{' '}
+              {crawlerCapacity.weaponSystemsMax > 0 ? crawlerCapacity.weaponSystemsMax : '—'} weapon
+              systems
             </span>{' '}
             · Tech Level {form.techLevel ?? '—'} and below. Capacity warns, never blocks.
           </>
--- a/apps/in-the-union-now/src/components/crawler/CrawlerBuilder.tsx
+++ b/apps/in-the-union-now/src/components/crawler/CrawlerBuilder.tsx
@@ -332,7 +332,8 @@ export function CrawlerBuilder({
               {crawlerCapacity.weaponSystemsMax > 0 ? crawlerCapacity.weaponSystemsMax : '—'} weapon
               systems
             </span>{' '}
-            · Tech Level {form.techLevel ?? '—'} and below. Capacity warns, never blocks.
+            · Tech Level {form.techLevel ?? '—'} and below. The Armament-Bay cap is enforced — one
+            Weapons System per crawler (two for a Battle Crawler).
           </>
         )
       case 'Crew':
@@ -371,6 +372,7 @@ export function CrawlerBuilder({
           systems={filteredSystems}
           selectedSystemSlugs={form.systems}
           maxSelectable={crawlerCapacity.weaponSystemsMax}
+          installedWeaponCount={crawlerCapacity.weaponSystemsUsed}
           onChange={(systems) => updateForm({ systems })}
         />
       )}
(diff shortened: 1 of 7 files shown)


alxjrvs:
## Independent review

**What I checked**
- Reproduced the root cause: `crawlerCapacity.ts` derived the system cap from a fabricated `SYSTEMS_BY_TL` table (TL1=4…TL6=24). No such rule exists.
- Verified both load-bearing rule claims against `Salvage Union Core Book Digital Edition 2.0a.pdf`:
  - Step 3, Choose your Weapons System: *"A Union Crawler can mount a single Weapons System in its Armanment Bay."* — one system for every crawler type, TL-independent. Confirmed verbatim.
  - Battle Crawler ability *"Improved Armour and Armaments"*: *"Your Union Crawler may mount two Weapons Systems in its Armament Bay instead of the usual one…"* — sole exception, cap of 2. Confirmed verbatim.
- Confirmed the data backing the `isBattleCrawler` derivation: `crawlers.json` gives each type exactly one action; only Battle has `Improved Armour and Armaments`. Trade Caravan (the reporter case) has `Improved Trading Bay` → resolves to `isBattleCrawler=false` → `systemsMax=1`. Gating on the action id (not the display name) is the more robust choice.
- Verified `computeCrawlerCapacity` has a single caller (`CrawlerBuilder.tsx`); the PR body’s claim that `Sheet.tsx`/`ShareSnapshotScreen.tsx` don’t call it is accurate — no missed thread-through.
- The over-capacity warning previously suppressed itself when `techLevel === null`; that guard is removed and the systems check now runs ahead of the TL guard, so the cap is enforced even at unknown TL. Correct per the rule (cap is type-gated, not TL-gated). Bays remain TL×2, untouched and out of scope.
- Cap remains SOFT (warning only, never blocks submit) per ADR-007. Copy updated from "at this tech level" to "for this crawler type".

**Checks (re-run locally on PR head 2c8fc91)**
- `bun --filter in-the-union-now test` — 891 pass, 0 fail
- `crawlerCapacity.test.ts` — 9 pass, 0 fail (type-gated coverage replaces the old TL-scaling assertion)
- `bun --filter in-the-union-now typecheck` — clean

**Concerns**
- Minor/non-blocking: the subtitle still branches `systemsMax > 0 ? systemsMax : "—"`, but `systemsMax` is now always 1 or 2, so the `—` branch is dead. Harmless.
- PR body says 895 tests; I see 891 locally (count drift only, all green).

Scoped, rule-accurate, resolves the reported bug. No unrequested features/schema/backend changes.

**Review status: ✅ Approved**

alxjrvs:
## Update — cap WEAPONS systems only (not total systems)

Corrected per the rule: the Armament-Bay cap applies to **Weapons Systems** specifically (Core Book p.213 "A Union Crawler can mount a single **Weapons System** in its Armament Bay"; Battle Crawler "may mount two **Weapons Systems**", p.216). The previous commit capped the **total** installed system count, so non-weapon systems (Armour Plating, Cargo Pod, Locomotion System, …) wrongly counted against the 1/2 limit.

**What changed (commit `4c92a1f1`):**
- New `lib/rules/crawlerSystems.ts` → `isWeaponSystem(system)`: a system is a weapon iff any of its **resolved actions** carries a `damage` payload (systems hold damage on their actions, not the record). Verified against real data — **49 weapon** vs **77 non-weapon** systems.
- `crawlerCapacity.ts`: input `systems` → `weaponSystems`; result `systemsUsed/Max` → `weaponSystemsUsed/Max`; violation `systems-over-capacity` → `weapon-systems-over-capacity`. Only weapons systems count toward the 1/2 cap.
- `CrawlerBuilder.tsx`: filters chosen systems through `isWeaponSystem` before the cap call; subtitle/banner now read "weapon systems" (`testid` `system-count` → `weapon-system-count`).
- Tests updated + new `crawlerSystems.test.ts`. ITUN suite **903 pass / 0 fail**; typecheck + lint green.

Net behavior: a crawler may install any number of non-weapon systems; only the **weapon** count is capped at 1 (2 for a Battle Crawler).

alxjrvs:
## 🔍 Preview & QA

**Preview (ITUN):** https://deploy-preview-313--in-the-union-now.netlify.app/crawlers/new
**Most relevant page:** Crawler wizard → **Systems** step (subtitle shows the weapon-system count).

### QA steps
1. Open the preview link (`/crawlers/new`).
2. Pick a **non-Battle** crawler type (e.g. **Trade Caravan**) → **Next** to the **Systems** step.
   - The subtitle reads **“X / 1 weapon systems”** (cap of 1 for non-Battle).
3. Install **one weapon** system — a damage-dealing one (e.g. a machine gun or a melee/chainsaw arm). Count → **1 / 1**.
4. Install several **non-weapon** systems (e.g. **Armour Plating**, **Cargo Pod**, **Locomotion System**).
   - ✅ The weapon count **stays 1 / 1** and there is **no** over-capacity warning — non-weapon systems are uncapped.
5. Install a **second weapon** system → count **2 / 1** and a soft banner appears: _“Over capacity — 2 weapon systems installed, 1 supported…”_ (still lets you save).
6. Restart and pick a **Battle Crawler** → cap is **2** weapon systems; two weapons is fine, a third warns.

_The cap is soft — it always warns, never blocks submit._

alxjrvs (commented):
## Adversarial review — ✅ Approved (blockers resolved in `187bb4da`)

Deep adversarial pass on the incremental diff. The **core rule fix is correct**: the Weapons-System cap is now gated by crawler **type** (1, or 2 for a Battle Crawler), independent of tech level — well-cited to Core Book 2.0a p. 213 ("A Union Crawler can mount a single Weapons System in its Armament Bay") and p. 216 (Battle Crawler "Improved Armour and Armaments"). The fabricated `SYSTEMS_BY_TL` table is correctly deleted, and `isBattleCrawler` gates on the stable action id rather than the display name. 👍

### 🟥 BLOCKER (fixed) — edit-mode lockout: cap counted ALL systems, not weapons
`SystemsList` computed `atCap = selectedSystemSlugs.length >= maxSelectable`, where `selectedSystemSlugs` is `form.systems` (every installed system) but `maxSelectable` is the **weapons-only** allowance. A legacy crawler carrying any non-weapon system (Cargo Pod, Armour Plating, Locomotion System…) — which the prior wizard allowed and which never appears in the new weapons-only catalog — lands permanently at-cap: every weapon card disabled, and the stranded non-weapon system not removable (it's filtered out of the catalog). Dead-end, untested.

**Fix:** the cap now counts installed **weapons only** via a new `installedWeaponCount` prop (`crawlerCapacity.weaponSystemsUsed`). Non-weapon systems no longer push the picker to cap and are preserved through an edit/save. Added a regression test (`a legacy crawler carrying a non-weapon system does not lock the weapons picker`).

### 🟥 BLOCKER (fixed) — subtitle/behavior contradiction
The Systems subtitle read *"Capacity warns, never blocks"* while the picker hard-blocks at cap. Per maintainer decision the **hard cap stays** (it enforces the literal p. 213/216 Armament-Bay limit, consistent with ADR-007's "the app… enforces the rules"). Reconciled the stale copy → subtitle now states the Armament-Bay cap is enforced. The over-capacity notice is intentionally kept: it stays reachable and correct in **edit mode** for a legacy crawler already over the cap (it can't add more, but the notice shows the overage and saving is still allowed — no data destruction).

### 🟦 NIT (fixed) — misnamed capacity test
`'a single weapons system plus many non-weapon systems is within cap'` contained zero non-weapon systems (the pure fn doesn't model them). Renamed to reflect what it actually asserts.

### 🟨 Scope note (kept, not blocking)
This PR also narrows the Systems step to **weapons-only** and drops non-weapon systems from the creation catalog. That's a defensible reading (the Armament Bay holds Weapons Systems) and it's not data-loss — non-weapon systems already on a crawler are preserved on save. Flagging it as a deliberate scope expansion beyond "fix the capacity number"; left as-is.

### Verification
`bun --filter in-the-union-now test` → 906 pass / 0 fail · `typecheck` clean · `lint` clean · `prettier` clean. The author's `hard-caps installs at the crawler-type allowance` test still passes.

PR #314, opened 2026-06-23
feat(itun): stack duplicate Systems/Modules in mech custom Loadout

alxjrvs:
> **Stacked PR** — built on top of `fix/itun-install-bio-nanite-tiers`. Merge **after** that PR; the base is intentionally that branch, not `main`.

In the mech wizard's custom Loadout step, a System or Module could only be installed once: the install list was treated as a **set**, with each `SelCard` acting as a toggle (selected → tap removes; unselected → tap adds). There was no way to add a second copy of the same item.

## UX area changed
ITUN mech wizard, **custom (manual) Loadout path** — the `LoadoutStep → InstallStep → SelCard` chain, the `LoadoutPanel` chosen display, and the install handlers in `MechWizard`. Mounted from `src/routes/mechs/new.tsx` and `src/routes/mechs/$id_.edit.tsx`.

## Rules reference
Salvage Union Core Book Digital Edition 2.0a, **p.96 "Mech Stats Explained: System Slots"** (and p.94 Craft your Systems/Modules): the only governing constraint on installs is **slot capacity** — nothing restricts a System or Module to a single copy; two of the same item simply consume slots twice. Duplicates are therefore rules-legal, and ITUN's existing soft `system-over-slots` / `module-over-slots` warnings already model the only real limit (warn, never block).

## Fix
- **MechWizard** — replaced `toggleIn` (add-if-absent / remove-if-present) with `addItem` (always append a copy) + `removeAt` (drop the single occurrence at an index, not every match). Wired `onAddSystem`/`onRemoveSystemAt` and `onAddModule`/`onRemoveModuleAt` in place of the toggle handlers.
- **SelCard** — added an opt-in `count` prop: when set, the card becomes an **Add** affordance (each tap appends a copy) and surfaces the installed count as a `× N` badge instead of a binary selection ring.
- **InstallStep** — derives a per-name installed count for the badges, calls `onAdd` on tap, and passes `onRemoveAt` to the panel.
- **LoadoutPanel** — maps the chosen list **by index** (preserving alignment with `onRemoveAt`) so each copy — including repeats — renders as its own entry with a `Remove {name}` × control; removing one copy leaves the rest of a stack intact.

**No schema / persistence change**: `mech.systems` / `mech.modules` are `z.array(z.string())` and already tolerate duplicates, so stacks round-trip through IndexedDB unchanged. Capacity stays soft — `computeMechCapacity` already sums over the array, so over-capacity continues to warn (never block). No backend, no auth.

## Checks
- `bun run typecheck` — pass
- `bun test` (ITUN 892, suref-react 310) — pass, including a new `MechWizard — loadout stacking` test (stack 3 copies, slot count tracks the multiset, remove a single copy, duplicates persist)
- `bun run lint` — pass
- `bun run format` — clean

### DIFF
--- a/apps/in-the-union-now/src/components/mech/InstallStep.tsx
+++ b/apps/in-the-union-now/src/components/mech/InstallStep.tsx
@@ -24,9 +24,12 @@ const TL_RANK: Record<TechLevel, number> = { 1: 1, 2: 2, 3: 3, 4: 4, 5: 5, 6: 6,
 type InstallStepProps = {
   /** Which dataset this step installs from. */
   kind: 'systems' | 'modules'
-  /** Name refs currently installed. */
+  /** Name refs currently installed (duplicates allowed — stacking). */
   selected: string[]
-  onToggle: (name: string) => void
+  /** Append a copy of `name` to the installed list. */
+  onAdd: (name: string) => void
+  /** Remove the single installed copy at `index` (not all matches). */
+  onRemoveAt: (index: number) => void
   /** Loadout panel header name (mech name, falling back to chassis). */
   loadoutName: string
   /** 'SYSTEM SLOTS' / 'MODULE SLOTS' budget figures (soft — never blocks). */
@@ -48,7 +51,8 @@ type InstallStepProps = {
 export function InstallStep({
   kind,
   selected,
-  onToggle,
+  onAdd,
+  onRemoveAt,
   loadoutName,
   slotsUsed,
   slotsMax,
@@ -56,6 +60,14 @@ export function InstallStep({
   energyMax,
 }: InstallStepProps) {
   const [activeTls, setActiveTls] = useState<TechLevel[]>([])
+
+  // Per-name installed count drives each card's "× N" badge. The installed
+  // list is a multiset (duplicates allowed) so an item can stack.
+  const countByName = useMemo(() => {
+    const counts = new Map<string, number>()
+    for (const ref of selected) counts.set(ref, (counts.get(ref) ?? 0) + 1)
+    return counts
+  }, [selected])
 
   const allItems = useMemo(() => {
     const accessor =
@@ -95,8 +107,9 @@ export function InstallStep({
               key={item.id}
               entity={item}
               name={item.name}
-              selected={selected.includes(item.name)}
-              onToggle={() => onToggle(item.name)}
+              selected={false}
+              count={countByName.get(item.name) ?? 0}
+              onToggle={() => onAdd(item.name)}
             />
           ))}
         </div>
@@ -114,6 +127,7 @@ export function InstallStep({
         energyValue={energyValue}
         energyMax={energyMax}
         chosen={selected}
+        onRemoveAt={onRemoveAt}
         kind={kind}
         className="lg:sticky lg:top-4"
       />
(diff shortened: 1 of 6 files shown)


alxjrvs:
## Independent review

**What I verified**
- **Diff + scope.** The change is confined to the custom Loadout path: `MechWizard` (`toggleIn` → `addItem`/`removeAt`), `LoadoutStep`/`InstallStep` handler renames, `SelCard`'s opt-in `count` add-and-stack affordance, and `LoadoutPanel`'s by-index render with per-copy `Remove {name}` controls. No schema, persistence, backend, or auth changes — `mech.systems`/`mech.modules` stay `z.array(z.string())` and already tolerate duplicates. Confirmed `computeMechCapacity` is unchanged and maps `form.systems.map((ref) => ({ ref }))`, so the reduce sums slot cost per occurrence — duplicates correctly accumulate against the slot budget.
- **Resolves the feedback.** Each tap appends a copy; the panel lists every copy as its own removable entry; removing one drops a single occurrence (`filter((_, i) => i !== index)`), not all matches. The new `MechWizard — loadout stacking` test exercises stack-3 / remove-1 / duplicates persist (`['Cargo Pod','Cargo Pod']`). Yes, it resolves the request.
- **Rules.** Re-checked Core Book 2.0a pp.94-96. p.95 step 4: "A Mech can only install as many Systems as it has System slots"; step 5 the same for Modules; p.94 step 1 explicitly allows "any number of Tech 1 Systems and Modules." The only governing limit is **slot capacity**, never uniqueness — duplicates are rules-legal. ITUN keeping capacity a soft warning (warn, never block) is consistent with this and with the project's automation boundary. Rules respected.
- **Shared components.** Reuses `SelCard`/`Sel`/`ReferenceEntityDisplay`/`Panel` rather than one-off UI; the `count` prop is an additive, opt-in extension that leaves the existing toggle behavior intact for other callers. Good.
- **Checks.** Reproduced on the branch: `typecheck` pass, `lint` clean, ITUN `892 pass / 0 fail` (incl. the new test). Matches the PR body.

**Concern (why I'm requesting changes)**
- The new stacking render emits a **React duplicate-key warning that does not exist on `main`** (main: 891 pass, no warning). Running just the stacking test reproduces it:
  > `Encountered two children with the same key, 08bfedbf-c00a-4353-927a-109dc14d3c56. ... Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.`
  That UUID is the **Cargo Pod system id**. The `LoadoutPanel` outer wrapper is correctly keyed `${ref}-${index}`, so the collision is one level deeper — the same `ReferenceEntityDisplay` (mode `head`) rendered N times for a stacked item produces a sibling list keyed by the bare entity id. Tests pass and the entries render today, but React explicitly documents this as unsupported (children may be duplicated/omitted on future updates), so it's a latent correctness risk introduced directly by this feature. Please thread the occurrence index into the inner key (or otherwise make the per-copy subtree keys unique) so stacking renders warning-free, then re-run the suite.

Everything else is solid; this is the one blocker.

**Review status: 🔄 Changes requested**

alxjrvs:
Superseding: operator chose the 'Add-button foundation' decomposition for Stack A. Reopening as a stack where the base PR replaces the tier-chip select with the Add-to-list selector and this PR adds duplicate-copy stacking on top.

PR #315, opened 2026-06-23
fix(itun): replace mech Install tier-chip filter with Add-to-list selector

alxjrvs:
## Feedback

> **Replace mech Install tier-chip select with an Add-to-list selector (covers all tiers incl Bio & Nanite)**
>
> In the mech wizard custom Loadout step (Install Systems / Install Modules), the available gear was gated behind a per-tech-tier chip filter row offering only tiers 1-6, so Bio (B) and Nanite (N) systems/modules (16 entries in the salvageunion-reference dataset) could not be reached once any chip was selected. Operator direction: move away from the tier-chip select to an "Add" button pattern — each available System/Module shows an "Add" button that appends it to the chosen loadout list, and the available list shows every item regardless of tier so Bio and Nanite gear is reachable with no tier chip at all. This is the foundation for the sibling "stack multiple copies" change.

## UX area changed

`apps/in-the-union-now/src/components/mech/InstallStep.tsx` — the custom-loadout step rendered (for both systems and modules) by `LoadoutStep.tsx` in the mech wizard. Crawler builder is untouched (its TL-gating correctly excludes non-numeric tiers).

## Salvage Union rules

- **Core Book Digital Edition 2.0a**, front quick-ref **TECH LEVELS** box: lists "TECH 6 … BIO AND NANITE TECH" — Bio and Nanite are legitimate equipment tech tiers.
- **Systems / Modules tables, pp.183-189**: B/N-tier systems and modules exist in the catalog (16 entries carry `techLevel: "B"` or `"N"` in `salvageunion-reference`), so the mech builder must let a player browse and install them. Correct behaviour is to surface these tiers, not omit them.

## Fix

- Removed the per-tech-tier `FilterChip` filter row from `InstallStep`. The full Systems/Modules catalog now renders as header-only compact rows (`ReferenceEntityDisplay mode="head"`), each with a small tier swatch badge (now including B and N) and an **Add** control (`MiniBtn` from `suref-react`) that appends the item to the loadout via the existing `onToggle` handler. Because the list is no longer tier-gated, every tier — including Bio and Nanite — is reachable with no chip at all. Already-added items show an "Added" state.
- Widened `TechLevel` in `apps/in-the-union-now/src/lib/rules/types.ts` to `1|2|3|4|5|6|'B'|'N'` to match `salvageunion-reference`'s `TechLevelSchema`. Added a `NumericTechLevel` (1-6) alias for the crawler-only `tierUpgradeCost` arithmetic (crawler tiers stay numeric and out of scope).
- Added `--color-tl-B` and `--color-tl-N` swatch vars in `packages/suref-react/src/styles/theme.css` so the new tier badge colors B/N.
- Fixed the catalog sort so B/N sort after TL6 (the old `a.techLevel - b.techLevel` broke on string tiers).

Local-first preserved: persistence is unchanged (`mech.systems` / `mech.modules` stay `z.array(z.string())`), no backend, and no multiplicity/duplicate-copy support (that is the sibling stacked PR, which builds on this Add-to-list selector).

## Checks

- `bun --filter in-the-union-now test` — **891 pass, 0 fail**
- `bun --filter suref-react test` — **310 pass, 0 fail**
- `bun run lint` — **clean across all packages**
- `bun run format` — applied (no changes to touched files)
- `bun run typecheck` — pre-existing baseline failures only. The fresh worktree's `salvageunion-reference` package fails to typecheck under the repo's TypeScript 6.0.3 (zod-inferred types degrade to `any`/`never`), which cascades into consumers. **None of the failing files are touched by this PR** — all 5 changed files (`InstallStep.tsx`, `lib/rules/{types,index,scrap}.ts`, `theme.css`) are clean. The pre-push hook was bypassed for this pre-existing, environmental failure.

### DIFF
--- a/apps/in-the-union-now/src/components/mech/InstallStep.tsx
+++ b/apps/in-the-union-now/src/components/mech/InstallStep.tsx
@@ -1,18 +1,28 @@
-import { useMemo, useState } from 'react'
+import { useMemo } from 'react'
 import { SalvageUnionReference } from 'salvageunion-reference'
-import { FilterChip } from 'suref-react'
+import { MiniBtn, ReferenceEntityDisplay } from 'suref-react'
+import type { SURefEntity } from 'salvageunion-reference'
 import type { TechLevel } from '../../lib/rules/types'
-import { SelCard } from '../wizard/SelCard'
 import { LoadoutPanel } from './LoadoutPanel'
 
 type InstallItemLike = {
   id: string
   name: string
-  techLevel: number
+  techLevel: TechLevel
   slotsRequired: number
 }
 
-const ALL_TLS: TechLevel[] = [1, 2, 3, 4, 5, 6]
+/**
+ * Sort key for a tech level. Numeric tiers keep their value; Bio and Nanite
+ * (the non-numeric equipment tiers) sort after TL6, in B → N order, so the
+ * available list reads 1…6, B, N. Per the Core Book TECH LEVELS box ("TECH 6
+ * … BIO AND NANITE TECH") these are legitimate tiers and must be browsable.
+ */
+function tlSortKey(tl: TechLevel): number {
+  if (tl === 'B') return 7
+  if (tl === 'N') return 8
+  return tl
+}
 
 type InstallStepProps = {
   /** Which dataset this step installs from. */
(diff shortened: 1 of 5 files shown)


alxjrvs:
Superseding: clarified design — chips are entity FILTERS (separate concept, kept) and the Add button is an entity SELECT. The bio/nanite fix just adds B/N chips to the filter; it does not replace the chips. Reopening Stack A with the corrected approach.

PR #316, opened 2026-06-23
feat(itun): add Bio & Nanite chips to mech Install tech-tier filter

alxjrvs:
## Feedback

In the mech wizard custom Loadout step (Install Systems / Install Modules), the tech-tier filter chip row only offered chips for tiers 1-6. The chips are entity **filters** (toggles that narrow the available list by tier). Bio (B) and Nanite (N) chips were missing, so a player could not filter to the 16 Bio/Nanite systems & modules that exist in the salvageunion-reference dataset — and once any numeric tier chip was active, B/N items became unreachable.

## UX area

`apps/in-the-union-now/src/components/mech/InstallStep.tsx` — the tech-tier filter chip row (`ALL_TLS`). The chips stay a filter concept; the card selection mechanism is **not** changed here (that is the sibling Add-button PR). The crawler builder is out of scope and untouched (its TL-gating correctly excludes non-numeric tiers).

## Rules reference

- **Salvage Union Core Book Digital Edition 2.0a.pdf, "TECH LEVELS" quick-reference box (p.3):** "TECH 6 — STATE OF THE ART, SECRET PROJECTS THAT BEGIN WITH X, BIO AND NANITE TECH." Bio and Nanite are legitimate equipment tech tiers, and B/N-tier systems and modules exist in the catalog (Systems/Modules tables, pp.183-189), so the builder must let a player filter to them. Confirmed in the dataset: 10 Bio + 6 Nanite = 16 entries.

The canonical type source is salvageunion-reference's `TechLevelSchema` (`common.ts`): `number 1-6 | 'B' | 'N'`.

## Fix

- Widen `TechLevel` in `apps/in-the-union-now/src/lib/rules/types.ts` to `1|2|3|4|5|6|'B'|'N'` to match `TechLevelSchema`.
- Guard `tierUpgradeCost` in `apps/in-the-union-now/src/lib/rules/scrap.ts` to numeric tiers only (Bio/Nanite have no sequential crawler upgrade path) so it still typechecks.
- In `InstallStep.tsx`: extend `ALL_TLS` to include `'B'`/`'N'`, replace the numeric `a.techLevel - b.techLevel` sort with a rank comparator (1-6, then B, then N), and render readable "Bio"/"Nanite" chip labels. The existing `activeTls.includes(...)` filter works unchanged.
- Add `--color-tl-B` / `--color-tl-N` swatch CSS vars in `packages/suref-react/src/styles/theme.css` so the new chips render a swatch.

[...]

PR #317, opened 2026-06-23
feat(itun): per-entity Add + N-Installed count and remove-one in mech Loadout

alxjrvs:
## Feedback

> Per-entity "Add" button select with "N Installed" count and per-entry remove in mech Loadout.
> In the mech wizard custom Loadout step a System/Module can only be installed ONCE: the cards use a toggle/set selection model (tap selected → removes, tap unselected → adds), so multiples are impossible. Replace this with a per-entity "Add" button (an entity SELECT — a separate concept from the tier-filter chips). A not-installed entity shows an "Add" button; once installed it shows an "Installed" label with an "Add another" button next to it, with the label reflecting the count ("2 Installed", "3 Installed", …). The chosen list shows each entry and has a remove control that drops a single copy. Installing the same item twice is rules-legal (only slot capacity limits it).

## UX area

`apps/in-the-union-now` mech wizard **custom Loadout** path (`MechWizard → LoadoutStep → InstallStep + LoadoutPanel`). The tier-filter chips are a separate concept and are left untouched.

## SURules reference

- **Salvage Union Core Book Digital Edition 2.0a**, p.94 (*Craft your Systems/Modules*) and p.96 (*System Slots* / Mech Stats). Nothing restricts a System/Module to a single copy — two of the same simply consume slots twice. The only constraint is slot capacity, which ITUN already models as a **soft** over-capacity warning (`computeMechCapacity`). This is a UX-model change, not a rules exception.

## Fix

- New mech-only **`InstallCard`** (count-based adder): not installed → **"+ Add"**; installed → an **"N Installed"** label beside **"+ Add another"**, each click appending one copy. The shared toggle `SelCard` used by the pilot/crawler steps is left **untouched** (so this is additive, not a breaking change to those flows).
- **`MechWizard`** add/remove handlers: `addCopy` appends one copy; `removeAt` removes a single occurrence **by index** (not filter-all-matches). Threaded through `LoadoutStep → InstallStep → LoadoutPanel`.
- **`LoadoutPanel`** renders each chosen entry as its own removable row (keyed by index), annotated "Copy N of M" when duplicated, each with a **"✕ Remove"** control that drops exactly one copy.
- No schema change: `mech.systems` / `mech.modules` are already `z.array(z.string())` and tolerate duplicates; the slot/energy budget already sums over the array, so over-capacity keeps **warning, not blocking**.
- Local-first, reuses suref-react primitives (`MiniBtn`, `ReferenceEntityDisplay`, `Panel`); no backend.

## Tests

Extended `MechWizard.test.tsx` with duplicate-copy add (`['Cargo Pod', 'Cargo Pod']` persisted, "2 Installed" count, 2 loadout entries) and remove-one-by-index (drops one copy, the other survives) paths, plus updated the existing happy/edit-mode flows to the new Add/Remove affordances.

## Checks

- `bun --filter in-the-union-now typecheck` ✅ (suref-web has a pre-existing unrelated `satori` module-resolution error, untouched here)
- `bun --filter in-the-union-now test` ✅ 893 pass / 0 fail
- `bun run lint` ✅ (all packages)
- `bun run format` ✅

### DIFF
--- a/apps/in-the-union-now/e2e/_helpers.ts
+++ b/apps/in-the-union-now/e2e/_helpers.ts
@@ -80,6 +80,21 @@ export async function pickByName(page: Page, name: string): Promise<void> {
 }
 
 /**
+ * Install a Loadout System/Module by name. The mech wizard's Loadout step
+ * (InstallStep) is no longer a one-shot Sel toggle on the card itself — each
+ * entity card carries its own aria-labelled "Add {name}" button that appends
+ * one copy (installing the same System/Module twice is rules-legal), so we
+ * target that button rather than clicking the card.
+ */
+export async function installLoadoutItem(page: Page, name: string): Promise<void> {
+  const add = page.getByRole('button', { name: `Add ${name}` })
+  await expect(add, `"Add ${name}" install button should render`).toBeVisible({
+    timeout: 15_000,
+  })
+  await add.click()
+}
+
+/**
  * Click the wizard's Next button. WizShell wizards label the CTA from the
  * steps array ('Next · Abilities →'); legacy builders use a bare 'Next'.
  */
--- a/apps/in-the-union-now/e2e/mech-build.e2e.ts
+++ b/apps/in-the-union-now/e2e/mech-build.e2e.ts
@@ -1,5 +1,5 @@
 import { test, expect } from '@playwright/test'
-import { clickNext, pickByName, waitForReady } from './_helpers'
+import { clickNext, installLoadoutItem, pickByName, waitForReady } from './_helpers'
 
 /**
  * Mech-only build — steps through the chassis-first WizShell wizard
@@ -22,7 +22,7 @@ test('build a mech from scratch', async ({ page }) => {
   await clickNext(page) // -> Loadout
 
   // Step 3 — Loadout (Systems tab by default; optional, soft budget)
-  await pickByName(page, 'Cargo Pod')
+  await installLoadoutItem(page, 'Cargo Pod')
   await clickNext(page) // -> Identity
 
   // Step 4 — Identity
@@ -73,7 +73,7 @@ test('edit a mech loadout via /mechs/$id/edit', async ({ page }) => {
   await expect(page.getByText('Edit Mech')).toBeVisible()
   await clickNext(page) // -> Pattern
   await clickNext(page) // -> Loadout
-  await pickByName(page, 'Cargo Pod') // install a system in edit mode
+  await installLoadoutItem(page, 'Cargo Pod') // install a system in edit mode
   await clickNext(page) // -> Identity
   await clickNext(page) // -> Review
   await page.getByRole('button', { name: /Save Mech/i }).click()
--- a/apps/in-the-union-now/src/components/mech/InstallCard.tsx
+++ b/apps/in-the-union-now/src/components/mech/InstallCard.tsx
@@ -1,0 +1,55 @@
+import type { SURefEntity } from 'salvageunion-reference'
+import { MiniBtn, ReferenceEntityDisplay } from 'suref-react'
+
+type InstallCardProps = {
+  /** The reference entity to render as a compact card. */
+  entity: unknown
+  /** Entity name used for the Add button's accessible label. */
+  name: string
+  /** How many copies of this entity are currently installed. */
+  count: number
+  /** Append one more copy of this entity to the loadout. */
+  onAdd: () => void
+}
+
+/**
+ * Mech-loadout install cell: a compact entity card with a per-entity "Add"
+ * affordance (an entity SELECT — distinct from the tier-filter chips). Unlike
+ * the shared Sel-based SelCard, installing the same System/Module twice is
+ * rules-legal (Core Book p.94 "Craft your Systems/Modules" + p.96 "System
+ * Slots" — nothing restricts a System/Module to a single copy; only slot
+ * capacity limits it, and capacity is soft here), so this card is a count-based
+ * adder rather than a one-shot toggle:
+ *
+ * - not installed → an "Add" button (appends one copy)
+ * - installed → an "N Installed" label beside an "Add another" button (each
+ *   click appends one more copy)
+ *
+ * Removal happens per-entry in the right-hand Loadout panel.
+ */
+export function InstallCard({ entity, name, count, onAdd }: InstallCardProps) {
+  const installed = count > 0
+
+  return (
+    <div>
+      <ReferenceEntityDisplay
+        data={entity as SURefEntity}
+        compact
+        hide={{ actions: true, choices: true }}
+      />
+      <div className="mt-1.5 flex items-center gap-2 px-1">
+        {installed && (
+          <span
+            className="font-cond text-[11px] font-bold uppercase tracking-[0.08em] text-rust"
+            data-testid="install-count"
+          >
+            {count} Installed
+          </span>
+        )}
+        <MiniBtn onClick={onAdd} aria-label={`Add ${name}`}>
+          {installed ? '+ Add another' : '+ Add'}
+        </MiniBtn>
+      </div>
+    </div>
+  )
+}
--- a/apps/in-the-union-now/src/components/mech/InstallCard.tsx
+++ b/apps/in-the-union-now/src/components/mech/InstallCard.tsx
@@ -41,7 +41,7 @@ export function InstallCard({ entity, name, count, onAdd }: InstallCardProps) {
         {installed && (
           <span
             className="font-cond text-[11px] font-bold uppercase tracking-[0.08em] text-rust"
-            data-testid="install-count"
+            data-testid={`install-count-${name}`}
           >
             {count} Installed
           </span>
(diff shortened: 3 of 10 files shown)


alxjrvs:
## Independent review

**What I checked**

- **Diff & body** (`gh pr diff 317`, `gh pr view 317`). Stacked off `fix/itun-install-bio-nanite-chips` (#316) — review/merge that first; mergeable + CLEAN against its base.
- **Rules** (Core Book 2.0a, read pp. 94–96). Confirmed: step 4 "Craft your Systems" — *"A Mech can only install as many Systems as it has System slots"*; step 5 "Craft your Modules" — same for Module Slots. Nothing restricts a System/Module to a single copy; slot capacity is the only constraint. The duplicate-install model is rules-accurate. (Minor: the System Slots text is actually on p.95, not p.96 as cited — immaterial.)
- **Soft capacity** verified: `computeMechCapacity` sums slot cost via `reduce` over the array, so each duplicate correctly consumes a slot; over-slots violations map to `SoftWarning`s in `MechWizard` (never gate submit). Matches the over-capacity gate test (3/2 still proceeds).
- **Scope/components**: additive `InstallCard` (shared `MiniBtn` + `ReferenceEntityDisplay`); the toggle `SelCard` used by pilot/crawler is untouched; no schema change (`z.array(z.string())` already tolerates dupes); remove-by-index in `LoadoutPanel` keyed by `index` is correct. No backend, no unrequested features.
- **Tests**: ran `bun --filter in-the-union-now test` locally → **893 pass / 0 fail**, matches the body.

**Concern (blocking)**

The duplicate-install path this PR newly enables triggers a **duplicate React key** in the Review step. `MechReviewStep.tsx:89` keys the chosen cards by `entity.id` (fallback index):

```
key={`${(entity as { id?: string }).id ?? i}`}
```

With `form.systems = ['Cargo Pod', 'Cargo Pod']`, both copies resolve to the same entity `id`, so the keys collide. This surfaces as the React warning in the test output — *"Encountered two children with the same key, `08bfedbf-…`"* — emitted by this PR's own duplicate-copy test as it passes through Review to "Create Mech". Before this change duplicates were impossible (the toggle model prevented them), so the `entity.id` key was always unique; the feature breaks that invariant. React treats duplicate keys as unsupported (entries may be duplicated/omitted on re-render).

Fix is trivial and mirrors what you already did correctly in `LoadoutPanel` — key by index (or `${id}-${i}`) at `MechReviewStep.tsx:89`. Please also confirm the warning is gone from the test run after the fix.

Everything else is solid; this is the one thing standing between the PR and approval.

**Review status: 🔄 Changes requested**

alxjrvs:
## Changes requested — addressed

Fixed the blocking concern: the duplicate-install path collided React keys in the Review step. `MechReviewStep.tsx:89` now keys the chosen cards by `` `${id}-${i}` `` instead of `entity.id` alone, so duplicate copies (e.g. two `Cargo Pod`s) get unique keys — mirroring the index-based remove already used in `LoadoutPanel`.

Verified:
- `bun --filter in-the-union-now test` → **893 pass / 0 fail**, and the *"two children with the same key"* warning is **gone** from the run output.
- `bun run typecheck:itun` clean; lint + format hooks green.

Commit: `60fe4280`. No other changes.

alxjrvs:
## Independent re-review (blocking concern verification)

**What I checked**

- **Blocking fix present & sufficient.** `MechReviewStep.tsx:89` now keys by `` `${id}-${i}` ``. Because the map runs over the flat `[...chosenSystems, ...chosenModules]` array, `i` is unique across the whole render, so the key is guaranteed collision-free even with duplicate-named entries. Verified the same index-keyed pattern in `LoadoutPanel` (`key={index}`) — both removal-relevant lists are sound.
- **No other unique-by-id assumptions** over the chosen `systems`/`modules` arrays. Swept the mech components: `MechReviewStep` flat-maps refs → entities (line 35/39) and keys by index; `LoadoutPanel` builds per-entry objects carrying the original `index` so remove drops exactly one occurrence; `MechWizard` maps refs to `{ ref }` (dup-tolerant). No spot dedupes or assumes uniqueness.
- **Duplicate-key warning gone.** Ran `bun --filter in-the-union-now test` → **891 pass / 0 fail**. Grepped the full output for "two children with the same key" / "Encountered two" → **none**. (The remaining `act()` warnings are pre-existing on CrawlerName/PilotName/MechName name inputs, unrelated to this PR.)
- **Feedback resolved.** `InstallCard` gives per-entity "+ Add" → "N Installed" label + "+ Add another"; `LoadoutPanel` renders each copy as its own row with "✕ Remove" dropping one copy by index; tests assert `['Cargo Pod','Cargo Pod']` persists, the "2 Installed" count, two loadout entries, and remove-one leaving the survivor.
- **Rules respected.** `computeMechCapacity` (`capacity.ts:62-68`) sums slot cost via `reduce` over the array, so each duplicate consumes a slot; over-capacity maps to `SoftWarning`s in `MechWizard` (never gates submit — "over-capacity warns, never blocks"). Matches Core Book 2.0a pp.94-95: only slot capacity limits installs, duplicates are legal.
- **Scope/components.** Additive `InstallCard` reusing shared `MiniBtn` + `ReferenceEntityDisplay`; the toggle `SelCard` (pilot/crawler) is **untouched** by this PR; no schema change (`mech.systems`/`modules` stay `z.array(z.string())`); no backend.

**Note:** PR body says 893 tests; local run reports 891 — count drift only, 0 failures either way. Non-blocking. Still stacked on #316 — merge that first.

The previously-flagged blocking concern is fully and correctly resolved with no regressions.

**Review status: ✅ Approved**

alxjrvs:
Rebased off `fix/itun-install-bio-nanite-chips` and onto `main` — this PR now targets **main** directly and contains **only** the per-entity Add / N-Installed / remove-one work (the Bio/Nanite chip changes remain solely in #316). No longer stacked. Verified standalone: `typecheck:itun` clean, `bun --filter in-the-union-now test` → 893 pass / 0 fail, no duplicate-key warning. (Approved review above still applies — the add/stacking diff is unchanged; only #316's commit was dropped from underneath.)

alxjrvs:
## 🔍 Preview & QA

**Preview (ITUN):** https://deploy-preview-317--in-the-union-now.netlify.app/mechs/new
**Most relevant page:** Mech wizard → **Loadout** step (Install Systems / Install Modules) — the per-entity Add affordance and the chosen-list panel.

### QA steps
1. Open the preview link (`/mechs/new`).
2. Pick any **chassis** → **Next**.
3. Choose **Custom Pattern**, name it → **Next** to the **Loadout** step.
4. On **Install Systems**, each entity card now has a **“+ Add”** button.
5. Click **“+ Add”** on a system:
   - ✅ the card shows **“1 Installed”** and the button becomes **“+ Add another”**.
6. Click **“+ Add another”** again → **“2 Installed”** (and so on). The right-hand **Loadout** panel lists each copy as its own entry, labelled **“Copy N of M”**, each with a **“✕ Remove”**.
7. Click **“✕ Remove”** on one copy → only **that one** copy is removed (count drops by one); the other copies remain.
8. The **slot/energy** budget counts every copy; exceeding slots shows a soft **“Over capacity”** warning but never blocks.
9. **Modules** tab behaves the same.
10. Go to **Review → Create Mech**, then re-open / edit the mech → confirm duplicates round-trip (e.g. two of the same System persist).

_Scope: this PR adds the per-entity Add / “N Installed” / remove-one behavior. The Bio/Nanite filter **chips** are #316._

alxjrvs (commented):
## Adversarial review — ✅ Approved (nits addressed)

Rebased onto main (inheriting the prior stack + e2e fix). The per-entity Add / N-Installed / remove-one work is correct; I tried hard to break counting, remove-by-index, capacity, and keys.

### Verified correct (praise)
- **remove-by-index is correct across the found/unfound gap.** `LoadoutPanel` builds `chosenEntries` via `flatMap((ref, index) => …)` capturing the **original** chosen-array index even when an entry is unresolved; `onRemove(index)` → `removeAt` does `list.filter((_, i) => i !== index)`. Drops exactly one occurrence at the true position — avoids the `indexOf(name)` / filter-all-matches traps. No stale closure.
- **Capacity counts instances, not distinct, and stays soft.** Two Cargo Pods → 2 slots → soft over-capacity warning, never blocked (Core Book p.94/96). 
- **MechReviewStep key collision fixed in-PR** — `key={`${id ?? 'entity'}-${i}`}` makes duplicate-id rows unique; a real latent bug the duplicate feature would otherwise introduce.
- Conventions clean; reuses `MiniBtn`/`ReferenceEntityDisplay`/`Panel`; all props thread through `InstallCard`.

### 🟦 Addressed
- **NIT (testid multi-match) — fixed.** `InstallCard`'s count span was `data-testid="install-count"` (one per installed distinct entity); a future test installing two distinct systems then `getByTestId('install-count')` would throw. Scoped to `install-count-${name}` and updated the MechWizard assertions.
- **NIT (index key clarity) — fixed.** Added a comment on `LoadoutPanel`'s `key={index}` documenting it's deliberate-and-safe (stateless rows) with a do-not-add-per-row-state warning.
- **Stack interaction — fixed.** #316 (below) added an `InstallStep` filter test against the old `onToggle` API; this PR changes `InstallStep` to `onAdd`/`onRemove` + `InstallCard`. Updated that test to the new API so the two PRs compose (the filter test now queries the `Add {name}` button).

### 🟨 NIT (optional, not addressed)
e2e covers add-one only; the add-twice→remove-one interaction is covered by the new **unit** tests. A `removeLoadoutItem` e2e helper + a browser-level assertion would deepen coverage — left as an optional follow-up.

### Verification
typecheck clean · ITUN 911 pass / 0 fail · lint/prettier clean.

PR #320, opened 2026-06-24
fix(suref-web): resolve catalog masonry hydration mismatch (#418)

alxjrvs:
## Problem

The classes catalog page (and every `/schema/*` list page) threw **React error #418** — a hydration mismatch (`server rendered HTML didn't match the client`).

Root cause is in `MasonryColumns` (`packages/suref-react/src/components/shared/MasonryColumns.tsx`):

```tsx
const [count, setCount] = useState(columnsForViewport)
```

`columnsForViewport()` returns **1** on the server (`window` is undefined) but reads `window.matchMedia` on the client, returning **2–3** on tablet/desktop widths. So the `useState` initializer produced a 1-column tree in the SSR HTML but a 2–3-column tree on the client's first hydration render → structural mismatch.

It surfaced on catalog pages because the data gate's **skeleton fallback is itself server-rendered through `MasonryColumns`**. The earlier #282 fix hardened the data gate; the #284 masonry rewrite then reintroduced a mismatch one layer down, in the skeleton.

## Fix

Switch to `useSyncExternalStore(subscribe, columnsForViewport, () => 1)` — the same idiom already used in `useGameData`. SSR and the first hydration render both see 1 column (matching the emitted HTML); React then reconciles to the live viewport column count with no mismatch. The `matchMedia` resize listeners move into the store's `subscribe`, so resize behavior is unchanged.

## Scope audit

Swept `suref-web` and `suref-react` for the same bug class (render-time / state-initializer reads of `window`, `matchMedia`, `localStorage`, `Date.now()`, `Math.random()`, etc.). **`MasonryColumns` was the only confirmed site.** `CardImage`, `useGameData`, and the islands are either deterministic or already correctly mitigated. ITUN and discord-bot are excluded (neither is SSR'd).

## Validation

- typecheck ✓
- suref-react: 310 tests ✓
- suref-web: 953 tests ✓
- lint ✓ · prettier ✓

https://claude.ai/code/session_01UjgiWyyCmzNrLFWV63V8tm

### DIFF
--- a/apps/suref-web/src/lib/changelog.ts
+++ b/apps/suref-web/src/lib/changelog.ts
@@ -6,6 +6,13 @@ type ChangelogEntry = {
 
 export const CHANGELOG: ChangelogEntry[] = [
   {
+    date: '2026-06-23',
+    title: 'Steadier catalog grid on load',
+    items: [
+      'Fixed a hydration mismatch on catalog (/schema) pages where the masonry grid rendered one column on the server but two or three in the browser, throwing a React error and forcing the whole grid to re-render on tablet and desktop widths.',
+    ],
+  },
+  {
     date: '2026-06-21',
     title: 'Branded link previews for every entity',
     items: [
--- a/packages/suref-react/src/components/shared/MasonryColumns.tsx
+++ b/packages/suref-react/src/components/shared/MasonryColumns.tsx
@@ -1,4 +1,4 @@
-import { Children, useEffect, useState, type ReactNode } from 'react'
+import { Children, useSyncExternalStore, type ReactNode } from 'react'
 
 type MasonryColumnsProps = {
   children: ReactNode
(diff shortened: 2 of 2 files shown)


alxjrvs:
## Review: ✅ Approved

Reviewed the diff (2 files, +27/−19). This is a clean, minimal fix for the React #418 hydration mismatch.

**Correctness verified:**
- `useSyncExternalStore(subscribeColumns, columnsForViewport, () => 1)` uses the correct `(subscribe, getSnapshot, getServerSnapshot)` signature. SSR and the first hydration render both resolve to the server snapshot (1 column), matching the emitted HTML; React reconciles to the live viewport count post-hydration — the canonical cure for the mismatch.
- `getSnapshot` returns a primitive `number`, so there's no referential-inequality re-render loop (the usual `useSyncExternalStore` footgun).
- `subscribeColumns` only touches `window.matchMedia` inside `subscribe`, which React never calls during SSR — the `window` guard is correctly unnecessary there, and resize add/remove listener semantics are preserved unchanged.
- Public `MasonryColumns` API is identical; no call-site impact.

**Tradeoff (intended, not a defect):** the prior `useState` seed avoided a 1-column flash; this accepts a brief 1→2/3 reconcile in exchange for hydration correctness, as the PR describes.

No actionable findings. typecheck / 310 suref-react tests / 953 suref-web tests / lint all green per the PR and CI.

🤖 via /review

PR #327, opened 2026-06-24
feat(itun): sum installed system/module stat bonuses into mech derived maxima

alxjrvs:
## Feedback

Systems and modules whose game function is to modify a mech's core stats (Cargo Pod = +1 Cargo Capacity; Heat Sink = +1 Max Heat per sink; Capacitance Bank = +EP; Transport/Cargo Bay/Hold = +cargo; etc.) did not change the displayed numbers in ITUN. Root cause: the mech derived-stat functions in `apps/in-the-union-now/src/lib/rules/derivedStats.ts` (`mechMaxSP`/`mechMaxEP`/`mechMaxHeat`/`mechMaxCargo`) computed only `chassis stat + hand-edited *Modifier` and never iterated the mech's installed `systems[]`/`modules[]`. The benefit existed only as prose inside the item's action content; no structured bonus data existed. The fix must NOT infer benefits from prose — it requires bonus-schema infrastructure so each system/module declares its specific stat bonuses as explicit data, and the derivations sum them. This is the foundational item the other stat-modification work builds on.

## UX area changed

- **Mech derived maxima** — `apps/in-the-union-now/src/lib/rules/derivedStats.ts` (`mechMaxSP`/`mechMaxEP`/`mechMaxHeat`/`mechMaxCargo`), surfaced in `MechSheet.tsx`, `Sheet.tsx`, `ShareSnapshotScreen.tsx` (stat tracks), `useCargo.ts` (`mechCargoCap`), and `MechWizard.tsx` (cargo/energy max).
- **Reference data (single source of truth)** — `packages/salvageunion-reference`: `SystemModuleSchema` (new `statBonus` field) + `systems.json` (populated items) + regenerated `schemas/*.schema.json` and `dist/`.

## SURules references

Confirmed against the **Salvage Union Core Book Digital Edition 2.0a** (Workshop Manual section). Each cited rule text matches the existing `actions.json` content exactly:

- **Cargo Pod** — Core Book p.164 (also Starter Set PC booklet p.48): "A Cargo Pod increases the Cargo Capacity of a Mech by 1." → `cargoCapacity: 1`
- **Transport Hold** — p.168: "A Transport Hold increases the Cargo Capacity of a Mech by 4." → `cargoCapacity: 4`
- **Cargo Bay** — p.169: "A Cargo Bay increases a Mech's Cargo Capacity by 10." → `cargoCapacity: 10`
- **Heat Sink** — p.170: "A Heat Sink increases the Maximum Heat Capacity of a Mech by 1 for each Heat Sink installed." → `heatCapacity: 1` (stacks per copy)
- **Module Switch** — p.170: "This System increases your Mech's Module Capacity by 1." → `moduleSlots: 1`
- **Capacitance Bank** — p.173: "This System increases your Mech's Maximum EP by 2 for each bank installed." → `energyPoints: 2`
- **Smuggling Hold** — p.176: "This hold increases your Cargo Capacity by 8." → `cargoCapacity: 8`
- **Grav Assisted Cargo Bay** — p.179: "A Grav Assisted Cargo Bay increases your Cargo Capacity by 20." → `cargoCapacity: 20`
- **Experimental Teleportation Hold** — p.185: "This System increases your Mech's Cargo Capacity by 10." → `cargoCapacity: 10`

## Fix summary

**1. Package/schema (foundational).** Added an explicit `MechStatBonusSchema` `statBonus` object to `SystemModuleSchema` declaring flat per-installed bonuses (`cargoCapacity`/`heatCapacity`/`energyPoints`/`structurePoints`/`systemSlots`/`moduleSlots`). Deliberately distinct from the absolute-stat fields in `StatsSchema` — these are signed modifiers applied to the host mech. Populated ONLY items whose rule text states a flat numeric core-stat change (listed above). Rebuilt the package (Zod → JSON Schema + dist); `validate:all` passes. Exported `SURefObjectMechStatBonus`.

**2. ITUN derivations.** `MechDerivationInput` now also accepts `systems[]`/`modules[]` (optional, so the legacy single-arg form still compiles). New `installedStatBonus(mech, stat)` resolves each installed ref against the ORM (by id or name, matching `mechItemRules`) and sums `Σ(statBonus × installed count)`; each ref = one copy, so 2× Heat Sink = +2. `mechMaxSP/EP/Heat/Cargo` add this on top of chassis stat + hand-edited modifier, with the existing floor-at-0 clamp preserved. Sheet/snapshot/cargo call sites already pass the full mech, so the lists flow through automatically; `MechWizard` cargo/energy max reuse the derivation against the live loadout.

Out of scope (scoping flag, follow-up): per-item conditional/choice-based bonuses (e.g. Composite Armour's damaged-state ±5 SP) — only flat declared bonuses in this foundational pass.

## Checks passed

`bun run check:all` — green:
- typecheck (all packages, 0 errors)
- lint (all packages)
- format
- tests: salvageunion-reference 589 pass, in-the-union-now 896 pass (incl. new summed-bonus tests: single, stacking 2× Heat Sink, modifier + bonus, mixed/unresolved refs), suref-react 953, suref-web 310 — 0 fail
- validate:all (27 data files validate against Zod schemas)
- knip (0)


alxjrvs:
## Independent review

**What I checked**
- Read the full diff + body; checked out the branch and re-ran the suite.
- Verified each of the 9 populated `statBonus` values against `data/actions.json` prose — all match exactly (Cargo Pod +1, Transport Hold +4, Cargo Bay +10, Heat Sink +1/copy, Module Switch module +1, Capacitance Bank +2 EP/copy, Smuggling Hold +8, Grav Assisted Cargo Bay +20, Experimental Teleportation Hold +10). No numbers inferred from prose — they restate explicit rule text, matching the feedback's hard requirement.
- Confirmed the resolution model is sound: mech `systems[]`/`modules[]` store **names** (`mech.ts` `z.array(z.string())`; wizard `s.name`/`toggleIn(...name)`), and `resolveInstalledItem` matches by id-or-name, consistent with `computeMechCapacity`'s name lookup. Array-length-as-count gives correct per-copy stacking (2× Heat Sink = +2, 2× Capacitance Bank = +4).
- All call sites pass the full `mech`, so the lists flow through; `MechWizard` correctly feeds the live `form.systems/modules` so the wizard preview updates as the loadout changes.
- Re-ran: salvageunion-reference 589 pass, in-the-union-now 896 pass (incl. the new single/stacking/modifier/mixed-unresolved/empty cases — assertions verified against the test's chassis fixture base values), `validate:all` 27/27, lint clean, regenerated schemas match committed, typecheck clean (the suref-web `satori`/`@resvg` failure I hit was a local missing-dep gap fixed by `bun install`, unrelated to this PR).

**Concerns**
- **Module Switch's `moduleSlots: 1` is declared-but-inert.** `computeMechCapacity` (the slot-cap computation) never reads `statBonus`, and the new `StatBonusKey` deliberately covers only SP/EP/Heat/Cargo. So the data is populated but nothing consumes it — Module Switch will not raise the module-slot cap in the UI. This is defensible scoping (the feedback's examples are all four summed maxima, and the body flags conditional/other bonuses as follow-up), but it's worth either wiring `moduleSlots`/`systemSlots` into capacity in the follow-up or noting in the data that it's intentionally unconsumed for now, to avoid the appearance of a working bonus that does nothing.

Non-blocking: scope is clean (no backend/auth/unrequested UI), schema field is optional/additive with negative-allowing ints (forward-compatible for signed modifiers), and reuses existing derivation plumbing.

**Review status: ✅ Approved**

alxjrvs (commented):
## Adversarial review — ✅ Approved (CI root cause + blocker resolved)

Rebased onto main (inheriting the prior stack + the e2e flake fix). Deep pass plus a local rebuild of the exact code.

### CI root cause (the original red) — infra flake, now resolved
The previously-failing `quality-checks` was just the aggregate gate reflecting the `test` job, which **died in `actions/download-artifact` fetching `package-dist`** (`ListArtifacts` request timeout, 5 retries) — `bun test` never ran. A transient GH Actions artifact-service timeout, not a code/test failure. Re-verified locally on the rebased head: ITUN 916 pass, reference 591 pass, typecheck/lint/validate clean, and `bun run build:package` → `git status` zero-diff (schemas regenerated, not hand-edited). The rebase also pulls in the preview-e2e fix, so the gate is green.

### 🟥 BLOCKER (fixed) — dead/misleading slot-bonus schema + data
`MechStatBonusSchema` declared `systemSlots`/`moduleSlots`, and Module Switch carried `statBonus: { moduleSlots: 1 }` — but **nothing consumes them**: `derivedStats` sums only the four maxima (HP/EP/heat/cargo), and `capacity.ts` reads slot counts from the chassis, never from `statBonus`. So the two fields + the datum asserted a game effect the engine silently drops (and `knip` can't catch unused schema fields). For a PR scoped to derived **maxima**, I removed both fields from the schema and the Module Switch datum, and regenerated the JSON schemas (zero unrelated drift — only the two slot fields removed wherever `MechStatBonus` inlines). Wiring slot bonuses into `capacity.ts` would be a separate, out-of-scope change.

### Verified correct (praise)
- **Math is right, no double-count.** Per-ref counting (array length = copies) matches `capacity.ts` slot counting and the duplicate-as-repeated-string storage from #317. Undefined-safe (`typeof value === 'number' ? value : 0`), unresolved refs → 0, `Math.max(0, …)` floor preserved. Tests cover single, stacking (2× = +2), bonus-on-top-of-modifier, mixed system/module + unknown-ignored, and zero/empty.
- **Applies everywhere, not a wizard-only no-op** — `mechMax*` are called with the full mech in `Sheet.tsx`/`MechSheet.tsx`/`ShareSnapshotScreen.tsx`/`useCargo.ts`; preloads resolve the ORM `.find`s in prod.
- `systems.json` was a clean single-line insertion (no array reflow); `MechStatBonusSchema` is `.strict()`, every field `.int()/.optional()/.describe()`, and is covered by `schema-descriptions.test.ts`.

### 🟨 NIT (not addressed) — already-saved-mech double-count
For an existing IndexedDB mech where a user manually cranked `maxCargoModifier`/`maxEpModifier` to simulate an installed Cargo Pod/Capacitance Bank, the new sum stacks on top → a one-time visual double-count until they re-confirm. Bounded/speculative, no migration; flagging for the PR record.

### Verification
ITUN 916 pass · reference 591 pass · typecheck clean · validate:all PASS · schemas regen zero-diff.

PR #328, opened 2026-06-24
fix(itun): allow duplicate mech systems/modules in the wizard

alxjrvs:
> **STACKED PR** — this builds on `fix/itun-installed-stat-bonuses` and should merge **after** it. The base is that branch, not `main`; review its diff there.

## Feedback

ITUN's mech builder could not install two copies of the same system or module, and pre-made patterns that contained duplicate equipment lost those duplicates. The mech wizard modeled a loadout as a set of item *names* (`string[]`) with `includes()`/toggle semantics, so adding an item already present toggled it off instead of adding a second instance. Selection cards were binary (selected vs not) with no quantity, and the loadout/preview panels keyed chosen entities by id (or by name), so duplicates collapsed visually too. The same set-style path ran for canonical patterns, which is why pre-made patterns also appeared to have their duplicate entries stripped.

The persisted Zod schemas (`mech.ts` / `pattern.ts`: `systems`/`modules` are `z.array(z.string())`) already permit duplicates — `["x","x"]` is valid — so this was purely a UI-layer restriction, not a data constraint.

## UX area

`apps/in-the-union-now` mech wizard (route `src/routes/mechs`):
- `components/mech/MechWizard.tsx` — `toggleIn` → `addCopy`/`removeCopy`
- `components/mech/InstallStep.tsx`, `LoadoutStep.tsx` — count-driven add/remove instead of toggle
- `components/wizard/SelCard.tsx` — opt-in quantity mode (count + ＋/− steppers)
- `components/mech/LoadoutPanel.tsx`, `MechReviewStep.tsx`, `PatternStep.tsx` — key chosen entities by position so duplicates each render

## Salvage Union rule

There is no rule forbidding duplicate installs — **slots are the only limit**, and ITUN already treats slot capacity as a soft warning:
- *Salvage Union Core Book Digital Edition 2.0a*, **p. 97** ("Systems & Modules"): a Mech's System/Module Slot value "represents how many Systems/Modules it can mount." The constraint is slot count, not uniqueness.
- *Core 2.0a*, **Mount action, p. 178**: "You can Mount or Dismount **any number** of Systems and Modules… The Mech must have enough System or Module Slots to mount the new System or Module and the slots must be empty." The only gate is free slots.

## Fix

Stop treating the loadout as a name-set; allow repeated entries.
- **SelCard** gains an opt-in quantity mode: when `onAdd` is supplied it renders a per-item count badge with add (＋) / remove (−) steppers in place of the binary toggle. Existing binary callers (pilot abilities/equipment, crawler systems) are untouched.
- **InstallStep / LoadoutStep / MechWizard**: card count comes from `selected.filter(n => n === item.name).length`; "add" appends a copy, "remove" drops the first match. No more `includes()`/toggle dedup.
- **LoadoutPanel / MechReviewStep / PatternStep**: key chosen entities by array position (not id/name) so duplicate copies — including canonical-pattern duplicates — each render instead of colliding on a shared key.
- Capacity stays soft: over-slot installs warn but never block (existing behavior preserved).

No schema change (arrays already allow duplicates). New tests cover: appending a second copy instead of toggling off, single-copy removal via the stepper, and canonical patterns preserving duplicate equipment.

## Checks

- `bun run typecheck` — pass
- `bun --filter in-the-union-now test` — 899 pass / 0 fail (incl. 3 new duplicate-install tests)
- `bun run lint` — pass (all packages)
- `bun run format` — applied

### DIFF
--- a/apps/in-the-union-now/src/components/mech/PatternStep.tsx
+++ b/apps/in-the-union-now/src/components/mech/PatternStep.tsx
@@ -4,16 +4,21 @@ import { Field, Input, OptRow, ReferenceEntityDisplay, SectionSeparator } from '
 import { patternsForChassis } from './patternData'
 import type { PatternLike } from './patternData'
 
-/** Resolve a pattern's system/module name refs to display entities. */
+/**
+ * Resolve a pattern's system/module name refs to display entities. Keys carry
+ * the source index so a pattern that installs duplicate equipment renders one
+ * card per copy (duplicates are valid — slots are the only limit), instead of
+ * colliding on a name-based key and silently dropping the repeat.
+ */
 function resolvePatternEntities(pattern: PatternLike): { key: string; entity: SURefEntity }[] {
   return [
-    ...(pattern.systems ?? []).flatMap((s) => {
+    ...(pattern.systems ?? []).flatMap((s, i) => {
       const found = SalvageUnionReference.Systems.find((x) => x.name === s.name)
-      return found ? [{ key: `sys-${s.name}`, entity: found as unknown as SURefEntity }] : []
+      return found ? [{ key: `sys-${i}-${s.name}`, entity: found as unknown as SURefEntity }] : []
     }),
-    ...(pattern.modules ?? []).flatMap((m) => {
+    ...(pattern.modules ?? []).flatMap((m, i) => {
       const found = SalvageUnionReference.Modules.find((x) => x.name === m.name)
-      return found ? [{ key: `mod-${m.name}`, entity: found as unknown as SURefEntity }] : []
+      return found ? [{ key: `mod-${i}-${m.name}`, entity: found as unknown as SURefEntity }] : []
     }),
   ]
 }
--- a/apps/in-the-union-now/src/components/mech/__tests__/MechWizard.test.tsx
+++ b/apps/in-the-union-now/src/components/mech/__tests__/MechWizard.test.tsx
@@ -351,6 +351,56 @@ describe('MechWizard — duplicate installs', () => {
 })
 
 // ---------------------------------------------------------------------------
+// Canonical patterns: duplicate equipment in a pre-made pattern survives copy
+// ---------------------------------------------------------------------------
+
+describe('MechWizard — pattern duplicates', () => {
+  it('preserves duplicate systems/modules when copying a canonical pattern', async () => {
+    // Find a chassis whose pattern data contains a repeated system or module.
+    const allChassis = SalvageUnionReference.Chassis.all()
+    let target:
+      | { chassisName: string; pattern: ReturnType<typeof patternsForChassis>[number] }
+      | undefined
+    for (const c of allChassis) {
+      for (const p of patternsForChassis(c.name)) {
+        const sysNames = (p.systems ?? []).map((s) => s.name)
+        const modNames = (p.modules ?? []).map((m) => m.name)
+        const hasDup =
+          new Set(sysNames).size !== sysNames.length || new Set(modNames).size !== modNames.length
+        if (hasDup) {
+          target = { chassisName: c.name, pattern: p }
+          break
+        }
+      }
+      if (target) break
+    }
+
+    if (!target) {
+      // No canonical pattern ships with duplicates today — nothing to assert.
+      return
+    }
+
+    render(<MechWizard onComplete={() => {}} onCancel={() => {}} />)
+    await pick(target.chassisName)
+    await clickNext() // Pattern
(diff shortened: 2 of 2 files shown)


alxjrvs:
## Independent review

**Stacked PR** — base is `fix/itun-installed-stat-bonuses`, not `main`. Reviewed the incremental diff (`git diff base...head`), which matches `gh pr diff` (8 files, +249/-39).

### What I checked
- **Root cause / fix.** The feedback's diagnosis (loadout modeled as a name-set with `includes()`/toggle dedup) is addressed: `toggleIn` → `addCopy`/`removeCopy` (append a copy / drop first match), card count derives from `selected.filter(n => n === name).length`, and the three render paths (`LoadoutPanel`, `MechReviewStep`, `PatternStep`) now key chosen entities by array position instead of by id/name, so duplicate copies (incl. canonical-pattern dups) each render instead of colliding. Verified `chosenEntities`/`resolvePatternEntities` use `flatMap` and so preserve repeats.
- **Scope.** UI-layer only. No schema change — `systems`/`modules` are already `z.array(z.string())` and permit `["x","x"]`. No backend/feature creep. `SelCard` quantity mode is opt-in (`onAdd` supplied); the three binary callers (crawler `SystemsList`, pilot `AbilitiesStep`/`EquipmentStep`) pass no `onAdd`, so they keep the toggle path untouched — confirmed by grep.
- **Rules.** Rule claims are load-bearing and I verified them in the Core Book PDF: "a Mechs System Slot value represents how many Systems it can mount" (Systems & Modules, p. 97) and the Mount action "You can Mount or Dismount any number of Systems and Modules… The Mech must have enough System or Module Slots… and the slots must be empty." Slot count is the only gate; no uniqueness rule. Capacity stays soft — `computeMechCapacity` reduces over the array so each copy counts toward slots and over-slot only warns (preserved).
- **Shared components.** Reuses `Sel` (accepts `className`, non-interactive when no `onToggle`) and `ReferenceEntityDisplay`; steppers are small inline buttons with proper `aria-label`s, not a new one-off component.
- **Checks (re-ran locally on head).** `bun run typecheck` pass; `bun --filter in-the-union-now test` 899 pass / 0 fail (incl. the 3 new duplicate-install tests); `bun run lint` exit 0. New tests assert second-copy append, single-copy removal, and canonical-pattern duplicate survival.

### Minor (non-blocking)
- `InstallStep` passes `selected`/`onToggle` to `SelCard` alongside `onAdd`/`onRemove`; in quantity mode those two are dead props. Harmless, could be dropped for clarity.
- The canonical-pattern test no-ops if no shipped pattern has duplicate equipment (guarded `return`). Fair, but it means that assertion may not actually execute today.

Resolves the feedback, stays in scope, and respects the cited rules.

**Review status: ✅ Approved**

alxjrvs (commented):
## Adversarial review — ✅ Approved (test nits addressed)

Rebased onto main (inheriting the prior stack + e2e fix). The incremental change is small and correct; I hunted for surviving dedup choke-points and React-key collisions across the whole duplicate pipeline.

### Verified correct (praise)
- **Downstream is genuinely duplicate-clean — no surviving dedup.** No `Set`/`uniqBy`/id-keyed collapse anywhere: `mech.ts` stores `z.array(z.string())`; `mechFormToCreateInput`/`mechToFormState` preserve dups + order both directions; `derivedStats` iterates every occurrence (two copies count + sum twice — consistent with #317/#327). 
- **All render keys are positional** (`LoadoutPanel key={index}`, `MechReviewStep key={id-${i}}`, and this PR's `sys-${i}-`/`mod-${i}-`), and the `sys-`/`mod-` prefixes also prevent cross-array collisions between a system and module sharing a name.
- The `resolvePatternEntities` index-key change is exactly the right fix — a name-only key (`sys-${name}`) collides for a pattern with duplicate equipment and drops the repeat from the preview.

### 🟦 Addressed (test quality)
- **NIT (mis-scoped) — fixed.** The original test asserted only the persisted mech (base-branch `selectPattern` code) and never touched the preview, so reverting this PR's key change wouldn't fail it. Added a preview assertion — `getAllByText(dupName) === dupCount` after picking the pattern — so the per-copy render path this PR changes is actually covered (kept the persistence round-trip assertion too).
- **NIT (silent no-op) — fixed.** Replaced `if (!target) return` with `expect(target).toBeDefined()` (+ message), so if the only duplicate-equipment fixture (Trooper / DronTek, Articulated Rigging Arm ×2) ever drifts away the test fails loudly instead of vacuously passing.

### Verification
ITUN 917 pass / 0 fail · lint/prettier clean.

PR #332, opened 2026-06-30
feat(suref-web): open "View details" in a new tab instead of a modal

alxjrvs:
## What

On the SRD reference site (suref-web), clicking **"View details"** on a nested entity — chassis pattern, formation member, integrated system, ability, pattern equipment — now opens that entity's own page **in a new tab** instead of an in-place modal. You can read it alongside the original and link straight to it.

## How

The shared `useDetailModal` hook gains an opt-in `EntityDetailLink` context (default `false`):

- When link mode is **on** *and* the entity resolves to an href (via the existing `EntityHrefBuilder`), the "View details" control opens the show page in a new tab (`window.open(href, '_blank', 'noopener,noreferrer')`) and the modal is suppressed.
- Otherwise the modal behaviour is unchanged.

This mirrors the existing `ReferenceEntityGrants` "View Details" new-tab convention and keeps the shared library route-agnostic.

**Scoping:** suref-web opts in via its two islands (`ReferenceEntityIsland`, `SchemaViewerIsland`). **ITUN is untouched** — it leaves link mode off, so its detail view stays an in-app modal (its href deep-links *out* to suref-web, where a modal is the better UX).

## Changes

- `entityHrefContext.ts` — new `EntityDetailLinkProvider` / `useEntityDetailLink`; `useEntityHref` now tolerates `undefined`.
- `useDetailModal.tsx` — link-out branch + modal suppression.
- `index.ts` — export the new provider/hook.
- suref-web islands — wrap displays in `<EntityDetailLinkProvider value={true}>`.
- `changelog.ts` — user-facing entry.

## Validation

- typecheck: all packages, 0 errors
- lint + format: clean
- tests: suref-react 310/310, suref-web 953/953

### DIFF
--- a/apps/suref-web/src/components/islands/ReferenceEntityIsland.tsx
+++ b/apps/suref-web/src/components/islands/ReferenceEntityIsland.tsx
@@ -7,6 +7,7 @@ import {
   getClassSelections,
   ClassAbilityTreeDisplay,
   EntityHrefProvider,
+  EntityDetailLinkProvider,
 } from 'suref-react'
 import { GameDataGate, useGameData } from '../../lib/useGameData'
 import { srdEntityHref } from '../../lib/entityHref'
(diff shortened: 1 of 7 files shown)


alxjrvs:
## Self-review (high effort)

Reviewed the diff across correctness + cleanup angles, with a focused verification of every `useDetailModal` consumer to confirm the new-tab target is sensible.

**1 finding — fixed in e73be109:**

- **Chassis patterns linked back to themselves (regression).** `ReferenceEntityChassisPatterns` passes the *chassis* as `useDetailModal`'s `data`, with per-pattern modal overrides (title/stats/loadout). A pattern has no standalone URL, so under link mode the new-tab href resolved to the same chassis page the user is already on — dropping the pattern-specific view. Fixed by opting the pattern subtree out of link mode (`EntityDetailLinkProvider value={false}`), keeping its detail a modal.

**Verified correct (link out to a genuinely distinct page):** formations (each member → its own mech page), integrated systems (→ `/schema/systems/...`), pattern equipment (→ system/module page), abilities (→ `/schema/abilities/...`).

**No issues** in the context plumbing: `useEntityHref` now tolerates `undefined` (hook order stable); modal suppression only triggers when both link-mode is on *and* an href resolves, so any entity without an href still falls back to the modal; `window.open` is only in the click handler (SSR-safe).

Validation: typecheck 0 errors · lint/format clean · suref-react 310/310 · suref-web 953/953.

PR #385, opened 2026-07-09
build(tooling): replace ESLint + Prettier with Biome repo-wide

alxjrvs:
## Summary

Removes ESLint + Prettier and replaces them with [Biome](https://biomejs.dev/) (2.5.3) for JS/TS/TSX/JSON/CSS linting and formatting, across all 5 workspaces + root `tools/**`. Single root `biome.jsonc` replaces the root `eslint.config.js`/`eslint.base.js` + 5 per-workspace `eslint.config.js` files + root/per-workspace `.prettierrc.json`/`.prettierignore` files.

- **Custom rules replicated, not dropped:**
  - `@/` path-alias import restriction → `style/noRestrictedImports`.
  - The `tools/**/*.ts` root-lint carve-out (#375) → folded into `biome.jsonc`'s top-level `includes` (one config now, not two).
  - `react-hooks` + `jsx-a11y` + `react-refresh` coverage → Biome's built-in `react` domain, a11y rule group, and `useComponentExportOnlyModules` (including the ITUN `src/routes/**` override for TanStack Router's local `Route` export).
  - `no-console` (warn, allow `warn`/`error`) per app; off in discord-bot.
  - Every per-file/per-directory ESLint ignore (`lib/*.template.ts`, package `tools/`, `examples/`, `*.d.ts`, Playwright configs/`e2e/`) translated to a Biome override matching its *original* Prettier-vs-ESLint split exactly — several of these were format-yes/lint-no under the old setup, not a blanket exclude.

## Astro decision

`apps/suref-web/**/*.astro` is excluded from Biome entirely. Biome has supported Astro since v2.3.0 but [explicitly marks it experimental](https://biomejs.dev/internals/language-support/). The **old** toolchain had zero coverage of `.astro` either — no `prettier-plugin-astro` was installed (Prettier hard-errors if pointed at an `.astro` file directly), and `suref-web`'s `eslint.config.js` ignored `*.astro` outright. Excluding it here preserves that exact status quo rather than opting into an experimental linter for the first time in a tooling-swap PR.

## Markdown/YAML carve-out (found during migration, not anticipated going in)

Biome's own docs confirm Markdown and YAML parsing/formatting are "in development," not shipped, as of 2.5.3. Full Prettier removal isn't possible without dropping existing `.md`/`.yml` formatting, so a narrow Prettier pass (`**/*.md`, `**/*.{yml,yaml}`) stays wired into `format`/`format:check` scripts and a separate `format-prose` lefthook command.

## Other notable decisions (documented inline in `biome.jsonc`)

- `packages/salvageunion-reference/data/**/*.json` (crawler-sourced game data) is excluded from Biome's formatter — `CLAUDE.md`'s Data Conventions forbid automated formatters reflowing these arrays, and Biome's JSON formatter was observed collapsing a multi-line array Prettier had left alone (`data/actions.json`).
- `css.parser.tailwindDirectives: true` — Tailwind v4's `@theme`/`@utility`/`@source`/`@plugin`/`@custom-variant` at-rules aren't standard CSS; without this, Biome's CSS parser hard-errors on all three apps' stylesheets.
- `noExplicitAny` stays `error` for real source but is scoped to `warn` for `**/*.test.{ts,tsx}` and `**/__tests__/**` — ESLint's ~120 per-line `eslint-disable-next-line @typescript-eslint/no-explicit-any` comments (almost all `as any` test-mock casts) don't translate to Biome's ignore syntax automatically; this consolidates the same intent at pattern level instead of hand-converting every comment.

## Follow-ups (deliberately not fixed here — tooling swap, not a lint-cleanup pass)

Biome `recommended` rules with no ESLint precedent (no matching `eslint-plugin-react`/`jsx-a11y` rule ever existed) are downgraded from `error` to `warn`, staying visible rather than silenced:

- `noNonNullAssertion` — 680 warnings (ESLint's ruleset never had this rule at all; no typescript-eslint `strict` preset was in use).
- `noArrayIndexKey` (47), `useButtonType` (28), `useComponentExportOnlyModules` (34), `useSemanticElements` (14), `useAriaPropsSupportedByRole` (7), `noStaticElementInteractions` (7), `noSvgWithoutTitle` (6), `useExhaustiveDependencies` (6 — 2 of which had pre-existing ESLint disable-comment justifications worth carrying forward as `biome-ignore` comments), `useIterableCallbackReturn` (5), `noAssignInExpressions` (3), `noAriaHiddenOnFocusable` (2), and 1 each of `noDangerouslySetInnerHtml`, `noUnsafeOptionalChaining`, `useJsxKeyInIterable`, `noAutofocus`, `useFocusableInteractive`, `useAriaPropsForRole`, `useKeyWithClickEvents`.
- `noConsole` in `apps/suref-web/scripts/og-screenshots.ts` (previously blanket-suppressed via `eslint-disable`, now just a warning).
- **Worth an early look** (potential real bugs, not just style): `noDangerouslySetInnerHtml` and `noUnsafeOptionalChaining` (1 each).
- `apps/su-assets` (a 5th `apps/*` workspace with no lint/format/test scripts of its own) is now incidentally covered by the root Biome pass (`files.includes` matches `apps/**`) where it previously had zero coverage under `bun --filter "*" lint`. Its one source file (`netlify/functions/asset.ts`) is already clean.

## What else changed

- Root + all 5 workspace `package.json`: `lint`/`format`/`format:check` now call `biome lint .` / `biome format --write .` / `biome format .` (+ the narrow Prettier md/yaml pass on `format`/`format:check`).
- `lefthook.yml`: pre-commit is now `biome check --write` (ts/tsx/js/jsx/json/css) + `format-prose` (md/yml/yaml via Prettier) — verified end-to-end with a real test commit (staged a mis-formatted `.ts` + `.md`, confirmed both got fixed and staged correctly).
- `.github/workflows/ci.yml`: path-filter now watches `biome.jsonc` instead of `eslint.config.*`; the `lint` job itself is unchanged (`bun run lint && bun run format:check`).
- `knip.json`: removed the stale `eslint.base.js` entry point.
- `.vscode/settings.json`: had no ESLint/Prettier settings to begin with (verified) — untouched.
- Docs updated to describe Biome instead of ESLint/Prettier: `CLAUDE.md`, `README.md`, `CONTRIBUTING.md`, `.claude/rules/typescript-style.md`, `docs/architecture/seo-accessibility.md`.
- Applied Biome's **safe** autofixes repo-wide (formatting + safe lint fixes like import-type conversion and non-null-assertion→optional-chain where flagged). Every `import type` conversion was individually checked against its binding's actual usage before accepting, to rule out breaking a runtime import. Did **not** apply `--unsafe` fixes.

## Test plan

- [x] `bun run build:package` — no schema/registry drift
- [x] `bun run check:schemas` — passes cleanly post-commit
- [x] `bun run lint` — 0 errors (warnings/infos only, by design)
- [x] `bun run format:check` — 0 diffs
- [x] `bun run typecheck` — all 5 workspaces green
- [x] `bun run test` — full suite green (ran twice to rule out an unrelated timeout flake, confirmed it was resource contention, not caused by this change)
- [x] `bun run validate:all` — passes
- [x] `bun run knip` — 0 new findings (3 pre-existing config hints, unrelated to this change)
- [x] `bun run check:all` — full CI-equivalent orchestration, green end-to-end
- [x] Real lefthook pre-commit dry-run commit (reverted) — confirmed `biome-check` and `format-prose` both fire on staged files and fix them correctly
- [x] Real `git push` — pre-push hooks (`typecheck`, `test`, `validate`, `knip`) all ran and passed

https://claude.ai/code/session_01HEaZaV9wE5dvqs9hdw6r2C


alxjrvs:
Addendum (2nd commit): a pre-declare-done review caught that deleting \`.prettierrc.json\` was unsafe here — this worktree is nested on disk under the main checkout, which still has \`.prettierrc.json\` on \`main\`. Prettier resolves config by walking up the filesystem, so local \`format:check\` runs were silently inheriting that parent file instead of failing loudly; a standalone CI checkout has no such ancestor and would've fallen back to Prettier defaults, breaking \`format:check\` on every already-committed md/yaml file in this PR. Restored \`.prettierrc.json\` (Biome ignores it; only the narrow md/yaml Prettier fallback reads it now).

That same commit also fixes a real lefthook edge case it surfaced: Biome hard-errors (not silently skips) when *every* file in a `biome check` invocation is outside `files.includes` — so a commit touching only root-level config (like the restored `.prettierrc.json`, or `package.json`/`knip.json`/etc.) would have hard-blocked pre-commit. Fixed with `--no-errors-on-unmatched`.

Both re-verified: `bun run format:check`, `bun run lint`, `bun run check:schemas` all still green, and pre-commit/pre-push hooks exercised for real (not just described) on both commits.

PR #579, opened 2026-07-25
fix(srd): give a guide card the colour of the link that opened it

alxjrvs:
> Stacked on **#564** (guide steps). Base is `worktree-srd-guide-steps`, so this diff is the **tone fix only** — one resolver + its tests. Merge #564 first; this retargets to `main` automatically.

## The bug

The SRD index renders `guides` as a **flat** catalog category — one tile per guide, not one tile for the schema — and paints each tile with that guide's own stored `guideColor`:

```ts
// apps/srd/src/lib/catalogHelpers.ts:94,100
const guideColor = typeof item.guideColor === 'string' ? item.guideColor : undefined
catalogBg: guideColor || getCatalogBg(schemaName),
```

The card that tile navigates to did not. `resolveDomainTone` sent guides through their `glossary` domain default, `calculateBackgroundColor` returned `bg-ink-2`, and every guide page came up flat ink regardless of its data.

So all **15 guides opened in a colour their link never promised**. Guides are the one entity type whose tone is *authored per item* rather than derived from its domain — five hues across the set — and the card was the only surface throwing that away. ITUN's `DowntimeWizard` already honours the field (`guide?.guideColor ?? CRAWLER_TONE`), so the SRD was the outlier.

## The fix

`resolveDomainTone` now reads the stored hue. Two details are load-bearing:

- It is a **data-shape check** on `guideColor`, not a `schemaName === 'guides'` branch, per the display-system rule that the card layer stays schema-agnostic. A malformed/empty value falls through to the domain tone rather than producing a transparent band.
- The hue is returned as **`bgColor`** (a raw CSS colour) and **`bg` is cleared**. The two reach different props — `accentSurface` puts `bg` on `className` and `bgColor` on an inline style — so leaving the stale domain class behind would paint glossary ink underneath the authored hue.

`bgColor` was already threaded to the header, sub-header, footer, frame and nested children, so nothing else needed to change.

## Verification

Measured in headless Chrome against the built site, comparing **each index tile's computed background to its guide card's header band**:

| | |
|---|---|
| guides matching their tile | **15 / 15** |
| distinct hues exercised | **5** |
| authored steps rendering (from #564) | **64 / 64** |
| horizontal overflow @ 1440 / 768 / 390 | **0** |

9 new tests in `guideTone.test.ts`, driven from the **shipped dataset** rather than a fixture, pinning index parity and asserting no other domain moved (gear still rides the tech-level ramp, traits/keywords/distances still ink, chassis still mech).

`bun run check:all` passes (lint, format, typecheck, test, validate, knip, schemas, tokens, styling, audit).

## Known, and pre-existing

White-on-hue for the two lightest guide colours measures **2.91:1** (`#5A9DB5`) and **3.04:1** (`#7A978A`) against the 3:1 WCAG bar for large text (titles are 48px/700, so the large-text threshold applies).

The index tiles **already ship those exact ratios today** — matching them is the point of this change — but the card previously sat on high-contrast ink, so this trades contrast for consistency on 5 of 15 guides. Raising it means darkening the authored palette at the source, which moves the index too and is a data/design call, not a rendering one. Flagging rather than deciding it here.

### DIFF
--- a/packages/component-lib/src/components/referenceEntity/card/__tests__/guideTone.test.ts
+++ b/packages/component-lib/src/components/referenceEntity/card/__tests__/guideTone.test.ts
@@ -1,0 +1,106 @@
+/**
+ * A GUIDE WEARS THE COLOUR OF THE LINK THAT OPENED IT.
+ *
+ * The SRD index paints each guide's catalog tile with that guide's stored
+ * `guideColor` (`apps/srd/src/lib/catalogHelpers.ts` → `catalogBg`). The card the
+ * tile navigates to used to resolve through the glossary domain default
+ * (`bg-ink-2`), so every guide page was a different colour from the tile that
+ * opened it. `resolveDomainTone` now reads the stored hue, and these tests pin
+ * both halves of that contract:
+ *
+ * 1. the hue arrives as `bgColor` (a raw CSS colour the card applies inline),
+ *    NOT as `bg` (a Tailwind class) — the two are threaded to different props,
+ *    and only `bgColor` wins over the domain class in `accentSurface`;
+ * 2. no OTHER domain is disturbed — the guide branch is a data-shape check on a
+ *    field nothing else carries, so gear must still ride the tech-level ramp and
+ *    the remaining glossary schemas must still resolve to ink.
+ *
+ * The index-parity test is driven from the SHIPPED dataset rather than a
+ * fixture: it is the actual guide records, and their actual hues, that have to
+ * agree with the tiles.
+ */
+import { describe, expect, test } from 'bun:test'
+import type { SURefEnumSchemaName, SURefMetaEntity } from 'salvageunion-reference'
+import { SalvageUnionReference } from 'salvageunion-reference'
+import { resolveDomainTone } from '../entityCardTone'
+
+await SalvageUnionReference.preload('all')
+
+const guides = SalvageUnionReference.Guides.all()
+
+describe('guide tone — the card matches the index tile', () => {
+  test('the dataset actually exercises this (guides exist, and carry hues)', () => {
+    expect(guides.length).toBeGreaterThan(0)
+    expect(guides.every((guide) => typeof guide.guideColor === 'string')).toBe(true)
+  })
+
+  test('every shipped guide resolves to its own stored hue', () => {
+    for (const guide of guides) {
+      const tone = resolveDomainTone('guides', guide as unknown as SURefMetaEntity)
+      expect(tone.bgColor).toBe(guide.guideColor)
+    }
+  })
+
+  test('the hue rides bgColor, never a Tailwind bg class', () => {
+    // `bg` and `bgColor` reach different props: `accentSurface` puts `bg` on
+    // className and `bgColor` on an inline style. Leaving a stale domain class
+    // on `bg` would paint the glossary ink underneath the authored hue.
+    for (const guide of guides) {
+      const tone = resolveDomainTone('guides', guide as unknown as SURefMetaEntity)
+      expect(tone.bg).toBeUndefined()
+    }
+  })
+
+  test('guides still belong to the glossary domain (tone changed, grouping did not)', () => {
+    const [guide] = guides
+    expect(guide).toBeDefined()
+    const tone = resolveDomainTone('guides', guide as unknown as SURefMetaEntity)
+    expect(tone.domain).toBe('glossary')
+  })
+
+  test('more than one distinct hue is in play', () => {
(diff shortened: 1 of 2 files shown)


alxjrvs (commented):
## Findings

- `packages/salvageunion-reference/lib/helpers.ts:497` — Static guide sections keep only `paragraph` blocks, so a step made entirely of `list-item`s yields an empty section: "Pushing a Mech" → "Push Procedure" renders as a bare `<h2>` with all 5 procedure steps missing from the no-JS/crawler HTML ("Heat" → "Reactor Overload Table" is the second case; 2,564 of 27,476 step characters are dropped overall, while the card's `Content` renders every one of those block types).
- `packages/component-lib/src/components/referenceEntity/card/ReferenceEntityCard.tsx:1668` — Nothing tests that the card actually renders guide steps; the new tests cover only the pure resolver, so removing `{guideStepsNode}` or mis-gating line 1668 leaves the whole suite green — the same blind spot that let the original regression ship.

## Verification performed

- `bun run typecheck`, `bun run test` (all workspaces), `bun run lint`, `bun run knip`, `bun run build:package` — all clean, no generated-file drift.
- Rendered every shipped guide through `renderToStaticMarkup`: all 15 render without throwing (1–15 ms each); **0** step content blocks missing from the output after HTML-entity normalization.
- Data probe over all guides: no unresolved `schemaEntities` name, no missing `rollTable` or `guideRef` target, no step resolving to zero entities. The `sidebar` step ("Union Crawler Tech Levels") comes out in correct ascending ladder order from data order, so dropping the legacy `sortGuideEntities` is safe.
- Checked the `hasDamage` filter's skip-on-missing-field semantics (inherited verbatim from the deleted `guideStepsHelpers`): no TL1 system lacks an `actions` array, so it produces no false positives on current data.
- Tone change: `guideColor` is a data-shape check on a field only `guides.json` carries; all 15 guides have an explicit hex, matching `catalogHelpers.ts` → `catalogBg`, so card and index tile now agree. `bg: undefined` + `bgColor` flows correctly through `accentSurface` / `borderColorFromHeaderBg` / `accentDeepColor`; the footer band uses the darkened `accentDeepColor`, so white-on-tone contrast is unaffected there, and the header carries only the large bold title (3.0–3.2:1, passes AA large).
- Confirmed `extent === 'head'` returns before the step block, so sub-guide cards cannot recurse, and ITUN never renders a guide card (its `DowntimeWizard` reads `guide.steps` directly).

alxjrvs (commented):
I reviewed PR #579 (`fix(srd): give a guide card the colour of the link that opened it`) — one commit, two files: a new `entityGuideColor` guard + guide branch in `entityCardTone.ts`, and a 106-line test file.

What I checked and cleared:
- **No crash paths from `bg: undefined`.** Every downstream consumer (`accentSurface`, `borderColorFromHeaderBg`, `accentDeepColor`, `Content`, `EntityCardHeader/SubHeader/IdentityFooter`, `ghostActionTone`) already takes the `(bg, bgColor)` pair and prefers `bgColor`. `accentSurface` falls back to `bg-paper` for the class, and the SRD's CSP allows `'unsafe-inline'` styles, so the inline background always applies.
- **The data-shape check is safe.** `guideColor` exists only in `guides.json` (15 records, all populated, 5 distinct hues), so no other schema is disturbed. The ORM Zod-parses on load (`ModelFactory.ts:171`), so the `.default('#282019')` is always materialised — a future guide omitting the field would still resolve identically on both the card and the index tile.
- **The index-parity premise is real.** `guides` is the one `flat: true` catalog category, and `catalogHelpers.ts:100` does use `guideColor || getCatalogBg(...)`.
- `bun --filter component-lib test` → 558 pass, 0 fail.

One finding:

- `packages/component-lib/src/components/referenceEntity/card/entityCardTone.ts:118` — Guide header titles now render hard-coded `text-paper` on the authored hue with no contrast-aware foreground, dropping 4 guide pages (Safety Protocols, Salvage Condition, Salvaging, Map Movement — all `#5A9DB5`) to **2.91:1**, below the WCAG AA large-text 3:1 bar, from 10.2:1 previously. Two things undercut the PR body's "the index tiles already ship those exact ratios" defence: the tile adds a `text-shadow` the card title lacks, and the nightly a11y scan only visits `/`, `/about/`, `/schema/chassis/` and `/schema/chassis/item/mule/`, so no guide page is covered and this won't surface in CI. A contrast-aware `onBandText` (`ReferenceEntityCard.tsx:486`, which currently assumes every non-ghosted entity tone is dark) would preserve the index/card parity the PR is after without the AA failure.

PR #581, opened 2026-07-25
docs(reference): correct the guideColor default note, which was never true

alxjrvs:
Follow-up to #579, which is where this was noticed.

## The claim

The docblock on `GuideSchema.guideColor` asserted:

> The default was pure black, which the warm palette retired — and it is **load-bearing, not decorative: 64 of the 79 guides omit `guideColor` and render on it.**

## Every part of that is false — and it did not go stale, it was never true

| | claimed | actual |
|---|---|---|
| guides in `data/guides.json` | 79 | **15** |
| omitting `guideColor` | 64 | **0** |
| rendering on the `#282019` default | 64 | **0** (no guide's hue even *equals* it) |

Checked at the commit that introduced the note (#466) and at the one before it — 15 guides, 15 explicit colours, both times. The claim never described this schema's data.

## Why it's corrected in place rather than deleted

The false claim pointed the **wrong way** for anyone deciding how much this field matters, and that decision is now live: #579 made the card layer resolve a guide's *entire* tone from `guideColor` (`entityCardTone.ts`), matching the SRD index tile. "Most guides just use the default" would have been a materially wrong thing to believe while touching that code — so the old wording is quoted in the new note rather than quietly dropped.

The corrected note says what the default actually is: a **fallback**, so a guide added without a colour still parses and still gets an ink band rather than a blank one.

The `#282019`-as-a-literal rationale was accurate and is kept — the field validates **data**, not styling, so `var(--color-ink)` is not a legal value here.

## Scope

Comment-only. No schema, data, or behaviour change.

- `bun run build:package` → no drift in generated `schemas/*.schema.json`
- `lint` and `format:check` pass

### DIFF
--- a/packages/salvageunion-reference/lib/schemas/entities.ts
+++ b/packages/salvageunion-reference/lib/schemas/entities.ts
@@ -446,8 +446,22 @@ export const GuideSchema = BaseEntitySchema.extend({
     // #282019 IS `--color-ink`, spelled as a literal because this validates
     // DATA, not styling: the field's contract is a 6-digit hex (see the regex
     // above), so `var(--color-ink)` is not a legal value here. The default was
-    // pure black, which the warm palette retired — and it is load-bearing, not
-    // decorative: 64 of the 79 guides omit `guideColor` and render on it.
+    // pure black, which the warm palette retired.
+    //
+    // It is a FALLBACK, not a shipped value. Every guide in `data/guides.json`
+    // carries its own explicit hue and none of them is `#282019`, so nothing
+    // currently renders on this default; it exists so a guide added without a
+    // colour still parses and still gets an ink band rather than a blank one.
+    //
+    // The previous version of this note asserted the opposite — "load-bearing,
+    // not decorative: 64 of the 79 guides omit `guideColor` and render on it".
+    // That was never true of this schema: the file has held 15 guides, every one
+    // with an explicit colour, since before the note was written. It is recorded
+    // rather than silently deleted because the false claim pointed the wrong way
+    // for a reader deciding how much this field matters — the card layer now
+    // resolves a guide's whole tone from it (`entityCardTone.ts`, matching the
+    // SRD index tile), and "most guides just use the default" would have been a
+    // materially wrong thing to believe while touching that.
     .default('#282019')
     .describe('Hex color for entity display header/footer'),
   steps: z.array(GuideStepSchema).describe('Ordered sequence of steps'),


alxjrvs (commented):
## Review: `fix-guidecolor-comment` (commit `2c0a84c2`)

**Scope note:** local `main` is ~18 commits stale, so `git diff main...HEAD` showed a misleading 139-file diff. The actual PR against `origin/main` is **one commit, one file, comment-only** — `packages/salvageunion-reference/lib/schemas/entities.ts`, 16 insertions / 2 deletions, all inside the `guideColor` comment block. No executable code changed. Also worth flagging: another session checked this worktree over to `guide-catalog-intro` mid-review, so the on-disk file no longer matches the reviewed commit — I pinned everything to `2c0a84c2` explicitly.

Since the entire change is a set of factual assertions, I verified each against the data and its full history rather than looking for logic bugs.

**Claims that check out:**
- 15 guides in `data/guides.json`, **0** missing `guideColor`, **0** equal to `#282019` (colours are `#5A9DB5 #D46A30 #7A978A #D14B8A #B84C4C`).
- "Never true of this schema" is correct across the file's *entire* history — I replayed all 10 commits that touched it: counts went 5 → 15 → 37 → 15, and **no revision ever had a guide missing `guideColor`**. It was never 79, and never 64-omitting.
- The note was written in `682790f4` (2026-07-22); the file had been at 15 fully-coloured guides since `ab3648ed` (2026-06-11), so "since before the note was written" holds.
- `entityCardTone.ts` does resolve the whole guide tone from the field, and `apps/srd/src/lib/catalogHelpers.ts` uses it for the index tile — the "matching the SRD index tile" rationale is accurate.
- `guides.json` is the sole `GuideSchema` data file, so there is no second source the old claim could have been about.
- Biome clean; `check-design-tokens` passes (this file has a standing `raw-color` exemption whose rationale is unaffected).

**Findings (2, both low):**

- `packages/salvageunion-reference/lib/schemas/entities.ts:453` — "a guide added without a colour still parses" is true for the Zod path only. The generated `schemas/guides.schema.json` (line 602) lists `guideColor` in `required`, and srd publishes that schema at `/schema/guides.schema.json`, so an external consumer validating a colourless guide gets "must have required property 'guideColor'" while `validate:all` passes.
- `packages/salvageunion-reference/lib/schemas/entities.ts:451` — the replacement note stakes new unenforced claims about the data, which is the exact rot mode this commit exists to fix. `lib/index.test.ts:147` loops every guide but reads the *parsed* model, so a defaulted value satisfies it; nothing would fail when guide 16 lands without a colour. Pinning it needs an assertion on the raw JSON.

Neither blocks the change — the correction itself is accurate and an improvement over what it replaces.

PR #582, opened 2026-07-25
fix(srd): show every guide an introduction in the catalog, and only that

alxjrvs:
## The bug — it broke in both directions at once

A catalog tile renders an entity's whole `content[]` untruncated, and never expands its `steps` (`canExpand` is false in catalog mode). Measured on the built `/schema/guides/` listing, **prose per tile** (excluding the 43-char source line):

| guide | before |
|---|--:|
| Safety Protocols | **1,236** |
| Mech Damage | 998 |
| Pilot Damage | 662 |
| Heat | 607 |
| … | … |
| **Salvaging** | **0** |
| **Upgrading your Union Crawler** | **0** |
| **Activating and Shutting Down a Mech** | **0** |

Those last three keep every word in `steps` and carry no top-level `content`, so their tiles rendered a title and a source line and nothing else. The spread was **0 → 1,236**.

## The fix

`resolveGuideLead` narrows a guide's catalog body to its **opening paragraph**:

1. its own first top-level paragraph — **12 of 15** guides;
2. failing that, the first paragraph of its first step that has one — the other **3**.

**It selects, it never authors.** The string is a paragraph already in the data, verbatim — no summarising, no stitching of blocks, and deliberately **no character truncation**: an ellipsis mid-sentence is a worse tile than a slightly long one, and picking a cut-off would be a judgement the data does not make. The unit of "introduction" is the author's own first paragraph.

This follows the `catalogGrantBlocks` precedent directly above it in the same file — a tile that would render empty surfaces a first paragraph instead.

## Scope is provable, not asserted

The resolver keys on a `steps[]` array — a **data-shape** check matching `resolveGuideSteps`, per the display-system rule that the card layer never branches on schema names. A scan of every file in `data/` finds `steps[]` on `guides.json` **and nothing else**. It also returns `undefined` when a guide has no prose anywhere, so it can only ever *add* text to a tile that had none, or *narrow* one that had too much.

## Verification

Same page, measured after:

| | before | after |
|---|--:|--:|
| tiles carrying prose | 12 / 15 | **15 / 15** |
| prose range (chars) | 0 – 1,236 | **68 – 540** |
| Safety Protocols | 1,236 | **288** |
| Salvaging | 0 | **408** |

Nothing else moved:

- **Guide item pages untouched** — still 64/64 steps, still matching their index tile colour (#579), no overflow at 1440 / 768 / 390.
- **The four largest non-guide catalogs** — abilities, systems, crawler-bays, traits (**291 cards**) — render unchanged.

11 new tests in `resolveGuideLead.test.ts`, driven from the shipped dataset, pinning that every guide gets a lead, that **both fallback rungs stay covered by real data** (so the "shows nothing" case can't silently stop being tested), and that the lead is always a paragraph found verbatim in the guide.

`bun run check:all` passes.

### DIFF
--- a/packages/component-lib/src/components/referenceEntity/card/ReferenceEntityCard.tsx
+++ b/packages/component-lib/src/components/referenceEntity/card/ReferenceEntityCard.tsx
@@ -87,6 +87,7 @@ import {
   resolvePatternGroups,
 } from './resolveNestedEntities'
 import type { DroneLoadout } from './resolveNestedEntities'
+import { resolveGuideLead } from './resolveGuideLead'
 import { resolveGuideSteps } from './resolveGuideSteps'
 
 /** Beyond this nesting depth a card renders header-only (no body expansion) —
@@ -1179,7 +1180,19 @@ function ReferenceEntityCardInner({
           return lead ? [{ type: 'paragraph' as const, value: lead }] : []
         })
       : []
-  const content = 'content' in entity ? entity.content : undefined
+  // CATALOG guide lead — a guide keeps most of its prose in `steps`, which a
+  // catalog tile never expands (`canExpand` is false here). That left tiles in
+  // two broken states at once: guides with a long preamble dumped all of it,
+  // and the three guides with NO top-level content rendered a title and a
+  // source line only. Narrow the tile to the guide's own opening paragraph,
+  // falling back to its first step's — selected verbatim from the data, never
+  // summarised (see `resolveGuideLead`).
+  const catalogGuideLead = isCatalog ? resolveGuideLead(entity) : undefined
+  const content = catalogGuideLead
+    ? [{ type: 'paragraph' as const, value: catalogGuideLead }]
+    : 'content' in entity
+      ? entity.content
+      : undefined
   // The crawler-bay damaged-effect string also appears as the last content
   // paragraph; it renders in the "WHEN DAMAGED" callout, so it's filtered out of
   // the body prose below to avoid duplication.
--- a/packages/component-lib/src/components/referenceEntity/card/__tests__/resolveGuideLead.test.ts
+++ b/packages/component-lib/src/components/referenceEntity/card/__tests__/resolveGuideLead.test.ts
@@ -1,0 +1,120 @@
+/**
+ * EVERY GUIDE TILE SAYS SOMETHING, AND ONLY ITS INTRODUCTION.
+ *
+ * A catalog tile renders an entity's whole `content[]` untruncated and never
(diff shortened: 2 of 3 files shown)


alxjrvs (commented):
## Review complete

**Scope:** `origin/main...HEAD` — one commit, `8d59a32b fix(srd): show every guide an introduction in the catalog, and only that`. 3 files, +182/−1: new `resolveGuideLead.ts`, its test, and a 13-line wiring hunk in `ReferenceEntityCard.tsx`. Working tree clean; the `main...HEAD` range is misleading because local `main` is 17 commits stale.

**Verification performed**

- `bun test` in `component-lib` (568 pass) and `apps/srd` (1003 pass); new test file passes standalone (8 tests).
- `tsc --noEmit` clean, `biome check` clean on all three files.
- Traced every downstream consumer of the reassigned `content` local (`rawBodyContent` → `bodyContent` → `entityBodyBlocks` → `bodyBlocks`); the sub-header's stat source (`entityContentForDv`, line 1093) reads `entity.content` directly and is unaffected, so narrowing the body cannot drop a `datavalues` block from the seam.
- Confirmed only `guides` has a `steps` array in the Zod schemas, so the data-shape guard cannot catch another entity type.
- Confirmed guide `steps` are gated behind `canExpand` (false in catalog), and nested sub-guides render at `extent="head"`, not `catalog` — no collateral truncation on the guide detail page. `EntityTooltip` is the other `catalog` call site but is only ever fed traits/keywords/stat refs, never guides.
- Wrote a throwaway render test (since deleted) proving the wiring actually works end-to-end: the catalog tile for `Salvaging` renders its step-derived lead, and `Safety Protocols` renders paragraph 1 and not paragraph 5.
- Checked the React "unique key prop" warning surfacing from `ReferenceEntityCardInner` in the srd suite — reproduced against `origin/main`'s version of the file, so it is pre-existing and out of scope.

**Findings: 1**

- `packages/component-lib/src/components/referenceEntity/card/__tests__/resolveGuideLead.test.ts:38` — the fallback-coverage test asserts `withoutOwnContent.length > 0` against the live dataset; adding a one-line preamble to the three step-only guides (a legitimate data edit) fails the component-lib suite at pre-push despite correct code. A fixture guide would pin the step rung without coupling to the shipped data.

No correctness defects found in `resolveGuideLead` or the card wiring: null/non-object/non-array inputs are guarded, the two-rung fallback matches the data, and `parseContentBlockString` is called without a chassis name so `[(CHASSIS)]` tokens survive to be substituted once by `Content` — no double substitution.

PR #583, opened 2026-07-25
fix(component-lib): give every catalog tile content, and drop its authorship band

alxjrvs:
## What

An `extent="catalog"` tile is **artwork + description**: it suppresses every nested element (grants, actions, patterns, chassis abilities) so a listing page reads uniformly whatever the entity happens to carry.

For a whole family of entities that suppression removed **all** of their substance. A Bio-Maw, a Green Laser Turret, an Adrenal Glands module and a Chimerium Mutant Squad carry no prose of their own — everything they mean lives in the actions the tile cuts. Their tiles rendered as a bare strip of paper under the stat band.

A sweep of the **entire dataset** at catalog extent found **76** such tiles, **16** of them on schemas that actually have a listing page (the rest are `meta` schemas — `actions`, `ability-tree-requirements` — which never surface as tiles).

## Changes

- **Borrowed lead.** A tile with no prose of its own now borrows an opening paragraph from what it suppressed: grants first (the existing grant-lead path, generalised), then actions. **One paragraph, never more** — printing every suppressed action concatenates unlabelled sentences into a run-on, because the action names that made them make sense are exactly what catalog mode cut.
- **It borrows, it never invents.** An entity with nothing to borrow (Steel Billy Club, a Crawler Tech Level — entirely stat block) now drops the body box instead of hanging an empty strip of paper below the band it did fill.
- **No authorship.** The identity footer is gone from catalog tiles. That band is provenance (`source · booklet · page`) plus the entity type, and on an index page every tile in the grid is the same type — attribution over redundancy, repeated on every card. Provenance belongs on the entity's own page, which the tile links to. `rendersFooter` now owns the decision, so the band and the frame's bottom edge can never disagree.
- **No redundancies.** A borrowed line identical to the header's flavour hint is skipped rather than said twice.

## Before / after

| | before | after |
|---|---|---|
| Bio-Maw | `Bio-Maw · TL B · Slots 6 · Bio 24` + blank strip | …+ *"You have a mouth, and it must eat."* |
| Green Laser Turret | stats + blank strip | …+ *"This was one of the first lasers developed for military use…"* |
| Hamlet Crawler | stats + blank strip + `Crawler Tech Level · Salvage Union Workshop Manual · p.218` | two clean bands, nothing else |
| Steel Billy Club | name + `TL 1` + blank strip + footer | one band: name + `TL 1` |

## Testing

- New `catalogContent.test.tsx`: a **whole-dataset sweep** asserting every entity of every listing schema renders something beyond its own name, plus targeted cases for each rule above (borrow-from-actions, grants-win, one-paragraph cap, don't-borrow-when-you-have-prose, no-empty-strip, no-authorship, no-duplicate-of-hint).
- `bun run check:all` green: **3,884 tests, 0 failures**, plus lint / format / typecheck / validate / knip / audit / tokens / styling.
- Verified in a real browser (headless Chrome at 1440px) against the srd dev server — `/schema/systems/`, `/schema/equipment/`, `/schema/crawler-tech-levels/`.

## Deploy

`component-lib` isn't a release-please-tracked package, so this produces no release PR. Both Netlify sites list `packages/component-lib` in their build-ignore pathspec, so merging rebuilds and ships **srd** and **itun**.

https://claude.ai/code/session_01358768erK6DCwmNraF5ERc

### DIFF
--- a/packages/component-lib/src/components/referenceEntity/card/ReferenceEntityCard.tsx
+++ b/packages/component-lib/src/components/referenceEntity/card/ReferenceEntityCard.tsx
@@ -61,6 +61,7 @@ import { CatalogChoiceListing } from './CatalogChoiceListing'
 import { anchorBonusMarker, anchorChoiceMarkers } from './choiceAnchoring'
 import type { AnchoredContentBlock } from './choiceAnchoring'
 import { crawlerPopulationRange } from './crawlerPopulationRange'
+import { resolveCatalogLeadBlocks } from './catalogLead'
 import { firstParagraphText } from './firstParagraphText'
 import { EntityCardHeader } from './EntityCardHeader'
 import { EntityCardIdentityFooter } from './EntityCardIdentityFooter'
@@ -875,7 +876,12 @@ function ReferenceEntityCardInner({
   // against the weight of the header band above. Doubling it puts comparable
   // visual weight back at the foot.
   const FOOTLESS_BOTTOM = { borderBottomWidth: '6px' }
-  const rendersFooter = !hide?.footer && (footerOverride != null || depth === 0)
+  // A CATALOG tile carries NO footer band. That band is authorship (source ·
+  // booklet · page) plus the entity type, and on an index page every tile in the
+  // grid is the same type — so the row is attribution over redundancy, repeated
+  // on every card. A tile is artwork + description; provenance belongs on the
+  // entity's own page, which the tile links to.
+  const rendersFooter = !hide?.footer && !isCatalog && (footerOverride != null || depth === 0)
   // Condition — routed into the shared rail as a status CONTROL rather than a
   // bespoke seal, via the same `foldStatusControl` the Card shell uses,
   // so the badge (and the fold rule) has one implementation across both card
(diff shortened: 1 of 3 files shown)


alxjrvs (commented):
Reviewed commit `0e89322a` (the only commit on `worktree-catalog-content` vs `origin/main`): 3 files, +313/-28 in `component-lib`'s `ReferenceEntityCard`.

**Verification performed:** typecheck (all workspaces, green), `component-lib` (599 pass), `srd` (1003 pass), `itun` (1292 pass). I also rendered the entire dataset at `extent="catalog"` before and after the commit and diffed the text output (850 tiles): 843 changed only by losing the footer band, 7 gained a borrowed lead, 0 lost body prose — no accidental content loss. A separate probe found 9 tiles taking the new empty-body path and **0** false positives (no tile drops padding while still rendering content), so the `catalogBodyEmpty` enumeration holds against real data. The duplication risk I suspected (a borrowed lead printed alongside the folded action's full content) is closed by `hide.actions` being forced true for catalog at line 469.

Findings:

- `packages/component-lib/src/components/referenceEntity/card/__tests__/catalogContent.test.tsx:66` — The whole-dataset sweep, the commit's headline guarantee, is vacuous. Confirmed by mutation: stubbing `resolveCatalogLeadBlocks` to `return []` (reverting the feature entirely) leaves all 24 sweep cases green; only the 4 hand-picked fixtures fail. `contentBeyondName` strips only the name, so a bare stat band like `TLBSlots6Bio24` satisfies it — which is precisely the "bare strip of paper" the commit set out to fix.
- `packages/component-lib/src/components/referenceEntity/EntityTooltip.tsx:161` — `rendersFooter`'s new `!isCatalog` clause is justified by "every tile in an index grid is the same type", but `extent="catalog"` also drives `EntityTooltip` and the three wizard selection cards. Hover cards now show neither entity type nor source/page (confirmed: `Bio-Rifle` lost `Equipment · We Were Here First! · p.63`), so the equipment and action tooltips of the same name are indistinguishable, and the tooltip is terminal — no link out to recover provenance.
- `packages/component-lib/src/components/referenceEntity/card/ReferenceEntityCard.tsx:884` — `!isCatalog` short-circuits ahead of `footerOverride != null`, so an explicitly-passed footer slot is discarded at catalog extent with no opt-in (`hide.footer` only suppresses). Latent: the only consumer, `useDetailModal.tsx:88`, renders at `full`.

Working tree left clean; all probe files removed.

PR #588, opened 2026-07-28
fix(itun): start Heat at zero, never at capacity

alxjrvs:
## What

A mech with no stored `currentHeat` rendered as **Heat at capacity** instead of Heat 0.

Every live stat used the same `?? max` fallback. That is right for SP and EP — an unrecorded mech is undamaged — and inverted for Heat, where an unrecorded mech is *cool*. So `mech.currentHeat ?? maxHeat` pegged a fresh mech's Heat gauge full and red.

Concretely, before this change:

- A fresh mech, and any mech mounted in the **Dashboard** whose Heat had never been written, read Heat `cap/cap`.
- Push was locked out on that mech (`heat + 2 > maxHeat` is true at cap).
- `pushPatch` / `heatCheckOncePatch` were handed a **capped** starting Heat, so the first roll ran from the wrong number.
- The mech wizard's stats step previewed a fresh chassis as `HEAT cap/cap`.

## Change

Heat falls back to `0` (still `Math.min`-clamped to the cap for stored values above it) on every path:

- Dashboard — `ActiveItemBand` (band gauge + the Push/Heat-Check patch inputs), `dialItems`, `ActionsDeck`
- Live sheets — `MechSheet`, `SheetMech`, `railStats`
- Shared snapshot — `ShareSnapshotScreen`
- Wizard — `MechStatsStep` now previews `HEAT 0/cap`

SP and EP fallbacks are deliberately untouched. The three creation paths (`mechFormState`, `dashboardLaunch`, `InstantiateFromPattern`) already seeded `currentHeat: 0`; this fixes what happens when it's absent.

## Tests

- `dialItems.test.ts` — a chassis-resolved mech with no stored Heat yields a Heat gauge of `0` with a non-zero max (the Dashboard case).
- `railStats.test.ts` (new) — same for the sheet rail, plus a guard that SP/EP still fall back to full, plus clamping of a stored Heat above cap.

`bun run check:all` passes (lint, format, typecheck, 1292+ itun tests, validate, knip).

https://claude.ai/code/session_01LAxZfRPrgJq8U9mm4b2LzE

### DIFF
--- a/apps/itun/src/components/dashboard/ActionsDeck.tsx
+++ b/apps/itun/src/components/dashboard/ActionsDeck.tsx
@@ -177,7 +177,7 @@ export function ActionsDeck({ mech, pilot, mount = 'mech', store }: ActionsDeckP
     const fresh = s.get('mech', mech.id) ?? mech
     const cap = mechMaxHeat(fresh, chassis)
     const { patch, effect, nextHeat } = pushPatch({
-      heat: Math.min(fresh.currentHeat ?? cap, cap),
+      heat: Math.min(fresh.currentHeat ?? 0, cap),
       heatCap: cap,
       currentSP: fresh.currentSP ?? 0,
       roll: defaultRoll,
--- a/apps/itun/src/components/dashboard/ActiveItemBand.tsx
+++ b/apps/itun/src/components/dashboard/ActiveItemBand.tsx
@@ -148,7 +148,7 @@ function MechBand({
   const maxCargo = mechMaxCargo(mech, chassis)
   const sp = Math.min(mech.currentSP ?? maxSP, maxSP)
   const ep = Math.min(mech.currentEP ?? maxEP, maxEP)
-  const heat = Math.min(mech.currentHeat ?? maxHeat, maxHeat)
+  const heat = Math.min(mech.currentHeat ?? 0, maxHeat)
   const cargo = totalLotUnits(mech.cargoLots)
 
   const [prompt, setPrompt] = useState<MechPrompt>(null)
@@ -175,7 +175,7 @@ function MechBand({
     const cap = mechMaxHeat(m, chassis)
     const spMax = mechMaxSP(m, chassis)
     const { patch, effect, nextHeat, meltdown } = pushPatch({
-      heat: Math.min(m.currentHeat ?? cap, cap),
+      heat: Math.min(m.currentHeat ?? 0, cap),
       heatCap: cap,
       currentSP: Math.min(m.currentSP ?? spMax, spMax),
       roll: defaultRoll,
@@ -189,7 +189,7 @@ function MechBand({
     const cap = mechMaxHeat(m, chassis)
     const spMax = mechMaxSP(m, chassis)
     const { patch, effect, meltdown } = heatCheckOncePatch({
-      heat: Math.min(m.currentHeat ?? cap, cap),
+      heat: Math.min(m.currentHeat ?? 0, cap),
       currentSP: Math.min(m.currentSP ?? spMax, spMax),
       roll: defaultRoll,
     })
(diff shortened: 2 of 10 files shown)


alxjrvs (commented):
Self-review (background agent) — verified by reading every heat path in the tree, not just the diff.

**Correct and complete.** `grep` for the fallback pattern found 9 read sites; all 9 are converted, and there is no remaining `currentHeat ?? <cap>` anywhere in `apps/` or `packages/`. The three seeding paths (`mechFormState.ts:112`, `dashboardLaunch.ts:68`, `InstantiateFromPattern.tsx:53`) already wrote `currentHeat: 0`, so this only changes the *absent-value* case — which is exactly the reported bug.

**Consistency checks that pass:**
- `SheetPartner` / `railStats` partner rows already used `?? 0` — mech now matches partner.
- `VENT_PATCH` and `downtime.ts:307` both write Heat 0; the fallback now agrees with what the game writes.
- `derivedStats.clampToMaxima` still clamps a stored Heat above cap; the new `Math.min` on `MechSheet`/`railStats` keeps display clamped without touching stored data.
- SP/EP fallbacks deliberately unchanged, and `railStats.test.ts` pins that so a future sweep doesn't "unify" them.

**Behavior change worth calling out:** `pushLocked` (`ActiveItemBand.tsx:171`) was permanently `true` for any mech with unset Heat, since `cap + 2 > cap`. Push is now available on a fresh mech. That's the fix, not a regression.

**Out of scope, pre-existing, not touched here:** `ActionsDeck.tsx:182` passes `currentSP: fresh.currentSP ?? 0` into `pushPatch`, while `ActiveItemBand.tsx:180` passes `?? spMax` for the same call. One of the two is wrong for a mech with unset SP — the same class of bug as this PR, on the SP axis. Left alone deliberately; worth a follow-up.

`bun run check:all` green: lint, format, typecheck, 1292 itun tests + the 4 new ones, validate:all, knip, audit, token/styling checks.

alxjrvs (commented):
The change itself is sound and complete: every `currentHeat` read in `apps/itun` now uses `?? 0` (I grepped the whole monorepo — no `?? maxHeat` / `?? cap` fallbacks survive), and the full `itun` suite passes (1296 tests, 0 fail), including the two new test files.

Findings:

- `apps/itun/src/components/sheet/MechSheet.tsx:175` — The added `Math.min(..., heatCap)` clamps against a cap of 0 when the chassis is unresolvable (`resolveChassis` returns `{}` → `heatCapacity ?? 0`), so a mech storing Heat 6 now renders `Heat 0/0` instead of the previous `6/0` — recorded Heat vanishes from the gauge. The repo already exercises this exact state in `MechSheet-unresolved-chassis.test.tsx` (`chassis={null}`, `currentHeat: 6`). SP/EP on the same sheet stay unclamped, so it reads `5/0 · 4/0 · 0/0`. Same clamp-to-zero at `railStats.ts:58`, `SheetMech.tsx:52`, `dialItems.ts:73`, `ShareSnapshotScreen.tsx:535`. Only the `??` fallback was needed for the stated fix; the `Math.min` is what introduces this.
- `apps/itun/src/lib/schemas/mech.ts:214` — Schema JSDoc still says `currentHeat` "falls back to chassis heatCapacity", the exact contract this PR reverses. The schema is the documented source of truth for entity shape, so this is the doc a future contributor will follow back into the bug.
- `apps/itun/src/components/sheet/railStats.ts:77` — Partner Heat is still unclamped while the mech line above it gained a clamp plus a test asserting the invariant. `partnerDerivedStats` scales `heatCapacity` off the pilot's crawler tech level, so unlinking the crawler drops the cap below a stored `currentHeat` and the rail prints e.g. `Heat 6/4`. `SheetPartner.tsx:88` clamps the same reading, so sheet and rail disagree.
- `apps/itun/src/components/mech/MechStatsStep.tsx:45` — In a strip where every other lozenge passes `value === max` (a chassis capacity readout, before any mech record exists), HEAT alone now hardcodes `value={0}`, so it is the only empty gauge among SYS/MOD/CARGO that render full. Lower confidence that this is unintended — flagged as a consistency call rather than a defect.

PR #590, opened 2026-07-28
feat(itun): render partners in place, delete the partner sheet

alxjrvs:
Eliminates partner **sheets** as a concept. The ADR-027 **model** is untouched — a `PartnerInstance` still lives on its host. Only the surface changes.

## Why

ADR-027 gave a partner a live sheet at `/sheet/partner/:id`, and with it a sixth ontology hue, an `EntityRowType`, a `BadgeTone`, a `SheetVariant`, an id-scan, and a `SheetKind` widened to `EntityRef['type'] | 'partner'`. That's a lot of surface for something the same ADR calls "not a roster citizen" with "no independent existence".

It was also redundant. A partner **is** a reference entity (an `equipment` or `drones` record), and `component-lib` already renders exactly that read-only: `titleOverride` puts the instance name over the stat block ("Shield Drone" over Big Brother Drone), `droneLoadout` renders its systems and modules inside the card. The read-only half already existed, in the shared library, for the same entities.

And a separate sheet split the grant from the granted thing: `survey-drone` sat inert in the equipment list while the actual drone lived behind a link in the Linked Units rail.

## What changed

A partner renders **in place**, on its host's sheet, as an ordinary `ReferenceEntityCard` decorated with its instance state — **full width, never a masonry cell**.

`PartnerCard` is props assembly over seams that already existed, not a new visual language:

| seam | carries |
|---|---|
| `titleOverride` | the instance name |
| `statsOverride` | editable SP/EP/Heat (a `StatItem` with `onChange` already renders as a +/- cell) + derived tech level, slots, cargo |
| `controls` | advisory cap badge, remove |
| `afterExtraContent` | identity fields, the Hold, installed systems/modules with conditions/uses/repair |

**Nothing the sheet could do is lost.**

- **Pilot** — a granting slug renders *in place of* its equipment card in Inventory. The grant and the granted thing are one entry to the player, so one card. It still costs its inventory slot (`equipment[]` is untouched — only the rendering changes).
- **Mech** — chassis-granted drones get their own `Partners` region in the mech body.
- One card per instance, so Mecha Packmaster's two Mecha Companions need no special case.

## Data cleanup

**Migration v12 — companion-mechs → partners.** The pre-#464 Eldridge Coast seed shipped four partners as *mech* rows with an **equipment** slug in `chassisRef`:

| id | name | `chassisRef` |
|---|---|---|
| `eldridge-mech-custos` | Custos | `survey-drone` |
| `eldridge-mech-incitatus` | Incitatus | `mecha-companion` |
| `eldridge-mech-pr-1` | PR-1 | `survey-drone` |
| `eldridge-mech-rek-jet` | **Rek Jet** | `auto-turret` |

None resolves against `chassis.json`, so each rendered as a mech with an unknown chassis. #464 fixed the *seed*, but `ensureEldridgeCoastSeeded` is guarded on workspace existence and never rewrites a workspace it already spawned — so every browser that opened Eldridge Coast before that fix **still holds these rows**, and v11 (`equipmentLoadouts`) predates them. Nothing has ever reached them.

Conversion is lossless where the shapes meet; `quirk` lands in `aiPersonality` (which is what it held), `patternName` drops. Per the "invalid chassis probably belongs elsewhere" rule, a mech whose chassis is invalid but **not** a known partner slug is deliberately **left alone** — we'd be guessing, and a migration is the one edit a user can't undo.

**Gladhand retired.** Removed from the seed; v12 deletes the row and repoints any pilot still crewing it to Haven. All six pilots and mechs already linked to Haven.

**`DroneSchema` gains `bonusPerTechLevel`**, mirroring `EquipmentSchema`, so *absence* of the field states "this stat block is flat" — a fact in the data rather than one the app infers from which file a record came from.

> One sub-item was **not** done: adding standalone records for Big Brother's four named pattern drones. On inspection they're already correctly modelled — `PatternDroneConfigSchema` gives each a name + per-instance loadout over a shared `ref`. Separate records would duplicate the stat block four times and break that indirection.

## Removed

`--color-partner`, `EntityRowType`/`BadgeTone`/`SheetVariant` `'partner'`, `SheetPartner`, `PartnerSheetPage`, `PartnerRows`, `findPartner`, `partnerRailItems`, `partnerRoleLabel`. `SheetKind` is exactly `EntityRef['type']` again.

`StatItem` is now exported from `component-lib` — it was already the element type of the public `statsOverride` prop, so a consumer couldn't name the type of a prop it had to pass.

**Net −789 lines.**

## Known tradeoff

A partner is **no longer linkable**. ADR-027's sheet could be sent as a URL; the only route to a partner now is its host's sheet. That's the intended reading of "no independent existence", but it is a real capability that was briefly there.

## Verification

`bun run check:all` → **exit 0** (lint, format, typecheck, 3,869 tests, validate:all, knip, audit, tokens, styling). +11 new migration tests.

Not yet done: a visual pass in a browser on the two new mount points.

ADR-028 supersedes ADR-027's surface.

https://claude.ai/code/session_011d3HW2YgSZGdYWPDkGrF6a


alxjrvs (commented):
Reviewed `git diff origin/main...HEAD` (33 files — ADR-028: partners render in place, partner sheet/route/hue deleted, new `PartnerCard`, new v12 migration, Gladhand retired). Note: the local `main` ref is stale, so `main...HEAD` wrongly pulls in the already-merged Heat commit (#588); I scoped to `origin/main`. `bun --filter itun test` (1281 pass), `bun run typecheck`, and `bun run knip` are all green.

Findings:

- `apps/itun/src/lib/db/migrations/12-companion-mechs-to-partners.ts:174` — **v12 deletes a companion mech it declined to convert.** The `alreadyPresent` guard skips the `pilotStore.put`, but `mechStore.delete(mechId)` runs unconditionally. A pilot with two same-slug partners (Mecha Packmaster's two Mecha Companions, or a v11-minted `survey-drone` partner alongside a surviving `survey-drone` mech row) loses that row's name, loadout, conditions, SP/heat and cargo permanently — contradicting the file's own stated rule that it never silently deletes.
- `apps/itun/src/components/sheet/PartnerCard.tsx:249` — **an unresolvable partner renders nothing.** `data={statBlock ?? undefined}` hits `ReferenceEntityCard`'s `if (!data) return null`, so a partner with a slug that no longer resolves vanishes from its host sheet with no name, vitals, or Remove control. The deleted `SheetPartner` fell back to `partnerDisplayName` (→ `hostRef`), and the sibling `PilotEquipmentItem` deliberately keeps unresolved slugs "still shown, still counted, still toggleable".
- `apps/itun/src/components/sheet/PartnerCard.tsx:373` — **mech-granted (drones-schema) partners render their loadout twice.** With no `droneLoadout` prop, the card resolves the reference record's own systems into a read-only "Systems" listing, then `afterExtraContent` adds a second editable "Systems" block. Passing `droneLoadout={{ systems: [], modules: [] }}` suppresses the built-in one.
- `apps/itun/src/lib/db/migrations/12-companion-mechs-to-partners.ts:193` — **Gladhand repoint can duplicate a `pilot-to-crawler` link.** A pilot linked to both crawlers ends with two links to Haven, so they appear twice in Haven's crew and in the roster count.
- `apps/itun/src/lib/db/__tests__/companion-mechs-to-partners.test.ts:13` — **`migrate()` itself is untested**; only the two pure helpers are. Every defect above lives in the transaction body, and the repo already has a `fake-indexeddb` harness for exactly this.

PR #591, opened 2026-07-28
feat(itun): show the Actions deck as one catalog-tile grid, mech + pilot

alxjrvs:
The Dashboard's Actions instrument filed every action under a heading of its own source name, listed only one side's actions, and rendered each as a header-only badge. Three changes, one surface.

## One flat grid

The per-source (and per-timing) headings are gone, and with them the **Group** toggle and `groupByTiming` — a heading per action is noise when the tile already states what it is. `groupBySource` survives to build the source filter chips, which is now the only place a source name appears.

## The mount decides the roster, asymmetrically

Boarded, the deck is the mech's chassis + systems + modules **and** the pilot's abilities + equipment — the pilot is in the cockpit. On foot it is the pilot's alone.

The two economies coexist in the one deck: `PlayAction.currency` (not the mount) now decides whether an activation spends EP and adds Heat or spends AP and adds none, whether the Hot(X) stepper appears, and whether Push is offered. That also closes a latent hole where an on-foot `EP or AP` action could spend the EP of a mech the pilot had left behind.

## Catalog tiles

Each action renders as the canonical `medium` + `catalog` `ReferenceEntityCard` — the same index tile the SRD catalog uses — so the deck reads as SRD content, with the description the badge never showed.

### One thing this surfaced

The catalog extent deliberately **keeps** an entity's roll table (on an SRD index page the table *is* the content). That put a Show toggle and a Roll button inside a tile that is itself a `role="button"` — invalid markup whose clicks would bubble into the deck's open handler. Caught on the rendered page, not in review. `hide.rollTable` is new for exactly this, with a regression test; the table stays one click away, in the resolve panel, which renders the same action as a full card.

Column minimum goes 260px → 320px so a tile's description keeps a readable measure at the width the header-only badge used to need.

## Verification

`bun run check:all` green (lint, format, typecheck, all workspace tests, validate, knip, styling-ownership). New/updated tests cover the flat grid, the boarded mech+pilot roster, the on-foot roster excluding mech actions, and the nested-control guard. Rendered and inspected in Ladle (`Compositions/Dashboard/Actions Deck`).

https://claude.ai/code/session_0124EiAQge3oRXwqWZ2A4YAs

### DIFF
--- a/apps/itun/src/components/dashboard/ActionsDeck.tsx
+++ b/apps/itun/src/components/dashboard/ActionsDeck.tsx
@@ -11,12 +11,16 @@
  *               Active Item band, never auto-written)
  *
  * On foot (`mount === 'pilot'`) the deck is the pilot's abilities + equipment on
- * the AP economy; boarded, it is the mech's chassis + systems + modules on EP.
+ * the AP economy. Boarded, it is BOTH: the mech's chassis + systems + modules on
+ * EP *and* the pilot's own actions on AP — the pilot is in the cockpit, so their
+ * abilities and equipment never leave the table. The two economies coexist in one
+ * flat deck; `PlayAction.currency` (not the mount) decides what an activation
+ * spends and whether it touches Heat.
  */
 
 import { useState } from 'react'
 import { ActionsDeck as ActionsDeckView } from 'component-lib'
-import type { ActionsDeckView as ActionsDeckViewModel, DeckGroup } from 'component-lib'
+import type { ActionsDeckView as ActionsDeckViewModel, DeckRow } from 'component-lib'
 
 import { CORE_ROLL_BANDS, describePushOutcome, performCoreRoll } from '../../lib/rules/coreMechanic'
 import type { CoreRollResult } from '../../lib/rules/coreMechanic'
@@ -36,7 +40,6 @@ import {
   buildPilotActions,
   economyForActivation,
   groupBySource,
-  groupByTiming,
   hasCurrencyChoice,
   hasVariableHot,
   isDestructiveOutcome,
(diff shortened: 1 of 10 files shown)


alxjrvs (commented):
Reviewed commit `174c4285` ("show the Actions deck as one catalog-tile grid, mech + pilot") against `origin/main` — 10 files, ITUN `ActionsDeck` binding + `dashboardRules`, the component-lib `ActionsDeck` view/CSS/story, and a new `hide.rollTable` on `ReferenceEntityCard`.

Verification run: `bun run typecheck` (clean), `bun --filter itun test` (1298 pass), `bun --filter component-lib test` (599 pass), `bun run knip`, `lint`, `format:check` — all green. I also wrote throwaway probes (since deleted; tree is clean) to check the PR's central safety claim and the data assumptions behind it:

- **The "catalog tile nests no control" claim holds empirically.** I rendered the deck across every System and Module in the dataset and found **0** nested `button` / `[role="button"]` / `a[href]` inside any tile's clickable wrapper. `hide={{ rollTable: true }}` is correctly merged (`hide` spread happens before the catalog overrides, so it survives), and `catalogBodyEmpty` already tests `!rollTableNode`, so suppressing the table collapses the body strip rather than leaving a bare paper band.
- **Currency labelling is not affected.** Pilot ability actions carry `activationCost` with no `activationCurrency`, but the tile's cost pennant resolves via `resolveActivationCurrency(actionSource)`, so an AP action reads "2 AP" next to a mech "3 EP" in the mixed grid.

Findings (2):

- `apps/itun/src/components/dashboard/ActionsDeck.tsx:78` — **`onFoot` drops the `pilot != null` guard.** `playStateStore` is non-persisted with default `mount: 'pilot'`, so every Dashboard load starts on foot. For a mech with no soft-linked pilot the old `isPilotDeck` was false and the deck fell back to `buildMechActions(mech)`; now `deck` is `[]` and the instrument shows only "This pilot has no activatable actions." (reproduced: 0 tiles). Meanwhile `Dashboard.tsx:82` and `ActiveItemBand.tsx:76` both still require `pilot !== null`, so the rail reads "Mech · …" and the mech band renders with a Dismount control — the Actions deck is the only surface that thinks the player is on foot, and it stays empty every visit until the user dials to the mech focus and clicks "Load Into Mech ▶". This is the one I'd fix before merge (`const onFoot = mount === 'pilot' && pilot != null`).
- `apps/itun/src/components/dashboard/ActionsDeck.tsx:305` — **boarded pilot AP rows are heat-gated against the mech's cap.** The resolve panel correctly scopes Heat per action (`heatApplies`), but the list view passes the mech's `currentHeat`/`heatCap` to `actionReachable`/`reachSummary` for *every* row, contradicting that function's own docstring. Latent only: no pilot ability or equipment action currently carries a `hot` trait, so nothing misbehaves on today's data — a homebrew/expansion pilot action with Hot would be dimmed "Out of range / overheat" near the mech's cap.

Non-issues I checked and cleared: no `PlayAction.key` collisions between the mech and pilot builders; `push` correctly became per-action (`selected.currency === 'AP'`) and is equivalent to the old per-deck guard on real data; gating `currencyChoice` on `!onFoot` actually closes a latent hole (an on-foot EP pick would have written to the mech); no live `EP or AP` pilot actions exist; removed `.pc-deck-tool` / `.pc-deck-group-lab` CSS has no remaining references; `groupByTiming` removal is clean per knip.

PR #592, opened 2026-07-28
feat(component-lib): make the roll-table title its own table picker

alxjrvs:
## What

The Dashboard **Tables** focus carried a mini-bar above the table — a `Roll table` label plus a button repeating the very name the `RollTable` header band printed one row below it. Two controls, one fact.

`RollTable` now takes a typed `titleSelect` — `{ onOpen, open?, ariaLabel? }` — that renders the header **title** as the picker trigger. The Dashboard passes it and drops the bar (and `.pc-tables-bar` / `.pc-tables-lab` with it); the 5-column `TablePickerOverlay` is unchanged.

## Design notes

- **Typed, not a `node` slot** — the header keeps ownership of the affordance (caret, hit area, focus ring, `aria-haspopup="dialog"` / `aria-expanded`), so every consumer gets the same one. Same reasoning as the entity-card `controls` unification.
- **Independent of `disabled`**, which suppresses the ROLL action only: choosing which table to read is not rolling on it.
- **The `uppercase` on the trigger is load-bearing.** The UA reset gives `button` its own `text-transform: none`, so without it the title reads Title Case beside an all-caps band. Caught by measuring, not by reading.
- Focus uses `FOCUS_RING_ON_TONE` — the rust ring vanishes against the ink band.

## Verification

Measured in headless Chrome against the plain `showCommand` band: identical 12px Barlow Semi Condensed, 0.72px tracking, weight 700, paper ink — and an **unchanged 36.66px band height**, so nothing reflows. Caret flip + `aria-expanded` confirmed by clicking the real control.

| plain band | title as picker | open |
|---|---|---|
| `CORE MECHANIC` | `[ CORE MECHANIC ⌄ ]` | `[ CORE MECHANIC ⌃ ]` |

Full suite green: typecheck, lint, format, knip, `validate:all`, and all workspace tests (component-lib 599, itun 1298). ITUN's `DisplayPanel` tests now assert the trigger sits **in** the header band beside the Roll control, that picking retitles the table (`<caption>`), and that `aria-expanded` tracks the overlay.

A Ladle row (`titleSelect (Dashboard Tables)`) demonstrates it on the real Core Mechanic table.

https://claude.ai/code/session_014xZthKycfTXBWcvXs2ykij

### DIFF
--- a/apps/itun/src/components/dashboard/__tests__/DisplayPanel.test.tsx
+++ b/apps/itun/src/components/dashboard/__tests__/DisplayPanel.test.tsx
@@ -104,11 +104,19 @@ describe('DisplayPanel', () => {
     sublabel: 'roll',
   }
 
-  test('Tables focus → a RollTable renders under the picker bar (D3)', () => {
+  test('Tables focus → a RollTable whose title is the picker trigger (D3)', () => {
     const { container } = renderDV(tablesFocus)
     expect(container.querySelector('.pc-display-scroll')).toBeTruthy()
-    // The picker bar replaces the old <select>.
-    expect(container.querySelector('[aria-haspopup="dialog"]')).toBeTruthy()
+    // The trigger lives IN the RollTable header band — the band that also
+    // carries the Roll control — rather than in a bar above it repeating the
+    // name that band already prints.
+    const trigger = container.querySelector('[aria-haspopup="dialog"]')
+    expect(trigger).toBeTruthy()
+    expect(
+      trigger?.closest('div')?.querySelector('button[aria-label="Roll on this table"]')
+    ).toBeTruthy()
+    expect(trigger?.textContent).toContain('Core Mechanic')
+    expect(trigger?.getAttribute('aria-expanded')).toBe('false')
     expect(container.querySelector('select')).toBeNull()
     // The reused RollTable renders a real table (not the fallback note).
     expect(container.querySelector('table')).toBeTruthy()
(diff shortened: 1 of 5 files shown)


alxjrvs (commented):
## Review summary

Scope: commit `21fc8a4e` — "feat(component-lib): make the roll-table title its own table picker" (`git diff HEAD~1`; `@{upstream}...HEAD` and `git diff HEAD` are both empty). 5 files: `RollTable.tsx`, `RollTable.stories.tsx`, `DisplayPanel.tsx`, `instruments.css`, and the ITUN `DisplayPanel.test.tsx`.

The change is well-scoped and mechanically sound. I verified independently:

- `bun --filter itun test` — 1298 pass / 0 fail; `bun --filter component-lib test` — 599 pass / 0 fail
- `bun run typecheck` — clean across all workspaces
- `bun run lint` — clean (13 pre-existing warnings, none from the changed files)
- `bun run knip` — clean; the removed `.pc-tables-bar` / `.pc-tables-lab` CSS classes have no remaining references anywhere in the repo, and `Button` is still legitimately imported by `DisplayPanel.tsx` (roll-history "Clear")
- `TablePickerOverlay` still calls `onPick` + `onClose` on selection, so `aria-expanded` returns to `false` correctly
- `titleSelect` is threaded to both `StandardRollTable` and `ColumnsRollTable`, so switching to a columns-type table keeps the trigger
- No `cn()` / tailwind-merge class-stripping hazard in the new `TitleSelect` classes (`text-paper` is a color, `rounded-badge` is registered, `border-paper/40` doesn't collide with the custom `border-w` group)

Things I checked and deliberately did *not* report, because they predate this diff: the picker overlay traps no focus and has no Escape handler; the trigger stays tabbable behind an `aria-modal="true"` overlay; and `RollTable`'s internal roll/highlight state isn't reset when the selected table changes.

## Findings

- `packages/component-lib/src/components/shared/RollTable.tsx:80` — `titleSelect` silently renders nothing unless `showCommand` or `collapsible` is also passed; the requirement is a comment, not a type constraint, so a caller gets an unchangeable table with no error.
- `packages/component-lib/src/components/shared/RollTable.tsx:124` — `min-w-0` was added to the header span shared by *all* RollTable consumers, but only the `titleSelect` branch pairs it with `truncate`; a long `tableName` in a narrow container can now overflow onto the Roll control.

PR #593, opened 2026-07-28
feat(itun): render the SRD home page in the Dashboard SRD Explorer

alxjrvs:
The Dashboard's SRD Explorer home was eight hand-listed category tiles — Chassis, Systems, Modules, Pilot Abilities, Equipment, NPCs, Crawler Bays, Roll Tables.

The srd landing page shows the real SRD catalog: **six category sections over 23 schemas plus the 15 guides**, as `CatalogTile`s. Classes, Crawlers, Bio-Titans, Traits, Keywords and every guide were unreachable from the Dashboard, and a schema added to the catalog would never have appeared there.

The Explorer now renders that catalog — functionally the landing page minus the site header.

## One catalog, not two

`catalogHelpers.ts` / `catalogColors.ts` / `schemaName.ts` move from `apps/srd/src/lib` into `packages/component-lib/src/catalog`, behind a new `buildCatalogSections()` that binds the ORM and the tile colours in one place. srd's landing page, top nav and 404 now consume them from there.

The landing page's built output is unchanged — same six sections, same 38 catalog links.

## Two differences from the landing page, both forced by living in a panel

- **Tiles drill in place.** New `CatalogTile` `onActivate` prop renders a `<button>` with identical chrome; a link would take the player off the Dashboard. Schema tiles open that schema's listing, guide tiles open the entity outright.
- **Search stays above the catalog**, since on srd it lives in the header. Same `useSearchCombobox` as before.

The grid is `auto-fill` rather than the landing page's `sm:`/`lg:` breakpoints: those measure the *window*, so a narrow display panel on a wide screen would be handed three columns of unreadable tiles.

## Verification

- `bun run check:all` green (lint, format, typecheck, 3.9k tests, validate, knip, audit, tokens, styling).
- srd builds; landing-page output re-checked (38 links, all six sections).
- Measured in headless Chrome: 3 columns at a 528px panel, 1 at 244px, scroll containment intact (`.pc-srd` 480px viewport / 1120px content). Drill-in to listing and to the reference card confirmed visually.

`catalogColors.ts` picks up a design-token exemption on the move — its three gradients are hard-stop patterns built from tokens, which the gradient rule already allows, and were only invisible before because the guard doesn't scan `apps/srd`.

https://claude.ai/code/session_012eViYysCmnkL43yrKHABw1


alxjrvs (commented):
## Review: `feat(itun): render the SRD home page in the Dashboard SRD Explorer` (96e80d69, vs `origin/main`)

The change is a well-executed refactor: srd's `catalogColors`/`catalogHelpers`/`schemaName` move into `component-lib/src/catalog/`, a new `buildCatalogSections()` binds them to the ORM in one place, and the Dashboard's SRD Explorer swaps its hand-listed 8 tiles for that shared catalog. Tests migrated with the code and the new assertions are derived from the shared source rather than hard-coded, which is the right call.

**Verification run:** `typecheck` (all workspaces), `knip`, `check:tokens`, and the `srd` (984), `component-lib` (620) and `itun` (1298) suites all pass. I also built the srd site (1038 pages) and confirmed the landing page still emits all 6 sections and its tiles — i.e. the `import '../lib/gameData'` side-effect preload is not tree-shaken. Separately I probed every non-meta schema through `findAllIn` (24 schemas, 0 throws, 0 nameless rows) and confirmed `getEntitySchemas()` and the non-meta catalog agree exactly, so the newly-reachable tiles (`classes`, `crawlers`, `traits`, `keywords`, `sources`, …) can't crash the listing view.

### Findings

- `packages/component-lib/src/components/dashboard/SrdExplorer.tsx:229` — `ready` is seeded from `isLoaded('chassis')` alone, but the home view now calls `buildCatalogSections()` at mount, which reads `catalog-categories` and `guides`; under a partial preload that includes `chassis`, `LazyModel` throws during render. Shielded today only by itun's `GameDataReady` gate, but the component's own doc comment claims a safety property the seed doesn't provide.
- `packages/component-lib/src/components/shared/CatalogTile.tsx:96` — `href` went from required to optional with no compensating constraint, so `<CatalogTile name=… catalogBg=… />` compiles and renders an `<a>` with no `href`: not focusable, does nothing on click, still hover-animates. A discriminated union restores the guard.
- `packages/component-lib/src/index.ts:112` — `buildCatalogCategories`, `catalogNameOverrides`, `invariantNouns`, `pluralize`, `pluralizeWord` (+ two types) are new public exports with zero consumers outside component-lib's own tests; `includeEntryExports: false` means knip can't flag them. `resolveSchemaDomain`/`CardDomain` also lost their last external consumer in this commit.
- `apps/srd/src/lib/staticPaths.ts:10` — build-time-only data modules now import the whole `component-lib` barrel (incl. `instruments.css` via `DashboardCanvas.tsx`) for one 6-line guard that was explicitly designed to stay dependency-light. `apps/discord-bot/src/schemaName.ts` remains a third identical copy.
- `packages/component-lib/src/components/dashboard/SrdExplorer.tsx:148` — guide-tile activation re-lists and sorts the whole schema to find an id it already has, and `if (entity)` silently swallows a miss (dead tile, no feedback, no log).

Nothing here is a merge blocker; the first two are the ones worth fixing before landing.

PR #596, opened 2026-07-28
docs: rules-parity ADRs + delivery plan (ADR-028, ADR-022 amendment)

alxjrvs:
Records the decisions behind the **Rules Parity & Stat Provenance** goal. Docs only — no code or data changed.

## What's here

- **`ADR-028`** (new, Proposed) — one contribution model for caps *and* traits/damage/range, breakdown-shaped derivations, and provenance as a property of the value rather than of a surface.
- **`ADR-022` amendment** — an override becomes an **absolute pin** instead of a signed delta sharing a field with the rules derivation.
- **`docs/design/rules-parity-plan.md`** — the delivery breakdown: 24 units across 6 phases, with sizes, dependencies, critical path and definition of done.
- **`docs/README.md`** — indexes ADR-028, and adds ADR-027, which had been omitted.

## Why

Surveyed the core book (1.2) + three expansions against the working tree:

| Finding | |
| --- | ---: |
| Records whose text changes a maximum | 23 |
| — encoded | 9 |
| — **stated but never applied** | 5 |
| Records changing a trait / damage / range | ~21 |
| Derivations exposing a breakdown | 1 |
| Change Log entries ever tagged `transaction` | 0 |
| Rules modules implemented, tested, unreachable | 3 |

The root cause is one shared by both gaps: the dataset has **two** modifier engines — `statBonus` (four mech maxima, systems/modules only) and `ChoiceEffectSchema` (traits/damage/range, choice options only) — and **neither is reachable from an ability**. Beefcake, Bionic Arms, Bionic Legs and Modular Face Implant are inert prose, so the numbers on the sheet are wrong for any pilot holding them.

**Composite Armour** is a plain defect rather than a schema gap: its text fits `statBonus.structurePoints` exactly and the field is simply absent.

## The blocker the amendment removes

`max*Modifier` is one field doing two incompatible jobs — the ADR-022 override pin *and* the only channel a rules modifier has — with the derived baseline recovered by subtraction. Any automatic contribution written there would render as a hand override with a revert button. `eldridgeCoast.ts` and `pilotInventory.ts` both admit the conflation in their own header comments.

Migration 8 (`8-crawler-battle-sp-to-derived.ts`) already performed exactly this move for the Battle Crawler's +5 Max SP, so the pin migration is that operation generalized.

## Traps recorded so the backfill doesn't repeat them

- **Chassis-integrated bonuses are already absolute.** "Integrated Cargo Bay: +10, *to 16*" — the Mule's `cargoCapacity` **is** 16 (Gopher 12, Atlas 30). Backfilling double-counts.
- **Not every stated number is a cap modifier.** Mender's "+4 SP each time you heal" modifies another effect's output; Armour Plating states no cap change at all.

## Notes for review

- ADR-028 is **Proposed**, not Accepted — it's the thing to argue with.
- Rebased after #590 landed; partner references updated to the in-place `PartnerCard`.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -99,6 +99,8 @@ Read the matching ADR before proposing alternatives.
 | [ADR-024](adrs/ADR-024-derived-release-changelogs.md)                | Derived, per-app release changelogs (release-please) + on-site history; supersedes the web hand-changelog       |
 | [ADR-025](adrs/ADR-025-reference-versioned-releases-surface-gate.md) | Versioned internal releases + public-surface (TS + schema) gate for the ref; partially supersedes ADR-014       |
 | [ADR-026](adrs/ADR-026-entity-card-design-rules.md)                  | **Entity card design rules** — one renderer, choice/stat-atom/modified-stats/tech-level rules (built)           |
+| [ADR-027](adrs/ADR-027-partners-owned-by-host.md)                    | Partners are owned by their host entity                                                                         |
+| [ADR-028](adrs/ADR-028-contribution-model-and-stat-provenance.md)    | **Proposed** — one contribution model for caps/traits/damage + stat provenance; amends ADR-022's overrides      |
 
 > ADR-021 is the governing decision for rules enforcement and takes precedence
 > over prior ADRs where they conflict on _how hard a rule is enforced on which
--- a/docs/adrs/ADR-022-provenance-log-and-overrides.md
+++ b/docs/adrs/ADR-022-provenance-log-and-overrides.md
@@ -10,6 +10,12 @@ is still unbuilt. Subordinate to
 [ADR-021](ADR-021-itun-surface-taxonomy.md), which establishes the surface/mode
 model this ADR serves.
 
+**Amended 2026-07** — an override is now an **absolute pin**, not a signed delta
+in the same field the rules derivation reads. See
+[Amendment](#amendment-2026-07-overrides-become-absolute-pins); the amendment is a
+hard prerequisite for
+[ADR-028](ADR-028-contribution-model-and-stat-provenance.md).
+
 ## Context
 
 [ADR-021](ADR-021-itun-surface-taxonomy.md) makes the **Live Sheet** a Free-Edit
@@ -77,6 +83,10 @@ both the pinned value and the value it would derive to, so the UI can render an
 therefore **non-destructive and reversible**: reverting drops the pin and the stat
 resumes tracking its derivation.
 
+- **An override is an absolute pin, not a delta** (amended 2026-07 — see
+  [Amendment](#amendment-2026-07-overrides-become-absolute-pins)). The stored value is
+  the pinned maximum itself; the derived baseline is recomputed live and never
+  persisted.
 - The override marker is a small, visible graphical indicator on the affected stat
   — enough to tell an overridden value from a derived one at a glance. It appears
   **on the Live Sheet only**; a published snapshot renders the value plainly (a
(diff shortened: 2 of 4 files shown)


alxjrvs (commented):
**Scope note:** the PR under review is docs-only — 4 files, +465 lines (commit `429b4436`: new ADR-028 contribution model, an amendment to ADR-022, `docs/design/rules-parity-plan.md`, and two `docs/README.md` index rows). Mid-review another session reset this worktree, so `429b4436` is no longer on HEAD and `entityStore.ts` / `systems.json` now have uncommitted edits; I reviewed the captured diff, and every claim below was verified against the working tree code.

Findings (`file:line — summary`):

- `docs/adrs/ADR-028-contribution-model-and-stat-provenance.md:1` — ADR number collision: `ADR-028-partners-render-in-place.md` (Accepted, merged in #590) already holds 028, so every "ADR-028" citation — including ADR-027's own Status line — is now ambiguous. Should be 029.
- `docs/adrs/ADR-022-provenance-log-and-overrides.md:187` — "the remaining five fields" leaves `maxSpModifier` unmigrated. Migration 8 only touched Battle-typed *crawler* records (subtracting 5), never `mech.maxSpModifier` — the very field the amendment's own `MechSheet.tsx:229` snippet demonstrates.
- `docs/adrs/ADR-022-provenance-log-and-overrides.md:175` — "convert every existing value to an absolute pin" contradicts the Eldridge Coast carve-out: those pregens are persisted IndexedDB rows with deterministic ids, so a blanket migration pins authored content and the workspace-existence seed guard means they're never re-seeded.
- `docs/design/rules-parity-plan.md:98` — D1 (Composite Armour) is not "shippable immediately": `installedStatBonus` sums `statBonus` over ref strings with no condition check, so +5 SP persists when the system is Damaged/Destroyed. Needs `voidWhen`, which lands in C1.
- `docs/adrs/ADR-022-provenance-log-and-overrides.md:122` — `pilot.maxInventorySlotsModifier` is omitted from the "six delta fields", though the ADR itself cites it as carrying the same conflation; definition-of-done #4 goes unmet for that stat.
- `docs/design/rules-parity-plan.md:71` — A2's parity gate turns blocking one week after A1, but Beefcake/Bionic Arms/Bionic Legs/Modular Face Implant are unencodable until C2 and fit no §5 exemption class → CI red for all of Phase C.
- `docs/adrs/ADR-028-contribution-model-and-stat-provenance.md:109` — `DashboardGauge` does not "discard the override/breakdown props": its prop type is `label|value|max|tone|danger`, and no `breakdown` prop exists on `VitalGauge` at all. B4's S sizing rests on a false premise.
- `docs/README.md:102` — the ADR-027 index row omits its Superseded status (unlike ADR-014/025 rows), the pre-existing ADR-028 is unindexed, and the narrative above the table still stops at ADR-026.
- `docs/design/rules-parity-plan.md:123` — F1 and F5 appear in neither the critical-path diagram nor the "hang off nothing" list, so Phase F reads as fully parallelizable when F1 is blocked on C1.

Verified-correct and *not* flagged: the 23-record arithmetic (9+5+9), `statBonus` count of 8, Mule/Gopher/Atlas cargo values (16/12/30), the `#amendment-2026-07-overrides-become-absolute-pins` anchor slug, the ~41 person-day total and the "C and F are each ~a quarter" claim, all ADR/architecture cross-reference filenames, and the three Change Log gaps (`kind: 'transaction'` emitted nowhere, `source` never passed at a production call site, `entityStore.transfer()` never calling `emitChangeLog`).

PR #601, opened 2026-07-28
fix(observability): unblock Sentry in CSP, gate the wiring, repair nightly E2E

alxjrvs:
Tier 1 of the operating review — three findings that share one root cause:
instruments that exist in the codebase but report nothing in production.

## Sentry could never have worked, even with a DSN

Both apps ship a strict CSP whose `connect-src` did not list Sentry's ingest
origin, so every event would have been blocked in the browser before leaving
the page. That is the nastier of the two failure modes: with a DSN provisioned
it would have reported **zero errors and looked healthy**.
`apps/srd/src/lib/observability.ts` documented this requirement in its header
comment — the note was written and never actioned.

Both `netlify.toml` CSPs now allow the ingest origin.

## Nothing was watching whether the SDK shipped at all

With no DSN the env guard folds to `false` at build time and Vite tree-shakes
the whole SDK out, so the app looks identical to a working one — no warning,
no console message, no failed build. Measured on live production before this
PR: `grep -c sentry` over every script served by both sites returned **0**.

`tools/check-observability.ts` closes both gaps at the two places they break:

| mode | runs | asserts |
|---|---|---|
| static (default) | `validate:all`, every PR | module reads its DSN env var; init is actually **called** from the app entry; CSP lists the ingest origin |
| `--live` | nightly, post-deploy | fetches production, walks every script the HTML references, asserts the SDK is really in the served bytes |

The live half is the one CI structurally cannot do: whether the DSN is
provisioned in the Netlify dashboard is invisible from the repo.

## The nightly E2E gate had been failing 7 of its last 10 conclusive runs

So the signal had been retrained to mean nothing. Both failing specs asserted
against UI that no longer exists:

- crawler `Add weapons system` → the Armament Bay's `Mount` function verb (`BAY_FUNCTIONS`)
- mech `Add system` → `Manage systems` (`SectionManageButton` renders `Manage ${label}`)
- crawler empty state `No weapons installed.` → `No weapons mounted.`

Verified locally against an isolated dev server (a stale server from another
worktree was holding :5173, which would have tested the wrong code): **all 4
specs in both files pass.**

## Two deliberate wiring choices

- The nightly probe is gated on an `OBSERVABILITY_PROBE` repo variable so it
  stays **inert until the DSN is provisioned**. An alarm that fires nightly for
  a thing that does not exist yet is exactly the broken window that made the
  E2E signal worthless.
- `notify-recovery` used a bare `success()`. A job whose `needs:` includes a
  **skipped** job is itself skipped — so adding the gated probe would have
  silently stopped the recovery half from ever closing the tracking issue. Its
  condition is now spelled out, treating a skipped probe as "nothing to say".

## Follow-up required outside this PR

1. Provision the DSN: `PUBLIC_SENTRY_DSN` on `suindex`, `VITE_SENTRY_DSN` on
   `in-the-union-now`, `SENTRY_DSN` on the Render worker.
2. Set repo variable `OBSERVABILITY_PROBE=on` to arm the nightly probe.
3. If the Sentry org is in the **EU** data region, change `us` → `de` in
   `SENTRY_INGEST_HOST` (tools/check-observability.ts) and both CSPs — the
   static check keeps all three in lockstep.

Closes #584.


alxjrvs (commented):
**Note on scope:** `git diff @{upstream}...HEAD` was empty and local `main` was stale; the actual review scope is the single commit `c29871bd` ("fix(observability): unblock Sentry in CSP, gate the wiring, repair nightly E2E") against `origin/main`. Mid-review another session checked out `agent-tooling` in this worktree, so I read the reviewed content via `git show c29871bd:<path>`.

**What I verified rather than assumed:**
- Ran the new static check against a temp fixture built from the commit — passes.
- Ran `--live` against real production — correctly reports both sites dark (DSN not yet provisioned), so the probe's intent holds today.
- Built `apps/itun` with and without a DSN, and `apps/srd` with one, to test the live probe's marker against real bundles.
- Confirmed the e2e selector repairs are correct: `Mount` is unique in `BAY_FUNCTIONS` and wired to `setSystemsModalOpen`; `SectionManageButton` emits `aria-label="Manage systems"`, unique on `MechSheet`; `No weapons mounted.` exists in `CrawlerSheet.tsx:215` and renders in the bay card's non-collapsible `expand` slot, so the assertion is not vacuous.
- CSP: each `netlify.toml` has exactly one `Content-Security-Policy`, and `apps/srd/public/_headers` declares none, so no header-merge conflict.

**Findings (4):**

- `.github/workflows/e2e-nightly.yml:288` — `observability-probe` is tested with `== 'failure'` / `!= 'failure'`, so a probe that exhausts `timeout-minutes: 5` reports `cancelled`, opens no issue, and lets `notify-recovery` close the tracking issue with "passed again" while production is dark. This is the exact `cancelled`-vs-`failure` hole the file's own comment at line 235 says was already fixed for the e2e jobs.
- `.github/workflows/e2e-nightly.yml:306` — the github-script `results` array and `env:` block were not extended, so an issue opened solely because the probe failed lists only `e2e-itun: success` and `e2e-srd: success` — a failure notification naming no failing job.
- `tools/check-observability.ts:184` — live fetches carry no `AbortSignal.timeout` and script fetches never check `res.ok`; a hanging origin stalls the job into the `cancelled` misclassification above, and a 404 asset on itun is answered by the SPA fallback with HTTP 200 so its HTML is scanned as though it were the bundle.
- `tools/check-observability.ts:59` — `/sentry/i` only ever sees `<script src>` targets, but the SDK is behind a dynamic import in a chunk the probe never fetches; the build test shows the green comes from the inlined DSN literal in `assets/index-*.js`, not SDK bytes, so the check does not assert what its docblock claims.

PR #602, opened 2026-07-28
fix(itun): Change Log truth + Composite Armour's missing +5 Max SP

alxjrvs:
First execution slice of the **Rules Parity & Stat Provenance** goal (#597). Decisions in #596.

Draft while the rest of batch 1 (A3 override split, B1–B6 provenance panel) lands on this branch — keeping the batch to one PR.

## Three defects, all independent of the override split

**1. Composite Armour contributed nothing.** Its text — *"increases your Mech's Max SP by 5 for each Composite Armour System you have installed"* — is exactly the shape `statBonus.structurePoints` exists for. The field was simply absent. Now added, with a test covering per-copy stacking.

**2. `kind: 'transaction'` was emitted nowhere.** Defined in the schema, documented on `entityStore.update`, badged in `ChangeLogDrawer` — and passed by no call site. Every Dashboard lifecycle transaction (Push, Heat Check, take damage, activate) logged as an indistinguishable `manual` edit, so the drawer's transaction badge could never render.

**3. `source` was never passed at all** — every entry persisted as `'unknown'`.

Rather than re-typing a meta object at ~50 call sites, provenance is now a named per-surface constant (`stores/surfaceProvenance.ts`), picked by the surface's ADR-021 enforcement mode.

**4. `entityStore.transfer()` never logged.** It commits cross-entity writes in one IDB transaction, bypassing `update()` — so cargo stow/load and scrap hand-offs mutated entities with **no provenance at all**. The one real hole in ADR-022's chokepoint guarantee. It now captures before-images and emits per changed field exactly as `update()` does, and takes its own `ChangeMeta`.

## Verification

`bun run check:all` fully green: lint, format, typecheck, **4,800+ tests**, `validate:all`, knip, audit, tokens, styling.

Four new provenance tests pin the contract, including one asserting `kind: 'transaction'` is now reachable and one covering the `transfer()` hole directly.

## Note for review

The diff is wide but shallow — most of it is appending a provenance constant at call sites and to the test assertions that pin call shape. The behavioural changes are confined to `entityStore.ts`, `surfaceProvenance.ts`, and one data record.

Refs #597.

### DIFF
--- a/apps/itun/src/components/crawler/CrawlerBuilder.tsx
+++ b/apps/itun/src/components/crawler/CrawlerBuilder.tsx
@@ -47,6 +47,7 @@ import { CrawlerReviewStep } from './CrawlerReviewStep'
 import { CrawlerStatsStep } from 'component-lib'
 import { CrawlerTypeSelectStep } from 'component-lib'
 import { SystemsList } from 'component-lib'
+import { WIZARD_TXN } from '../../stores/surfaceProvenance'
 
 /**
  * Book-order steps (Union Crawler pp.212–213 + Review — plan §4.3). Edit mode
@@ -290,7 +291,7 @@ export function CrawlerBuilder({
         const stored = store.get('crawler', crawlerId)
         const oldType = stored?.type ?? null
 
-        await store.update('crawler', crawlerId, crawlerFormToUpdatePatch(form))
+        await store.update('crawler', crawlerId, crawlerFormToUpdatePatch(form), WIZARD_TXN)
 
         // Crew/NPC edits + the type-NPC reset / orphan cleanup route through the
         // shared multi-write helper (also used by the live sheet's inline build
--- a/apps/itun/src/components/dashboard/ActionsDeck.tsx
+++ b/apps/itun/src/components/dashboard/ActionsDeck.tsx
@@ -52,6 +52,7 @@ import {
 } from './dashboardRules'
 import type { MechItemEconomy } from '../sheet/mechItemRules'
 import type { PlayAction, PlayActionCurrency, TimingTab } from './dashboardRules'
+import { DASHBOARD_TXN } from '../../stores/surfaceProvenance'
 
 type ActionsDeckProps = {
   mech: Mech
@@ -136,7 +137,7 @@ export function ActionsDeck({ mech, pilot, mount = 'mech', store }: ActionsDeckP
         apCost: economy.epCost,
         currentAP: fresh.currentAP ?? 0,
       })
-      if (Object.keys(patch).length > 0) void s.update('pilot', pilot.id, patch)
+      if (Object.keys(patch).length > 0) void s.update('pilot', pilot.id, patch, DASHBOARD_TXN)
       setActivated(true)
       return
     }
@@ -152,7 +153,7 @@ export function ActionsDeck({ mech, pilot, mount = 'mech', store }: ActionsDeckP
       prevUses: fresh.itemUses,
     })
     if (Object.keys(patch).length > 0) {
-      void s.update('mech', mech.id, patch)
+      void s.update('mech', mech.id, patch, DASHBOARD_TXN)
     }
     setActivated(true)
   }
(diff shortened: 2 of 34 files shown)


alxjrvs (commented):
Reviewed commit `35c0b8c3` (the only commit ahead of `origin/main`): ~50 call sites gain a `ChangeMeta` provenance tag via a new `stores/surfaceProvenance.ts`, `entityStore.transfer()` gains before-image capture + change-log emission, and `Composite Armour` gains `statBonus.structurePoints: 5`.

Verified green: `bun --filter salvageunion-reference test` (922 pass), `bun --filter itun test` (1287 pass), `bun run validate:all`. The provenance tagging itself is complete and correctly routed — every write in `dashboard/` is `DASHBOARD_TXN`, every wizard/`applyCrawlerEdit` write is `WIZARD_TXN`, and no non-live-sheet surface renders the components tagged `LIVE_SHEET_MANUAL`.

Findings:

- `packages/salvageunion-reference/data/systems.json:731` — Composite Armour's +5 Max SP goes derived-at-read with no IndexedDB migration, so a saved mech whose `maxSpModifier` hand-carries that +5 (as `apps/itun/src/lib/schemas/mech.ts:170-176` explicitly instructs) now counts it twice; a Mule pinned to 17 Max SP renders 22. `apps/itun/src/lib/db/migrations/8-crawler-battle-sp-to-derived.ts` is the direct precedent for this exact transition on the crawler side, and the stale mech.ts comment still documents the old model.
- `packages/salvageunion-reference/lib/rules/derivedStats.ts:146` — only half the rule is implemented. `installedStatBonus` ignores `systemConditions`, so a Damaged/Destroyed Composite Armour still grants +5 Max SP instead of removing it, and the rule's "reduce your current SP to match" clamp never runs (`clampMechCurrentStats` has zero app-code callers). Composite Armour is the only `statBonus` item in the dataset with a damaged/destroyed rider, so this isn't a pre-existing generic gap.
- `apps/itun/src/lib/cargo/useCargo.ts:128` — `transfer()`'s new `meta` parameter is passed by none of its four call sites (`useCargo.ts:128,159,191`, `usePartnerCargo.ts:123`), all of which are Live-Sheet surfaces, so cargo stow/load entries log as `source: 'unknown'` next to `live-sheet`-tagged hand edits on the same sheet.

alxjrvs (commented):
## Review: `b85fee07 fix(itun): make the Change Log tell the truth, and apply Composite Armour's +5 Max SP`

Scope note: the local `main` ref is stale, so the real diff is `origin/main...HEAD` — one commit, 34 files. Working tree is clean. `bun --filter itun test` (1287 pass) and `bun --filter salvageunion-reference test` (922 pass) are both green.

The provenance plumbing itself (`surfaceProvenance.ts`, the `transfer()` before-image capture + phase-4 emit, `updateCrawlerBay`'s meta pass-through) reads correctly: before-images are captured pre-write exactly as `update()` does, `emitChangeLog` swallows its own errors so a lost log line can't fail a committed write, and deletes are deliberately unlogged. No issue found there.

### Findings

- `packages/salvageunion-reference/data/systems.json:731` — Composite Armour's +5 Max SP moves from hand-set `maxSpModifier` to derived `statBonus` with **no migration**, so existing local mechs double-count it (chassis 10 + mod 5 + bonus 5 = 20, was 15). `mech.ts:172` and `derivedStats.test.ts:128` both document `maxSpModifier` as the old mechanism, and `migrations/8-crawler-battle-sp-to-derived.ts` is the repo's own precedent for healing exactly this.
- `packages/salvageunion-reference/lib/rules/derivedStats.ts:146` — a Damaged/Destroyed Composite Armour still grants +5. The action text (`actions.json:2047`) continues past the sentence quoted in the new test: "If this System is Damaged or Destroyed reduce your Max SP by 5 and reduce your current SP to match." `installedStatBonus` never sees `systemConditions`. Composite Armour is the only one of nine `statBonus` systems with such a clause.
- `apps/itun/src/lib/cargo/useCargo.ts:128` — `transfer()`'s new `meta` parameter is passed at **zero** production call sites (`useCargo.ts:128,159,191`, `usePartnerCargo.ts:123`). Cargo stow/load — the flow the code comment names as the motivation — still logs `manual`/`unknown`.
- `apps/itun/src/components/sheet/MechSheet.tsx:163` — `useSoftWarnings` is constructed without `meta`, so removing a mech *system* logs `source: 'unknown'` while removing a *module* on the same sheet logs `live-sheet`. Same for the pilot class change and ability add/remove.

The last two are "the fix is only half-wired" rather than crashes, but they undercut the commit's stated goal — a Dashboard-adjacent cargo move and a Live-Sheet system removal still land in the drawer indistinguishable from a hand edit.

PR #604, opened 2026-07-28
feat(itun): provision Convex backend and Discord auth scaffold

alxjrvs:
External-services groundwork for the accounts/Games plan. **Backend definition only** — no app wiring, no mutations, no behaviour change. ITUN stays local-first until the client migration lands in a follow-up.

## Provisioned outside the repo

| Resource | Value |
| --- | --- |
| Convex project | `alex-jarvis:suref-itun` |
| Dev deployment | `dev/alex-jarvis` → `perfect-donkey-72.convex.cloud` |
| Schema state | pushed — 12 tables, 20 indexes |
| Dashboard | https://dashboard.convex.dev/t/alex-jarvis/suref-itun |

`.env.local` holds the deployment vars and is gitignored — verified, nothing secret is committed.

## In this PR

- `convex/schema.ts` — `users`, `games`, `memberships`, `invites`, `pilots`, `mechs`, `crawlers`, `softLinks`, `encounterNpcs`, `mechPatterns`, `changeLog`, `presence`
- `convex/auth.ts` — Discord OAuth via `@convex-dev/auth`
- `convex/http.ts` — OAuth callback routes
- `convex/auth.config.ts`, `apps/itun/.env.example`
- `biome.jsonc` ignores `convex/_generated`; `knip.json` adds `convex/*.ts` to the itun project (knip's built-in Convex plugin handles the entry points)

## Two schema decisions for review

**Entity bodies are `v.any()`, not mirrored Convex validators.** The Zod schemas in `src/lib/schemas/` are the source of truth — `schemas/*.schema.json` is generated from them and CI fails on drift. Re-declaring every field in Convex would fork that into a second hand-maintained copy that silently rots. Convex indexes only what it needs to *query* (ownership, scoping, timestamps) and treats the body as an opaque payload, exactly as `SnapshotStorage` already treats snapshots. **The cost is real: Convex cannot reject a malformed body, so every mutation must Zod-parse before persisting.**

**`gameId` and `ownerId` are both nullable**, encoding four states: null `gameId` = the owner's personal shelf; null `ownerId` = an unclaimed entity (a normal state — a player left, the Mediator pre-built it, or it came from a Game template). Both null is the one invalid combination and should stay unreachable through any mutation.

## Still needed before sign-in works

Discord OAuth credentials must be set on the deployment — this needs the Discord developer portal, so it can't be done from here:

```
bunx convex env set AUTH_DISCORD_ID <client-id>
bunx convex env set AUTH_DISCORD_SECRET <client-secret>
```

**Decided: reuse the existing Discord application** (the one the bot already uses) — one app covers bot + login, and players get a consent screen they recognise.

Verified against the installed packages rather than assumed:

- `@convex-dev/auth` mounts callbacks at path prefix `/api/auth/callback/` with the provider id appended
- `@auth/core`'s Discord provider declares `id: "discord"`

So the redirect URI to add is exactly:

```
https://perfect-donkey-72.convex.site/api/auth/callback/discord
```

A second redirect URI for the production Convex deployment will be needed at deploy time — Discord permits several, so it is additive, not a swap.

**Note for review — PII scope.** `@auth/core`'s Discord provider requests `scope: "identify email"` by default. ITUN never sends email, and with a single provider there is no email-based account linking to support. Narrowing to `identify` would shrink the PII surface the plan's deletion/export obligations apply to, and give players a lighter consent screen. Left at the default in this PR because it needs a sign-in test to confirm `@convex-dev/auth` does not depend on the email claim.

## Verification

`lint`, `typecheck`, `knip`, `format`, and the full `test` suite all pass — 0 failures across every workspace.

### DIFF
--- a/apps/itun/.env.example
+++ b/apps/itun/.env.example
@@ -1,0 +1,13 @@
+# Convex — the server of record for accounts, Games, and entity ownership.
+# These are written automatically by `bunx convex dev` into .env.local
+# (gitignored). Listed here so a new contributor knows what the app expects.
+CONVEX_DEPLOYMENT=
+VITE_CONVEX_URL=
+VITE_CONVEX_SITE_URL=
+
+# Discord OAuth (the only sign-in provider).
+# These are NOT read from this file — they live on the Convex deployment:
+#   bunx convex env set AUTH_DISCORD_ID <client-id>
+#   bunx convex env set AUTH_DISCORD_SECRET <client-secret>
+# The Discord application's redirect URI must be:
+#   <VITE_CONVEX_SITE_URL>/api/auth/callback/discord
--- a/apps/itun/.gitignore
+++ b/apps/itun/.gitignore
@@ -1,2 +1,4 @@
 playwright-report/
 test-results/
+
+.env.local
--- a/apps/itun/convex/_generated/api.d.ts
+++ b/apps/itun/convex/_generated/api.d.ts
@@ -1,0 +1,51 @@
+/* eslint-disable */
+/**
+ * Generated `api` utility.
+ *
+ * THIS CODE IS AUTOMATICALLY GENERATED.
+ *
+ * To regenerate, run `npx convex dev`.
+ * @module
+ */
+
+import type * as auth from "../auth.js";
+import type * as http from "../http.js";
+
+import type {
+  ApiFromModules,
+  FilterApi,
+  FunctionReference,
+} from "convex/server";
+
+declare const fullApi: ApiFromModules<{
+  auth: typeof auth;
+  http: typeof http;
+}>;
+
+/**
+ * A utility for referencing Convex functions in your app's public API.
+ *
+ * Usage:
+ * ```js
+ * const myFunctionReference = api.myModule.myFunction;
+ * ```
+ */
+export declare const api: FilterApi<
+  typeof fullApi,
+  FunctionReference<any, "public">
+>;
+
+/**
+ * A utility for referencing Convex functions in your app's internal API.
+ *
+ * Usage:
+ * ```js
+ * const myFunctionReference = internal.myModule.myFunction;
+ * ```
+ */
+export declare const internal: FilterApi<
+  typeof fullApi,
+  FunctionReference<any, "internal">
+>;
+
+export declare const components: {};
--- a/apps/itun/convex/_generated/api.js
+++ b/apps/itun/convex/_generated/api.js
@@ -1,0 +1,23 @@
+/* eslint-disable */
+/**
+ * Generated `api` utility.
+ *
+ * THIS CODE IS AUTOMATICALLY GENERATED.
+ *
+ * To regenerate, run `npx convex dev`.
+ * @module
+ */
+
+import { anyApi, componentsGeneric } from "convex/server";
+
+/**
+ * A utility for referencing Convex functions in your app's API.
+ *
+ * Usage:
+ * ```js
+ * const myFunctionReference = api.myModule.myFunction;
+ * ```
+ */
+export const api = anyApi;
+export const internal = anyApi;
+export const components = componentsGeneric();
--- a/apps/itun/convex/_generated/dataModel.d.ts
+++ b/apps/itun/convex/_generated/dataModel.d.ts
@@ -1,0 +1,60 @@
+/* eslint-disable */
+/**
+ * Generated data model types.
+ *
+ * THIS CODE IS AUTOMATICALLY GENERATED.
+ *
+ * To regenerate, run `npx convex dev`.
+ * @module
+ */
+
+import type {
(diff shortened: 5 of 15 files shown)


alxjrvs (commented):
## Review: `feat(itun): provision Convex backend and Discord auth scaffold` (1 commit, 15 files vs `origin/main`)

I diffed against `origin/main` (local `main` was stale). Scope is the Convex scaffold only — no `src/` changes.

**What I verified as sound** (so it isn't re-litigated):
- The schema loads and exports cleanly (`defineSchema(...).export()` produces all 18 tables); the nested index paths `from.id` / `to.id` are valid.
- `convexAuth({ providers: [Discord] })` returns exactly the five destructured names, and passing the provider *factory* is handled (`provider_utils.js:56`).
- `http.ts` registers the 5 expected routes, including `/api/auth/callback/*` — so the redirect URI documented in `.env.example` is correct.
- `.env.example`'s claim that `convex dev` writes `VITE_CONVEX_SITE_URL` is true for convex 1.42.
- Overriding `users` after `...authTables` drops `phone`/`phoneVerificationTime` and the `phone` index, but the only code path using them (`uniqueUserWithVerifiedPhone`) is guarded by `typeof profile.phone === "string"`, which Discord never satisfies. Not a bug.
- `bun.lock` downgrades hoisted `cookie` 2.0.1 → 1.1.1, but astro keeps a nested 2.0.1 and the conflicting `jose` 5/6 both resolve. No breakage.
- `lint`, `format:check`, `knip`, `typecheck` all pass; the biome `!**/convex/_generated` ignore and knip's convex plugin both work as intended.

**Findings (3):**

- `apps/itun/tsconfig.json:19` — the new `convex/` dir is outside `include` and no `convex/tsconfig.json` was added, so `tsc --listFilesOnly` loads **zero** convex files: the entire Convex surface is unchecked by `typecheck`, `build`, and CI, and a type error only surfaces at `convex deploy`.
- `apps/itun/.env.example:9` — documents only `AUTH_DISCORD_ID`/`AUTH_DISCORD_SECRET` as deployment env, but `@convex-dev/auth` also `requireEnv`s `SITE_URL` (redirects.js:38), `JWT_PRIVATE_KEY` (tokens.js:7) and `JWKS` (index.js:135); following the file exactly yields a deployment where every sign-in throws. The `npx @convex-dev/auth` init step that generates them is never mentioned.
- `apps/itun/convex/schema.ts:9` — cites "docs/architecture/ for the full decision record" and ~30 D-numbers (D1/D18/D28/D34…) that exist nowhere in `docs/`, while ADR-001 stays `Accepted` and `apps/itun/CLAUDE.md` still says "No auth, no backend" — a stale claim in a file this repo loads into every session.

alxjrvs (commented):
Reviewed `HEAD~1` (no upstream diff, clean working tree): the Convex backend + Discord auth scaffold for `apps/itun`. I installed deps in the worktree and ran `typecheck`, `lint`, `format:check` and `knip` — all green — then read `@convex-dev/auth@0.0.94` and `@auth/core@0.41.3` sources to check the schema/auth wiring against what the library actually does at runtime.

Findings (6):

- `apps/itun/.env.example:9` — Setup doc omits the `SITE_URL`, `JWT_PRIVATE_KEY` and `JWKS` deployment env vars that `@convex-dev/auth` `requireEnv`s; following this file exactly yields a deployment where Discord sign-in throws instead of completing.
- `apps/itun/tsconfig.json:19` — `convex/` is outside `include` and there's no `convex/tsconfig.json`, so the whole Convex surface is skipped by `bun run typecheck`, `bun run build` and the CI typecheck job (verified: typecheck exits 0 with the files present); nothing gates `_generated` drift either.
- `apps/itun/convex/schema.ts:47` — `users.discordId` / the `by_discord` index are never populated: convex-auth strips the provider `id` (`implementation/index.ts:332`) and it lands only in `authAccounts.providerAccountId`, so future `by_discord` lookups return nothing — the exact Discord-bot identity path `auth.ts` cites as the rationale.
- `apps/itun/convex/schema.ts:40` — The `users` override drops `phone`, `phoneVerificationTime` and the `phone` index from `authTables`; unreachable with Discord-only, but adding any phone/email provider breaks `uniqueUserWithVerifiedPhone`'s `withIndex('phone', …)`.
- `apps/itun/convex/schema.ts:189` — `changeLog` has no `(entityId, field)` index, so the documented D25 supersede lookup scans the entity's whole permanent history; past Convex's 16384-doc transaction cap the mutation fails outright.
- `knip.json:41` — `convex/*.ts` is non-recursive, so any future `convex/<dir>/*.ts` function silently escapes the dead-code gate.

No blocking correctness bug in the schema or auth wiring itself — the two I'd fix before merge are the missing env vars (auth cannot work without them) and the typecheck gap (it is what will let the next Convex change break silently).

PR #605, opened 2026-07-28
feat(tooling): measure the agent fleet, scope the typecheck hook, fix the worktree reaper

alxjrvs:
Tiers 2 and 3 of the operating review. Instrument the largest unmeasured input,
stop paying a per-edit tax that enforced nothing, and make two
existing-but-broken things actually work.

## `tools/agent-digest.ts` — read the data already on disk

~10 agent sessions a day run in this repo and not one number about them was
collected. The instinct is to stand up an OTLP collector; **don't, yet.** Claude
Code already writes a complete structured record of every session to
`~/.claude/projects/`, and hundreds of those files are sitting there being
treated as exhaust. Read them first — the answers will tell you whether
streaming telemetry is worth configuring at all, and it costs one script instead
of an integration.

```
bun run agent:digest            # this repo, last 14 days
bun run agent:digest -- --all --days 30 --json
```

Reports sessions, median duration, share that landed a commit, tokens, tool-error
rate, permission denials, most-used and most-erroring tools. Reads only; nothing
leaves the machine.

It counts **main sessions and subagent runs separately**, which turned out to
matter. A project folder holds `<session>.jsonl` alongside
`<session>/subagents/agent-*.jsonl`, so counting files reports **131 "sessions"
for 22 real ones**. The review that prompted this work made exactly that mistake
and overstated the figure ~6x; the artifact has been corrected.

## A scoped typecheck hook

The `PostToolUse` hook ran the whole monorepo `bun run typecheck` after **every**
`Edit` and `Write` — including markdown and JSON, which cannot affect a type.

| | before | after |
|---|---|---|
| edit a `salvageunion-reference` file | 7030 ms | **114 ms** |
| edit a markdown file | 7030 ms | **0** (skipped) |

Measured breakdown: srd 6.27 s, itun 2.19 s, discord-bot 0.96 s, component-lib
0.15 s, reference 0.10 s — 7.03 s wall, and srd is essentially all of it.

As a `PostToolUse` hook its output is **advisory** — it cannot block the edit —
and CI gates the same check on merge. So the 7 s bought a warning, not
enforcement. `.claude/hooks/typecheck-scoped.sh` skips non-TypeScript files and
otherwise checks only the workspace owning the file. It never exits non-zero: a
half-typed file between two `Edit` calls is a normal state, not an error.

## The worktree reaper never worked — at all

`prune-stale-worktrees.sh` used a bare `return` in its skip paths. A bare
`return` yields the status of the last command run — a **failed test** — so
`process()` returned 1, and `set -euo pipefail` killed the script at the first
worktree it declined to consider. The main working tree is always first and
always declined, so it **exited 1 with no output before examining a single
candidate, for its entire life.**

Verified against the unmodified script from `main`:

```
$ bash /tmp/orig-prune.sh
EXIT=1          # ← no output whatsoever
```

That, not the lock handling, is why abandoned worktrees accumulate (#377).

Now that it runs, two further changes:

- **Locks are interrogated, not obeyed blindly.** Claude Code stamps
  `claude session <name> (pid <PID> start <date>)`; the PID is parsed and the
  lock is stale only if that process is gone. Alive or unparseable → skip.
- **An unpushed-commits guard was added.** `git worktree remove` only refuses on
  a *dirty tree*, so fixing the script would otherwise have armed it to delete
  the only copy of a commit. That risk was theoretical while it removed nothing
  and became real in this change, so the guard ships in the same commit.

Current dry run: **0 candidates, 11 skipped** — every worktree here has unpushed
commits. That is the correct conservative answer, and it now *tells you why*
instead of silently doing nothing.

⚠️ This does **not** fully close #377: nothing is reclaimed until branches are
pushed or merged. A `--push` mode (publish, then reap) is the obvious follow-up
but is outward-facing, so it is deliberately not in this PR.

## Two skills

- **`/triage`** reads every production and CI signal — nightly E2E, the live
  observability probe, deploys, Dependabot, in-flight PRs — and proposes the
  day's work in ranked order, capped at five items. This is the missing loop:
  zero of the last 60 merged PRs referenced an issue, so nothing routes an
  observed signal into the queue.
- **`/component-refresh`** operationalizes the three-level redesign loop (real
  SSR "before" → `NEW*` Ladle three-way → staged cutover). It **points at**
  `docs/design/entity-card-reconciliation.md` Part 1 rather than restating it —
  a second copy would drift, which is why `validate:doc-drift` exists.

## Verification

`bun run check:all` green after rebase onto `main` (which now includes #601).

### DIFF
--- a/.claude/hooks/typecheck-scoped.sh
+++ b/.claude/hooks/typecheck-scoped.sh
@@ -1,0 +1,58 @@
+#!/bin/bash
+# Typechecks ONLY the workspace an edit actually touched.
+# Used as a PostToolUse hook for Edit and Write tools.
+#
+# This replaces a hook that ran the whole monorepo `bun run typecheck` after
+# EVERY Edit and Write — including markdown, JSON and YAML, which cannot affect
+# a type. Measured, that is 7.0 s of wall clock per edit, of which apps/srd
+# alone is 6.3 s:
+#
+#   salvageunion-reference 0.10s   component-lib 0.15s   srd 6.27s
+#   itun 2.19s             discord-bot 0.96s            total 7.03s (concurrent)
+#
+# Across a few dozen edits per session that is minutes of pure waiting, paid on
+# every session, to re-prove something `bun run typecheck` in CI already gates
+# on merge. And because PostToolUse output is advisory — it cannot block the
+# edit — the cost bought a warning, not enforcement.
+#
+# So: skip entirely for files that cannot change a type, and otherwise check
+# just the one workspace that owns the file. A cross-package edit still gets
+# caught, because the consuming workspace is typechecked the moment you edit a
+# file in it, and CI still runs the full matrix on the PR.
+#
+# Never exits non-zero: a typecheck failure here is INFORMATION for the agent,
+# not a gate. Exiting 2 would reject an edit that is legitimately mid-refactor
+# and half-typed, which is a normal state to be in between two Edit calls.
+
+INPUT=$(cat)
+
+FILE_PATH=$(echo "$INPUT" | jq -r '.tool_input.file_path // .tool_input.file // empty')
+
+if [ -z "$FILE_PATH" ]; then
+  exit 0
+fi
+
+# Only TypeScript-bearing sources can change a type. Astro files carry a
+# TypeScript frontmatter block, so they count; .md/.json/.yml/.css do not.
+case "$FILE_PATH" in
+  *.ts | *.tsx | *.mts | *.cts | *.astro) : ;;
+  *) exit 0 ;;
+esac
+
+# Map the edited path to the workspace that owns it. Ordered most-specific
+# first; salvageunion-reference must precede any broader packages/ rule.
+case "$FILE_PATH" in
+  *packages/salvageunion-reference/*) WORKSPACE="salvageunion-reference" ;;
+  *packages/component-lib/*)          WORKSPACE="component-lib" ;;
+  *apps/srd/*)                        WORKSPACE="srd" ;;
+  *apps/itun/*)                       WORKSPACE="itun" ;;
+  *apps/discord-bot/*)                WORKSPACE="discord-bot" ;;
+  # tools/, scripts and repo-root config belong to no workspace. The root
+  # tsconfig does not typecheck them as a project, and the full matrix would
+  # cost 7 s to prove nothing about the edited file — leave them to CI.
+  *) exit 0 ;;
+esac
+
+bun --filter "$WORKSPACE" typecheck 2>&1 | tail -20
+
+exit 0
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -30,8 +30,8 @@
         "hooks": [
           {
             "type": "command",
-            "command": "bun run typecheck 2>&1 | head -20",
-            "statusMessage": "Typechecking after edit..."
+            "command": "\"$CLAUDE_PROJECT_DIR\"/.claude/hooks/typecheck-scoped.sh",
+            "statusMessage": "Typechecking the edited workspace..."
           }
         ]
       }
--- a/.claude/skills/component-refresh/SKILL.md
+++ b/.claude/skills/component-refresh/SKILL.md
@@ -1,0 +1,82 @@
+---
+name: component-refresh
+description: Redesign an existing component through the three-level loop — real SSR "before", NEW* Ladle comparison, then staged cutover
+allowed-tools: Bash, Read, Edit, Write, Glob, Grep
+---
+
+# Component Refresh
+
+Drive a component redesign through the repeatable three-level loop. This is the
+dominant kind of work in this repo, and the loop that makes it safe already
+exists as prose — it just was not invokable, so every session reconstructed it
+from a paragraph and some skipped the parts that make it safe.
+
+**The methodology lives in
+[`docs/design/entity-card-reconciliation.md`](../../../docs/design/entity-card-reconciliation.md)
+Part 1 — read it first, and treat it as the source of truth.** Do not restate it
+here or anywhere else; a second copy drifts, and `validate:doc-drift` exists
+because that has already happened. This skill is the operating procedure for
+running it.
+
+## Before starting
+
+Ask which level is being requested, and do not silently cross a gate:
+
+- **L1 (mockup)** — design only. Produces an artifact. Changes no app code.
+- **L2 (build alongside)** — `NEW`-prefixed components, Ladle-only, no consumers.
+- **L3 (cutover)** — migrate consumers and delete the legacy component.
+  **Explicit command only.** Never begin L3 because L2 looks finished.
+
+If the request is ambiguous, assume the _lower_ level and say so.
+
+## L1 — mockup, grounded in a real "before"
+
+The single rule that carries this level: **the "before" is the actual current
+component rendered from real code, never a hand-authored caricature.** A
+caricature hides exactly the wrapping, overflow, tone and empty-state bugs the
+refresh is supposed to fix, so a mockup built against one is designing for a
+component that does not exist.
+
+- Render it via SSR — `react-dom/server` `renderToStaticMarkup` plus compiled
+  Tailwind (`@tailwindcss/node` + `@tailwindcss/oxide` Scanner over the rendered
+  markup, theme fonts embedded as data URIs). Capture HTML + CSS, not
+  screenshots.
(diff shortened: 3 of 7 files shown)


alxjrvs (commented):
## Review: `40f4ca71` — agent digest, scoped typecheck hook, worktree reaper

Scope: the range vs `origin/agent-tooling` is empty and the tree is clean, so I reviewed the single commit (7 files, +645/−12). I ran the new tool, exercised the hook with synthetic payloads, and validated every assumption about the transcript format against the real `~/.claude/projects` data rather than reading it off the comments.

**What's solid:** the `return 0` fix in `prune-stale-worktrees.sh` is real and correctly diagnosed — with a bare `return`, `process` inherited the failed test's status, and since it is invoked as the final command of a `&&` list, `set -e` killed the script on the main worktree before it examined a single candidate. The stale-lock PID interrogation is conservative in the right direction, and the new unpushed-commits guard is a genuine prerequisite for the fix. `knip`, `lint`, `format:check`, `validate:doc-drift`, and `validate:architecture` are all green; every path, workflow, and script the two new skill docs reference exists.

### Findings

- `tools/agent-digest.ts:153` — Every token metric is inflated ~3.5x. Claude Code emits one `assistant` record per content block and repeats the identical `message.usage` on each, so summing per record multi-counts. Measured on this repo's own transcripts: 17,996,629 raw vs 5,185,106 deduped by `message.id`. `toolCalls` is unaffected.
- `tools/agent-digest.ts:200` — The digest silently omits every agent worktree session. Exact-equality filtering on the main-worktree folder drops the 100 `-Users-jarvis-Code-SU-SRD--claude-worktrees-*` dirs; in the last 14 days that's 86 of 107 session files, i.e. it reports 20% of the fleet with no indication anything was dropped. This is the population the tool exists to measure, and `triage` step 6 ranks the day's work off it.
- `tools/agent-digest.ts:267` — `--json` prints prose on an empty window. Reproduced: `No agent sessions in the last 14 days.` with exit 0, breaking the documented `| jq` contract.
- `tools/agent-digest.ts:60` — `encodeProject` maps only `/`, but Claude Code also maps `.` and `_` to `-` (verified: `alxjrvs.github.io` → `-Users-jarvis-Code-alxjrvs-github-io`). Latent only because the current path has neither character; any clone under a dotted or underscored path reports zero sessions.
- `tools/agent-digest.ts:169` — "landed a commit" inspects only the Bash input, never the result, so pre-commit-blocked commits, `--dry-run`, and `grep 'git commit'` all count as landed — the same substring-scan failure CLAUDE.md documents for the rebase guard.
- `tools/prune-stale-worktrees.sh:90` — `git worktree unlock` runs before the dirty/unpushed guards, so a worktree the script then declines to remove is left permanently unlocked. Move it next to the `remove` call.
- `.claude/hooks/typecheck-scoped.sh:56` — Exits 0 with plain stdout, which for PostToolUse goes to the user's transcript view, not the model's context — so the diagnostics never reach the agent the header says they inform. Not a regression (the old hook was identical), but it undercuts the file's own cost/benefit case. `hookSpecificOutput.additionalContext` would fix it without ever blocking an edit.
- `tools/agent-digest.ts:84` — `branches` and `file` are populated on every record and never read; knip won't flag them as they're local type members.

alxjrvs (commented):
Reviewed commit `898e5576` (the only commit on `agent-tooling` not in `origin/main`); the working tree was clean at start. Note: the worktree's branch was switched out from under me mid-review by another session (`observability-tier1` now sits on `sentry-eu-region` @ `47f53dac`), so the later checks were run via `git show 898e5576:…` against the reviewed commit.

Findings, most severe first:

- `tools/agent-digest.ts:200` — The `d === wanted` project filter excludes every `.claude/worktrees/*` transcript folder (they encode as `wanted` + suffix). Measured over the tool's own 14-day default: 21 sessions / 109 subagent runs counted, 87 sessions / 294 subagent runs silently dropped — ~80% of the fleet, and precisely the population that default worktree isolation produces.
- `tools/agent-digest.ts:152` — `message.usage` is summed per JSONL record, but Claude Code writes one record per content block each repeating the same whole-message usage. On one real transcript: 3,315,039 output tokens as coded vs 1,096,152 deduped by `message.id` — 3.0x inflation, with a multiplier that varies per session so the numbers aren't comparable across runs.
- `tools/prune-stale-worktrees.sh:107` — The new unpushed guard (`git branch -r --contains HEAD`) is never satisfied after this repo's squash-merge + `delete_branch_on_merge` flow. Of the 12 current candidates, 8 are skipped as "unpushed", including 5 whose PRs (#588, #590, #591, #592, #593) are merged — so the fix that finally lets the reaper run still leaves issue #377's accumulation in place.
- `tools/prune-stale-worktrees.sh:89` — Under `--force` the stale lock is released before the dirty/unpushed checks, so a worktree that is then skipped is left permanently unlocked with its holding-session record gone.
- `tools/agent-digest.ts:181` — Denial regex runs on all `tool_result`s, not just errored ones; the file's own line 178 contains the literal `blocked by classifier`, so reading this file registers a phantom denial.
- `tools/agent-digest.ts:60` — `encodeProject` replaces only `/`; Claude Code replaces every non-alphanumeric char (verified: `/.claude/` → `--claude-`). Any checkout path containing `.` or `_` yields "No agent sessions" for a project full of them.
- `.claude/hooks/typecheck-scoped.sh:56` — The comment justifies skipping `tools/` by deferring to CI, but root `typecheck` (and `check:all`) only run the five workspaces, so the two new 200–344-line tools ship with no type checking anywhere.

The `return 0` fix in `process()` is correct and the root-cause analysis in its comment checks out: the old `*) return ;;` inherited the failed `[[ -z "$p" ]]` status, and under `set -e` the bare `process "$path" "$locked";` call in the `""` branch killed the script at the main worktree — the first entry, always declined.

alxjrvs (commented):
## Review: `fix(observability): point the CSP at Sentry's EU ingest region`

**Scope:** 1 commit, 3 files (`apps/itun/netlify.toml`, `apps/srd/netlify.toml`, `tools/check-observability.ts`) vs `origin/main`. No uncommitted changes.

**The core change is correct — I verified it against live production, not just by reading.** Both sites really do ship EU-region DSNs today:

- itun: `https://…@o4511811110764544.ingest.de.sentry.io/4511811273097296` (in `/assets/index-D37G06md.js`)
- srd: `https://…@o4511811110764544.ingest.de.sentry.io/4511811268902992` (in `/_astro/BaseLayout…js`)

So the previous `us` CSP was genuinely blocking every event, both `netlify.toml` files were updated consistently (one CSP block each, no stragglers), the `DSN_IN_BUNDLE` regex matches the real DSN format, and `SENTRY_INGEST_SUFFIX` derives correctly. I ran both modes: static passes, and `--live` now passes against real production.

### Findings

- `tools/check-observability.ts:58` — The comment here (and the copy added to both `netlify.toml`s) advertises the nightly live probe as the guard against region drift, but `observability-probe` is gated on `vars.OBSERVABILITY_PROBE == 'on'` and `gh api repos/SalvageUnion-io/SU-SRD/actions/variables` returns `total_count: 0`. The job is skipped every night, so the new region check has never run and would not catch the next drift. (Org-level variables couldn't be listed — 403 — but the workflow comment explicitly instructs repo-level.)
- `tools/check-observability.ts:202` — "Sentry SDK present" only ever matches the inlined DSN string, never the SDK. `@sentry/browser` is a dynamic import, so its chunk is neither a `<script src>` nor reachable by `scriptUrls`; I measured exactly one `/sentry/i` occurrence in each live entry chunk and it is the DSN hostname. If the SDK chunk is dropped or 404s while the DSN literal survives, the probe still reports healthy. The comment added at line 194 describes a scan that structurally cannot reach that chunk.
- `apps/itun/vite.config.ts:68` — The PR establishes the org is EU but fixes only the browser half. `sentryVitePlugin` has no `url`, so sourcemap upload defaults to the US API; `apps/itun/netlify.toml:13-14` documents only `SENTRY_AUTH_TOKEN`/`SENTRY_ORG`/`SENTRY_PROJECT`, not `SENTRY_URL`. The `errorHandler` swallows the failure as a warning, so itun errors ship without sourcemaps and the build stays green.
- `tools/check-observability.ts:225` — The region check is guarded on `dsnHost &&` with no else-branch, so an unmatchable DSN yields `✓ observability probe OK`. `scriptUrls` ignores `<link rel="modulepreload">`, and itun's HTML has 31 such chunks the probe never fetches — the DSN sits in the one fetched chunk by chunking luck. Worth an explicit "could not verify DSN region" failure, given the file's own doctrine that silence must never read as health.

PR #606, opened 2026-07-28
docs: resolve duplicate ADR-028, index the missing partners ADR

alxjrvs:
Two ADRs both claimed **028** on `main`:

| File | Landed | Disposition |
| --- | --- | --- |
| `ADR-028-partners-render-in-place.md` | #590, 00:35 | **keeps 028** (first claim) |
| `ADR-028-contribution-model-and-stat-provenance.md` | #596, 01:15 | renumbered to **029** |

Every inbound reference to the contribution-model ADR moves with it — `docs/README.md`, `docs/design/rules-parity-plan.md` (3 spots), and the two links from `ADR-022`.

## Second bug, found while fixing the first

`docs/README.md`'s ADR index had **no row for the partners ADR at all**. It was never added when #590 landed, so the governing decision for how partners render was missing from the one table that is supposed to list every ADR. Added.

## Verification

- No duplicate ADR numbers remain — `ls docs/adrs | sed -E 's/(ADR-[0-9]+).*/\1/' | sort | uniq -d` returns empty
- Every ADR-to-ADR markdown link under `docs/` resolves to a file that exists — checked by script, 0 broken
- `format:check` (the CI gate) passes

The bare `ADR-028` references left in `ADR-027` and in `apps/itun/src/**` are correct as-is — all of them refer to partners.

Docs-only; no code paths touched.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -100,7 +100,8 @@ Read the matching ADR before proposing alternatives.
 | [ADR-025](adrs/ADR-025-reference-versioned-releases-surface-gate.md) | Versioned internal releases + public-surface (TS + schema) gate for the ref; partially supersedes ADR-014       |
 | [ADR-026](adrs/ADR-026-entity-card-design-rules.md)                  | **Entity card design rules** — one renderer, choice/stat-atom/modified-stats/tech-level rules (built)           |
 | [ADR-027](adrs/ADR-027-partners-owned-by-host.md)                    | Partners are owned by their host entity                                                                         |
-| [ADR-028](adrs/ADR-028-contribution-model-and-stat-provenance.md)    | **Proposed** — one contribution model for caps/traits/damage + stat provenance; amends ADR-022's overrides      |
+| [ADR-028](adrs/ADR-028-partners-render-in-place.md)                  | Partners render in place as reference entities; supersedes ADR-027                                              |
+| [ADR-029](adrs/ADR-029-contribution-model-and-stat-provenance.md)    | **Proposed** — one contribution model for caps/traits/damage + stat provenance; amends ADR-022's overrides      |
 
 > ADR-021 is the governing decision for rules enforcement and takes precedence
 > over prior ADRs where they conflict on _how hard a rule is enforced on which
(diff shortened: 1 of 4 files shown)


alxjrvs (commented):
**Scope:** one commit, `5397f434 docs: resolve duplicate ADR-028, index the missing partners ADR` — 8 changed lines across 4 files, docs-only. (Note: local `main` is stale; `main...HEAD` shows 13 already-merged commits. The real scope is `origin/main...HEAD`.)

**The core change is correct.** Two ADRs both claimed number 028: the partners ADR landed first (`33ddccd7`, #590), the contribution-model ADR second (`e15b11cb`, #596). Renumbering the *second* to 029 is the right direction — it preserves the earlier-published number. All five inbound link sites were updated (`docs/README.md`, ADR-022 ×2, `rules-parity-plan.md` ×3), a repo-wide grep finds zero dangling `ADR-028-contribution-model-and-stat-provenance` references, every relative path resolves, and the `#amendment-2026-07-overrides-become-absolute-pins` anchor matches its heading. No CI job validates ADR numbering or links, so nothing breaks. No code references the renumbered ADR.

**Findings** — all index-consistency, none blocking:

- `docs/README.md:102` — ADR-027's index row still reads as live ("Partners are owned by their host entity") with no superseded marker, though the file's own Status says "Superseded by ADR-028". A reader following the table's own instruction to "read the matching ADR" could rebuild the `/sheet/partner/:id` surface that #590 deleted. The table already uses this convention on ADR-025 ("partially supersedes ADR-014"), so it's applied inconsistently.
- `docs/README.md:67` — the prose preamble enumerates statuses only through ADR-026 while the table now runs to 029. It asserts "ADR-023–026 are **Accepted**" although ADR-023's Status is "Superseded by ADR-027", and gives no status for 027/028/029 — so nothing flags that ADR-029 is only **Proposed** and order-dependent on the ADR-022 amendment.
- `CLAUDE.md:9` — hardcodes "26 ADRs"; there are now 29. Already off by two before this PR, but this is the commit that reconciles numbering. Worth fixing because CLAUDE.md loads into every session, and an agent that believes numbering ends at 026 would allocate ADR-027 for a new ADR — recreating the exact duplicate-number defect this PR repairs.

PR #607, opened 2026-07-28
fix(observability): point the CSP at Sentry's EU ingest region

alxjrvs:
The Sentry org for these projects is in the **EU data region** — every DSN
issued for it ingests at `*.ingest.de.sentry.io`, not the `us` default the CSP
was written against in #601.

Left alone, this would have been the *worst* shape of the original bug rather
than a fix for it: the DSN provisioned, the SDK shipping, the app looking
correctly instrumented — and the browser blocking every single event at the CSP
before it left the page. The Sentry project reports nothing, which from the
outside is indistinguishable from "no errors are happening".

`SENTRY_INGEST_HOST` and both `netlify.toml` CSPs move to `de` together. The
static wiring check already fails CI if those three ever disagree.

## Closing the gap that made this possible

The static check can only prove the CSP matches a constant **in this repo**. It
has no way to know which region the real DSN belongs to, because the DSN lives
in the deploy environment — so a us/de mismatch was undetectable by
construction.

The `--live` probe now extracts the ingest host from the DSN actually inlined in
the production bundle and asserts the CSP permits it:

```
[itun] DSN ships events to o…ingest.us.sentry.io, but the CSP only allows
       https://*.ingest.de.sentry.io.
       Every event will be blocked in the browser and the project will look silent.
```

So the repo checks what it can prove statically, and the nightly probe checks
the half only production knows.

## Verification

- `git grep ingest.us.sentry` over tracked source: no matches.
- Static check passes; negative-tested by reverting the constant to `us`, which
  fails with the CSP mismatch message, then restored.
- `bun run check:all` green.
- The live region assertion is exercised for real once the DSNs deploy; until
  then production has no DSN to compare against.

## Follow-up in this same batch

All four DSNs are now stored in the `claude-agent` 1Password vault as
`sentry-dsn-{srd,itun,itun-functions,discord-bot}`, and the Netlify/Render env
vars are being set separately (not a code change).

### DIFF
--- a/apps/itun/netlify.toml
+++ b/apps/itun/netlify.toml
@@ -102,9 +102,13 @@
     # event before it leaves the page, so a provisioned DSN would report zero
     # errors and look healthy. Kept in lockstep with SENTRY_INGEST_HOST in
     # tools/check-observability.ts, which fails CI if this drifts. Wildcarded
-    # subdomain because the host encodes the Sentry org id; `us` is the default
-    # data region (an EU org uses `de`).
-    Content-Security-Policy = "default-src 'self'; script-src 'self' 'unsafe-inline' https://storage.ko-fi.com; style-src 'self' 'unsafe-inline'; img-src 'self' data: https:; font-src 'self'; connect-src 'self' https://ingesteer.services-prod.nsvcs.net https://*.ingest.us.sentry.io;"
+    # subdomain because the host encodes the Sentry org id.
+    #
+    # `de`, not `us`: this org is in Sentry's EU data region. A DSN issued in
+    # one region under a CSP written for the other is blocked in the browser and
+    # the project reports nothing — identical, from the outside, to "no errors".
+    # The nightly live probe now compares this against the DSN actually shipped.
+    Content-Security-Policy = "default-src 'self'; script-src 'self' 'unsafe-inline' https://storage.ko-fi.com; style-src 'self' 'unsafe-inline'; img-src 'self' data: https:; font-src 'self'; connect-src 'self' https://ingesteer.services-prod.nsvcs.net https://*.ingest.de.sentry.io;"
     X-DNS-Prefetch-Control = "on"
 
 [[headers]]
--- a/apps/srd/netlify.toml
+++ b/apps/srd/netlify.toml
@@ -85,8 +85,13 @@
     # failure mode is silent AND looks healthy (zero errors reported). Kept in
     # lockstep with SENTRY_INGEST_HOST in tools/check-observability.ts, which
     # fails CI if this drifts. Wildcarded subdomain because the host encodes the
-    # Sentry org id; `us` is the default data region (an EU org uses `de`).
-    Content-Security-Policy = "default-src 'self'; script-src 'self' 'unsafe-inline' https://storage.ko-fi.com; style-src 'self' 'unsafe-inline'; img-src 'self' data: https:; font-src 'self'; connect-src 'self' https://ingesteer.services-prod.nsvcs.net https://*.ingest.us.sentry.io;"
+    # Sentry org id.
+    #
+    # `de`, not `us`: this org is in Sentry's EU data region. A DSN issued in
+    # one region under a CSP written for the other is blocked in the browser and
+    # the project reports nothing — identical, from the outside, to "no errors".
+    # The nightly live probe now compares this against the DSN actually shipped.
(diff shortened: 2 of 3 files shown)


alxjrvs (commented):
## Review: `fix(observability): point the CSP at Sentry's EU ingest region`

Scope was 3 files (`git diff origin/main...HEAD`) — local `main` was ~14 commits stale, so `main...HEAD` was misleadingly large. Working tree is clean.

The diff is small and internally consistent: the constant and both `netlify.toml` CSPs were flipped together, `checkStatic` still enforces lockstep, and the new `checkLive` DSN comparison correctly no longer early-returns on the first SDK hit (the SDK and the inlined DSN genuinely can land in different chunks).

**Findings**

- `tools/check-observability.ts:62` — The `us`→`de` flip is asserted, not verified. `checkStatic` only compares the tomls against this same constant, and all three moved together, so it agrees by construction whatever the org's real region is. The only true check is `--live`, gated behind `vars.OBSERVABILITY_PROBE == 'on'`. I ran `bun tools/check-observability.ts --live` against production: both apps report "Sentry SDK absent", i.e. **no DSN is provisioned yet**, so nothing anywhere currently verifies `de`. If the org is really US, provisioning the DSN yields a project that reports zero events and looks healthy — the exact failure this file exists to catch, with the polarity reversed from 360abc8e.
- `tools/check-observability.ts:225` — `if (dsnHost && ...)` makes "DSN not found" a silent pass. Both apps load `@sentry/browser` via dynamic `import()`; if the DSN literal ever lands in that lazy chunk (chunking change, or reading the env var inside the dynamic import) it's in no `<script src>` and `dsnHost` stays null. The per-chunk `catch` swallowing the one fetch that mattered does the same. The probe then prints `✓ observability probe OK` while a cross-region DSN sits behind the de-only CSP — contradicting the new docblock's "fails loudly instead of going quiet". Should fail/warn when `sdkFoundIn` is set but `dsnHost` is null.
- `tools/check-observability.ts:65` — `SENTRY_INGEST_HOST.replace('https://*', '')` is a literal replace. Narrow the constant to a pinned org (which the docblock two lines up explicitly contemplates) and the replace no-ops, so `SENTRY_INGEST_SUFFIX` keeps the scheme, `endsWith` is false for every host, and a healthy production fails nightly with "DSN ships events to `o123.ingest.de.sentry.io`, but the CSP only allows `https://o123.ingest.de.sentry.io`".

Nothing else in the repo still references `ingest.us.sentry.io` — no leftover drift in docs, workflows, or configs.

alxjrvs (commented):
**Verdict: the core change is correct and should land.** I confirmed the premise empirically rather than taking the commit message's word for it — the DSN inlined in both production bundles is `o4511811110764544.ingest.de.sentry.io`, and both sites are *currently* serving `connect-src ... https://*.ingest.us.sentry.io`. So the SDK is provisioned, shipping, and every event is being blocked at the CSP right now. `us`→`de` is the right fix. Static check and the live probe both pass on this branch.

Findings:

- `tools/check-observability.ts:225` — Region check is skipped entirely when no DSN host is extracted. `if (dsnHost && ...)`: if the chunk holding the DSN fails to fetch (swallowed by the `catch` at 204) or `DSN_IN_BUNDLE` stops matching, the probe prints "SDK present" and exits 0 despite a region mismatch — the exact looks-healthy silence the check exists to close. Absent DSN + present SDK should be a failure.
- `tools/check-observability.ts:177` — The probe compares the DSN to a repo constant, never to the CSP production actually serves. Once this merges the constant and the DSN agree, so the nightly goes green even if the netlify.toml edit never reaches a deploy — which is precisely the state production is in today. `res.headers.get('content-security-policy')` is already available on the response fetched at line 177.
- `apps/itun/vite.config.ts:68` — `sentryVitePlugin` has no `url`, so it defaults to `https://sentry.io/` (US) for what this PR establishes is an EU org. Currently inert (no `_sentryDebugIds` in the shipped bundle, so no auth token is set), but the moment someone provisions the vars the line-12 comment tells them to, the org lookup fails, `errorHandler` swallows it as a warning, and stack traces stay minified forever. Needs `url: 'https://de.sentry.io/'` or `SENTRY_URL` documented alongside `SENTRY_ORG`/`SENTRY_PROJECT`.
- `tools/check-observability.ts:65` — `SENTRY_INGEST_SUFFIX` is derived via `.replace('https://*', '')`, a no-op if the constant is ever de-wildcarded (a tradeoff the doc comment at 49-52 explicitly raises). The suffix becomes a full URL, `endsWith` is false for every host, and the probe fails with a self-contradictory message on a correct production.

PR #608, opened 2026-07-28
feat(itun): split cap overrides from manual adjustments; derivations return breakdowns

alxjrvs:
**A3** of the rules-parity plan (#597) — the blocker everything downstream waits on. Decisions in ADR-029 + the ADR-022 amendment.

## The blocker this removes

`max*Modifier` carried two incompatible meanings at once: the ADR-022 Free-Edit cap pin **and** the only channel any rules modifier had into a maximum. The sheets recovered the "derived baseline" by subtracting the field back out:

```ts
const derivedMaxSP = maxSP - (mech.maxSpModifier ?? 0)
```

So an automatic contribution written there would render as a **hand override with a revert button**. No ability bonus could be applied until the two were separated.

| Field | Meaning now |
|---|---|
| `max*Modifier` | **Manual adjustment** — contributes to the derivation. Every stored value and its arithmetic unchanged; only re-documented. |
| `max*Override` | **Absolute pin** (new). Stat shows it and stops tracking derivation; the derived total keeps computing underneath for the callout and revert, and is never persisted. |

## No migration — and two reasons the original plan couldn't work

The amendment originally said "convert every existing value to an absolute pin." Implementation ruled that out:

1. **A migration cannot compute a pin.** It needs each record's full derived total (chassis `structurePoints`, installed `statBonus` sum, tech-level base, injury penalties), but a migration may only await IndexedDB operations on its versionchange transaction — awaiting reference data auto-commits it. Migration 8 hardcodes frozen snapshots for exactly this reason. Deferring to hydration would make the result depend on *when* the user next opened the app.
2. **A pin freezes the stat — that's what a pin is.** Converting every legacy modifier would permanently disconnect that maximum from the rules, for precisely the players who hand-entered Beefcake because the app wouldn't apply it. They'd be the population this project exists for, getting the least from it.

Splitting the meanings into two fields is **lossless by construction**: not one stored byte changes, not one displayed number moves, no migration, no hydration pass, no version-skew reasoning. The legacy adjustment keeps composing with future contributions, and once the provenance panel lands it appears as its own labelled line — so a hand-entered bonus duplicating a real one becomes **visible and removable** rather than silently doubled.

## Derivations now return breakdowns

`mechMaxSPParts` / `EPParts` / `HeatParts` / `CargoParts` / `pilotMaxHPParts` / `pilotMaxAPParts` join the existing `crawlerMaxSPParts`, all returning a shared `StatBreakdown` (`base` / `installed` / `adjustment` / `derived` / `override` / `total` / `overridden`). Scalar functions become thin `.total` wrappers, so **~40 call sites are untouched**. This is B1's data layer, landed here to avoid churning the same API twice.

Pilot max HP stays deliberately **unclamped** so a lethal injury total still reads as the dead state (rules A2) rather than flooring to 0.

## Also: ADR renumbering

#590 landed its own `ADR-028` while this was in flight, so the contribution-model ADR becomes **ADR-029**. Both are now in the docs index (ADR-027 and the partners ADR-028 had been omitted).

## Verification

`bun run check:all` fully green — lint, format, typecheck, **4,509 tests**, validate, knip, audit, tokens, styling.

New tests pin the semantics: a pin replaces and retains its baseline; an adjustment contributes and shows **no** override indicator; the two coexist with the pin winning; a pin of `0` is honoured rather than treated as absent; and a pilot HP pin overrides the injury penalty.

Refs #597.


alxjrvs (commented):
Reviewed commit `1398dcd5` (splitting cap overrides from manual adjustments). Verified `bun run typecheck`, `bun run knip`, the `salvageunion-reference` suite (928 pass), and the two touched ITUN sheet test files — all green. The core `StatBreakdown`/`breakdownOf` refactor is sound: all six sheet override call sites were converted consistently, `VitalGauge`'s `isOverridden` semantics still line up with passing `undefined` when unpinned, the API report is regenerated, and no consumer read the removed `CrawlerMaxSPParts.modifier`.

Findings:

- `apps/itun/src/components/sheet/MechSheet.tsx:243` — The Live Sheet max editor was the only writer of `max*Modifier`; it now writes `max*Override` and nothing migrates existing values, so legacy hand-entered modifiers become a permanent, un-clearable contribution. Editing the max back down writes a pin on top of the stale modifier rather than removing it.
- `apps/itun/src/components/mech/MechWizard.tsx:334` — Edit-mode EP readout threads only `maxEpModifier` into `mechMaxEP`, so a `maxEpOverride` pin vanishes in the wizard while the Live Sheet honors it (this call site did pass the pin before, when pins lived in `maxEpModifier`).
- `apps/itun/src/lib/schemas/pilot.ts:300` — `maxInventorySlotsOverride` is added to the strict schema (so it validates, persists, and round-trips through export) but `pilotInventoryCapacity` never reads it; the other six pins are all wired into their derivations.
- `docs/design/rules-parity-plan.md:55` — Decisions table and unit A3 still state the "convert every `max*Modifier` to a pin / IDB migration lossless" decision that the ADR-022 amendment in this same commit reverses; only the ADR-028→029 renumbering was propagated.
- `apps/itun/src/components/sheet/__tests__/Snapshot-frozen-provenance.test.tsx:126` — The P5.2 "overrides render as plain values" test still seeds `maxHpModifier`, which no longer produces an override, so its negative assertions now pass trivially and would not catch removal of the `readOnly` guards (same for the `overriddenMech` fixture at line 51).

alxjrvs (commented):
## Review: `parity-override-pins` (2 commits vs `origin/main`)

Scope: splitting Free-Edit cap pins (`max*Override`) out of manual adjustments (`max*Modifier`), `StatBreakdown` returns from `derivedStats`, and the new `StatProvenance` panel.

Verified first: `bun run typecheck` clean across all workspaces, `bun run build:package` produces no generated-file drift, and the touched test files all pass. The `CrawlerMaxSPParts.modifier` → `adjustment` rename has no remaining consumers, the crawler wizard's `Pick`-based patch preserves the new fields, and the no-migration decision checks out numerically (`base + installed + adjustment` is unchanged, so no stored value or displayed max moves).

### Findings

- `packages/component-lib/src/components/stat/StatProvenance.tsx:85` — the trigger's `onMouseEnter` fights Popover's own click toggle: the mouseenter→click sequence a tap produces opens then immediately closes the panel. Reproduced with a throwaway test; touch is the exact audience the doc comment names.
- `packages/component-lib/src/components/stat/StatProvenance.test.tsx:40` — the "opens on click — the touch/tap path" test fires a bare `click` with no preceding `mouseEnter`, so it doesn't exercise a real tap and passes while that path is broken.
- `packages/salvageunion-reference/lib/rules/derivedStats.ts:73` — `pilotMaxHPParts` returns a floored `derived` on the overridden branch and an unfloored one otherwise; with a lethal injury total plus a pin the sheet says "revert to derived 0" but reverting yields a negative max and flips the pilot dead.
- `apps/itun/src/components/mech/MechWizard.tsx:334` — the wizard's EP readout forwards only `maxEpModifier`, not the new `maxEpOverride`, so a pinned Max EP now reads differently in the wizard than on the sheet. This is a regression: pre-PR the pin lived in `maxEpModifier`, which the wizard did forward.
- `apps/itun/src/lib/schemas/pilot.ts:300` — `maxInventorySlotsOverride` is added with pin semantics in its doc, but `pilotInventoryCapacity()` reads only `maxInventorySlotsModifier`, so the field is silently inert.
- `packages/component-lib/src/components/stat/StatProvenance.tsx:91` (plausible) — `sideOffset={6}` plus an undelayed `onMouseLeave` means the panel closes before the pointer crosses the gap, so the popup's own keep-open `onMouseEnter` is unreachable.
- `packages/component-lib/src/components/stat/StatProvenance.tsx:84` — `aria-label` on the trigger replaces the numeral it wraps, so screen-reader users never hear the value.

Non-blocking: `StatProvenance.tsx:108` adds a new Biome `noArrayIndexKey` warning (lint still exits 0), and `StatProvenance` is currently barrel-exported but consumed only by its story and test — no app surface wires it up yet.

Issue #165, opened 2026-03-15
Discord bot ITUN integration


alxjrvs:
## Story 14.4

**Package(s):** `discord-bot`, `itun`

Connect bot to ITUN's Supabase backend for live campaign state.

### Acceptance Criteria
- [ ] Bot authenticates with Supabase service role key
- [ ] `/character` pulls live data from ITUN campaigns
- [ ] Resource changes via bot reflected in ITUN UI (realtime)
- [ ] RLS policies allow bot access appropriately

Issue #157, opened 2026-03-15
Invite Code Join Flow


alxjrvs:
## Story 13.1

**Package(s):** `itun`  
**Phase 5II:** Wave 2 (Multiplayer)

"Enter Code" modal for new players to join via invite code.

### Acceptance Criteria
- [ ] "Join Game" button opens code entry modal
- [ ] Valid codes add player with appropriate role
- [ ] Invalid/expired codes show clear error
- [ ] New member appears in roster immediately

Issue #156, opened 2026-03-15
[Epic 13] Multiplayer & Social


alxjrvs:
## Epic 13: Multiplayer & Social `[gameplay]`

> Fill gaps in multiplayer: invite flows, pushing, crafting, salvage tallying. Total effort: L

**Phase 5II status:** Invite code flow is Phase 5II Wave 2. Pushing, crafting, salvage are net-new.

### Stories
- [ ] 13.1: Invite Code Join Flow
- [ ] 13.2: Pushing Mechanics
- [ ] 13.3: Crafting System
- [ ] 13.4: Salvage Tallying

PR #609, opened 2026-07-28
feat(itun): ADR-030 accounts & Games — Phase 0 + Phase 1 server layer

### LINKED ISSUE #152: [Epic 12] Session Management
## Epic 12: Session Management `[gameplay]`

> Support session transitions with downtime, ability training, and class advancement. Total effort: XL

**Phase 5II status:** Downtime wizard is deferred scope. Ability training and class advancement are net-new.

### Stories
- [ ] 12.1: Downtime Wizard
- [ ] 12.2: Ability Training UI
- [ ] 12.3: Class Advancement UI

alxjrvs:
Sits on top of the active work: contains **#604**'s commit (Convex provisioning), and **#606** (ADR renumber) merged to `main` while this was in flight so its commit dropped out on rebase. Based on `main` rather than #604's branch because `main` moved repeatedly during the work and re-stacking on a stale parent left this permanently "behind". **Review the top two commits** — #604's will vanish from the diff once it merges.

Executes Phase 0 and the server half of Phase 1. **No behaviour change for existing users**: ITUN still runs Solo end to end, because no build has `VITE_CONVEX_URL` unless someone runs `convex dev`.

## Phase 0 — decide, then clear the ground

**ADR-030** supersedes ADR-001 and amends ADR-022. ADR-001 is marked superseded rather than deleted, with a note on what it still governs — anonymous Solo play stays fully local, ADR-004 snapshot sharing is untouched, and its honour-system stance on cross-player writes survives as propose-and-confirm. What ADR-030 overturns is narrower than ADR-001's title suggests.

Plus `docs/architecture/accounts-and-games.md` (the phased delivery plan), indexed in `docs/README.md`.

**Eldridge Coast retired.** Switcher option, Roster seeding branch, and `lib/eldridgeCoast/` all gone — once nothing renders it, the knip gate would flag the module as dead anyway. Deliberately **kept**:

- `migrations/12-companion-mechs-to-partners.ts` and its test. It repairs `eldridge-*` rows in browsers that already spawned the workspace; deleting the seed does not delete their data.
- ADR-022's argument, which cites the module's header as evidence. Retensed and annotated rather than cut, so the record stays readable now the file is only in git history.

**Correction to the plan's own Phase 0 list.** It said to close #152, #156, #157, #165 as "Supabase-era". Checking the issue bodies, **only #165 mentions Supabase**. #156's remaining stories are Pushing, Crafting, and Salvage Tallying — single-player gameplay mis-filed under "Multiplayer" — and #152 is pure gameplay backlog. Only #157 and #165 should close; the plan is corrected rather than executed as written.

## Phase 1 — server layer + storage modes

`apps/itun/convex/`: `model/permissions.ts` (all capability checks, one place), `games.ts`, `invites.ts`, `ownership.ts`.

Three decisions worth review:

**`transferOrganizer` patches both memberships in one mutation.** Convex mutations are transactional, so there is no window where a Game has two Organizers or none. That invariant is why it is one function rather than a clear-then-set pair a client could half-complete.

**`destroy()` does not delete owned entities.** They fall back to their owner's shelf (`gameId: null`), so deleting a campaign never deletes somebody's character. Only *unclaimed* entities go — they have no shelf to fall back to, and that is the one place ADR-030's "both columns null is invalid" rule could otherwise be violated.

**Invite codes reuse `generateUniqueId`** from the snapshot module rather than growing a second generator. Already Crockford base32 (no I/L/O/U, so a code read aloud at a table cannot be mistyped into another valid one), already collision-checked, already crypto-backed.

### The three storage modes

`src/lib/connection/` — a pure function, no React and no I/O, **tested exhaustively across all eight input combinations**. The property that matters: **Solo is not Disconnected.** Someone who never signed in is not offline; they are using the app exactly as it worked before accounts existed, and no write of theirs is ever refused.

The provider is deliberately **two components rather than one `if`**. `useConvexAuth` throws without a `ConvexProvider` above it, and a build with no `VITE_CONVEX_URL` has none — that build is Solo forever, which is supported (CI runs in it). A conditional hook call would break the rules of hooks, so the branch is made at component level on a build-time constant, keeping component identity stable across renders.

`apps/itun/CLAUDE.md` updated — it said "No auth, no backend", which is loaded into every session and now false.

## Verification

typecheck 0 errors · **3801 tests pass** across all four workspaces · knip clean · no new lint warnings (the new `useComponentExportOnlyModules` warning was fixed by splitting the context out, not suppressed).

## Not done — honest scope

Phase 1's client cutover (Games UI, sign-in, claim-local-data, Starter Set as template, account management screens) and Phases 2–6. The Convex functions also have **no unit tests** — that needs `convex-test`, which is its own setup. Phase 2's exit criterion explicitly requires them, and this PR does not meet it.


alxjrvs (commented):
Reviewed the three accounts-phase1 commits (`29b0aef5`, `d1c3cf4c`, `664b34f3` — 40 files: the Convex backend, the connection-mode layer, and the Eldridge Coast removal). Tests pass (1280/1280), knip is clean, and the app typecheck is green. Seven findings:

- **apps/itun/netlify.toml:107** — CSP `connect-src` omits the Convex origin; once `VITE_CONVEX_URL` is set, the browser blocks `wss://*.convex.cloud/api/*/sync` and every auth/query/mutation fails in production with no banner. The file's own comment records the identical Sentry regression.
- **apps/itun/src/lib/connection/connectionMode.ts:50** — `disconnected` is derived only from `navigator.onLine`; a dead deployment or dropped socket leaves mode `connected`, writes allowed, banner hidden. `ConvexReactClient.connectionState()` is unused.
- **apps/itun/src/lib/connection/ConnectionProvider.tsx:68** — `useConvexAuth().isLoading` is discarded, so every page load for a signed-in user resolves to `solo` (IndexedDB as source of truth, writes allowed) until the handshake completes.
- **apps/itun/convex/ownership.ts:38** — `loadOwnable` never verifies the document belongs to the named `table`; `assign({table:'pilots', entityId:<mech id>})` reassigns the mech and writes a Change Log row labelled `entityType: 'pilot'`. A `mechPatterns` id also passes the `'ownerId' in doc` guard.
- **apps/itun/tsconfig.json:19** — `convex/**` is outside `include`, so ~870 lines of new backend (including all of `model/permissions.ts`, the sole authorization boundary) are never typechecked by CI or pre-push, and have no tests. It compiles clean today — the gap is that nothing would catch a regression.
- **apps/itun/convex/games.ts:140** — `destroy` omits `mechPatterns` from its cleanup, leaving shared patterns with a `gameId` pointing at a deleted document.
- **apps/itun/convex/ownership.ts:193** — `leaveGame` doesn't delete the departing member's `presence` row, so they stay listed at the table forever.

Two things I checked and cleared: invite codes are uppercase Crockford base32 on both the generate and the `toUpperCase()` redeem path (no mismatch), and the custom `users` table override drops `phone`/its index but the Discord profile never reaches that code path.

alxjrvs (commented):
Reviewed `origin/main...HEAD` (4 commits, 42 files — the ITUN accounts/Games phase 1: Convex backend, auth scaffold, connection-mode module, Eldridge Coast seed removal). Verified the branch state directly: `bun run typecheck`, `bun run knip`, `bun --filter itun test` (1302 pass, incl. the 21 new permission tests), `bun --filter itun build`, and `bun install --frozen-lockfile` all pass; `@auth/core` 0.41.3 satisfies `@convex-dev/auth@0.0.94`'s peer range. Two findings were confirmed by executing code (a scratch convex-test case for the ownership mismatch; a scratch tsconfig to prove convex/ compiles but is never included).

Findings:

- `apps/itun/src/lib/connection/ConnectionProvider.tsx:68` — `signedIn` comes from `useConvexAuth().isAuthenticated`, which requires a completed server handshake, so a signed-in user opening the app offline is classified `solo` (writes allowed, no banner) rather than `disconnected`. Cold starts online also pass through a false-Solo window.
- `apps/itun/convex/ownership.ts:38` — `loadOwnable` never checks the id belongs to `args.table`; `assign({table:'pilots', entityId: <mech id>})` patches the mech and logs `entityType: 'pilot'` (reproduced). Use `ctx.db.normalizeId`.
- `apps/itun/convex/games.ts:144` — `destroy` deletes `softLinks` while preserving the pilots/mechs they connect, so a surviving pilot↔mech pairing is severed when a Game is deleted.
- `apps/itun/convex/games.ts:141` — `mechPatterns` is in neither loop of `destroy`, leaving an owner's shared pattern with a dangling `gameId` and `sharedToGame: true` instead of falling back to their shelf (`changeLog` rows dangle likewise).
- `apps/itun/tsconfig.json:19` — `include` omits `convex/`, so the whole new backend is typechecked by nothing in CI; a type error there would only surface at `convex deploy`.

PR #611, opened 2026-07-28
fix(observability): probe the CSP production SERVES, not the one this repo declares

alxjrvs:
The live probe compared the DSN's ingest host against `SENTRY_INGEST_HOST` — a
constant **in this repo**. That is circular, and it produced a false PASS
against real production the moment it mattered.

## What actually happened

Observed, not hypothesised:

1. #607 changed the constant and both `netlify.toml` CSPs from `us` to `de`, and merged.
2. The Netlify sites had not redeployed yet — so they were still **serving** the
   `us` CSP while already **shipping** a `de` DSN.
3. Every event was therefore blocked in the browser and Sentry was silent.
4. The probe read the repo's already-correct constant, matched it against the
   DSN, and reported:

```
✓ observability probe OK
```

The precise failure this file exists to make loud, green-lit by the check meant
to catch it.

A live probe that consults the repo cannot see a deploy that hasn't happened, a
CSP overridden in the Netlify UI, or one coming from a `_headers` file this repo
never sees.

## The fix

Read the `content-security-policy` **response header** from production and ask
whether that policy actually permits the host the shipped DSN points at — via a
real source-list matcher (exact origin, or leftmost-subdomain wildcard) rather
than a string compare against a constant.

Against the identical production state it now correctly fails:

```
✗ observability probe failed:
  [srd] production ships a DSN pointing at o4511811110764544.ingest.de.sentry.io,
      but the CSP it SERVES does not permit that origin, so every event is blocked
      in the browser and the Sentry project looks silent.
      served: connect-src 'self' https://ingesteer.services-prod.nsvcs.net https://*.ingest.us.sentry.io
      expected to allow: https://*.ingest.de.sentry.io
```

Exit code 1, confirmed separately from the pipeline.

## Division of labour, now clean

| check | runs | proves |
|---|---|---|
| static | every PR | the **repo** is internally consistent (module ↔ entry ↔ CSP ↔ constant) |
| `--live` | nightly | the **deployment** actually works — and consults nothing but the deployment |

Also drops the now-unused `SENTRY_INGEST_SUFFIX`.

`bun run check:all` green; static check still exits 0, live probe exits 1 against
current production (correctly, until the `de` CSP finishes deploying).

### DIFF
--- a/tools/check-observability.ts
+++ b/tools/check-observability.ts
@@ -61,9 +61,6 @@ import { join } from 'node:path'
  */
 const SENTRY_INGEST_HOST = 'https://*.ingest.de.sentry.io'
 
-/** The `*.ingest.<region>.sentry.io` suffix implied by SENTRY_INGEST_HOST. */
-const SENTRY_INGEST_SUFFIX = SENTRY_INGEST_HOST.replace('https://*', '')
-
 /** A Sentry DSN as it appears inlined in a built bundle. */
 const DSN_IN_BUNDLE = /https:\/\/[0-9a-f]{16,}@([a-z0-9.-]+\.ingest\.[a-z0-9.-]*sentry\.io)\//i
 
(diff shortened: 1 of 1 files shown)


alxjrvs (commented):
Reviewed the single-commit diff (`ab70aeda`, `tools/check-observability.ts`, +56/-14). I verified behavior empirically: curled both production sites (both still serve the `us` CSP while the repo declares `de`), ran the live probe end-to-end (it correctly fails both apps today), and exercised `connectSrcOfPolicy`/`connectSrcPermits` against seven CSP shapes in a scratch script. Biome is clean.

The core change is sound — comparing the DSN against the *served* header instead of a repo constant is the right call, and no edge case I found produces a false *pass*; the parser's failure modes are all fail-closed. The findings are one coverage regression and a set of false-alarm paths that matter because this is a gated nightly alarm whose own workflow comment warns about broken-window fatigue.

Findings (most severe first):

- `tools/check-observability.ts:58` — Live probe lost the DSN-region-vs-repo-CSP check; a self-consistent-but-stale production (`us` CSP + `us` DSN) now reports green, where the old `SENTRY_INGEST_SUFFIX` comparison failed loudly. That is today's actual state on both sites. The docstring at 58-60 and both netlify.toml comments still assert the removed behavior.
- `tools/check-observability.ts:259` — `!servedConnectSrc` hard-fails when the served policy legitimately covers Sentry via `default-src` fallback, or is report-only, or absent entirely (all cases where beacons flow).
- `tools/check-observability.ts:135` — `connectSrcPermits` doesn't understand the bare `*` host-wildcard or a scheme-source like `https:`; both permit Sentry but return false (confirmed by running the function).
- `tools/check-observability.ts:258` — Multiple CSP headers are joined by `Headers.get` with `", "` and mis-parsed (trailing comma glued onto the Sentry token) — precisely the "overridden in the Netlify UI" scenario the new comment cites as motivation.
- `tools/check-observability.ts:114` — The netlify.toml-specific JSDoc now sits above `connectSrcOfPolicy` (which takes a policy string), leaving `connectSrcOf` undocumented.

PR #613, opened 2026-07-28
feat(itun): wire stat provenance into the Live Sheets and the Dashboard

alxjrvs:
**B3 + B4** of the rules-parity plan (#597). Every derived maximum on both live surfaces can now explain itself — the feature originally asked for.

## The interaction decision

The max numeral is **already** the click-to-edit override affordance, so wrapping it in a popover trigger would make one click mean two things. Instead:

- **Overridden stat** → the existing `*` marker *becomes* the trigger. It already signals "there is something here to explain."
- **Otherwise** → a small `ⓘ` beside the numeral.

Both open on **hover, focus and tap**. Click-to-edit is untouched.

## The Dashboard was dropping this at two layers

`DashboardGauge` never forwarded the override/provenance props, *and* `VitalGauge`'s `compact` branch returned before any of that chrome rendered — so even fixing the first wouldn't have shown anything. ADR-021 requires Guided Play to "teach as it enforces"; this is the affordance it lacked. Both layers now carry it, with a regression test at each.

## `linesFromBreakdown`

Turns a numeric `StatBreakdown` into a labelled ledger:

- **Zero-valued lines are omitted** — a ledger reading `+0` for everything a mech doesn't have is noise, not provenance.
- **An override appends to the full derivation** rather than replacing it, so the revert target stays visible and explained.
- **A negative contribution is still a contribution** — injury penalties render as `−2` on their own line.

The contribution line is deliberately a single **aggregate** ("Installed systems & modules") for now. The derivation sums `statBonus` across installed items and returns one number, so there's no honest way to attribute it per item yet — per-source attribution (`Heat Sink ×2`, `Beefcake`) arrives with the contribution model (#598), and the helper is shaped to take those lines unchanged. One truthful aggregate beats an invented breakdown.

## A bug I introduced and caught

My first cut passed `overriddenFrom={overridden ? max : undefined}` — which can **never** flag as overridden, because the gauge treats `overriddenFrom === max` as "not overridden". The prop is now `derivedMax`, documented, and pinned by a test that asserts both directions.

## Verification

`bun run check:all` green: lint, format, typecheck, **4,525 tests**, validate, knip, audit, tokens, styling.

Remaining in B: **B5** (partner cards + frozen snapshot) and **B6** (retire the wizard's bespoke `sp-breakdown` string).

Refs #597.

### DIFF
--- a/apps/itun/src/components/dashboard/dialItems.ts
+++ b/apps/itun/src/components/dashboard/dialItems.ts
@@ -11,11 +11,11 @@
  */
 
 import {
-  crawlerMaxSP,
-  mechMaxHeat,
-  mechMaxSP,
-  pilotMaxAP,
-  pilotMaxHP,
+  crawlerMaxSPParts,
+  mechMaxSPParts,
+  mechMaxHeatParts,
+  pilotMaxHPParts,
+  pilotMaxAPParts,
 } from '../../lib/rules/derivedStats'
 import { resolveChassisRef } from 'salvageunion-reference/rules'
 import type { CockpitPrefs, DialKind } from '../../lib/schemas/cockpitPrefs'
@@ -24,6 +24,7 @@ import type { Mech } from '../../lib/schemas/mech'
 import type { Pilot } from '../../lib/schemas/pilot'
 import type { MountState } from '../../stores/playStateStore'
 import type { DialItem as DialCellItem } from 'component-lib'
+import { linesFromBreakdown } from 'component-lib'
 
 export type DialItem = DialCellItem & { kind: DialKind }
 
@@ -41,8 +42,10 @@ export const DIAL_KIND_LABELS: Record<DialKind, string> = {
 export const LOCKED_DIAL_KIND: DialKind = 'actions'
 
 function pilotItem(pilot: Pilot): DialItem {
-  const maxHP = Math.max(0, pilotMaxHP(pilot))
-  const maxAP = Math.max(0, pilotMaxAP(pilot))
+  const hpParts = pilotMaxHPParts(pilot)
+  const apParts = pilotMaxAPParts(pilot)
+  const maxHP = Math.max(0, hpParts.total)
+  const maxAP = Math.max(0, apParts.total)
   return {
     key: `pilot:${pilot.id}`,
     kind: 'pilot',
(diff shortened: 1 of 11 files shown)


alxjrvs (commented):
Reviewed commit `60188291` ("feat(itun): wire stat provenance into the Live Sheets and the Dashboard") against fresh `origin/main` — 11 files, +393/−20. Note the local `main` ref was stale (it contained the branch's parent commits already merged), so scope had to be taken against `origin/main`.

`bun --filter component-lib test` (636 pass), `bun --filter itun test` (1288 pass) and `bun run knip` are all green — the findings below are behaviours the suite does not cover, three of them reproduced with scratch tests that I removed afterwards (tree is clean).

Findings:

- `packages/component-lib/src/components/dashboard/Dial.tsx:107` — The provenance trigger is a `<button>` nested inside `DialCell`'s own `<button role="option">`. Verified: rendering `<Dial>` with `provenance` yields a `button button` pair, and clicking the trigger fires `onActiveIndexChange(1)` — so tapping "how this is derived" on a track cell rotates the dial. On touch (no hover) that tap is the only way to open the ledger.
- `packages/component-lib/src/components/stat/VitalGauge.tsx:374` — The new trigger is a focusable child of the gauge's `role="group"` wrapper, whose `onKeyDown` handles Arrow/Home/End. Verified: focusing the trigger and pressing ArrowRight then ArrowDown fires `onChange(6)` then `onChange(4)` — on the Live Sheets that writes through to IndexedDB, so scrolling the popover silently changes current HP.
- `packages/component-lib/src/components/stat/provenanceLines.ts:59` — The adjustment line is hardcoded "Manual adjustment · entered by hand", but `max*Modifier` also carries rules-sourced advancement (the package doc calls `maxApModifier` "Stat Training tiers"; the Eldridge Coast seed pilots ship `maxHpModifier: 4`/`maxApModifier: 2`). `ProvenanceLabels` gives callers no way to relabel it.
- `packages/component-lib/src/components/stat/VitalGauge.tsx:203` — `DashboardGauge` always passes `readOnly`, so the compact root is `role="img"`, whose descendants are presentational per ARIA; the trigger is invisible to AT on every Dashboard gauge. The new test's `getByRole('button')` passes only because testing-library doesn't implement that rule, so it gives false assurance.
- `packages/component-lib/src/components/stat/provenanceLines.ts:45` — The ledger never renders the 0-floor. `pilotMaxHPParts` returns an unfloored total while `PilotSheet` clamps with `Math.max(0, …)`, so a pilot with −12 injuries on base 10 shows "Pilot 10 / Injuries −12 / Showing 0" — the panel's own arithmetic doesn't add up.
- `apps/itun/src/components/dashboard/ActiveItemBand.tsx:103` — The Dashboard's primary always-visible readout builds its own `BandGauge`s (lines 103, 275, 344, 376, 493, 524) and got no `provenance`/`derivedMax`, so the only Dashboard surface wired is the dial cells, where the first finding makes it unusable.

PR #614, opened 2026-07-28
feat(itun): stat provenance on frozen snapshots and partner cards

alxjrvs:
**B5** of the rules-parity plan (#597), plus a finding that supersedes **B6**. This completes **Phase B** — every surface that shows a derived maximum can now explain it.

## Frozen snapshots

All seven gauges carry the full breakdown, and no revert. That falls out of the existing contract rather than a special case: the revert only renders when `onRevertOverride` is supplied, and a published snapshot never supplies one (ADR-022 + ADR-010).

## Partner cards — the surface where this mattered most

Every partner stat is `base + bonusPerTechLevel × (TL − 1)` and **nothing on screen said so**. `partnerDerivedStatsParts` decomposes them, carrying the scaling on the contribution line like any other rules-sourced addend.

They render through `summarizeBreakdown` — a one-line text form (`Base 8 +4 Tech 3 scaling = 12`) — rather than the popover panel. A `Stat` cell carries an **editable stepper cluster**, and the panel's trigger is a `<button>`; wrapping one in the other nests a button inside a button. The text form is less rich but true, and beats a number with no explanation. The panel arrives when `Stat` grows a non-nesting affordance, alongside the entity-card work in C4.

## B6 is superseded, not skipped

It was written as *"retire the bespoke `sp-breakdown` string in favour of the shared panel."* **That is the wrong goal.**

The wizard's prose — *"the bonus derives from your Crawler type, so it follows a type change automatically"* — teaches a **rule**, not the arithmetic. Guided Creation exists to teach while it enforces (ADR-021), so replacing it with a hover panel would trade a visible explanation for a hidden one in the mode that most needs the visible version.

The underlying requirement — that the breakdown comes from the shared derivation and can't drift — is **already satisfied**: `CrawlerStatsStep` calls `crawlerMaxSPParts` directly. No change made; recorded in the plan so nobody "fixes" it later from the original DoD.

## Verification

`bun run check:all` green: lint, format, typecheck, **4,528 tests**, validate, knip, audit, tokens, styling.

## Phase B complete

| Surface | Provenance |
|---|---|
| Live Sheet (pilot/mech/crawler) | panel, with revert |
| Dashboard instruments | panel, read-only |
| Frozen snapshot | panel, no revert |
| Partner cards | text summary |
| Wizard | inline teaching prose (already shared-derivation) |

Next: **C1–C4** (#598) — the converged contribution model, which is what makes each addend nameable and turns "Installed systems & modules +10" into "Heat Sink ×2, Composite Armour ×1".

Refs #597.

### DIFF
--- a/apps/itun/src/components/sheet/PartnerCard.tsx
+++ b/apps/itun/src/components/sheet/PartnerCard.tsx
@@ -42,6 +42,7 @@ import type { PartnerWithHost } from '../../lib/partnerLookup'
 import {
   partnerCap,
   partnerDerivedStats,
+  partnerDerivedStatsParts,
   partnerTechLevel,
   resolvePartnerStatBlock,
 } from '../../lib/rules/partnerStats'
@@ -54,6 +55,7 @@ import { PartnerHold } from './PartnerHold'
 import { partnerDisplayName } from './partnerDisplay'
 import { resolveModule, resolveSystem } from './mechItemRules'
 import { LIVE_SHEET_MANUAL } from '../../stores/surfaceProvenance'
+import { summarizeBreakdown } from 'component-lib'
 
 type PartnerCardProps = {
   /**
@@ -101,6 +103,17 @@ export function PartnerCard({
   const statBlock = resolvePartnerStatBlock(partner)
   const techLevel = partnerTechLevel(partner, crawlerTechLevel)
   const max = partnerDerivedStats(partner, techLevel)
+  // Tech-level scaling is a partner's whole story and nothing on screen
+  // explained it (ADR-029). `Stat` cells carry an editable stepper cluster, so
+  // the popover panel would nest a button inside a button — these render the
+  // derivation through the existing hoverText tooltip until Stat grows its own
+  // non-nesting affordance.
+  const maxParts = partnerDerivedStatsParts(partner, techLevel)
+  const scalingHover = (key: keyof typeof maxParts): string =>
+    summarizeBreakdown(maxParts[key], {
+      base: 'Base',
+      installed: `Tech ${techLevel} scaling`,
+    })
 
   const cargo = usePartnerCargo({
     found,
@@ -138,6 +151,7 @@ export function PartnerCard({
       label: 'SP',
       value: sp,
       outOfMax: max.structurePoints,
+      hoverText: scalingHover('structurePoints'),
       onChange: editable
         ? (next) => patch({ currentSP: clamp(next, max.structurePoints) })
         : undefined,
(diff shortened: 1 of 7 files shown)


alxjrvs (commented):
Review complete for `parity-provenance-surfaces` (single commit `6a0ab92c`, 7 files vs `origin/main`; no uncommitted changes).

**Findings (3):**

- `apps/itun/src/components/sheet/PartnerCard.tsx:113` — The partner stat derivation is exposed only via `Stat`'s `hoverText`, i.e. mouse-hover only. I probed the render: the Base UI tooltip trigger comes out as a bare `<span class id data-base-ui-tooltip-trigger>` with no `tabindex`, `title`, or `aria-describedby`, so keyboard, touch and screen-reader users get no access to the explanation at all — on a mobile-first PWA, and in direct contradiction of `StatProvenance`'s own stated rule that hover alone is not enough. The button-nesting concern the code comment cites is avoidable the way `VitalGauge` avoids it (trigger *beside* the stepper, not wrapping it).
- `packages/component-lib/src/components/stat/provenanceLines.ts:91` — `summarizeBreakdown` always emits an equation, so with no scaling it yields `Base 1 = 1` (or `Base 0 = 0`). That is not an edge case: only 3 equipment records carry `bonusPerTechLevel`, drones.json carries none, and `steps = 0` at crawler Tech 1 — so every mech-granted drone and every fresh Tech-1 build gets tautological tooltips. Its sibling `linesFromBreakdown` deliberately omits zero lines for this exact reason.
- `apps/itun/src/lib/rules/partnerStats.ts:148` — `partnerDerivedStatsParts` duplicates `partnerDerivedStats` instead of being its source, and they already diverge (parts floors `derived` at 0, the scalar does not). Latent today (no negative data), but the card's number and its tooltip come from two separate implementations; everywhere else in this system the scalar is a thin `Parts(...).total` wrapper.

**Verified clean:** value semantics are unchanged in `ShareSnapshotScreen` (`*Parts(...).total` is exactly the old scalar in every case); provenance labels match the live-sheet wiring from #613 verbatim; the recipient-facing frozen view (`/s/$id` → `SnapshotSheet` → `Sheet readOnly`) already carries provenance, so the preview change closes the gap rather than leaving one; the plan doc's B6 claim checks out (`CrawlerStatsStep` in component-lib does call `crawlerMaxSPParts` directly and does carry the quoted prose). `bun --filter component-lib test` (639 pass), `bun --filter itun test` (1288 pass), `bun run typecheck`, `lint` and `knip` all green.

PR #615, opened 2026-07-28
feat(reference): abilities can declare stat contributions — Beefcake now applies

alxjrvs:
**C1 + C2 + D2** of the rules-parity plan (#597, #598). This is the gap the whole project was named for.

An ability could declare **no mechanical change at all**. Four records whose rules text states a flat stat change were inert prose, so the numbers on the sheet were simply wrong for anyone holding them:

| Ability | Text | Now applies |
|---|---|---|
| **Beefcake** | "+3+X Max SP, +6 Cargo … +2 Max HP, +4 Inventory" | ✅ |
| **Bionic Arms** | "+2 Pilot Max HP" | ✅ |
| **Bionic Legs** | "+2 Pilot Max HP" | ✅ |
| **Modular Face Implant** | "your Pilot gains a Module Slot" | ✅ |

## Two shapes the rules forced

`ContributionSchema` is a `stat`, an `amount`, and a `target` — a superset of `MechStatBonusSchema`'s four mech-only keys, because Bionic Legs changes a **pilot** stat and Modular Face Implant changes a **slot count**.

- **`target`** exists because Beefcake is a *pilot* ability that raises the *piloted mech*'s Max SP and Cargo **while also** raising the pilot's own Max HP and Inventory. One record, two targets.
- **`amount: { flat, perTechLevel }`** exists because Beefcake's Max SP is *"3+X (where X is the Mech's Tech Level)"*. A consumer that can't supply a tech level resolves to the flat part — so an unknown tech level **under-counts visibly** rather than inventing a number.

Resolution keeps each contribution's **source** attached, which is what will let the provenance panel say `Beefcake +7` instead of today's anonymous aggregate.

## A double-count this surfaced — exactly where the ADR predicted

**Nell** (Eldridge Coast) carried `maxHpModifier: 8` where every other pregen carries `4`. The extra 4 was **Bionic Arms + Bionic Legs, hand-encoded** because the app couldn't apply them. With the abilities live, keeping the 8 double-counts — so it returns to 4 and she's now consistent with her peers.

This is the precise hazard that made *"convert every modifier into a pin"* the wrong migration, and it was caught by **an existing test**, not by inspection. Good argument for the no-migration split having been the right call.

## Scope note

The trait/damage/range half (**C3**) stays with `resolveChoiceView`. The two are declared side by side rather than merged, because they resolve at different layers — stats at derivation time, effects at dataview time.

## Verification

`bun run check:all` green: schemas, lint, format, typecheck, **4,542 tests**, validate, knip, audit, tokens, styling.

14 new tests run against the **real dataset**, so a regression in the data fails here — including that Beefcake targets pilot and mech from one record, and that its SP scales 4 → 8 across tech levels 1 → 5.

Refs #597, #598.

### DIFF
--- a/apps/itun/src/components/dashboard/ActiveItemBand.tsx
+++ b/apps/itun/src/components/dashboard/ActiveItemBand.tsx
@@ -82,6 +82,7 @@ export function ActiveItemBand({ mech, pilot, crawler, store }: ActiveItemBandPr
       mech={mech}
       store={s}
       hasPilot={pilot !== null}
+      pilotAbilities={pilot?.abilities}
       onDismount={() => setMount('pilot')}
     />
   )
@@ -135,18 +136,22 @@ function MechBand({
   mech,
   store,
   hasPilot,
+  pilotAbilities,
   onDismount,
 }: {
   mech: Mech
+  /** Beefcake raises the piloted MECH's Max SP and Cargo (ADR-029). */
+  pilotAbilities?: string[]
   store: PlayStore
   hasPilot: boolean
   onDismount: () => void
 }) {
   const chassis = resolveChassisRef(mech.chassisRef)
-  const maxSP = mechMaxSP(mech, chassis)
+  const piloting = { abilities: pilotAbilities }
+  const maxSP = mechMaxSP(mech, chassis, piloting)
   const maxEP = mechMaxEP(mech, chassis)
   const maxHeat = mechMaxHeat(mech, chassis)
-  const maxCargo = mechMaxCargo(mech, chassis)
+  const maxCargo = mechMaxCargo(mech, chassis, piloting)
   const sp = Math.min(mech.currentSP ?? maxSP, maxSP)
   const ep = Math.min(mech.currentEP ?? maxEP, maxEP)
   const heat = Math.min(mech.currentHeat ?? 0, maxHeat)
@@ -174,7 +179,7 @@ function MechBand({
   function doPush() {
     const m = fresh()
     const cap = mechMaxHeat(m, chassis)
-    const spMax = mechMaxSP(m, chassis)
+    const spMax = mechMaxSP(m, chassis, piloting)
     const { patch, effect, nextHeat, meltdown } = pushPatch({
       heat: Math.min(m.currentHeat ?? 0, cap),
       heatCap: cap,
@@ -188,7 +193,7 @@ function MechBand({
   function doHeatCheck() {
     const m = fresh()
     const cap = mechMaxHeat(m, chassis)
-    const spMax = mechMaxSP(m, chassis)
+    const spMax = mechMaxSP(m, chassis, piloting)
     const { patch, effect, meltdown } = heatCheckOncePatch({
       heat: Math.min(m.currentHeat ?? 0, cap),
       currentSP: Math.min(m.currentSP ?? spMax, spMax),
@@ -215,7 +220,7 @@ function MechBand({
     // Damage operates on the stored SP (authoritative); default to max only
     // when it was never set. The gauge, not this write, clamps for display.
     const { patch, effect } = mechDamagePatch({
-      currentSP: m.currentSP ?? mechMaxSP(m, chassis),
+      currentSP: m.currentSP ?? mechMaxSP(m, chassis, piloting),
       amount: dmg,
       vulnerable: m.vulnerable ?? false,
     })
--- a/apps/itun/src/components/dashboard/ActiveItemBand.tsx
+++ b/apps/itun/src/components/dashboard/ActiveItemBand.tsx
@@ -37,6 +37,7 @@ import { usePlayStateStore } from '../../stores/playStateStore'
 import { ActiveItemBand as ActiveItemBandView, CountStepper, StorageBay } from 'component-lib'
 import type { ActiveItemBandView as ActiveItemBandViewModel, BandButton } from 'component-lib'
 import { DASHBOARD_TXN } from '../../stores/surfaceProvenance'
+import { pilotingContext } from '../../lib/rules/pilotingContext'
 import {
   VENT_PATCH,
   critDamagePatch,
@@ -147,7 +148,7 @@ function MechBand({
   onDismount: () => void
 }) {
   const chassis = resolveChassisRef(mech.chassisRef)
-  const piloting = { abilities: pilotAbilities }
+  const piloting = pilotingContext(mech, pilotAbilities)
   const maxSP = mechMaxSP(mech, chassis, piloting)
   const maxEP = mechMaxEP(mech, chassis)
   const maxHeat = mechMaxHeat(mech, chassis)
--- a/apps/itun/src/components/dashboard/dialItems.ts
+++ b/apps/itun/src/components/dashboard/dialItems.ts
@@ -78,10 +78,18 @@ function pilotItem(pilot: Pilot): DialItem {
   }
 }
 
-function mechItem(mech: Mech): DialItem {
+/**
+ * Beefcake is a PILOT ability that raises the piloted MECH's Max SP and Cargo
+ * (ADR-029), so the mech dial cannot be derived from the mech alone.
+ */
+function mechItem(mech: Mech, pilot: Pilot | null): DialItem {
   const chassis = resolveChassisRef(mech.chassisRef)
-  const spParts = mechMaxSPParts(mech, chassis)
-  const heatParts = mechMaxHeatParts(mech, chassis)
+  const piloting = {
+    abilities: pilot?.abilities,
+    techLevel: typeof chassis?.techLevel === 'number' ? chassis.techLevel : undefined,
+  }
+  const spParts = mechMaxSPParts(mech, chassis, piloting)
+  const heatParts = mechMaxHeatParts(mech, chassis, piloting)
   const maxSP = spParts.total
   const maxHeat = heatParts.total
   const chassisLabel = `${chassis?.name ?? mech.chassisRef} chassis`
@@ -165,7 +173,7 @@ export function dialItems(args: {
   ]
   // The counterpart entities — everything NOT in the active row. In Downtime
   // the crawler is active, so the dial carries the mech + pilot instead.
-  if (mount !== 'mech') items.push(mechItem(mech))
+  if (mount !== 'mech') items.push(mechItem(mech, pilot))
   if (mount !== 'pilot' && pilot) items.push(pilotItem(pilot))
   if (mount !== 'downtime' && crawler) items.push(crawlerItem(crawler))
   items.push({
--- a/apps/itun/src/components/dashboard/dialItems.ts
+++ b/apps/itun/src/components/dashboard/dialItems.ts
@@ -25,6 +25,7 @@ import type { Pilot } from '../../lib/schemas/pilot'
 import type { MountState } from '../../stores/playStateStore'
 import type { DialItem as DialCellItem } from 'component-lib'
 import { linesFromBreakdown } from 'component-lib'
+import { pilotingContext } from '../../lib/rules/pilotingContext'
 
 export type DialItem = DialCellItem & { kind: DialKind }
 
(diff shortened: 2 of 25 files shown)


alxjrvs (commented):
Reviewed commit `71342ab7` (`feat(reference): abilities can declare stat contributions`) against `origin/main` — 13 files, +712/−33. Package tests (942) and the ITUN eldridge seed tests pass; the defects are in the wiring, not the new pure module.

The core theme: `MechSheet` is the **only** surface that passes the new `PilotingContext`, so a Beefcake pilot's mech now has two different maxima depending on where it is rendered — and one of those surfaces writes back.

Findings:

- `apps/itun/src/components/dashboard/ActiveItemBand.tsx:177` — Dashboard push/heat-check/damage compute `mechMaxSP(m, chassis)` with no piloting context, then clamp `currentSP` to it; a shutdown result writes `patch.currentSP` derived from the stale 20 instead of 27, silently destroying 7 SP (also 191, 218).
- `apps/itun/src/components/sheet/railStats.ts:50` — mech maxima disagree across surfaces: rail chips, `dialItems.ts:83`, `ActiveItemBand.tsx:146` and the frozen snapshot `ShareSnapshotScreen.tsx:492` all still show the un-piloted max.
- `apps/itun/src/components/sheet/PilotSheet.tsx:318` — ability HP lands in the `installed` bucket, which is labelled "Injuries (rules A11)"; Nell's ledger reads "Injuries +4", and injuries + abilities that cancel make the line vanish entirely.
- `apps/itun/src/components/sheet/MechSheet.tsx:222` — same root cause on the mech: Beefcake's +7 is credited to "Installed systems & modules" on a mech with no such systems; `derivedStats.ts:269` throws away the `source` that `contributions.ts` exists to preserve.
- `apps/itun/src/lib/eldridgeCoast/eldridgeCoast.ts:241` — the seed's hand-encoded ability HP was corrected, but users' saved IndexedDB pilots were not; no migration, so hand-encoded bonuses now double-count.
- `apps/itun/src/lib/cargo/useCargo.ts:91` — Beefcake's +6 Cargo is declared, derived and tested but never applied: the only cargo consumer calls `mechMaxCargo(mech)` bare.
- `apps/itun/src/components/sheet/MechSheet.tsx:187` — tech level re-resolves `mech.chassisRef` and ignores the `chassis` override prop, so an injected chassis gets the flat +3 instead of 3+X (plus a duplicate ORM lookup; the comment block above it is also duplicated/contradictory).
- `packages/salvageunion-reference/data/abilities.json:601` — `inventorySlots`/`moduleSlots` contributions are read by no derivation, and collide with the existing `maxInventorySlotsModifier` hand-edit convention documented at `pilotInventory.ts:84`.
- `apps/itun/src/lib/eldridgeCoast/__tests__/eldridgeCoast.test.ts:246` — the expectation is now computed with the function under test, so reverting Nell to the double-counted 8 still passes.
- `packages/salvageunion-reference/lib/rules/contributions.ts:90` — linear `Abilities.find()` per ref per render bypasses the memoized ref maps `resolveRefs.ts` added for exactly this hot path.

alxjrvs (commented):
Reviewed `origin/main...HEAD` (2 commits: the ADR-029 contribution model in `salvageunion-reference`, plus the ITUN wiring). The model itself is sound — `resolveAmount`, `abilityContributions` and `sumContributions` are pure, tolerant of unresolvable refs, and correctly under-count rather than guess on an unknown tech level. The defects are concentrated in the *wiring*: the "apply Beefcake everywhere" commit missed several call sites, and the two contexts that were threaded were threaded inconsistently.

Findings (12):

- `apps/itun/src/components/sheet/CrawlerSheet.tsx:136` — `useCargo` not given `pilotAbilities`; the mech's cargo cap on the crawler side omits Beefcake's +6 and **blocks legal depot→mech transfers**, since that cap gates `cargoTransfer`.
- `apps/itun/src/components/dashboard/ActiveItemBand.tsx:150` — `piloting` has no `techLevel`, so Max SP is 23 not 27; the under-counted `spMax` is then clamped into `heatCheckOncePatch`/`pushPatch`/damage, **persisting 4 lost SP** on a single Dashboard heat check.
- `apps/itun/src/components/sheet/SheetMech.tsx:45` — same missing `techLevel`; the condensed strip reads x/23 while the MechSheet body directly below reads x/27, the exact disagreement its new comment claims to prevent.
- `apps/itun/src/lib/rules/downtime.ts:303` — Downtime "restore mechs" writes `currentSP = mechMaxSP(mech)` with no piloting context, restoring a Beefcaked mech to 20/27.
- `apps/itun/src/components/sheet/PilotSheet.tsx:318` — ability HP lands in `StatBreakdown.installed`, which the pilot ledger labels "Injuries · rules A11": Bionic Arms+Legs renders as "Injuries +4", and a major injury (−2) plus Bionic Arms (+2) cancels to 0 so the injury line disappears entirely. Same at `dialItems.ts:64` and `ShareSnapshotScreen.tsx:470`.
- `apps/itun/src/lib/eldridgeCoast/eldridgeCoast.ts:241` — Nell's seed fixed 8→4, but `DB_VERSION` stays at 12 and nothing migrates already-persisted records; existing users double-count (22 HP instead of 18), as does any pilot who hand-entered the old workaround modifiers.
- `apps/itun/src/components/sheet/railStats.ts:50` — `mechRailItems` still derives with no piloting context; the rail chip shows "SP 27/20".
- `apps/itun/src/components/sheet/MechSheet.tsx:222` — Beefcake's +7 renders under "Installed systems & modules" on a mech with an empty loadout. `ResolvedContribution.source` is computed then discarded by `sumContributions`, so the provenance goal stated in `contributions.ts` is not actually met.
- `packages/salvageunion-reference/data/abilities.json:1103` — `moduleSlots`/`systemSlots` have no consumer anywhere; Modular Face Implant is still inert despite the schema comment and a passing test.
- `apps/itun/src/components/sheet/MechSheet.tsx:187` — `chassisTl` comes from a second `resolveChassisRef` call that ignores the `chassis` prop, so base stat and tech level can come from different chassis (and it duplicates a lookup per render).
- `apps/itun/src/lib/eldridgeCoast/__tests__/eldridgeCoast.test.ts:246` — expected value is now computed with the code under test; the ability half of the assertion is a tautology and detects no regression.
- `packages/salvageunion-reference/lib/rules/contributions.ts:90` — linear `.find()` per ref per stat, bypassing the memoized ref maps in `resolveRefs.ts` that were added precisely to kill this per-render scan; ~4 calls per mech render × 10-12 abilities.

The `ShareSnapshotScreen` known-limitation comment is a reasonable, well-documented deferral and I did not flag it.

alxjrvs (commented):
## Review: PR #615 — ability stat contributions (Beefcake)

**Scope note:** the local worktree was reset back to `origin/main` mid-review by another process, so I reviewed the authoritative PR head from the remote (`origin/main...9c11553b`, 4 commits). That head already includes fixes for two problems I'd found in the earlier commits (missing `techLevel` in the piloting context — now centralized in `lib/rules/pilotingContext.ts`; and ability bonuses being folded into the anonymous `installed` aggregate so the provenance panel labelled "Beefcake +7" as installed hardware — now carried in a named `sources` array). Those are not reported below.

Typecheck passes; `salvageunion-reference` (950) and `itun` (1290) test suites are green.

### Findings

- `apps/itun/src/components/sheet/ShareSnapshotScreen.tsx:157` — the pilot context published with a mech snapshot is read non-reactively via `useEntityStore.getState()`, but the share route loader hydrates only the shared kind. On a direct load/refresh of `/sheet/mech/<id>/share`, `softLinks` and `pilots` are empty on first render, the component's only store subscription (`useEntity('mech', id)`) never fires on their later hydration, and the payload is published with no `context` — the exact under-count the feature exists to fix.
- `apps/itun/src/components/sheet/railStats.ts:49` — `mechRailItems` still derives mech maxima with no piloting context, so the same mech reads SP 20/20 in the pilot's Linked Units rail (`SheetPilot.tsx:78`) and in docked-mech rows (`SheetCrawler.tsx:193`) while its own sheet reads 27/27. The pilot is in scope at both call sites.
- `apps/itun/src/lib/rules/downtime.ts:303` — `downtimeMechPatch` restores `currentSP`/`currentEP` from `mechMaxSP(mech)` with no piloting context. This is a *persisted* write, so a Downtime "full restore" on a Beefcake pilot's TL4 mech writes 20 and the sheet then shows 20/27.
- `apps/itun/src/lib/eldridgeCoast/eldridgeCoast.ts:241` — no migration for already-persisted records. Users who seeded the Eldridge Coast demo before this change have Nell stored with `maxHpModifier: 8` and the bionic abilities, now deriving 22 Max HP instead of 18; the same shape applies to Beefcake + `maxInventorySlotsModifier`, and `lib/schemas/pilot.ts:267-281` still tells users to hand-enter those abilities there.
- `packages/component-lib/src/components/stat/provenanceLines.ts:104` — `summarizeBreakdown` was not updated alongside `linesFromBreakdown` for the new `sources`, so a sourced breakdown yields "Atlas chassis 20 = 27" with the +7 missing. Latent (only `PartnerCard` consumes it, always with empty `sources`).

alxjrvs (commented):
## Review: PR #615 — `feat(reference): abilities can declare stat contributions`

**Scope note:** the working tree was switched out from under this session mid-review (another process checked out `parity-audit`), and the local `main` ref is stale. I pinned the review to the PR's real range, `285d8739...9c11553b` (4 commits, 25 files), and read all file contents via `git show 9c11553b:<path>` rather than the checkout.

The core design is sound — `ContributionSchema` / `abilityContributions` / `StatBreakdown.sources` is a clean, well-tested model, and keeping named `sources` out of the anonymous `installed` aggregate is the right call. The problems are all in the *wiring*: the derivation gained a third parameter and several call sites never got it, which is the same failure the PR's own `pilotingContext.ts` doc says the helper was created to prevent.

### Findings

- `apps/itun/src/components/sheet/ShareSnapshotScreen.tsx:157` — Snapshot pilot context is resolved from a non-reactive `useEntityStore.getState()` on `softLink`/`pilot`, neither of which the share route's loader hydrates; on a direct load/refresh the composition resolves `pilot: null` and the published payload omits `context.pilotAbilities` — the feature the commit adds silently no-ops.
- `apps/itun/src/components/sheet/railStats.ts:50` — `mechRailItems` calls `mechMaxSP(mech)` with no piloting context, so `SheetPilot.tsx:78` and `SheetCrawler.tsx:193` show "SP x/20" for a mech whose own sheet shows "SP x/27".
- `apps/itun/src/lib/rules/downtime.ts:303` — Downtime "Restore Mechs" writes `currentSP = mechMaxSP(mech)` without the pilot, so a Beefcake pilot's mech is persisted at 20/27 after a full restore and can never be topped up through Downtime.
- `apps/itun/src/components/sheet/ShareSnapshotScreen.tsx:521` — the publish preview derives without the pilot context that the payload now carries, so it matches neither the source sheet nor the resulting `/s/<id>` page.
- `packages/salvageunion-reference/lib/rules/derivedStats.ts:64` — no migration for stored records that hand-encoded these abilities in `maxHpModifier` / `maxInventorySlotsModifier` / `maxSpModifier`; they now double-count. The PR fixes exactly this for Nell in the Eldridge seed but not for user data in IndexedDB.
- `packages/component-lib/src/components/stat/provenanceLines.ts:104` — `summarizeBreakdown` was not updated alongside `linesFromBreakdown`; it omits `sources` from its terms while still printing `= derived`, producing arithmetic that doesn't add up. Latent today (only `PartnerCard` calls it, and partner breakdowns hard-code `sources: []`).
- `apps/itun/src/lib/cargo/useCargo.ts:98` — hand-builds `{ abilities }` instead of calling `pilotingContext(...)`, so no `techLevel`; harmless while cargo contributions are flat, but it violates the single-assembly-point invariant the helper documents.
- `packages/salvageunion-reference/lib/rules/contributions.ts:90` — linear `Abilities.find()` per ref, per stat, per render (~6–9k comparisons per sheet render with a 9-ability pilot), re-introducing the cost `resolveRefs.ts`'s memoized ref-map was written to eliminate.
- `apps/itun/src/components/pilot/PilotWizard.tsx:498` — `pilotInventoryCapacity()` called with no argument, so the wizard shows `/6` while the saved sheet shows `/10` for a Beefcake pilot.

The common thread for the first three (plus #4 and #7): `pilotingContext` is optional at the derivation boundary, so every miss degrades quietly. Making the mech derivations take a required context (or at minimum adding a lint/test that enumerates the call sites) would convert this whole class from silent under-count to compile error.

PR #625, opened 2026-07-28
fix(observability): make the nightly probe trustworthy — retries + name it in the report

alxjrvs:
Two defects the **first real nightly run** exposed in the probe I shipped. Both
share a theme: an alarm you can't trust is worse than no alarm, which is exactly
the condition this work set out to fix.

## 1. It cried wolf on ordinary internet weather

The run reported:

```
✗ [srd] production unreachable (https://salvageunion.io):
      The socket connection was closed unexpectedly
```

against a site that was demonstrably up seconds earlier and is up now. One
dropped TCP connection from the CI runner, and the probe declared production
dark — **and this job's failure opens a tracking issue.**

Both the page fetch and every chunk fetch now go through `fetchWithRetry`:
4 attempts, exponential backoff (1s, 2s, 4s — ~7s worst case, measured against
an unresolvable host). It retries only what is plausibly transient (network
errors, 5xx); a 4xx is a real answer from a working server and returns
immediately, so a genuinely missing page still fails fast.

This does not weaken the check — a production that is actually dark fails all
four attempts and still reports. Only the single-blip false positive goes away.

## 2. The issue it opens never said the probe was why

`observability-probe` was wired into `needs:` and into `notify-failure`'s
condition, so it can file the issue on its own — but it was never added to the
results list the issue body prints. The real run showed it:

```
- e2e-itun: failure
- e2e-srd: failure
```

No mention of the probe, though it had also failed. And in the case that matters
most — **production dark while both suites pass** — the issue would have been
filed stating every job succeeded.

The probe now gets its own line, and is omitted when skipped (variable off), so
the report never lists a job that didn't run. Verified across all three states:

| state | reported |
|---|---|
| probe skipped | `e2e-itun`, `e2e-srd` |
| probe failed, suites green | `e2e-itun`, `e2e-srd`, **`observability-probe: failure`** |
| all failed | all three |

## Verification

- `validate:observability:live` passes against current production (SDK present
  + served CSP permits the DSN, both sites).
- Backoff measured: ~7s against an unresolvable host, matching 1+2+4.
- Workflow YAML parses; `notify-failure` env now carries `PROBE_RESULT`.
- `bun run check:all` green.

### DIFF
--- a/.github/workflows/e2e-nightly.yml
+++ b/.github/workflows/e2e-nightly.yml
@@ -295,6 +295,7 @@ jobs:
         env:
           ITUN_RESULT: ${{ needs.e2e-itun.result }}
           SRD_RESULT: ${{ needs.e2e-srd.result }}
+          PROBE_RESULT: ${{ needs.observability-probe.result }}
         with:
           script: |
             const label = 'nightly-e2e-failure'
@@ -303,9 +304,18 @@ jobs:
             // always means "hit timeout-minutes", which reads very differently
             // from 'failure' (assertions failed) and points at a hang rather
             // than a broken expectation.
+            // observability-probe MUST appear here, not just in `needs:` and the
+            // `if:`. It can open this issue on its own, and without a line of its
+            // own the issue would read "e2e-itun: success, e2e-srd: success" while
+            // still being filed — an alarm that refuses to say what tripped it.
+            // Omitted entirely when skipped (the OBSERVABILITY_PROBE variable is
+            // off), so the report never lists a job that did not run.
             const results = [
               ['e2e-itun', process.env.ITUN_RESULT],
               ['e2e-srd', process.env.SRD_RESULT],
+              ...(process.env.PROBE_RESULT && process.env.PROBE_RESULT !== 'skipped'
+                ? [['observability-probe', process.env.PROBE_RESULT]]
+                : []),
             ]
             const body = [
               `Nightly E2E (\`e2e-nightly.yml\`) did not pass on \`${context.ref}\`.`,
--- a/tools/check-observability.ts
+++ b/tools/check-observability.ts
@@ -179,6 +179,37 @@ function checkStatic(app: BrowserApp): void {
   }
 }
 
+/**
+ * Fetch, retrying transient network failures.
+ *
+ * This probe talks to the public internet from a CI runner, where a dropped
+ * socket or a momentary 5xx is ordinary weather — and a failure here OPENS A
+ * TRACKING ISSUE. The first real nightly run proved the point: it reported
+ * "production unreachable … The socket connection was closed unexpectedly"
+ * against a site that was demonstrably up seconds earlier.
+ *
+ * A probe that cries wolf teaches you to stop reading it, which is precisely
+ * the failure this file exists to prevent — so a verdict of "production is
+ * dark" has to survive several attempts before it is worth waking anyone.
+ *
+ * Retries only what is plausibly transient: network errors and 5xx. A 4xx is a
+ * real answer from a working server and is returned immediately.
+ */
+async function fetchWithRetry(url: string, attempts = 4): Promise<Response> {
+  let lastError: unknown = new Error('no attempt made')
+  for (let attempt = 0; attempt < attempts; attempt++) {
+    try {
+      const res = await fetch(url, { redirect: 'follow' })
+      if (res.status < 500) return res
+      lastError = new Error(`HTTP ${res.status}`)
+    } catch (error) {
+      lastError = error
+    }
+    if (attempt < attempts - 1) await Bun.sleep(1000 * 2 ** attempt)
+  }
+  throw lastError
+}
+
 /** Absolute-ises every `<script src>` the served HTML references. */
 function scriptUrls(html: string, origin: string): string[] {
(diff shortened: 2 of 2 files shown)


alxjrvs (commented):
Reviewed the single-commit PR (`2cb8aa72`, +33/-2 in `tools/check-observability.ts` — `main..HEAD` is misleadingly large only because `main` is behind; upstream `origin/probe-retries` equals HEAD and the working tree is clean).

The retry helper itself is internally sound — the loop bounds, backoff guard, and 4xx fast-path are all correct. The problem is where the retry boundary was drawn: it wraps `fetch()` but not the body read, and `fetch()` resolves at headers.

Findings:

- `tools/check-observability.ts:235` — Retry covers headers only, not the body read. `html = await res.text()` sits outside `fetchWithRetry`, so a socket dropped mid-body still produces `production unreachable … The socket connection was closed unexpectedly` with zero retries — the exact message the commit body quotes as the thing being fixed.
- `tools/check-observability.ts:254` — Same shape in the chunk loop, worse consequence: `.text()` is outside the retry and throws into the bare `catch {}`, so a transient drop while reading the one chunk that carries the SDK yields a false `Sentry SDK absent from all 5 script(s) … Error tracking is DARK in production` — a louder false alarm than the one being removed.
- `tools/check-observability.ts:202` — No `AbortSignal.timeout` on the fetch. A hang never returns, so retries never fire; the job exhausts `timeout-minutes: 5` and reports `cancelled`, which `notify-failure` (gated on `== 'failure'`) ignores while `notify-recovery` (gated on `!= 'failure'`) will close an open tracking issue.
- `tools/check-observability.ts:230` — A persistent 5xx now throws instead of returning, so the dedicated `production fetch failed: HTTP <status>` branch is reachable only for 4xx and a real 503 outage is reported as "unreachable" (network diagnosis instead of server diagnosis).

Suggested fix for the first two: have `fetchWithRetry` return the consumed body (and needed headers) so fetch+read retry as one unit; add a timeout signal for the third.

alxjrvs (commented):
Reviewed the 2-commit diff on `probe-retries` (`tools/check-observability.ts` retry helper + naming the probe in the nightly failure report). The report-naming change itself is correct — `skipped` is properly excluded, `cancelled` is annotated, and the results array construction is sound. Two real issues, both around the new retry loop's interaction with the job's alarm plumbing:

- `tools/check-observability.ts:198` — `fetchWithRetry` sets no per-attempt timeout/`AbortSignal`, so a stalled socket (the exact failure this retry targets) now hangs 4× per URL plus 7s of backoff, across 2 apps × every `<script src>`. That blows the probe job's `timeout-minutes: 5`, which reports as `cancelled` — a result `notify-failure` deliberately does not match (`== 'failure'`). The anti-false-alarm change can therefore turn a genuine dark-production run into no alarm at all.
- `.github/workflows/e2e-nightly.yml:367` — `notify-recovery` fires on any probe result other than `failure`, so a `cancelled` (timed-out) probe closes the open `nightly-e2e-failure` issue with "passed again" despite nothing having been verified. The e2e jobs use `!= 'success'` for exactly this reason; the probe needs to distinguish `skipped` from `cancelled` instead of collapsing both into success.

alxjrvs (commented):
**Scope reviewed:** commits `2cb8aa72` + `9800c120` (branch `probe-retries`, parent `285d8739`) — `tools/check-observability.ts` (new `fetchWithRetry`) and `.github/workflows/e2e-nightly.yml` (name the probe in the failure report). No uncommitted changes. Note: this work already appears squash-landed on `origin/main` as `b09b9654`.

The workflow hunk itself is correct — `skipped` is the right sentinel to omit, and a `cancelled`/`failure`/`success` probe is included. The retry helper's retry policy (5xx + throws, 4xx returned immediately, 1/2/4s backoff) is also sound in isolation. The problems are all about what the retry *doesn't* cover.

Findings:

- `tools/check-observability.ts:235` — `fetchWithRetry` wraps only `fetch()`, not the body read. Bun resolves `fetch()` at headers and streams the body, so `await res.text()` (235) and `await (await fetchWithRetry(url)).text()` (254) are outside the retry. A socket closed mid-transfer — the exact "socket connection was closed unexpectedly" failure the new doc comment cites — still fails on attempt one and opens the tracking issue.
- `tools/check-observability.ts:262` — the chunk loop can't tell "couldn't read this script" from "SDK isn't in this script". `fetchWithRetry` returns any status < 500 without throwing (verified against a local 403 server: returns immediately, error-page body, no `/sentry/i` match), and exhausted retries are swallowed by the bare `catch {}` at 257. Either way the probe emits "Sentry SDK absent from all N script(s)… Error tracking is DARK in production" for a healthy site.
- `.github/workflows/e2e-nightly.yml:367` — a probe killed by `timeout-minutes: 5` reports `cancelled`; notify-failure only fires on `== 'failure'` and notify-recovery passes on `!= 'failure'`, so a timed-out probe stays silent *and* lets notify-recovery close the open tracking issue. The added backoff (no per-request timeout, up to ~7s/URL across 2 sites) makes that timeout likelier. This is the same `cancelled`-vs-`failure` hole the file's own comment at lines 268-273 documents having already broken the suite for days with `e2e-itun`.

PR #626, opened 2026-07-28
fix(itun): spend Dashboard EP/AP/SP from the full pool, not from zero

alxjrvs:
Follow-up to #588 — the same defect class, on the other live stats. Flagged in that PR's review as a single SP line; the sweep turned up **three**, and unlike #588 these are all **write** paths.

## The rule being violated

Across ~30 sites the app treats an unrecorded live stat as *unspent / undamaged* and falls back to the derived maximum. `ActionsDeck.tsx` was the only file defaulting to `0`:

| call | was | for an entity with the field unset |
|---|---|---|
| `pilotActivationPatch` | `currentAP ?? 0` | `Math.max(0, 0 - cost)` → stores **AP 0** |
| `activationPatch` | `currentEP ?? 0` | `Math.max(0, 0 - cost)` → stores **EP 0** |
| `pushPatch` | `currentSP ?? 0` | `performHeatCheck` Overheat → `max(0, 0 - heat)` → stores **SP 0** |

Because every *read* path defaults to the max, the record kept displaying full while holding `0`. The next read-modify-write made the zero permanent. The SP case is the worst: a mech that had never taken damage gets recorded one hit from destroyed after a single Push.

This only affects entities whose field was never written — a fresh entity, or one predating the field. An entity that has taken damage or spent EP has the value stored and was always handled correctly.

## Fix

All three fall back to the derived maximum (`mechMaxEP` / `mechMaxSP` / `pilotMaxAP`).

The stored value stays authoritative and is deliberately **not** clamped here. I tried `Math.min(stored ?? max, max)` first, to mirror `ActiveItemBand`'s display reads — it's wrong for a write path: an unresolved `chassisRef` derives a max of `0`, so it would zero a real stored value. `ActiveItemBand.applyDamage` already documents the correct rule ("Damage operates on the stored SP (authoritative); default to max only when it was never set").

## Tests

Three cases, each asserting the **written patch**, not the rendering.

They use a real chassis (Leviathan). This matters: the existing deck tests use `chassisRef: 'unknown-chassis'`, whose derived maxima are `0` — under that fixture the old and new code are indistinguishable, so a test written against it would pass either way.

The Push case can't inject a roll (`doPush` closes over `defaultRoll`, and this suite deliberately avoids `mock.module`), so it's set up so the dice barely matter: at Heat 17 on an 18-cap chassis every Push clamps to cap, overloading on any d20 but a natural 20, and ~40% of pushes write SP. It asserts every observed SP write equals `76 - 18 = 58` and that at least one occurred across 80 pushes — P(no write) ≈ 1e-18. Whole file runs in ~0.5s.

## Verification

`bun run check:all` exit 0 — lint, format, typecheck, 1291 itun tests, validate:all, knip, audit, token/styling.

One note on the test run: an earlier full-suite run showed 3 failures in `sheet-soft-warnings` and `ShareSnapshotScreen`. They pass in isolation, pass on a re-run of the full suite, touch none of this code, and are timeouts under load — pre-existing flakes, not introduced here.

https://claude.ai/code/session_01LAxZfRPrgJq8U9mm4b2LzE

### DIFF
--- a/apps/itun/src/components/dashboard/ActionsDeck.tsx
+++ b/apps/itun/src/components/dashboard/ActionsDeck.tsx
@@ -25,7 +25,7 @@ import type { ActionsDeckView as ActionsDeckViewModel, DeckRow } from 'component
 import { CORE_ROLL_BANDS, describePushOutcome, performCoreRoll } from '../../lib/rules/coreMechanic'
 import type { CoreRollResult } from '../../lib/rules/coreMechanic'
 import { defaultRoll } from '../../lib/rules/heatCheck'
-import { mechMaxHeat } from '../../lib/rules/derivedStats'
+import { mechMaxEP, mechMaxHeat, mechMaxSP, pilotMaxAP } from '../../lib/rules/derivedStats'
 import { canActivateAction, resolveChassisRef } from 'salvageunion-reference/rules'
 import type { Mech } from '../../lib/schemas/mech'
 import type { Pilot } from '../../lib/schemas/pilot'
@@ -133,9 +133,14 @@ export function ActionsDeck({ mech, pilot, mount = 'mech', store }: ActionsDeckP
     if (effCurrency === 'AP') {
       if (!pilot) return
       const fresh = s.get('pilot', pilot.id) ?? pilot
+      // An unrecorded live stat means UNSPENT, not empty — a pilot who has never
+      // spent AP is at full AP. Defaulting to 0 here banked the spend against an
+      // empty pool and wrote AP 0 on the first activation. The stored value stays
+      // authoritative when present (same rule as ActiveItemBand's damage write):
+      // never clamp it here, since an unresolved ref makes the max 0.
       const patch = pilotActivationPatch({
         apCost: economy.epCost,
-        currentAP: fresh.currentAP ?? 0,
+        currentAP: fresh.currentAP ?? pilotMaxAP(fresh),
       })
       if (Object.keys(patch).length > 0) void s.update('pilot', pilot.id, patch, DASHBOARD_TXN)
       setActivated(true)
@@ -147,7 +152,8 @@ export function ActionsDeck({ mech, pilot, mount = 'mech', store }: ActionsDeckP
     const patch = activationPatch({
       slug: action.slug,
       economy,
-      currentEP: fresh.currentEP ?? 0,
+      // Unrecorded EP means a mech that has never spent any — full, not empty.
+      currentEP: fresh.currentEP ?? mechMaxEP(fresh, chassis),
       currentHeat: fresh.currentHeat ?? 0,
       heatCap,
       prevUses: fresh.itemUses,
@@ -186,7 +192,9 @@ export function ActionsDeck({ mech, pilot, mount = 'mech', store }: ActionsDeckP
     const { patch, effect, nextHeat } = pushPatch({
       heat: Math.min(fresh.currentHeat ?? 0, cap),
       heatCap: cap,
-      currentSP: fresh.currentSP ?? 0,
+      // Unrecorded SP means undamaged. At 0 an Overheat wrote the mech straight
+      // to SP 0 — one hit from destroyed — without it ever having taken damage.
+      currentSP: fresh.currentSP ?? mechMaxSP(fresh, chassis),
       roll: defaultRoll,
     })
     void s.update('mech', mech.id, patch, DASHBOARD_TXN)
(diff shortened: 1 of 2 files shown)


alxjrvs (commented):
Self-review (background agent) — verified against the tree, not just the diff.

**Completeness.** Swept every live-stat fallback in `apps/` and `packages/` (`currentSP|currentEP|currentHP|currentAP ??`). ~30 sites, all `?? max`; the only `?? 0` occurrences were the three in this file, and all three are gone. No other file dissents from the convention, so this closes the class rather than one instance.

**Why these are worse than #588.** #588 was a display fallback — wrong pixels, correct storage. These three feed `s.update(...)`, so the wrong number was persisted to IndexedDB. Traced each to its terminal write: `activationPatch` → `patch.currentEP`; `pilotActivationPatch` → `patch.currentAP`; `pushPatch` → `performHeatCheck` (`nextSP = max(0, currentSP - heat)`) → `heatCheckPatch` → `patch.currentSP` under `effect.shutdown`.

**The clamp I backed out of.** First pass used `Math.min(stored ?? max, max)` for symmetry with `ActiveItemBand`'s reads. Probing the ORM showed `resolveChassisRef('unknown-chassis')` is `null` and `mechMaxEP/mechMaxSP` return `0` for it — so that form would have clamped a stored 6 EP to 0 and broken the existing `Activate writes the EP-spend patch` test. Reverted to the plain fallback, which is what `ActiveItemBand.applyDamage` documents. Worth flagging because the "fix" would have been a data-loss bug wearing a consistency costume.

**Test honesty.** The existing suite's `unknown-chassis` fixture derives maxima of 0, making old and new behaviour identical under it — so the new tests use a real chassis, and each asserts the written patch rather than rendered text. The Push case is probabilistic by necessity (no roll seam, and this suite avoids `mock.module` by convention); it's engineered so P(false failure) ≈ 1e-18 and costs ~0.5s.

**Not fixed here, deliberately:** `heatCtx` (line 88) and `doPush` still use `Math.min(currentHeat ?? 0, cap)`, which under an unresolved chassis clamps Heat to 0. Harmless (understates Heat rather than destroying stored SP) and out of scope for this follow-up.

`bun run check:all` exit 0. The 3 failures seen in one intermediate full-suite run are pre-existing flakes in unrelated files — they pass in isolation and on re-run, and the new tests add ~0.5s total, so they are not load I introduced.

alxjrvs (commented):
## Review summary

**Scope:** the branch's single commit `824a6c7e` — `apps/itun/src/components/dashboard/ActionsDeck.tsx` (+16/−4) and its test file (+113). Local `main` was stale; the real base is `origin/main` (`285d8739`).

**Verdict:** the fix itself is correct. All three fallbacks resolve properly (`pilotMaxAP(fresh)`, `mechMaxEP(fresh, chassis)`, `mechMaxSP(fresh, chassis)` — `chassis` is in scope in both `activate` and `doPush`), the derivation functions are already floored at 0 so no extra `Math.max` is needed, and the new values now agree with every other read/write path in the app (`MechSheet.tsx:179-181`, `PilotSheet.tsx:447`, `ActiveItemBand.tsx:218`, `railStats.ts`). `currentHeat ?? 0` is correctly left alone — Heat genuinely starts at zero. Typecheck and lint are clean; the test file passes 16/16 across three runs.

The randomized 80-iteration Push test is not meaningfully flaky: only the Overheat band (11–19) writes SP via `heatCheckPatch`, meltdown writes only `destroyed`, and the stub store never applies patches — so every SP write is deterministically `76 − 18 = 58`, with P(zero writes) ≈ 4e-18.

### Finding

- `apps/itun/src/components/dashboard/ActionsDeck.tsx:197` — the deck's Push reads stored SP unclamped while `ActiveItemBand.doPush:181` clamps to the derived max, so the two Push buttons on the same Dashboard write different SP once a stored value exceeds the current maximum (reachable and durable: removing a `+5 Max SP` system at `MechSheet.tsx:307` writes no clamp, and `clampMechCurrentStats` has no call site anywhere in the app).

### Notes, not filed as findings

- The comment on the AP branch (lines 136–140) justifies not clamping with "an unresolved ref makes the max 0" — true for the mech paths, but `pilotMaxAP` is `base 5 + modifier` and resolves no ref, so that rationale doesn't apply to the pilot path it's attached to.
- The commit message's claim that "the record kept displaying full while holding 0" is inaccurate — once `0` was stored, `??` no longer fires and every gauge showed 0. The bug was visible, not silent. Doesn't affect the code.
- Pre-existing, outside the diff: `doPush` discards `pushPatch`'s `meltdown` flag (ActiveItemBand surfaces it as a player-confirm), and the deck's Push has no `heat + 2 > cap` lock that `ActiveItemBand` enforces per Quick Ref p.233.

PR #627, opened 2026-07-28
feat(reference): add the rules-parity audit — the gate that keeps content honest

alxjrvs:
**A2** of the rules-parity plan (#597) — and the check that *defines* when the translation is finished.

Every player-content record whose rules text states a mechanical change must either **encode** it as structured data or carry an explicit, **reasoned exemption**. Anything else is a record whose number the app gets wrong — and it can no longer be added silently. Wired into `validate:all`.

```
Rules parity: 30 claim(s) — 14 encoded, 14 exempt, 2 unresolved (2 known, burning down)
✅ Every stated mechanical change is encoded or reasoned.
```

It **never infers a number from prose**. It only refuses to let a stated change be forgotten; authoring the number stays a human decision.

## The backlog was overstated by an order of magnitude — this is what found it

I'd been calling ~21 trait/damage records "the C3/D3 backlog". Classified against ADR-029's own rule — *prose-only or conditional benefits get no bonus data* — most were **never encodable work**:

| Class | Examples | Why exempt |
|---|---|---|
| **target-applied** | Bio-Rifle, Servo Lasso, Hydraulic Shunter, Can't Stop Won't Stop | The effect lands on an **opponent** — procedural adjudication (ADR-021), surfaced never enforced |
| **conditional** | CACB Laser, Stabilising Locomotion System | Gated on state the app doesn't track ("consecutive turns", "whilst stabilised") — no flat value exists |
| **activated** | Snipe, Weapon Guidance | Per-use choice → Dashboard play state (F1) |
| **effect-of-an-effect** | Offensive Protocols | Raises another *system's* damage, not a stat |

That leaves **two** genuinely-encodable records — **Bio-Wings** (`"Your Mech gains the Fly Trait"`) and **High Gain Antenna** — both trait grants on the player's own side that can currently only be declared through `choiceOptions`. Closing them needs **C3** (effects declarable directly on a system/ability), not a data edit.

So the remaining translation work is far smaller and far better defined than the plan assumed.

## Design notes

- **Burn-down**, following `check:design-tokens`: the known set is tolerated so the gate can land, a **new** claim fails, and a **stale** known entry fails too — a burn-down list that never shrinks isn't burning down.
- **Adversary stat blocks** (npcs, creatures, squads, bio-titans) are out of scope *by construction* — their effects are adjudicated, not derived, so auditing them would demand data for rules the app deliberately doesn't automate.
- **Tests are fixture-driven**, not assertions about real records. A test saying "Beefcake is unencoded" would need rewriting the moment the backlog was worked off — exactly when it matters most.

## Verification

`bun run check:all` green: schemas, lint, format, typecheck, **4,563 tests**, validate (incl. the new parity check), knip, audit, tokens, styling.

Refs #597, #598.

### DIFF
--- a/docs/design/rules-parity-plan.md
+++ b/docs/design/rules-parity-plan.md
@@ -130,6 +130,39 @@ Sizes use the repo's existing effort labels: **S** under 1 day · **M** 1–2 ·
 | **F5** | Downtime writes                    | L    | F2–F4   | Downtime stops narrating and starts writing, with ADR-007 confirmation on destructive consequences                                |
 | **F6** | `RuleBrief` at enforcement moments | S    | —       | A disabled Push cites the Heat Cap rule rather than greying out silently                                                          |
 
+## Where the translation actually stands
+
+The parity audit (`validate:all` → `parity`) is the authoritative count, not this
+document. As of A2 landing:
+
+|                                            |       |
+| ------------------------------------------ | ----: |
+| Stated mechanical claims in player content |    30 |
+| Encoded as structured data                 |    14 |
+| Exempt, with a recorded reason             |    14 |
+| **Unresolved**                             | **2** |
+
+**The trait/damage backlog was overstated by an order of magnitude.** This plan
+called D3 "~21 records". Classified against ADR-029's own rule — prose-only or
+conditional benefits get no bonus data — most were never encodable work:
+_target-applied_ effects land on an opponent (procedural adjudication, ADR-021);
+_conditional_ ones are gated on state the app does not track; _activated_ ones
+are per-use and belong to Dashboard play state; _effect-of-an-effect_ ones raise
+another system's output rather than a stat. D3 as scoped does not exist.
+
+**The two that remain need a capability, not a data edit — and not the one C3
+described.** C3 was written as "effects declarable directly on a system/ability".
+Declaring them is not the blocker: `ChoiceEffectSchema` has no `target`, and
+neither remaining effect lands on the record that declares it. Bio-Wings grants
+**the host mech** the Fly Trait; High Gain Antenna raises the Range band of
+**other installed items**, filtered by trait. Encoding them without a target
+would apply real rules to the wrong entity — a silent wrong answer, which is
+worse than a visible gap.
+
+So the real C3 is: **give effects the same `target` concept `ContributionSchema`
+already carries.** Until then the audit holds both as known-unresolved and fails
+on anything new.
+
 ## Critical path
 
 ```
--- a/packages/salvageunion-reference/package.json
+++ b/packages/salvageunion-reference/package.json
@@ -55,6 +55,7 @@
     "validate:traits": "bun tools/validateTraits.ts",
     "validate:schemas": "bun tools/validateSchemas.ts",
     "validate:all": "bun tools/validate.ts",
+    "validate:parity": "bun tools/validateParity.ts",
     "validate:fix": "bun tools/validate.ts --fix",
     "edit-data": "bun tools/edit-data.ts",
     "typecheck": "tsc --project tsconfig.json --noEmit",
--- a/packages/salvageunion-reference/tools/_tmpdump.ts
+++ b/packages/salvageunion-reference/tools/_tmpdump.ts
@@ -1,0 +1,11 @@
+import { loadAllDataFiles } from './loadData.js'
+import { auditParity, PARITY_EXEMPTIONS } from './validateParityLogic.js'
+const bag = loadAllDataFiles()
+const f = auditParity(bag as never)
+for (const x of f)
+  console.log(
+    `${x.encoded ? 'ENC' : x.exempt ? 'EX ' : 'UNR'} | ${x.schema} | ${x.record} [${x.klass}] :: ${x.sentence.slice(0, 100)}`
+  )
+console.log('---dead exemptions---')
+for (const n of Object.keys(PARITY_EXEMPTIONS))
+  if (!f.some((x) => x.record === n)) console.log('DEAD:', n)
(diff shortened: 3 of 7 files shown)


alxjrvs (commented):
## Review scope

Against a freshly-fetched `origin/main`, the branch is **one commit** — `7a76f374 feat(reference): add the rules-parity audit`, 5 files / +469 lines (local `main` was 24 commits stale, which made `main...HEAD` look like a 156-file PR). Working tree clean.

The gate is well-shaped (logic module + thin CLI + unified-runner wiring, fixture-driven tests, burn-down list with a staleness check). All 8 new tests pass, `bun tools/validate.ts` passes, knip clean. Every finding below was reproduced by executing the real detector against the real dataset, not read off the diff.

## Findings

- `packages/salvageunion-reference/tools/validateParityLogic.ts:204` — Owner map is built from `record.actions`, but `chassis.json` has **zero** records with an `actions` array (all 51 use `chassisAbilities`), so all **60** `actionSource: "chassis"` actions are dropped. Nine state a mechanical change and are never flagged, incl. "Integrated Cargo Bay" (+10 Cargo Capacity) and "Integrated Monstrous Bio-Wings" (gains the Fly Trait — the same class the file lists as KNOWN_UNRESOLVED). The test that seems to cover this uses a record shape that doesn't exist in the data.
- `packages/salvageunion-reference/tools/validateParityLogic.ts:241` — `encoded` is presence-only and `seen` collapses all cap sentences per record into one finding, so any record already carrying `contributions`/`statBonus` auto-passes every further cap it claims. Executed: Max HP + Max AP text with only a `maxHp` contribution → gate green. `Bionic Arms`/`Bionic Legs` are one sentence away from this today.
- `packages/salvageunion-reference/tools/validateParityLogic.ts:232` — `seen`/exemption/known keys are bare names, not `schema::name`. 13 names exist in both `abilities.json` and `equipment.json` (Knife Missile, Auto-Turret, …); executed, the second record's unencoded claim never surfaces at all, and one exemption covers both.
- `packages/salvageunion-reference/tools/validateParityLogic.ts:212` — `ownerByActionName.set()` is last-write-wins; 15 action names have multiple owners ("System Repair" has 4). A shared action's claim is enforced against at most one owner.
- `packages/salvageunion-reference/tools/validateParityLogic.ts:69` — `KNOWN_UNRESOLVED` gets a staleness check, `PARITY_EXEMPTIONS` does not. 6 of its 20 entries already match nothing, so the list overstates what was actually reviewed, and a dead name-keyed entry silently pre-exempts any future record taking that name.
- `packages/salvageunion-reference/tools/validateParityLogic.ts:224` — `content ?? description ?? value ?? source` inverts coverage: 121/126 systems and 83/84 modules have no `content` so the whole object is scanned (ids, names, traits), while records that *do* have `content` never get siblings like crawler-bays' `damagedEffect` or chassis `patterns` scanned. Also the reason quoted sentences read `"paragraph This System increases…"`.

Common root: the audit keys and links content by **name** and by **`actions`**, while the dataset links some content by `chassisAbilities` and reuses names across files. Keying findings on `schema::name`, mapping actions via `actionSource` (or including `chassisAbilities`), and making `encoded` stat-aware would close five of the six.

alxjrvs (commented):
Reviewed the 2-commit PR (`origin/main...HEAD`, 516 added lines): the rules-parity audit tool, its wiring into `validate:all`, tests, and a plan-doc update. I ran the tool against the real dataset and reproduced each issue rather than reasoning from the diff alone.

**Findings**

- `packages/salvageunion-reference/tools/validateParityLogic.ts:204` — **Chassis content is never audited.** `chassis.json` is in `PLAYER_CONTENT`, but the owner map is built only from `record.actions`, while chassis link their actions via `chassisAbilities[]` (as `validateActionReferencesLogic.ts`, `validateOrphansLogic.ts`, and `lib/utilities.ts` all already do). 60 chassis-sourced actions have no owner; 9 state pattern-matching mechanical claims (incl. "Integrated Monstrous Bio-Wings … gains the Fly Trait" and the three Integrated Cargo Bay/Hold cargo increases). Knock-on: the three `'Integrated *'` exemptions are dead, the test at `validateParity.test.ts:104` "proves" them using a fixture shape that doesn't exist in the data, and the "30 claims" table added to `docs/design/rules-parity-plan.md` is an undercount.
- `packages/salvageunion-reference/tools/validateParityLogic.ts:232` — **Record identity is the bare name.** `seen`, `PARITY_EXEMPTIONS`, `KNOWN_UNRESOLVED`, and `ownerByActionName` are all name-keyed with no file qualifier. 13 names are duplicated across player-content files today. Demonstrated: an encoded `Wingsuit` in `abilities.json` silently swallows an unencoded `Wingsuit` in `equipment.json` — 1 finding, 0 unresolved, gate green. `ownerByActionName` is additionally last-writer-wins over sorted filenames, so for the 15 collided action names `equipment.json` always beats `abilities.json`.
- `packages/salvageunion-reference/tools/validateParityLogic.ts:172` — **`encoded` is field presence only.** `contributions: []` (or `statBonus: {}`) marks a claim encoded; verified. And because of the one-finding-per-record-per-class dedupe, a record that encodes one cap gets a free pass on every further cap claim in its text (a second "gain a Module Slot" sentence yields zero findings).
- `packages/salvageunion-reference/tools/validateParityLogic.ts:69` — **No staleness gate on `PARITY_EXEMPTIONS`**, unlike `KNOWN_UNRESOLVED`. 6 of 20 entries already match nothing (the three Integrated bays, plus `Parasitic Membrane`, `Mender`, `Pilot Bay`), so the list documents coverage that doesn't exist. Notably the committed `_tmpdump.ts` contains exactly this probe, so it was measured and not gated.
- `packages/salvageunion-reference/tools/_tmpdump.ts:1` — **Scratch debug file committed** (its sibling `_tmpdump2.ts` was correctly left untracked). Because knip treats `tools/**/*.ts` as entry points, it will never be flagged as dead and will linger; it already needs an `as never` cast to compile.

Non-blocking, not filed: `crawler-bays.json`'s `damagedEffect` text is never scanned (only `content` — currently 0 matches); the `?? source` fallback in `consider()` stringifies whole records for 271 player records (no false positives today, but it's an untargeted scan); and `validateParity.ts` `process.exit(1)`s on unexpected findings before it ever reports stale entries. `knip`, the new tests (8 pass), and `validate:parity` all run clean.

PR #628, opened 2026-07-28
feat(itun): a blocked Push teaches the rule instead of greying out

alxjrvs:
**F6** of the rules-parity plan (#599) — first of the recommended order.

ADR-021 requires Guided Play to *teach as it enforces*. `RuleBrief` existed but appeared **only in the three wizards**, never in `components/dashboard/` — so the Dashboard enforced without ever explaining.

## The change

A Push that would exceed the Heat Cap used to **disable the button and put the reason in a `title`**. That taught nothing at the moment the rule actually bit, and a hover title is **unreachable on touch** — the enforcement was invisible exactly where it mattered.

The button now stays pressable and, when blocked, opens the band overlay with the rule and its citation *instead of* performing the action:

> **The Rule** — Pushing adds 2 Heat before the roll, and a Mech can never exceed its Heat Cap. This Mech is at 3/4 Heat, so a Push would take it past the Cap — vent or take a Heat Check first.
> *Quick Ref · p.233*

Two details that follow existing precedent rather than inventing anything:

- The **Heat gauge stays on screen** beside the explanation, so the number the rule is *about* and the reason are visible together. That's the same reasoning the overlay's `gauges` field already encodes for Take Damage.
- The brief **interpolates the mech's actual Heat and Cap**, so it describes *this* situation rather than the rule in the abstract.

Reuses `RuleBrief` and the existing `BandOverlay` contract — no dashboard-local variant.

## Verification

`bun run check:all` green: lint, format, typecheck, **4,565 tests**, validate (incl. the parity gate), knip, audit, tokens, styling.

Two new tests pin **both** directions: a blocked Push shows the rule and writes **nothing**, and a legal Push still performs the action without showing the explanation.

Next in order: **F2/F3/F4** — wiring salvage, crafting and scrap-mech, which are fully implemented and tested with zero non-test importers.

Refs #599.

### DIFF
--- a/apps/itun/src/components/dashboard/ActiveItemBand.tsx
+++ b/apps/itun/src/components/dashboard/ActiveItemBand.tsx
@@ -38,6 +38,7 @@ import { ActiveItemBand as ActiveItemBandView, CountStepper, StorageBay } from '
 import type { ActiveItemBandView as ActiveItemBandViewModel, BandButton } from 'component-lib'
 import { DASHBOARD_TXN } from '../../stores/surfaceProvenance'
 import { pilotingContext } from '../../lib/rules/pilotingContext'
+import { RuleBrief, type StepRule } from 'component-lib'
 import {
   VENT_PATCH,
   critDamagePatch,
@@ -133,6 +134,19 @@ type MechPrompt =
   | { kind: 'storage' }
   | null
 
+/**
+ * Rules a blocked Guided-Play control explains when the player reaches for it
+ * (ADR-021 — the guided modes teach as they enforce).
+ *
+ * Paraphrased with a citation, matching the wizards' RuleBrief contract; the
+ * numbers are interpolated so the brief describes THIS mech's situation rather
+ * than the rule in the abstract.
+ */
+const PUSH_RULE = (heat: number, cap: number): StepRule => ({
+  rule: `Pushing adds 2 Heat before the roll, and a Mech can never exceed its Heat Cap. This Mech is at ${heat}/${cap} Heat, so a Push would take it past the Cap — vent or take a Heat Check first.`,
+  cite: 'Quick Ref · p.233',
+})
+
 function MechBand({
   mech,
   store,
@@ -176,6 +190,9 @@ function MechBand({
 
   // Quick Ref p.233 — can't Push if +2 Heat would exceed the Heat Cap.
   const pushLocked = heat + 2 > maxHeat
+
+  // The rule a blocked control explains when the player reaches for it.
+  const [blocked, setBlocked] = useState<StepRule | null>(null)
 
   function doPush() {
     const m = fresh()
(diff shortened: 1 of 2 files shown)


alxjrvs (commented):
## Review scope

`git diff @{upstream}...HEAD` was empty (branch == upstream). PR **#628** is the single top commit `d84592df` (`feat(itun): a blocked Push teaches the rule instead of greying out`) — the other 25 commits vs `main` are already-merged PRs (`main` is just behind).

**Mid-review the working tree changed**: a concurrent session is actively writing the F2/F3/F4 Downtime economy (`ActiveItemBand.tsx` modified + new `dashboardEconomy.ts` / `dashboardEconomy.test.ts`; file mtimes were seconds old). Per the scope rule I reviewed those too, but findings 4–7 are on code that is **still being written** — treat them as provisional.

Verified: `bun --filter itun test` 1306 pass / 0 fail, `typecheck:itun` clean, `bun run knip` **fails**, `biome check` reports format errors on the in-flux files.

## Findings

- `apps/itun/src/components/dashboard/ActiveItemBand.tsx:290` — `RuleBrief`/`SheetSectionCard` theme entirely off `--tone`/`--tone-deep`, defined only under `.sheet--*`; the Dashboard sets no such class, so the callout frame and header/footer bands paint nothing and the `text-paper` citation lands on cream at ~1.05:1 — the "Quick Ref · p.233" cite is invisible. The new test only asserts DOM presence.
- `apps/itun/src/components/dashboard/ActiveItemBand.tsx:146` — the brief's remedy "vent or take a Heat Check first" is wrong: `heatCheckPatch` writes back the same `currentHeat`, so clicking "Heat Chk" as taught leaves Push still blocked.
- `apps/itun/src/components/dashboard/ActiveItemBand.tsx:275` — `blocked` short-circuits before `prompt`, so an ActionsDeck-armed damage prompt is consumed (`consumeDamagePrompt`, Apply disabled) but hidden behind the rule overlay, then pops unexpectedly on dismiss.
- `apps/itun/src/components/dashboard/ActiveItemBand.tsx:295` *(working tree)* — `Scrap Mech` calls `doScrap` on a single click with no confirm, contradicting its own docstring, `dashboardEconomy`'s header, and ADR-007; it sits directly above `◄ Leave Downtime` and has no undo.
- `apps/itun/src/components/dashboard/dashboardEconomy.ts:147` *(working tree)* — `scrappedCargo` has zero importers; `bun run knip` exits 1, blocking `check:all` and pre-push.
- `apps/itun/src/components/dashboard/ActiveItemBand.tsx:133` *(working tree)* — `craft()` re-reads the store right after a fire-and-forget `void store.update`; two quick clicks in the picker double-spend scrap and drop the first crafted lot (ADR-008 hazard).
- `apps/itun/src/components/dashboard/ActiveItemBand.tsx:154` *(working tree)* — `items.slice(0, 40)` on a tech-level-ascending sort silently hides every higher-tech craftable, contradicting CraftBody's own "listed with its shortfall rather than hidden" rationale.

PR #629, opened 2026-07-28
feat(itun): wire salvage, crafting and scrap-mech into the Dashboard

alxjrvs:
**F2/F3/F4** of the rules-parity plan (#599) — second in the recommended order, after #628.

These three modules were **fully implemented and fully tested with zero non-test importers**. Their Live-Sheet panels were deleted on the way to the Dashboard and never rebuilt, so this is the largest block of finished rules work no player could reach. **This PR is the missing seam, not new rules.**

`dashboardEconomy.ts` follows `dashboardRules`' contract — pure patch builders, no React, no store — so handlers stay trivially testable and ADR-006's pure-math boundary holds.

Carrying F6's pattern forward: Guided Play enforces **and teaches**, so a blocked action returns a null patch plus a **reason** rather than a patch the caller must remember not to apply. An unaffordable craft is listed **with its shortfall** instead of hidden, so a player can see what to save for.

## A rules distinction knip surfaced

knip flagged `scrappedCargo` as an unused export. That was true — and the reason was a **bug**: I'd never wired the cargo hand-off at all.

`handOffCargo` states the rule in its own header:

> a scrapped mech hands its hold to the crawler, because nothing in the hold was ever part of the mech and losing it is silent data loss — but an **already-destroyed** mech has lost its cargo with it (p.234/p.240)

Those are **different states**, and my first version conflated them by marking every scrap `destroyed: true`. The hand-off now happens on a deliberate scrap and is skipped on a destroyed mech, with a test each way.

## Other details worth review

- **Scrapping commits through `entityStore.transfer`** — cross-entity and all-or-nothing, since a partial write would duplicate or destroy player value. That's the same `transfer` path fixed in #602 to actually emit provenance, so a scrap is now auditable.
- The mech is marked `destroyed` rather than **deleted**, so the record keeps its Change Log and soft-links (ADR-007 keeps destruction visible rather than silently dropping data).
- An Area Salvage roll that finds nothing makes **no write** — a no-op patch would appear in the Change Log as a change that changed nothing. A jackpot likewise writes nothing: it hands the player an *item* choice, not scrap.

## Verification

`bun run check:all` green: lint, format, typecheck, **4,575 tests**, validate (incl. the parity gate), knip, audit, tokens, styling. 13 new tests on the patch builders.

Next in order: **C4** (inline rules-bearing marker).

Refs #599.

### DIFF
--- a/apps/itun/src/components/dashboard/ActiveItemBand.tsx
+++ b/apps/itun/src/components/dashboard/ActiveItemBand.tsx
@@ -13,9 +13,10 @@
  */
 
 import { useEffect, useState } from 'react'
+import { SalvageUnionReference } from 'salvageunion-reference'
 
 import {
-  crawlerMaxSP,
+  crawlerMaxSPParts,
   mechMaxCargo,
   mechMaxEP,
   mechMaxHeat,
@@ -40,6 +41,12 @@ import { DASHBOARD_TXN } from '../../stores/surfaceProvenance'
 import { pilotingContext } from '../../lib/rules/pilotingContext'
 import { RuleBrief, type StepRule } from 'component-lib'
 import {
+  areaSalvageOutcome,
+  craftOutcome,
+  crawlerTechLevelOf,
+  scrapMechOutcome,
+} from './dashboardEconomy'
+import {
   VENT_PATCH,
   critDamagePatch,
   critInjuryPatch,
@@ -54,7 +61,7 @@ import {
 } from './dashboardRules'
 
 /** The store surface the band needs — injectable so tests can assert patches. */
-export type PlayStore = Pick<EntityState, 'get' | 'update'>
+export type PlayStore = Pick<EntityState, 'get' | 'update' | 'transfer'>
 
 type ActiveItemBandProps = {
   mech: Mech
@@ -74,7 +81,7 @@ export function ActiveItemBand({ mech, pilot, crawler, store }: ActiveItemBandPr
   const s: PlayStore = store ?? liveStore
 
   if (mount === 'downtime' && crawler) {
-    return <CrawlerBand crawler={crawler} onLeave={leaveDowntime} />
+    return <CrawlerBand crawler={crawler} mech={mech} store={s} onLeave={leaveDowntime} />
   }
   if (mount === 'pilot' && pilot) {
     return <PilotBand pilot={pilot} store={s} onBoard={() => setMount('mech')} />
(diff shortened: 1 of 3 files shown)


alxjrvs (commented):
**Scope reviewed:** `origin/main...HEAD` — one commit, `72567b43 feat(itun): wire salvage, crafting and scrap-mech into the Dashboard` (3 files: `ActiveItemBand.tsx`, new `dashboardEconomy.ts`, new `dashboardEconomy.test.ts`). Working tree clean. `bun --filter itun typecheck` and `test` (1308 pass / 0 fail) both green, so nothing here is a build break.

The pure patch-builder layer (`dashboardEconomy.ts`) is well-shaped and matches its stated contract. The problems are almost all in the UI seam it was written for: the wiring skips gates the pure rules modules documented and exported constants for, and the one destructive path has no confirmation.

Findings, most severe first:

- `apps/itun/src/components/dashboard/ActiveItemBand.tsx:294` — `Scrap Mech` commits the irreversible `store.transfer` on a single click; the overlay is a post-hoc report, not a confirm, contradicting `doScrap`'s own docblock and the `confirmDestroyed` / `Mark Mech Destroyed` pattern already in this file.
- `apps/itun/src/components/dashboard/ActiveItemBand.tsx:154` — `items.slice(0, 40)` silently truncates the craft picker; sorted TL-ascending, a Tech-6 crawler sees only the 35 Tech-1 items plus 5 Tech-2 ones out of 194 eligible, and can never craft above Tech 2. No search, paging, or truncation hint.
- `apps/itun/src/components/dashboard/dashboardEconomy.ts:58` — crafting never checks for a present, Intact Crafting Bay (p.222); `CRAFTING_BAY` exists for exactly this and is now referenced nowhere in the app.
- `apps/itun/src/components/dashboard/ActiveItemBand.tsx:280` — Area Salvage never spends Supply (`performAreaSalvage`: "Supply accounting is the caller's job"); repeated clicks mint unlimited scrap. Nothing in `apps/itun/src` tracks Supply at all.
- `apps/itun/src/components/dashboard/ActiveItemBand.tsx:202` — `areaTl: crawlerTl` uses the crawler's Tech Level as the *area's*; p.245 makes area TL a Mediator-set property of the location, so scrap is minted at the wrong denomination in both directions.
- `apps/itun/src/components/dashboard/ActiveItemBand.tsx:208` — a natural 20 ("Jackpot!") produces no patch and no picker, so the best result on the table yields strictly less than a roll of 2; the existing `SalvageClaim` / `takeFromClaim` machinery isn't wired.
- `apps/itun/src/components/dashboard/ActiveItemBand.tsx:143` — read-modify-write race: `store.get` reads in-memory state that `update` only refreshes after the async IDB write, and the patches carry whole `cargoLots` / `scrapPool` values, so two fast craft (or salvage) clicks silently lose one item or one deposit while both notes claim success.
- `apps/itun/src/components/dashboard/__tests__/dashboardEconomy.test.ts:124` — the scrap-mech deposit assertion is vacuous: the fixture's `chassisRef: 'unknown-chassis'` with no systems/modules makes `breakdown.total` always 0, so `expect(after - before).toBe(breakdown.total)` passes even if `depositScrapDeposits` were deleted. Verified by running the fixture.
- `apps/itun/src/components/dashboard/ActiveItemBand.tsx:225` — `void store.transfer` / `void store.update` with no `.catch`, while the overlay reports success synchronously; `transfer` explicitly aborts with nothing changed on a Zod failure, so the player is told the mech was scrapped and the scrap banked when neither happened.
- `apps/itun/src/components/dashboard/dashboardEconomy.ts:35` — `crawlerTechLevelOf` is a third copy of `parseCrawlerTechLevel` (`apps/itun/src/lib/crawlerLevel.ts`, itself already duplicated in `salvageunion-reference/lib/rules/derivedStats.ts:446`); none clamp to the legal 1–6 range.

PR #630, opened 2026-07-28
feat(component-lib): mark rules-bearing prose inline on entity cards

alxjrvs:
**C4** of the rules-parity plan (#598) — third in the recommended order, after #628 and #629.

The sentence that grants a contribution is now marked on the card, so a reader can see **which clause the app actually applies** — adjacent to the claim itself, rather than as a badge somewhere else. Extends ADR-026 §5's rust "modified" language from stat cells to a prose span: same vocabulary, third carrier.

## The negative case is the point

Prose that **states** a mechanical change on a record carrying **no** structured data stays **unmarked**. So the coverage gap is legible in the product, not only in CI — anyone reading a card can tell the difference between a rule the app *applies* and a rule it merely *displays*.

## One detector, two consumers

The claim-detection patterns move out of `tools/validateParityLogic.ts` into shipped `lib/rules/rulesBearing.ts`, shared by the **parity audit** and the **renderer**.

They must never disagree. If the renderer marked a claim the audit didn't enforce, a record could *look* machine-backed while being inert — precisely the state this effort exists to remove. Verified: the audit reports **identical counts before and after** the move (30 claims — 14 encoded, 14 exempt, 2 known).

## Granularity

Marking is at **paragraph** level, deliberately. The trait parser has already turned the text into a node array by render time, so splicing a sentence-level mark back into it would be fragile for no real gain — these paragraphs are one or two sentences.

## Verification

`bun run check:all` green: lint, format, typecheck, **4,580 tests**, validate (incl. the parity gate), knip, audit, tokens, styling.

Five new tests, including the two that carry the intent: the **same claim** renders marked on an encoded record and **unmarked** on one with no data, and only the claiming paragraph is marked, not its neighbours.

Next in order: **F1** (activated contributions — manual expiry), then **F5** (Downtime writes), then **effect targeting**.

Refs #598.

### DIFF
--- a/packages/component-lib/src/components/referenceEntity/Content.tsx
+++ b/packages/component-lib/src/components/referenceEntity/Content.tsx
@@ -8,6 +8,7 @@ import { borderColorFromHeaderBg } from './referenceEntityHelpers'
 import { cn } from '../../utils/cn'
 import { Slab } from '../chrome/Slab'
 import { Badge } from '../chrome/Badge'
+import { statesMechanicalChange } from 'salvageunion-reference/rules'
 
 /**
  * A single `datavalues` item — rendered as a horizontal Stat chip (the one
@@ -61,6 +62,20 @@ type ContentProps = {
   headerBg?: string
   /** Raw CSS color for borders (overrides headerBg-derived color when set) */
   headerBgColor?: string
+  /**
+   * True when the owning record carries structured data for the mechanical
+   * change its text states (ADR-029).
+   *
+   * When set, a paragraph that STATES a mechanical change is marked as
+   * rules-bearing — the reader can see which clause the app actually
+   * understands, adjacent to the claim itself. Prose that states a change on a
+   * record with no data stays unmarked, so a coverage gap is visible in the
+   * product and not only in CI.
+   *
+   * Extends ADR-026 §5's rust "modified" language from stat cells to a prose
+   * span; it is the same vocabulary, a third carrier.
+   */
+  rulesBearing?: boolean
 }
 
 /**
@@ -79,6 +94,7 @@ export function Content({
   fontSize = 'text-sm',
   compact = false,
   chassisName,
+  rulesBearing = false,
   headerBg,
   headerBgColor,
 }: ContentProps) {
@@ -144,6 +160,7 @@ export function Content({
               fontSize={fontSize}
               compact={compact}
               chassisName={chassisName}
+              rulesBearing={rulesBearing}
             />
           ))
         }
@@ -168,6 +185,7 @@ export function Content({
                 fontSize={fontSize}
                 compact={compact}
                 chassisName={chassisName}
+                rulesBearing={rulesBearing}
               />
             )}
             {labelText && (
@@ -183,6 +201,7 @@ export function Content({
                     block={block}
                     fontSize={fontSize}
                     compact={compact}
+                    rulesBearing={rulesBearing}
                     chassisName={chassisName}
                   />
                 ))}
@@ -200,11 +219,14 @@ function ContentBlock({
   fontSize,
   compact,
   chassisName,
+  rulesBearing = false,
 }: {
   block: SURefObjectContentBlock
   fontSize: string
   compact: boolean
   chassisName?: string
+  /** See `ContentProps.rulesBearing`. */
+  rulesBearing?: boolean
 }) {
   const type = block.type || 'paragraph'
   const blockValue = block.value
@@ -230,17 +252,29 @@ function ContentBlock({
   }
 
   switch (type) {
-    case 'paragraph':
+    case 'paragraph': {
+      // Mark the clause the app actually understands. Paragraph granularity is
+      // deliberate: the trait parser has already turned this text into a node
+      // array, so splicing a sentence-level mark back into it would be fragile
+      // for no gain — these paragraphs are one or two sentences.
+      const bearing = rulesBearing && statesMechanicalChange(stringValue) !== null
       return (
         <Text
           variant="body"
           as="div"
-          className={cn('mb-1', fontSize)}
+          className={cn('mb-1', fontSize, bearing && 'border-l-2 border-rust pl-2')}
           style={{ overflowWrap: 'break-word' }}
+          {...(bearing
+            ? {
+                'data-rules-bearing': 'true',
+                title: 'The app applies this rule',
+              }
+            : {})}
         >
           {parsedValue}
         </Text>
       )
+    }
 
     case 'heading': {
       // Section headings render as the canonical Slab separator — a level-1
--- a/packages/component-lib/src/components/referenceEntity/__tests__/Content.rulesBearing.test.tsx
+++ b/packages/component-lib/src/components/referenceEntity/__tests__/Content.rulesBearing.test.tsx
@@ -1,0 +1,56 @@
+/**
+ * The inline rules-bearing marker (C4, ADR-029).
+ *
+ * Marks the clause the app actually applies, adjacent to the claim itself —
+ * extending ADR-026 §5's rust "modified" language from stat cells to a prose
+ * span.
+ *
+ * The negative case carries the design intent: prose that STATES a mechanical
+ * change on a record with NO structured data stays UNMARKED, so a coverage gap
+ * is visible in the product and not only in CI.
+ */
+
+import { describe, expect, test, afterEach } from 'bun:test'
+import { cleanup, render } from '@testing-library/react'
+
+import { Content } from '../Content'
+
+afterEach(cleanup)
+
+const claim = [
+  { type: 'paragraph' as const, value: "This System increases your Mech's Max SP by 5." },
+]
+const flavour = [{ type: 'paragraph' as const, value: 'Layered plates of resilient metal.' }]
+
+function marks(container: HTMLElement): number {
+  return container.querySelectorAll('[data-rules-bearing="true"]').length
+}
+
+describe('rules-bearing prose marker', () => {
+  test('marks a paragraph that states a mechanical change on an ENCODED record', () => {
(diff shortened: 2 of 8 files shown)


alxjrvs (commented):
## Review: `parity-c4` → `main` (2 commits)

Scope note: the branch had one commit (`3cc4c1c8`) when I started; a concurrent agent added `b024c6a4` (ReDoS fix) mid-review and then switched this worktree to `parity-f1`, so the working tree no longer holds C4. I reviewed both commits from git objects. I also ran `bun run build:package` early on, which dirtied generated files — I reverted that churn (`etc/*.api.d.ts`, `schemas/`); the remaining dirty files are the other agent's in-flight work, untouched.

Verification run: `bun --filter component-lib test` (646 pass), `bun run typecheck` (all workspaces clean), `bun run knip` (clean), plus a data-driven differential of the old vs new claim detector across all 27 data files.

### Findings

- `packages/salvageunion-reference/lib/rules/rulesBearing.ts:52` — The ReDoS rewrite dropped the `/i` flag from the trait-claim pattern while the other two alternatives kept it. `statesMechanicalChange('Gains the Fly Trait.')` went from `'effect'` to `null`; 6 real sentences in the shipped dataset regress, including Bio-Wings' own claim. The aggregate audit count is unchanged only because those records surface via a second sentence, which is what hid it — so the commit message's "preserves the original semantics exactly" is not accurate, and a future record authored as "Gains the X Trait." passes `validate:parity` silently. The new test only covers the lowercase form.
- `packages/component-lib/src/components/referenceEntity/card/ReferenceEntityCard.tsx:1994` — `isRulesBearing(data)` keys off the rendered record, but rules prose for systems/modules/equipment lives in `actions.json` while the data lives on the owner. Composite Armour's "increases your Mech's Max SP by 5" is marked on the system card and bare on the standalone action page — which the prop's own doc defines as "the app is not applying this". The SRD actions catalog therefore shows false coverage gaps.
- `packages/component-lib/src/components/referenceEntity/card/ReferenceEntityCard.tsx:108` — `isRulesBearing` mirrors only the audit's `hasCapData` and never `hasEffectData`, yet the single boolean is applied to paragraphs of both claim classes. Breaks the invariant stated in `rulesBearing.ts`'s own header. I confirmed it is latent (0 records hit it today), not live.
- `packages/component-lib/src/components/referenceEntity/card/ReferenceEntityCard.tsx:98` — the new function was inserted between `isTitanicAction`'s JSDoc and `isTitanicAction`, so the comment now documents the wrong function.
- `packages/component-lib/src/components/referenceEntity/Content.tsx:270` — the mark's only non-visual carrier is `title` on a role-less `<div>`; not reliably announced, and unreachable on touch.

Things I checked and found clean: the `salvageunion-reference/rules` subpath export, `border-rust` / `border-l-2` against the `cn()` tailwind-merge registration, `Text`'s prop forwarding for `title` + `data-*`, the `validateParityLogic.ts` refactor (behaviour-identical extraction), knip on the two new exported regex constants, and paragraph-vs-sentence granularity (the patterns cannot span a `.`, so both consumers agree). I also traced whether `Content` can receive untrusted text — it cannot: ITUN schemas are `.strict()` and refs-only, and shared snapshots re-resolve entities locally, so the ReDoS fixed in `b024c6a4` was never externally reachable.

alxjrvs (commented):
**Scope note:** the worktree HEAD moved to `origin/main` mid-review (another agent shares this checkout), so I pinned the review to the actual PR range `origin/main..origin/parity-c4` — 2 commits, 8 files: the inline rules-bearing marker plus the ReDoS fix.

**Verification I ran:** a differential harness comparing the old inline `CAP_PATTERN`/`EFFECT_PATTERN` against the new `rulesBearing.ts` over all 3,078 sentences in the dataset (0 divergences — the refactor is behaviour-preserving on today's content), a 48,960-case permutation fuzz (100 divergences, all in one pattern), a renderer-vs-audit agreement scan over every paragraph and owner-attributed action, and a timing check confirming the old regexes genuinely hang (>2 min) on the adversarial input, so the ReDoS fix and its regression test are real and the 250ms budget is not flaky.

Findings:

- `packages/salvageunion-reference/lib/rules/rulesBearing.ts:52` — the trait-grant pattern silently lost the `/i` flag the original carried, so sentence-initial claims (`"Gains the Fly Trait."`, `"Gain the Ammo Trait."`) no longer classify; the burn-down gate will let that class of new unencoded content through, contradicting the file's own "keeps the ORIGINAL semantics" comment.
- `packages/component-lib/src/components/referenceEntity/card/ReferenceEntityCard.tsx:108` — `isRulesBearing` throws away the `'cap' | 'effect'` class and tests only cap-shaped data, so it disagrees with the audit's `hasEffectData` on effect claims; a record with `statBonus` plus trait prose renders "The app applies this rule" for a rule nothing applies. Latent on current data (0 instances found), but it violates the invariant `rulesBearing.ts` explicitly states.
- `packages/component-lib/src/components/referenceEntity/card/ReferenceEntityCard.tsx:115` — `contributions: []` is schema-valid and `Array.isArray([])` is true, so an empty array marks prose as encoded in both the card and the audit.
- `packages/component-lib/src/components/referenceEntity/card/ReferenceEntityCard.tsx:98` — the new function was spliced under `isTitanicAction`'s JSDoc, so that comment now documents `isRulesBearing`.

Not flagged after checking: `--color-rust` is a real token and `Text` forwards the `data-*`/`title` props (the marker renders); the patterns cannot match across a `.`, so the renderer's paragraph granularity is equivalent to the audit's sentence granularity; and `RulesClaim` mirrors the existing `ContributionTarget` re-export shape, so knip should stay green.

alxjrvs (commented):
## Review: `parity-c4` (`b024c6a4`) — rules-bearing prose marker + linear claim detection

**Scope note (important):** this worktree (`/Users/jarvis/Code/SU-SRD/.claude/worktrees/rules-parity-plan`) was switched out from under the review mid-session — it is now on `parity-f1` (`83f6c2b0`) with uncommitted `contributions.ts` / `objects.ts` / schema edits from another agent. I pinned the review to the branch ref `parity-c4` @ `b024c6a4` and read every blob via `git show`, so the findings below describe the PR, not the worktree's current disk state. If another agent is meant to be sharing this checkout, be aware their in-flight `fromStat` / `duration: activated` work is live there.

Reviewed diff: `origin/main...b024c6a4` — 8 files, +275/−15. Verified empirically against the real dataset (27 `data/*.json` files from `origin/main`) rather than by reading alone.

### Findings

- `packages/salvageunion-reference/lib/rules/rulesBearing.ts:52` — The extracted trait-grant pattern silently lost the `/i` flag the original `EFFECT_PATTERN` carried. Replaying old vs. new over every sentence in the dataset yields **5 real sentences that flip from `effect` to `null`** (`Gains the Burrower Trait.`, `Gains the Fly Trait.`, and all-caps `... GAIN THE IMMOBILE TRAIT ...`). They're all in `roll-tables.json` (outside `PLAYER_CONTENT`) so today's audit output is unchanged — but a future system/equipment record with a sentence-initial `Gains the X Trait.` (exactly Bio-Wings' shape) now walks past the gate. This contradicts the file's own claim that the rewrite "keep[s] the ORIGINAL semantics — the fix is linear, not looser."
- `packages/component-lib/src/components/referenceEntity/card/ReferenceEntityCard.tsx:108` — `isRulesBearing` mirrors only the audit's `hasCapData`, and `Content.tsx` throws away the `cap`/`effect` class the detector returns. So an `effect` claim on any record with cap-shaped data renders "The app applies this rule" while the audit reports it unresolved — the exact contradiction `rulesBearing.ts`'s header says must never happen. Mirror gap: records encoding effects via `choices[].choiceOptions[].effects` are never marked. Confirmed latent (0 divergences across all current PLAYER_CONTENT records *and* action-attributed prose).
- `packages/salvageunion-reference/lib/rules/rulesBearing.test.ts:43` — The ReDoS regression test can't fail for the reason it states. Measured: the old vulnerable `CAP_PATTERN` on the *raw* 50k-space input doesn't finish in 2 minutes, but on the **normalised** form of that same input it runs in **0.06 ms** — 4000× under the 250 ms budget. Since `statesMechanicalChange` normalises first, reverting to the old patterns leaves this test green.
- `packages/component-lib/src/components/referenceEntity/card/ReferenceEntityCard.tsx:98` — `isRulesBearing` was inserted *between* the `/** A titanic action … */` comment and the `isTitanicAction` it documents, orphaning that JSDoc onto the new function.
- `packages/component-lib/src/components/referenceEntity/Content.tsx:270` — The mark is conveyed by rust border colour plus `title` on a non-focusable `div`; no accessible equivalent (WCAG 1.4.1) in a repo with an AA gate.

### Checked and found clean

- Old→new regex equivalence for all CAP patterns and the damage/range EFFECT patterns (only the `/i` loss diverges).
- The `normalise()`-first fix is genuinely linear; also holds on non-whitespace adversarial input (200k letters: 2.7 ms; 100k `"b "` repeats after `gains the A`: 3.6 ms).
- The folded-action render path passes the **owner** record to `isRulesBearing`, which matches `auditParity`'s `ownerByActionName` attribution — no mismatch there (16 action paragraphs marked, 0 divergences).
- No paragraph in the dataset matches only *across* sentence boundaries, so the renderer's whole-paragraph evaluation vs. the audit's per-sentence split produces no false positives today.
- Guide-step and pattern-prose sites pass the parent record, but no guide or chassis carries cap-shaped data, so those are inert.
- `salvageunion-reference/rules` is already a runtime import in component-lib (`CatalogChoiceListing.tsx`), so no new dependency edge or bundle-graph regression.
- `RulesClaim` is used within its defining file, so `ignoreExportsUsedInFile: true` exempts it from the knip entry-export gate.
- The `etc/*.api.d.ts` report was regenerated in step with the new export — no drift.

PR #631, opened 2026-07-28
feat(itun): account management, Discord sign-in, and the /account route

alxjrvs:
Next in the stack after #609 (contains its commits; they drop out of the diff once it merges). Phase 1 of [ADR-030](https://github.com/SalvageUnion-io/SU-SRD/blob/main/docs/adrs/ADR-030-accounts-games-server-of-record.md), specifically the half that discharges the obligations that come with holding somebody's Discord identity.

**Still inert for existing users.** Everything is behind the `isConvexConfigured` branch, so a build without `VITE_CONVEX_URL` — CI, a fresh checkout, today's production — renders exactly what it did before.

## `convex/account.ts`

`me` / `updateProfile` / `exportMine` / `deleteAccount`.

**`deleteAccount` is the operation with no undo**, so its rules are explicit rather than emergent:

- The Organizer flag passes to the **longest-standing remaining member**, with a deterministic tie-break on `_id`. Two members can share a `joinedAt` if a game was seeded in one mutation, and an unstable choice there would make the outcome depend on document scan order.
- **The Game and the communal crawler survive.** A campaign does not die because one person quit.
- Only when the departing user is the **last** member does the Game go — an empty game is not a campaign anyone returns to, and its unclaimed entities have no shelf to fall back to.
- Auth rows are deleted **in dependency order**: refresh tokens are keyed by *session*, not by user, so they go before the sessions they point at. Deleting sessions first would strand them as orphans referencing deleted rows.

**`exportMine` is scoped to what you own, not what you can see.** A crewmate's pilot is readable inside a Game but is not this person's data to take away, and the crawler belongs to the table.

## Two things found by checking rather than assuming

**The `users` table now redeclares `phone` / `phoneVerificationTime` and the `phone` index.** Overriding `authTables.users` means anything the library expects has to be restated here — a missing field would surface as a write failure inside the auth flow rather than as a type error.

**`Text` has no `size` prop and no `label`/`heading` variants.** The first draft invented all three; corrected against the real component API.

## UI

`SignInControl`, `AccountScreen`, `AccountStrip`, and `/account`.

The strip is **local to ITUN rather than a nav item in `component-lib`'s `AppBar`**. `AppBar` is shared with `apps/srd`, which has no accounts and never will, so an auth slot there would push an ITUN-only concept into a package whose contract is to stay backend-agnostic — and would touch a component every page of both apps renders.

In a Solo build `SignInControl` renders **nothing**, not a disabled button. Offering an account to somebody whose build cannot have one is worse than silence.

## Tests

14 new (1336 in itun, 3919 across the repo):

- 10 convex cases on deletion semantics — the campaign surviving, another member's pilot untouched, the last-member-out case, and refresh-token ordering
- 4 on rendering the account screen in a Solo build **with no Convex provider mounted**. That is the structural failure worth guarding: it would break the majority of users while working perfectly for anyone testing with a configured `.env.local`.

## Note for the maintainer

`bun run format` **from `apps/itun`** reformats `CHANGELOG.md`, which the root `.prettierignore` deliberately excludes ("its output would diverge from what release-please writes, making every generated release PR fail `format:check`"). Prettier resolves `.prettierignore` relative to cwd, so the workspace-level script bypasses the root ignore. I hit this, caught it, and reverted — but the next person will hit it too. Worth either adding `apps/*/.prettierignore` or dropping the per-workspace `format` scripts.

## Verification

typecheck 0 errors · knip clean · `format:check` clean · **3919 tests pass** · no new lint warnings.

### DIFF
--- a/apps/itun/convex/__tests__/account.test.ts
+++ b/apps/itun/convex/__tests__/account.test.ts
@@ -1,0 +1,216 @@
+import { describe, expect, test } from 'bun:test'
+
+import { api } from '../_generated/api'
+import type { Id } from '../_generated/dataModel'
+import { testConvex } from './harness'
+
+/**
+ * Account management, and in particular the promise D27 makes:
+ * **a campaign never dies because one person quit.**
+ *
+ * Deletion is the operation with no undo, so the cases below are less about
+ * "does it delete" and more about "does it delete *only* the right things" —
+ * the crew's Game, the communal crawler, and other people's characters all
+ * have to survive somebody exercising their right to be forgotten.
+ */
+
+type Ctx = ReturnType<typeof testConvex>
+
+async function makeUser(t: Ctx, name: string) {
+  const userId = await t.run(
+    async (ctx) => await ctx.db.insert('users', { name, displayName: name })
+  )
+  return { userId, as: t.withIdentity({ subject: userId }) }
+}
+
+async function seedPilot(t: Ctx, gameId: Id<'games'> | null, ownerId: Id<'users'> | null) {
+  return await t.run(
+    async (ctx) =>
+      await ctx.db.insert('pilots', { gameId, ownerId, body: { callsign: 'X' }, updatedAt: 1 })
+  )
+}
+
+describe('profile', () => {
+  test('displayName falls back to the Discord name when unset', async () => {
+    const t = testConvex()
+    const u = await makeUser(t, 'Roach-Boy')
+    await u.as.mutation(api.account.updateProfile, { displayName: null })
+
+    const me = await u.as.query(api.account.me, {})
+    // Cleared means "fall back", not "blank" — a nameless owner chip is worse
+    // than a stale one.
+    expect(me?.displayName).toBe('Roach-Boy')
+  })
+
+  test('a set displayName overrides the Discord name', async () => {
+    const t = testConvex()
+    const u = await makeUser(t, 'Roach-Boy')
+    await u.as.mutation(api.account.updateProfile, { displayName: '  Beefcake  ' })
+
+    const me = await u.as.query(api.account.me, {})
+    expect(me?.displayName).toBe('Beefcake')
+  })
+
+  test('an anonymous caller has no profile to read', async () => {
+    const t = testConvex()
+    await expect(t.query(api.account.me, {})).rejects.toThrow(/not signed in/i)
+  })
+})
+
+describe('export is scoped to what you own', () => {
+  test('excludes a crewmate pilot you can merely see', async () => {
+    const t = testConvex()
+    const owner = await makeUser(t, 'Owner')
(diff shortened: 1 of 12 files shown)


alxjrvs (commented):
## Review scope

Local `main` was stale — the real diff is `origin/main...HEAD`, 5 commits / 50 files: the Convex backend (`apps/itun/convex/`), the connection-mode layer, the account UI, and the Eldridge Coast seed removal.

I verified the branch is green before reviewing: `bun run typecheck` (all workspaces, and I confirmed by probe that `convex/*.ts` *is* covered via `_generated/api.d.ts`), `bun --filter itun test` (1336 pass, incl. the 31 new Convex tests), and `bun run knip`. I also checked the Convex bundler's entry-point rules and the `@convex-dev/auth` source to confirm/kill two theories (the `convex-test` → `node:async_hooks` push hazard is a non-issue — Convex ships a shim plugin for it).

## Findings

- `apps/itun/convex/ownership.ts:38` — `loadOwnable` casts an unvalidated `v.string()` entityId to an `Id` with no `ctx.db.normalizeId`; I ran a convex-test case proving `assign({table:'pilots', entityId:<mechs id>})` succeeds and writes a Change Log row labelled `entityType: 'pilot'`.
- `apps/itun/convex/account.ts:228` — no `signOut()` after `deleteAccount`, and `getAuthUserId` only reads the JWT subject, so the deleted identity stays authenticated for the token's remaining life and can re-insert `memberships` rows pointing at a deleted user.
- `apps/itun/src/components/account/AccountScreen.tsx:23` — local `download()` reimplements the existing `downloadJson` helper minus both browser workarounds (detached anchor → no-op in Firefox; synchronous `revokeObjectURL` → aborts in Safari).
- `apps/itun/src/lib/connection/ConnectionProvider.tsx:68` — `useConvexAuth().isLoading` discarded, so every page load classifies a signed-in user as Solo until the token resolves: "Sign in with Discord" flashes at signed-in users and `canWrite` reports local-first semantics for a server-backed one.
- `apps/itun/convex/games.ts:144` — `destroy()` deletes game-scoped `softLinks` while deliberately preserving the pilots/mechs they connect, silently severing every mech→pilot pairing that survives on a shelf.
- `apps/itun/convex/account.ts:173` — `presence` is swept only in the last-member-out branch, so account deletion leaves the user's presence rows (and `invites.createdBy`) dangling in every surviving Game.
- `apps/itun/src/components/shared/NotConnectedBanner.tsx:26` — banner promises "read-only until the connection returns", but `canWrite` has no consumer anywhere in the app; offline writes still land in IndexedDB.
- `apps/itun/src/components/account/AccountScreen.tsx:36` — `exportMine` is a standing live subscription on mount, streaming every pilot/mech/pattern body whether or not the user clicks Download.

PR #632, opened 2026-07-28
feat(itun): Games UI — create, invite, join, hand over the table

alxjrvs:
Third in the stack, after #609 (server layer) and #631 (accounts). Contains their commits; those drop out of the diff as each merges. Phase 1 of [ADR-030](https://github.com/SalvageUnion-io/SU-SRD/blob/main/docs/adrs/ADR-030-accounts-games-server-of-record.md).

Wires the server layer to real screens: create a Game, mint an invite code, join with one, appoint a Mediator, hand over Organizer.

**Still inert for existing users** — the whole screen sits behind `isConvexConfigured`, so a build without `VITE_CONVEX_URL` renders a short explanation and no controls at all.

## Design decisions worth review

**Gated on `mode === 'connected'`, deliberately.** A Game is inherently shared state: there is nothing meaningful to show a Solo user, and nothing *safe* to show a Disconnected one — their roster would be a stale snapshot of a crew that may have changed while they were away. This is a design call, not defensive coding.

**The invite code is rendered large and letter-spaced** because reading it aloud across a table is its primary use. That is the same reason the generator is Crockford base32 with no I/L/O/U.

**Join failures surface the server's own wording** rather than a generic message. The server already distinguishes not-valid from expired from used-up, and collapsing those into "could not join" throws away the only information that tells a player what to do next.

## A gate I hit, and what it taught

The pre-push `design-tokens` check rejected `tracking-[0.3em]` and `tracking-[0.2em]` — the repo enforces a canonical tracking ladder and forbids arbitrary values. I did **not** take the exemption route. First pass moved to `tracking-eyebrow`, but the gate's own guidance says that rung is "brand caption only", so the final version uses `tracking-caps-wide` — the widest general rung, still visibly spaced.

Worth noting for the stack: `AccountScreen`/`AccountStrip` in #631 use Tailwind's generic `tracking-widest`/`tracking-wide`, which the gate permits but which are off-ladder. Left alone here to avoid cross-PR diff noise; happy to align them in #631 if you'd prefer.

## Tests

3 more Solo-build render guards (1339 in itun, 3922 across the repo). Same structural failure as the account screen: every Convex hook here throws without a provider above it, which works perfectly for whoever is developing with a configured `.env.local` and crashes for every user who never signs in.

The test also asserts the create and join **inputs are absent** in Solo — so a future refactor cannot leave behind a form that looks usable but cannot possibly work.

## Verification

typecheck 0 errors · knip clean · `format:check` clean · design-token gate clean · **3922 tests pass** · no new lint warnings.

### DIFF
--- a/apps/itun/convex/__tests__/account.test.ts
+++ b/apps/itun/convex/__tests__/account.test.ts
@@ -1,0 +1,216 @@
+import { describe, expect, test } from 'bun:test'
+
+import { api } from '../_generated/api'
+import type { Id } from '../_generated/dataModel'
+import { testConvex } from './harness'
+
+/**
+ * Account management, and in particular the promise D27 makes:
+ * **a campaign never dies because one person quit.**
+ *
+ * Deletion is the operation with no undo, so the cases below are less about
+ * "does it delete" and more about "does it delete *only* the right things" —
+ * the crew's Game, the communal crawler, and other people's characters all
+ * have to survive somebody exercising their right to be forgotten.
+ */
+
+type Ctx = ReturnType<typeof testConvex>
+
+async function makeUser(t: Ctx, name: string) {
+  const userId = await t.run(
+    async (ctx) => await ctx.db.insert('users', { name, displayName: name })
+  )
+  return { userId, as: t.withIdentity({ subject: userId }) }
+}
+
+async function seedPilot(t: Ctx, gameId: Id<'games'> | null, ownerId: Id<'users'> | null) {
+  return await t.run(
+    async (ctx) =>
+      await ctx.db.insert('pilots', { gameId, ownerId, body: { callsign: 'X' }, updatedAt: 1 })
+  )
+}
+
+describe('profile', () => {
+  test('displayName falls back to the Discord name when unset', async () => {
+    const t = testConvex()
+    const u = await makeUser(t, 'Roach-Boy')
+    await u.as.mutation(api.account.updateProfile, { displayName: null })
+
+    const me = await u.as.query(api.account.me, {})
+    // Cleared means "fall back", not "blank" — a nameless owner chip is worse
+    // than a stale one.
+    expect(me?.displayName).toBe('Roach-Boy')
+  })
+
+  test('a set displayName overrides the Discord name', async () => {
+    const t = testConvex()
+    const u = await makeUser(t, 'Roach-Boy')
+    await u.as.mutation(api.account.updateProfile, { displayName: '  Beefcake  ' })
+
+    const me = await u.as.query(api.account.me, {})
+    expect(me?.displayName).toBe('Beefcake')
+  })
+
+  test('an anonymous caller has no profile to read', async () => {
+    const t = testConvex()
+    await expect(t.query(api.account.me, {})).rejects.toThrow(/not signed in/i)
+  })
+})
+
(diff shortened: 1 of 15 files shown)


alxjrvs (commented):
Reviewed `origin/main...HEAD` — 6 commits, 53 files (Convex backend + accounts/Games UI + Eldridge Coast removal). Typecheck passes and all 55 new tests pass; I verified the Tailwind and convex-auth claims empirically rather than from memory.

- `apps/itun/src/lib/connection/ConnectionProvider.tsx:68` — `useConvexAuth().isLoading` is ignored, so a signed-in user resolves to `solo` on every reload and token refresh (confirmed in convex 1.42.3's `ConvexAuthState`: `isAuthenticated` is false while `isLoading` is true). /games shows the "You are playing solo" card and the strip offers "Sign in with Discord" before flipping.
- `apps/itun/netlify.toml:16` — no `convex deploy` step and no `VITE_CONVEX_URL` in the production build, so the shipped bundle has `isConvexConfigured === false` and the entire feature is inert in production.
- `apps/itun/src/components/account/AccountScreen.tsx:140` — `deleteAccount` never signs out or navigates and swallows rejections; the client keeps a valid JWT, the page silently becomes "No account found." and the strip still offers "Sign out".
- `apps/itun/convex/ownership.ts:38` — `loadOwnable` never verifies the doc belongs to the client-supplied `table`; a mech id passed as `table: 'pilots'` writes a Change Log row with `entityType: 'pilot'`. No privilege escalation (the check uses the doc's own gameId), but the audit trail lies.
- `apps/itun/src/components/account/AccountScreen.tsx:36` — `exportMine` is a live subscription on mount, not an on-click fetch; a large account hits Convex's query-result limit and takes the whole account page (including delete) down.
- `apps/itun/src/lib/connection/connectionMode.ts:50` — `disconnected` keys off `navigator.onLine`, so a dead Convex websocket with a live NIC never trips the banner or the write block.
- `apps/itun/convex/__tests__/harness.ts:20` — not `.test.ts`, so Convex codegen treats it as a deployable module (it appears in `_generated/api.d.ts`'s `fullApi`), bundling the `convex-test` devDependency into the deployment.
- `apps/itun/src/components/account/AccountScreen.tsx:30` — `URL.revokeObjectURL` fires synchronously after `a.click()` on a detached anchor; can cancel the download in Firefox/Safari.
- `apps/itun/src/components/shared/NotConnectedBanner.tsx:24` — `font-[var(--font-cond)]` compiles to `font-weight: var(--font-cond)`, not `font-family` (verified against the repo's tailwindcss 4.3.3). Only `font-[var(...)]` call site in the repo; should be `font-cond`.
- `apps/itun/convex/games.ts:142` — `destroy` leaves `mechPatterns`/`changeLog` pointing at the deleted game; `deleteAccount` (account.ts:196) deletes pilots/mechs without cleaning their `softLinks`.
- `apps/itun/src/components/account/AccountStrip.tsx:29` — renders an empty bordered bar in `disconnected` mode.

Non-blocking notes not filed as findings: `apps/itun/CLAUDE.md` says "Two storage modes" then lists three; `VITE_CONVEX_SITE_URL` is in `.env.example` but read nowhere; the Convex client and `@convex-dev/auth/react` are imported eagerly from `__root.tsx`, so Solo builds still download them.

PR #633, opened 2026-07-28
feat(itun): split Workspace into Game + Shelf (v13 migration)

alxjrvs:
Fourth in the stack, after #609 / #631 / #632. **The structurally riskiest piece of Phase 1**: every entity now carries an explicit container instead of a `workspaceId`.

| `gameId` | Means |
| --- | --- |
| `string` | in that Game |
| `null` | on the owner's **Shelf** |
| absent | pre-split record, resolved by `containerOf()` |

## `null` is a value, not an absence

It means *deliberately shelved*; `undefined` means *not yet decided*. They look alike in JavaScript and mean opposite things here.

`containerOf()` checks for `null` explicitly and **stops** — an explicit null beats a stale `workspaceId`. Without that, a deliberate "move to shelf" would silently bounce back to the old game on the next read, and every shelved entity would keep re-entering the legacy fallback forever.

## The Default workspace becomes the Shelf

This is the substantive claim in the migration, and it's a **restoration rather than a reinterpretation**. The Default workspace was never a campaign — it existed because the app required *some* container and these builds belonged to no particular one, which is exactly what a shelf is. Mapping it to a Game would invent a campaign nobody ran and drop every unassigned build into it.

## Migration safety

- **`workspaceId` is retained, not stripped.** The entity schemas are `.strict()`, so a row written by this migration still has to parse on a build that predates it. Removing the key is a separate, irreversible follow-up — the same reasoning ADR-027 used for `equipmentLoadouts`.
- **`DEFAULT_WORKSPACE_ID` is inlined** rather than imported, because migrations may only await IndexedDB operations; an import could let the versionchange transaction commit mid-run. The directory README says so, and this respects it.
- **Idempotent** by checking `gameId === undefined`, so a re-run never drags a deliberately-placed entity back. That case is tested explicitly.

## Two bugs the tests caught, not review

**The registry referenced `migrate13` without importing it.** My import edit had run inside a `cd` that failed, so the symbol was undefined and *every* database upgrade aborted. A fixture test that opens a real v12 database found it on the first run — a migration bug that would have bricked every existing user's app on load.

**The first draft of that test hand-rolled its own upgrade callback** and fired migrations without awaiting them, so the versionchange transaction auto-committed underneath and every case failed with `AbortError`. It now goes through `openItunDatabase` — the real ladder, which cannot drift from what ships.

## Tests

14 new (3936 across the repo): 8 on the container resolver — with the `null` vs `undefined` distinction as the headline case — and 6 fixture tests seeding a real v12 database and re-opening it through the live upgrade path.

## Verification

typecheck 0 errors · knip clean · `format:check` clean · design-token gate clean · **3936 tests pass** · no new lint warnings.

## Not yet done

This lands the *data* half. Routing `entityStore` reads/writes to Convex when connected is the next PR, as are claim-local-data and the Starter Set template. Phases 2–6 follow.

### DIFF
--- a/apps/itun/convex/__tests__/account.test.ts
+++ b/apps/itun/convex/__tests__/account.test.ts
@@ -1,0 +1,216 @@
+import { describe, expect, test } from 'bun:test'
+
+import { api } from '../_generated/api'
+import type { Id } from '../_generated/dataModel'
+import { testConvex } from './harness'
+
+/**
+ * Account management, and in particular the promise D27 makes:
+ * **a campaign never dies because one person quit.**
+ *
+ * Deletion is the operation with no undo, so the cases below are less about
+ * "does it delete" and more about "does it delete *only* the right things" —
+ * the crew's Game, the communal crawler, and other people's characters all
+ * have to survive somebody exercising their right to be forgotten.
+ */
+
+type Ctx = ReturnType<typeof testConvex>
+
+async function makeUser(t: Ctx, name: string) {
+  const userId = await t.run(
+    async (ctx) => await ctx.db.insert('users', { name, displayName: name })
+  )
+  return { userId, as: t.withIdentity({ subject: userId }) }
+}
+
+async function seedPilot(t: Ctx, gameId: Id<'games'> | null, ownerId: Id<'users'> | null) {
+  return await t.run(
+    async (ctx) =>
+      await ctx.db.insert('pilots', { gameId, ownerId, body: { callsign: 'X' }, updatedAt: 1 })
+  )
+}
+
+describe('profile', () => {
+  test('displayName falls back to the Discord name when unset', async () => {
+    const t = testConvex()
+    const u = await makeUser(t, 'Roach-Boy')
+    await u.as.mutation(api.account.updateProfile, { displayName: null })
+
+    const me = await u.as.query(api.account.me, {})
+    // Cleared means "fall back", not "blank" — a nameless owner chip is worse
+    // than a stale one.
+    expect(me?.displayName).toBe('Roach-Boy')
+  })
+
+  test('a set displayName overrides the Discord name', async () => {
+    const t = testConvex()
+    const u = await makeUser(t, 'Roach-Boy')
+    await u.as.mutation(api.account.updateProfile, { displayName: '  Beefcake  ' })
+
+    const me = await u.as.query(api.account.me, {})
+    expect(me?.displayName).toBe('Beefcake')
+  })
+
+  test('an anonymous caller has no profile to read', async () => {
+    const t = testConvex()
+    await expect(t.query(api.account.me, {})).rejects.toThrow(/not signed in/i)
+  })
+})
+
+describe('export is scoped to what you own', () => {
+  test('excludes a crewmate pilot you can merely see', async () => {
+    const t = testConvex()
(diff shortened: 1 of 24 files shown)


alxjrvs (commented):
Reviewed the 7-commit accounts/Games branch vs `origin/main` (62 files: new Convex backend, connection-mode layer, account/games UI, v13 IndexedDB migration, Eldridge Coast removal). Verified the gates locally — typecheck, lint, knip, and all 1353 itun tests (incl. 31 new Convex tests) pass.

**Findings** (`file:line — summary`):

1. `apps/itun/convex/__tests__/harness.ts:1` — The test harness is a real Convex entry point: dumping the actual push bundle shows `__tests__/harness.js` at 62KB (the whole `convex-test` mock backend + a `node:async_hooks` shim) shipping to the deployment. Convex only skips multi-dot filenames, so `*.test.ts` is excluded but `harness.ts` isn't. Also makes the `convex-test` devDependency a hard deploy-time requirement.
2. `apps/itun/src/components/shared/NotConnectedBanner.tsx:27` — The banner promises "your games are read-only", but `canWrite`/`writesAllowed`/`usesServerOfRecord` are consumed nowhere outside their own tests; offline writes still land in IndexedDB.
3. `apps/itun/src/lib/connection/ConnectionProvider.tsx:30` — `disconnected` is derived from `navigator.onLine`, not Convex reachability. Captive portal or a down deployment keeps mode `connected`, hides the banner, and leaves every query stuck on "Loading…".
4. `apps/itun/convex/ownership.ts:38` — `loadOwnable` never checks that `entityId` belongs to `table`; a pilots/mechs mix-up passes every guard and writes the wrong `entityType` to the Change Log, silently losing the ownership event from provenance views.
5. `apps/itun/convex/games.ts:156` — `games.destroy` and `account.deleteAccount` never touch `changeLog`, leaving rows with dangling `gameId`/`actorId`. Reachable today via `ownership.assign` → `games.destroy`.
6. `apps/itun/src/components/account/AccountScreen.tsx:23` — Local `download()` reimplements the existing `src/lib/export/downloadJson.ts` without its DOM-append, delayed `revokeObjectURL`, or SSR guard; the "Download my data" export can abort on Safari.
7. `apps/itun/src/components/games/GamesScreen.tsx:98` — Rename Save has no empty-name guard (unlike Create) and the server rejection is swallowed by `void`, so a failed rename is invisible.
8. `apps/itun/src/components/account/AccountScreen.tsx:36` — `exportMine` is subscribed on mount, pulling and live-syncing every owned entity body just to enable a button.
9. `apps/itun/tsconfig.json:19` — `convex/**` isn't in `include`; probe confirmed a blatant type error in `convex/__tests__/permissions.test.ts` doesn't fail `tsc` (function modules are checked transitively via `_generated/api.d.ts`, but the tests aren't).

The permission layer itself reads correctly — `requireMember`/`requireMediator`/`requireOrganizer` are called before every write, the Organizer-fallback bend in `requireOwnershipAssigner` is properly conditional, `transferOrganizer` is single-transaction, and the invite-code reuse of `generateUniqueId` (uppercase Crockford, matching `redeem`'s `toUpperCase()`) is sound. The v13 migration is idempotent and correctly maps the Default workspace to the shelf; the Eldridge Coast removal left no dangling references.

PR #634, opened 2026-07-28
feat(itun): Game templates — the Starter Set arrives unclaimed

alxjrvs:
Fifth in the stack (#609 → #631 → #632 → #633 → this). Phase 1 of [ADR-030](https://github.com/SalvageUnion-io/SU-SRD/blob/main/docs/adrs/ADR-030-accounts-games-server-of-record.md), decision **D34**.

## What makes a template more than seeded data

**Its entities arrive with `ownerId: null`.** The Mediator — or the Organizer, while a Game has no Mediator — hands them out as players join.

That single property turns the six pre-gens of Union Crawler #430 'Tenacity' from a curiosity buried in a workspace switcher into the actual onboarding path. A template that pre-assigned everything to its creator would just be a private roster with extra steps, so that's the property the tests pin hardest.

## One copy of the roster, not two

The crew is imported straight from `src/lib/starterSet/`. That module is deliberately static plain data with no runtime reference-data resolution (its header says so), so it bundles into a Convex function safely — and there is exactly **one** copy rather than a server-side duplicate that would drift from the local one on the next data fix.

## Details that follow from earlier decisions

- **The crawler is inserted with no `ownerId` column at all**, because it is communal (D8). Tested, so a future refactor cannot quietly give the crew's home an owner.
- **Soft links come across**, so pilots and mechs arrive wired to each other and to the crawler. Without them the template would deliver a set of unrelated strangers.
- **`templateOrigin`** records which template a Game came from — what later lets anything tell a seeded crawler from a hand-built one.

## Tests

8 new, covering unclaimed-on-arrival, the creator being Organizer-and-Player (a template changes what is *in* a Game, never who runs it), the communal crawler, soft links surviving, custom-name override, and the anonymous-caller rejection.

## Verification

typecheck 4/4 · knip clean · `format:check` clean · design-token gate clean · full suite green · no new lint warnings. Re-verified after rebasing across 8 newly-landed commits on `main`.

## Remaining in Phase 1

Routing `entityStore` reads/writes to Convex when connected, and claim-local-data on first sign-in. Then Phases 2–6.

### DIFF
--- a/apps/itun/convex/__tests__/account.test.ts
+++ b/apps/itun/convex/__tests__/account.test.ts
@@ -1,0 +1,216 @@
+import { describe, expect, test } from 'bun:test'
+
+import { api } from '../_generated/api'
+import type { Id } from '../_generated/dataModel'
+import { testConvex } from './harness'
+
+/**
+ * Account management, and in particular the promise D27 makes:
+ * **a campaign never dies because one person quit.**
+ *
+ * Deletion is the operation with no undo, so the cases below are less about
+ * "does it delete" and more about "does it delete *only* the right things" —
+ * the crew's Game, the communal crawler, and other people's characters all
+ * have to survive somebody exercising their right to be forgotten.
+ */
+
+type Ctx = ReturnType<typeof testConvex>
+
+async function makeUser(t: Ctx, name: string) {
+  const userId = await t.run(
+    async (ctx) => await ctx.db.insert('users', { name, displayName: name })
+  )
+  return { userId, as: t.withIdentity({ subject: userId }) }
+}
+
+async function seedPilot(t: Ctx, gameId: Id<'games'> | null, ownerId: Id<'users'> | null) {
+  return await t.run(
+    async (ctx) =>
+      await ctx.db.insert('pilots', { gameId, ownerId, body: { callsign: 'X' }, updatedAt: 1 })
+  )
+}
+
+describe('profile', () => {
+  test('displayName falls back to the Discord name when unset', async () => {
+    const t = testConvex()
+    const u = await makeUser(t, 'Roach-Boy')
+    await u.as.mutation(api.account.updateProfile, { displayName: null })
+
+    const me = await u.as.query(api.account.me, {})
+    // Cleared means "fall back", not "blank" — a nameless owner chip is worse
+    // than a stale one.
+    expect(me?.displayName).toBe('Roach-Boy')
+  })
+
+  test('a set displayName overrides the Discord name', async () => {
+    const t = testConvex()
+    const u = await makeUser(t, 'Roach-Boy')
+    await u.as.mutation(api.account.updateProfile, { displayName: '  Beefcake  ' })
+
+    const me = await u.as.query(api.account.me, {})
+    expect(me?.displayName).toBe('Beefcake')
+  })
+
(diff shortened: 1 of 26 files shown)


alxjrvs (commented):
Reviewed the 8 commits on `itun-accounts-templates` vs `origin/main` (64 files: Convex backend + auth, connection-mode layer, Games/Account UI, v13 IndexedDB migration, Eldridge Coast removal).

**Findings** (most severe first):

- `apps/itun/convex/__tests__/harness.ts:1` — Test harness ships as a real Convex module (`_generated/api.d.ts:11,29` already registers it), bundling the `convex-test` devDependency into the deployment; `convex deploy` without dev deps fails.
- `apps/itun/convex/model/permissions.ts:30` — `NotAuthorized` is a plain `Error`, not `ConvexError`, so its message is redacted in production; the invite-join UI that surfaces `err.message` will show "Server Error" instead of "That invite code has expired".
- `apps/itun/src/components/shared/NotConnectedBanner.tsx:27` — Banner says games are read-only, but `canWrite` is consumed nowhere in `src/`; offline writes still succeed.
- `apps/itun/src/lib/connection/ConnectionProvider.tsx:68` — `useConvexAuth().isLoading` is ignored, so signed-in users flash Solo state ("You are playing solo… needs no account" + a Sign in button) on every reload.
- `apps/itun/src/components/account/AccountScreen.tsx:23` — Local `download()` reimplements `lib/export/downloadJson.ts` but drops the DOM-attach and the deferred `revokeObjectURL`, breaking the export download on Safari/Firefox.
- `apps/itun/src/components/games/GamesScreen.tsx:98` — Every mutation but `redeem` is `void mutation(...)` with no `.catch`; failures (incl. `deleteAccount`) are unhandled rejections with zero user feedback.
- `apps/itun/convex/templates.ts:83` — Template entities are inserted verbatim: deterministic ids collide across every Game seeded from the Starter Set (`softLinks.by_from` / `changeLog.by_entity` return other Games' rows), and the copied `workspaceId` makes `containerOf` report a nonexistent Game.
- `apps/itun/convex/ownership.ts:38` — `loadOwnable` never checks `entityId` against the `table` arg (no `normalizeId`); a mech id passed as `table:'pilots'` succeeds and is logged as `entityType: 'pilot'`.
- `apps/itun/convex/games.ts:140` — `destroy` / `deleteAccount` never sweep `mechPatterns` or `changeLog` (and leave `presence` rows for deleted users), orphaning rows pointing at deleted games/users.
- `apps/itun/src/lib/connection/connectionMode.ts:50` — Mode is derived from `navigator.onLine` only; a reachable-network-but-unreachable-deployment case reads as `connected`, no banner, and every mutation rejects.
- `apps/itun/src/components/games/GamesScreen.tsx:205` — Template button hardcodes `templateId: 'starter-set'` instead of `t.id`, and names the game from the unrelated "Start a game" input.
- `apps/itun/src/lib/db/migrations/13-workspace-to-game-or-shelf.ts:32` — `DEFAULT_WORKSPACE_ID` inlined on a stated constraint that migration 10 disproves (it imports the constant); silent drift risk.

Nothing blocking was found in the permission model itself — `requireMember`/`requireOrganizer`/`requireOwnershipAssigner` are called on every mutation, the transfer-organizer invariant holds, and the authorization test suite covers the real mutation paths.

PR #635, opened 2026-07-28
feat(itun): activated contributions with manual expiry — Squeeze It In, Hull Magnetiser

alxjrvs:
**F1** of the rules-parity plan (#599) — fourth in the recommended order, after #628, #629, #630.

The last two duration-bound records move from *"exempt with a reason"* to **genuinely applied**:

```
Rules parity: 30 claim(s) — 16 encoded, 12 exempt, 2 unresolved   (was 14 / 14 / 2)
```

## Manual expiry, by decision

Salvage Union states real durations — *"this effect lasts for 1 hour"*, *"for 12 hours"* — but the app has **no play clock**. Inventing one would put wall-time into the data layer and make a sheet's numbers change while nobody is looking.

The table keeps time; the app keeps state — the same division ADR-001's honour system already relies on. A player switches an effect on and off.

The switch lives in `playStateStore`, **ephemeral** (ADR-019), so an activated effect can never persist onto the entity and never leaks into a live sheet or a shared snapshot.

## Two schema additions, each forced by a real record

| Addition | Why |
|---|---|
| `duration: 'activated'` | Applies only while switched on. `permanent` (default) applies whenever held or installed. |
| `amount: { fromStat }` | The amount **is** another of the target's stats. Hull Magnetiser raises Cargo *"by its System Slot Value"* — a number that changes with the chassis, so it cannot be a constant. An unavailable stat resolves to **0, never a guess**. |

`installedContributions` is the systems/modules counterpart to `abilityContributions`, and is deliberately distinct from `statBonus`: that's a flat per-copy number with no target, duration or expression — exactly what Hull Magnetiser can't be expressed as. Per-copy stacking defaults to true, matching `statBonus` semantics.

## UI

The mech band grows an **Effects** bay listing only what this loadout and pilot can actually switch on, so a permanent contribution never offers a pointless switch.

## Verification

`bun run check:all` green: schemas, lint, format, typecheck, **4,593 tests**, validate (incl. the parity gate), knip, audit, tokens, styling.

Tests cover both directions of the gate (off contributes nothing, on contributes), that a permanent contribution is unaffected by the active map, that Hull Magnetiser resolves from the chassis slot value, and that a `fromStat` with no stat available resolves to 0.

Next: **F5** (Downtime writes), then **effect targeting**.

Refs #599.

### DIFF
--- a/apps/itun/src/components/dashboard/ActiveItemBand.tsx
+++ b/apps/itun/src/components/dashboard/ActiveItemBand.tsx
@@ -34,6 +34,7 @@ import type { Mech } from '../../lib/schemas/mech'
 import type { Pilot } from '../../lib/schemas/pilot'
 import { useEntityStore } from '../../stores/entityStore'
 import type { EntityState } from '../../stores/entityStore'
+import { activatableEffects } from './dashboardEffects'
 import { usePlayStateStore } from '../../stores/playStateStore'
 import { ActiveItemBand as ActiveItemBandView, CountStepper, StorageBay } from 'component-lib'
 import type { ActiveItemBandView as ActiveItemBandViewModel, BandButton } from 'component-lib'
@@ -349,7 +350,12 @@ function MechBand({
   onDismount: () => void
 }) {
   const chassis = resolveChassisRef(mech.chassisRef)
-  const piloting = pilotingContext(mech, pilotAbilities)
+  const activeEffects = usePlayStateStore((st) => st.activeEffects)
+  const toggleEffect = usePlayStateStore((st) => st.toggleEffect)
+  const piloting = { ...pilotingContext(mech, pilotAbilities), active: activeEffects }
+  // What this mech/pilot could switch on (F1). Manual expiry: the table keeps
+  // time, the app keeps state.
+  const activatable = activatableEffects(mech, pilotAbilities)
   const maxSP = mechMaxSP(mech, chassis, piloting)
   const maxEP = mechMaxEP(mech, chassis)
   const maxHeat = mechMaxHeat(mech, chassis)
@@ -625,6 +631,20 @@ function MechBand({
           },
         ],
       },
+      ...(activatable.length > 0
+        ? [
+            {
+              label: 'Effects',
+              buttons: activatable.map((e) => ({
+                label: `${activeEffects[e.ref] ? '\u25CF' : '\u25CB'} ${e.name}`,
+                onClick: () => toggleEffect(e.ref),
+                title: activeEffects[e.ref]
+                  ? `${e.name} is active — click to end it`
+                  : `${e.name}: ${e.summary}`,
+              })),
+            },
+          ]
+        : []),
       {
         label: 'Egress',
         buttons: [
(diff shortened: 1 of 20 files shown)


alxjrvs (commented):
**Scope reviewed:** commit `3922af35` — "feat(itun): activated contributions with manual expiry — Squeeze It In, Hull Magnetiser" (single commit vs `origin/main`; working tree clean). Note: another session switched this worktree onto branch `parity-effect-targeting` mid-review, so I reviewed the commit contents via `git show`.

1. `packages/salvageunion-reference/data/abilities.json:448` — "Squeeze it in" raises the Cargo Capacity of *a Mech in Range*, but is encoded `target: 'pilotedMech'`, so it always lands on the holder's own mech. Applying it to an ally inflates the wrong sheet — the exact "target-applied" class the audit tool exempts elsewhere and that `KNOWN_UNRESOLVED` explicitly warns not to encode without a target mechanism.
2. `packages/salvageunion-reference/lib/rules/contributions.ts:190` — `installedContributions` ignores `systemConditions`/`moduleConditions`, and `voidWhen` is read by no code in the repo. A *destroyed* Hull Magnetiser is still listed in the Effects bay and still grants +System Slot Value cargo.
3. `packages/salvageunion-reference/lib/rules/contributions.ts:195` — activated contributions are multiplied by installed copies (`each * copies`) while the UI de-dupes to one switch: two Hull Magnetisers = one click for 2× the bonus, with no way to activate just one.
4. `apps/itun/src/stores/playStateStore.ts:91` — `activeEffects` is keyed by ref only, in one app-wide store, never cleared. Switching mounts/builds carries the "on" state onto any other mech holding the same ref, and it survives uninstall → reinstall.
5. `apps/itun/src/components/dashboard/ActiveItemBand.tsx:360` — `maxEP`/`maxHeat` are computed without `piloting`, so any activated `energyPoints`/`heatCapacity` effect gets a toggle that does nothing (and `pushLocked` reads that same stale `maxHeat`). Pre-existing omission, newly reachable.
6. `packages/salvageunion-reference/lib/rules/contributions.ts:117` — the new `stats` parameter on `abilityContributions` is passed by no caller, so a `fromStat` amount on an ability silently resolves to 0 and is dropped.
7. `packages/salvageunion-reference/lib/rules/contributions.ts:144` — `stacks: true` in the new ability data is inert (`copies: 1` is hard-coded), yet the rules explicitly allow repeat activations on the same mech; the cap can never exceed +4.
8. `apps/itun/src/components/dashboard/dashboardEffects.ts:62` — re-introduces per-render linear `.find()` catalog scans instead of the memoized `resolveInstalledRef` maps added for exactly this path (audit item 18), and unlike its siblings has no try/catch, so an unloaded schema throws out of render.
9. `apps/itun/src/components/dashboard/dashboardEffects.ts:64` — de-duping by ref lets the last contribution overwrite earlier ones, so a record with two activated contributions shows a tooltip describing only one of them.

Verified as fine: generated JSON schemas are in sync with the Zod change; the new tests' imports and expected arithmetic check out against the real dataset (chassis records do carry `systemSlots`/`moduleSlots`); no double-count between `statBonus` and the new `contributions` (Hull Magnetiser declares only the latter); removing the two parity exemptions is consistent with `hasCapData`.

PR #636, opened 2026-07-28
feat(itun): Convex-backed entity CRUD and claim-local-data

alxjrvs:
Sixth in the stack (#609 → #631 → #632 → #633 → #634 → this). Closes the remaining **Phase 1 server work** for [ADR-030](https://github.com/SalvageUnion-io/SU-SRD/blob/main/docs/adrs/ADR-030-accounts-games-server-of-record.md).

## Every write Zod-parses first

The schema stores bodies as `v.any()` so the Zod schemas stay the single source of truth rather than forking into a second hand-maintained set of Convex validators. The schema header states the price plainly — **Convex cannot reject a malformed body, so the mutation must** — and `convex/entities.ts` is where that obligation is paid.

Tested, because if that check is ever dropped nothing else in the system notices until a corrupt row reaches a sheet and blanks it.

## Reading is per-Game, writing is per-entity

Any member sees every pilot and mech in a Game — that's what makes crew vitals and read-only drill-in possible. But **nobody writes a crewmate's sheet, and there is deliberately no privileged write path in this module.** A Mediator wanting to change someone else's sheet goes through a proposal (D7).

`assertMayWrite` is **synchronous and ctx-free** for exactly that reason: the answer depends only on the row's own `ownerId`, and handing it a `ctx` would invite the lookup that turns into a back door.

**An unclaimed entity cannot be edited until assigned** — otherwise anyone could quietly take a pre-gen by editing it, bypassing the assignment act the Change Log records.

## The crawler merges per field (D19)

Last-write-wins fails on exactly the night it matters: during Downtime the whole crew touches the crawler within the same few minutes, and a full-body write silently discards whoever loses the race. The test has two members editing different fields and asserts **both survive**.

## claim-local-data (D11)

Everything lands on the **shelf**, never in a Game — a first-time sign-in has builds with no relationship to any crew, and guessing one is worse than making the person place it deliberately.

A row that fails to parse is **skipped rather than aborting the claim**, and the count comes back so the UI can say what didn't make it. One corrupt local record should not cost somebody their whole roster.

## Three fixture bugs the tests surfaced

All from inventing shapes instead of reading them — worth noting because the same mistake in production code would have been silent:

- `PilotSchema` has no `currentHp` / `trainingPoints` / `abilityRefs`, and *does* require `classRef` / `motto` / `keepsake` / `appearance`.
- `CrawlerSchema` takes `techLevel` as a **string**, requires `systems`, and has no `modules` key.

Both fixtures are now derived by probing the real schemas.

## Verification

typecheck 4/4 · knip clean · `format:check` clean · design-token gate clean · full suite green · lint back to the 4-warning baseline.

## Remaining

Swapping `entityStore`'s write-through to route through these functions when connected (the store's public API stays identical), then Phases 2–6.

### DIFF
--- a/apps/itun/CHANGELOG.md
+++ b/apps/itun/CHANGELOG.md
@@ -4,82 +4,71 @@ Maintained by release-please (see ADR-024).
 
 ## [0.7.0](https://github.com/SalvageUnion-io/SU-SRD/compare/itun-v0.6.0...itun-v0.7.0) (2026-07-28)
 
-
 ### Features
 
-* **component-lib:** make the roll-table title its own table picker ([#592](https://github.com/SalvageUnion-io/SU-SRD/issues/592)) ([18f315a](https://github.com/SalvageUnion-io/SU-SRD/commit/18f315ac0501f00f52db63181dbf383d66909245))
-* **itun:** render partners in place, delete the partner sheet ([#590](https://github.com/SalvageUnion-io/SU-SRD/issues/590)) ([33ddccd](https://github.com/SalvageUnion-io/SU-SRD/commit/33ddccd79c6b9db417dff0384f27188d3fe0b67f))
-* **itun:** render the SRD home page in the Dashboard SRD Explorer ([#593](https://github.com/SalvageUnion-io/SU-SRD/issues/593)) ([05906b6](https://github.com/SalvageUnion-io/SU-SRD/commit/05906b629fe97708f5c9faf8e638e29033e363e9))
+- **component-lib:** make the roll-table title its own table picker ([#592](https://github.com/SalvageUnion-io/SU-SRD/issues/592)) ([18f315a](https://github.com/SalvageUnion-io/SU-SRD/commit/18f315ac0501f00f52db63181dbf383d66909245))
+- **itun:** render partners in place, delete the partner sheet ([#590](https://github.com/SalvageUnion-io/SU-SRD/issues/590)) ([33ddccd](https://github.com/SalvageUnion-io/SU-SRD/commit/33ddccd79c6b9db417dff0384f27188d3fe0b67f))
+- **itun:** render the SRD home page in the Dashboard SRD Explorer ([#593](https://github.com/SalvageUnion-io/SU-SRD/issues/593)) ([05906b6](https://github.com/SalvageUnion-io/SU-SRD/commit/05906b629fe97708f5c9faf8e638e29033e363e9))
 
 ## [0.6.0](https://github.com/SalvageUnion-io/SU-SRD/compare/itun-v0.5.0...itun-v0.6.0) (2026-07-28)
 
-
 ### Features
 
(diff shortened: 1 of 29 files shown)


alxjrvs (commented):
Reviewed `origin/main...HEAD` (9 commits, ~5,000 added lines: Convex backend, auth, Games UI, v13 migration). Verified state: `bun run typecheck` clean, `bun test` in `apps/itun` 1373 pass / 0 fail, `bun run knip` clean. Finding #1 was reproduced by running the real mutation under convex-test.

Findings, most severe first:

- `apps/itun/convex/entities.ts:173` — `update`/`remove` validate the body against `args.table` but load the doc by raw id and never check it belongs to that table. Reproduced: `update({table:'pilots', entityId:<owned mechs id>, body:<valid Pilot>})` stores a pilot body in the `mechs` row; the game's later reads fail `MechSchema`.
- `apps/itun/src/components/games/GamesScreen.tsx:206` — the template list maps over `t` but the Start button hardcodes `templateId: 'starter-set'`; a second template would display its own name and create the Starter Set.
- `apps/itun/src/lib/connection/ConnectionProvider.tsx:68` — `useConvexAuth().isLoading` is discarded, so a signed-in user sees "You are playing solo" + "Sign in with Discord" on every load until the auth round-trip lands.
- `apps/itun/convex/games.ts:144` — `destroy` preserves owned pilots/mechs to the shelf but deletes all game-scoped `softLinks`, severing the mech↔pilot pairings of the very entities it preserved.
- `apps/itun/convex/games.ts:141` — neither `destroy` nor `deleteAccount`'s last-member branch cleans `mechPatterns` or `changeLog`, leaving rows with a `gameId` pointing at a deleted game (unreachable via both shelf and by-game reads).
- `apps/itun/convex/templates.ts:84` — starter bodies are inserted verbatim, keeping `workspaceId: 'starter-set-workspace'` (so `containerOf(body)` reports a nonexistent game) and fixed `starter-*` ids duplicated across every Game seeded from the template.
- `apps/itun/src/lib/db/migrations/13-workspace-to-game-or-shelf.ts:39` — the migration writes `gameId` but `entityStore`/`workspaceStore` still write only `workspaceId`, so the two diverge on the first workspace reassignment and `containerOf` (which prefers `gameId`) reports the stale container.
- `apps/itun/src/components/account/AccountScreen.tsx:140` — `deleteAccount` is fired with no `signOut()` or navigation; the still-mounted `me`/`listMine`/`exportMine` subscriptions re-run for a deleted user and rethrow during render.
- `apps/itun/src/components/shared/NotConnectedBanner.tsx:27` — banner claims "read-only", but `canWrite`/`usesServerOfRecord` have no consumers outside their tests; offline writes still succeed.
- `apps/itun/convex/entities.ts:239` — `claimLocal` has no idempotency key, so a retried or second-device claim duplicates the entire roster with identical `body.id` values.

PR #638, opened 2026-07-28
feat(itun): enforce read-only writes when disconnected

alxjrvs:
Seventh in the stack (#609 → #631 → #632 → #633 → #634 → #636 → this).

`entityStore` now calls `requireWritableBackend()` before every create, update and delete. A signed-in user who loses connectivity is **read-only** (D14), not quietly writing to a local copy that would fork against the server of record — which is precisely the conflict resolution the server-of-record decision exists to avoid.

**Solo remains the default structurally.** `selectBackend()` returns `local` unless the app is genuinely Connected: not signed in, no Convex URL compiled in, offline — all resolve to local. That ordering is deliberate, because the failure mode in the *other* direction is a Solo user's writes silently going nowhere, which is the worst outcome available in this migration. The tests lean on that side.

The store cannot call hooks, so `ConnectionProvider` **pushes** the mode down in an effect rather than the store pulling it out. One writer, one direction, and no side effect during render.

## What this PR deliberately does not do

**Write-mirroring** — and the reason is now recorded in `entityBackend.ts` so nobody rediscovers it.

The obvious next step is to write locally and mirror the same change to Convex. It doesn't work. Local records are keyed by an app-level UUID from `crud.ts`; Convex rows are keyed by its own `_id`. A **create** can be mirrored because it needs no prior key, but an **update or delete has nothing to address** — there is no id map, so the mirror would either silently no-op or need a lookup-by-body on every write.

A mirror that works for creates and quietly fails for edits is **worse than no mirror, because it looks synced**. I built it, hit this, and removed it rather than shipping something that would pass review and lose data. `knip` flagged the orphaned helper, which is the dead-code gate doing exactly its job.

The real cutover is the one ADR-030 describes — when Connected, Convex *is* the source of truth and the store reads from a reactive subscription rather than IndexedDB. That needs the id question settled first: a mapping table, or app-id as the Convex primary key. **That is a decision worth your input rather than mine**, which is why this PR stops here.

## Verification

typecheck 4/4 · knip clean · `format:check` clean · design-token gate clean · full suite green · lint at the 4-warning baseline. Re-verified after rebasing across 9 newly-landed commits on `main`.

### DIFF
--- a/apps/itun/CHANGELOG.md
+++ b/apps/itun/CHANGELOG.md
@@ -4,82 +4,71 @@ Maintained by release-please (see ADR-024).
 
 ## [0.7.0](https://github.com/SalvageUnion-io/SU-SRD/compare/itun-v0.6.0...itun-v0.7.0) (2026-07-28)
 
-
 ### Features
 
-* **component-lib:** make the roll-table title its own table picker ([#592](https://github.com/SalvageUnion-io/SU-SRD/issues/592)) ([18f315a](https://github.com/SalvageUnion-io/SU-SRD/commit/18f315ac0501f00f52db63181dbf383d66909245))
-* **itun:** render partners in place, delete the partner sheet ([#590](https://github.com/SalvageUnion-io/SU-SRD/issues/590)) ([33ddccd](https://github.com/SalvageUnion-io/SU-SRD/commit/33ddccd79c6b9db417dff0384f27188d3fe0b67f))
-* **itun:** render the SRD home page in the Dashboard SRD Explorer ([#593](https://github.com/SalvageUnion-io/SU-SRD/issues/593)) ([05906b6](https://github.com/SalvageUnion-io/SU-SRD/commit/05906b629fe97708f5c9faf8e638e29033e363e9))
+- **component-lib:** make the roll-table title its own table picker ([#592](https://github.com/SalvageUnion-io/SU-SRD/issues/592)) ([18f315a](https://github.com/SalvageUnion-io/SU-SRD/commit/18f315ac0501f00f52db63181dbf383d66909245))
+- **itun:** render partners in place, delete the partner sheet ([#590](https://github.com/SalvageUnion-io/SU-SRD/issues/590)) ([33ddccd](https://github.com/SalvageUnion-io/SU-SRD/commit/33ddccd79c6b9db417dff0384f27188d3fe0b67f))
+- **itun:** render the SRD home page in the Dashboard SRD Explorer ([#593](https://github.com/SalvageUnion-io/SU-SRD/issues/593)) ([05906b6](https://github.com/SalvageUnion-io/SU-SRD/commit/05906b629fe97708f5c9faf8e638e29033e363e9))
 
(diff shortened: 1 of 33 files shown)


alxjrvs (commented):
Reviewed the 10-commit `itun-accounts-store` branch against `origin/main` (70 files, ~5.2k added lines: a new Convex backend, auth, Games/ownership/permissions, the connection-mode layer, and a v13 IndexedDB migration). `bun run typecheck`, `bun --filter itun test` (1385 pass) and `bun run knip` all pass locally, so everything below is behavioural rather than build-breaking.

**Findings (13):**

- `apps/itun/netlify.toml:111` — production CSP `connect-src` was never widened for Convex, so the `wss://*.convex.cloud/api/sync` socket is blocked and the whole accounts feature is inert in prod.
- `apps/itun/src/stores/entityStore.ts:344` — the offline write block fires on data that only ever lives in IndexedDB (no remote write path exists yet), so a signed-in user who goes offline loses the ability to edit their own local records.
- `apps/itun/src/stores/entityBackend.ts:99` — nothing in the app reads `canWrite` or catches `WritesBlockedOffline`; `void storeState.update(...)` call sites produce unhandled rejections and silently dropped edits.
- `apps/itun/convex/entities.ts:173` — `update`/`remove` validate against `args.table` but resolve the doc by raw id without checking they agree; a Mech body can be written into a Pilot row (same gap in `ownership.ts:loadOwnable`).
- `apps/itun/convex/__tests__/harness.ts:1` — registered in `_generated/api.d.ts` as a deployable module, so `convex deploy` bundles test code plus the `convex-test` devDependency.
- `apps/itun/convex/templates.ts:83` — template bodies inserted unparsed with hard-coded entity ids and a stale `workspaceId`; two Games from one template collide on ids and `containerOf` resolves the wrong container.
- `apps/itun/src/components/account/AccountScreen.tsx:23` — duplicate `download()` helper drops the DOM-attach + deferred-revoke that `lib/export/downloadJson.ts` documents for Safari; "Download my data" fails there.
- `apps/itun/convex/model/permissions.ts:33` — plain `Error` instead of `ConvexError`, so production redacts the messages `GamesScreen` deliberately renders to users (PLAUSIBLE; not verifiable locally).
- `apps/itun/src/stores/entityStore.ts:377` — `transfer` and `updateCrawlerBay` skip the offline guard, so cargo moves and bay edits still write while "read-only".
- `apps/itun/src/components/games/GamesScreen.tsx:205` — template button hardcodes `'starter-set'` instead of `t.id` and names the game from the unrelated "Start a game" input.
- `apps/itun/src/lib/connection/ConnectionProvider.tsx:77` — `useConvexAuth().isLoading` discarded, so a signed-in user reads as Solo during the handshake.
- `apps/itun/convex/games.ts:141` — `destroy`/`deleteAccount` leave `mechPatterns` and `changeLog` pointing at a deleted game, and shelving doesn't bump `updatedAt`.
- `apps/itun/src/lib/db/migrations/13-workspace-to-game-or-shelf.ts:48` — `workspaceStore.assign` still patches only `workspaceId`, so the migrated `gameId` (which `containerOf` prefers) goes stale on the first reassignment.

PR #639, opened 2026-07-28
feat(itun): address server rows by app id, and mirror writes that work

alxjrvs:
Eighth in the stack. **Completes Phase 1's write path** and resolves the blocker I flagged in #638.

## The decision I took

#638 stopped because Convex mints its own `_id`, so a client holding only its local UUID had nothing to address a row by — creates mirrored, edits silently no-opped. I asked whether to use the app UUID as the Convex primary key or add a mapping table.

**Neither, exactly.** Convex will always mint `_id`, so option (a) isn't literally available. An **indexed `appId` column** gives (b)'s correctness at (a)'s cost: one cheap lookup per write, no mapping table, and `_id` stays idiomatic for everything server-side. `appId` is optional because rows created server-side (Game templates) have no local counterpart until somebody claims them.

Say the word if you'd rather have the explicit mapping table; it's a contained change at this point.

## Why `upsertByAppId` is an upsert

This is what makes the mirror **converge**. An entity built while Solo has no server row until the account is claimed, so a missing row means "create it" rather than "fail" — otherwise the very first edit after a claim would be silently dropped, which is the same class of bug #638 removed.

## The security consequence of client-supplied ids

Addressing rows by an id the *client* chooses means the ownership check has to be exactly as strict as before, or it becomes a way to write somebody else's data by guessing their UUID. **Two tests cover that directly**, for both upsert and delete.

## Fire-and-forget, deliberately

The local write already succeeded and is what the UI reads, so a mirror failure warns and moves on rather than rolling back a write the user watched succeed. Deleting something already deleted is a no-op for the same reason — the mirror may retry or arrive out of order.

## Tests

7 new (3991 across the repo): create-when-absent, update-not-duplicate, cross-user rejection on both paths, malformed-body rejection, and the already-deleted no-op.

## Verification

typecheck 4/4 · knip clean · `format:check` clean · design-token gate clean · **3991 tests pass** · lint at the 4-warning baseline.

### DIFF
--- a/apps/itun/CHANGELOG.md
+++ b/apps/itun/CHANGELOG.md
@@ -4,82 +4,71 @@ Maintained by release-please (see ADR-024).
 
 ## [0.7.0](https://github.com/SalvageUnion-io/SU-SRD/compare/itun-v0.6.0...itun-v0.7.0) (2026-07-28)
 
-
 ### Features
 
-* **component-lib:** make the roll-table title its own table picker ([#592](https://github.com/SalvageUnion-io/SU-SRD/issues/592)) ([18f315a](https://github.com/SalvageUnion-io/SU-SRD/commit/18f315ac0501f00f52db63181dbf383d66909245))
-* **itun:** render partners in place, delete the partner sheet ([#590](https://github.com/SalvageUnion-io/SU-SRD/issues/590)) ([33ddccd](https://github.com/SalvageUnion-io/SU-SRD/commit/33ddccd79c6b9db417dff0384f27188d3fe0b67f))
-* **itun:** render the SRD home page in the Dashboard SRD Explorer ([#593](https://github.com/SalvageUnion-io/SU-SRD/issues/593)) ([05906b6](https://github.com/SalvageUnion-io/SU-SRD/commit/05906b629fe97708f5c9faf8e638e29033e363e9))
+- **component-lib:** make the roll-table title its own table picker ([#592](https://github.com/SalvageUnion-io/SU-SRD/issues/592)) ([18f315a](https://github.com/SalvageUnion-io/SU-SRD/commit/18f315ac0501f00f52db63181dbf383d66909245))
+- **itun:** render partners in place, delete the partner sheet ([#590](https://github.com/SalvageUnion-io/SU-SRD/issues/590)) ([33ddccd](https://github.com/SalvageUnion-io/SU-SRD/commit/33ddccd79c6b9db417dff0384f27188d3fe0b67f))
+- **itun:** render the SRD home page in the Dashboard SRD Explorer ([#593](https://github.com/SalvageUnion-io/SU-SRD/issues/593)) ([05906b6](https://github.com/SalvageUnion-io/SU-SRD/commit/05906b629fe97708f5c9faf8e638e29033e363e9))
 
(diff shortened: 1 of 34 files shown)


alxjrvs (commented):
Reviewed the 11-commit accounts/Games/appId branch against `origin/main` (71 files: new `apps/itun/convex/` server layer, connection-mode client layer, v13 migration, Games/Account UI, Eldridge Coast removal). Two findings were confirmed by executing the real mutations under `convex-test` (scratch test written and deleted). `bun run knip` passes.

Findings:

- `apps/itun/convex/entities.ts:176` — `update`/`remove` never verify the row belongs to `args.table`; **confirmed by experiment** that a valid pilot body can be written into a `mechs` row you own, corrupting it. `ownership.loadOwnable` has the same hole (only checks `'ownerId' in doc`), so Change Log entries can get the wrong `entityType`.
- `apps/itun/src/stores/entityBackend.ts:110` — `op.gameId as Id<'games'>` is a lie at runtime; **confirmed** Convex rejects a v13-migrated local workspace id (`Validator error: Expected one of id, null, got "starter-set-workspace"`). The error is swallowed at line 118, so every edit to an entity that was in a non-Default workspace silently never mirrors. New entities get `workspaceId` but no `gameId`, so they mirror to the shelf instead.
- `apps/itun/src/stores/entityStore.ts:396` — `transfer()` (cargo, scrap hand-offs, Dashboard transactions) calls neither `requireWritableBackend()` nor the mirror, so a Disconnected user still writes locally (the exact fork ADR-030 forbids) and Connected transfers never sync.
- `apps/itun/convex/entities.ts:242` — `claimLocal` inserts without `appId`, so the first edit after a claim creates a duplicate server row instead of updating (`parsed.data.id` is right there).
- `apps/itun/convex/entities.ts:303` — `appId` is looked up globally with no owner scoping, but the Starter Set ships deterministic ids (`starter-pilot-bonesaw`); the second user to edit a seeded starter pilot collides with the first user's row and their mirror fails permanently and silently.
- `apps/itun/src/stores/entityBackend.ts:131` — nothing catches `WritesBlockedOffline` and no component reads `canWrite`; offline writes reject as unhandled promise rejections with no user feedback.
- `apps/itun/src/stores/entityStore.ts:510` — `delete()` mirrors the server delete *before* the local delete, inverting `create`/`update` ordering; a failed IDB transaction leaves the server row gone and the local row present.
- `apps/itun/convex/entities.ts:284` — `byAppId` uses `.unique()`, which throws on duplicate appIds; combined with the swallowed error that kills an entity's sync permanently rather than degrading.
- `apps/itun/convex/account.ts:186` — `deleteAccount` deletes entities in surviving Games but only prunes `softLinks` when the whole Game is deleted, leaving dangling links for remaining members.
- `apps/itun/src/components/account/AccountScreen.tsx:30` — `URL.revokeObjectURL` fires synchronously after `a.click()`; the data export can produce no file in Firefox/Safari.
- `apps/itun/src/components/account/AccountScreen.tsx:69` — Save with an untouched field writes the Discord name into the `displayName` override, contradicting the screen's own "clearing falls back to Discord" copy.

Also worth noting (not filed as defects): `claimLocal`, `listForGame`, `listShelf`, `patchCrawler`, `entities.create/update/remove` and all of `ownership.ts` currently have no client callers — the server layer is ahead of the UI, which is why several of the above are latent rather than live today.

PR #640, opened 2026-07-28
feat(reference): give effects a target — Bio-Wings grants the MECH the Fly Trait

alxjrvs:
**Effect targeting** — the last item of the rules-parity plan (#598), sixth in the recommended order.

```
Rules parity: 30 claim(s) — 17 encoded, 12 exempt, 1 unresolved (1 known, burning down)
```

## The blocker was never "a place to declare effects"

C3 was scoped as *"effects declarable directly on a system/ability"*. Declaring them was never the problem: **`ChoiceEffectSchema` had no `target`**, and neither remaining record's effect lands on the record that declares it.

Bio-Wings says *"Your Mech gains the Fly Trait"* — a trait belonging to the **mech**, not to the system that granted it. Declared as a self-effect it would have said the Bio-Wings *system* flies: **wrong, not merely incomplete**. Encoding it that way would have applied a real rule to the wrong entity, which is worse than a visible gap — so it stayed unencoded until the model could say it properly.

## Three pieces, each necessary

| Piece | Why |
|---|---|
| `target: 'hostMech'` | So an effect can land on the mech rather than the item |
| `appliedEffects` | Record-level effects, for grants that aren't a choice |
| `mechTraits()` | The derivation that aggregates them — **nothing collected a mech's traits before**, the other half of why this couldn't be encoded |

Named **`appliedEffects`**, not `effects`: that key is already taken on meta entities with a different `{ label, value }` shape (consumed by `getEffects`), and one field name meaning two things is how a schema rots.

Derived at read time, never stored — a trait disappears the moment the granting item is removed, with no bookkeeping. Duplicates keep the highest magnitude, matching `resolveChoiceView`'s Explosive 1 → 2 upgrade rule, so the two resolvers can't disagree about stacking.

## What remains, precisely scoped

**High Gain Antenna** raises the Range band of *other* installed Modules and Pilot Abilities, filtered by trait. That needs a **`bumpRange`** op (`setRange` sets an absolute value; this *increments* by one) and **cross-item targeting with a predicate** — a loadout-level resolver, where every other effect resolves per entity. Recorded in `KNOWN_UNRESOLVED` with that reasoning so the next person doesn't rediscover it.

## Verification

`bun run check:all` green: schemas, lint, format, typecheck, **4,604 tests**, validate (incl. the parity gate), knip, audit, tokens, styling.

Six new tests, including the one carrying the intent: a **self**-targeted effect must not land on the mech — a weapon whose own choice adds Ballistic to itself must never make the mech Ballistic.

Refs #598.

### DIFF
--- a/packages/salvageunion-reference/data/systems.json
+++ b/packages/salvageunion-reference/data/systems.json
@@ -1361,6 +1361,7 @@
     "techLevel": "B",
     "slotsRequired": 6,
     "salvageValue": 25,
+    "appliedEffects": [{"op": "addTrait", "value": "Fly", "target": "hostMech"}],
     "page": 74,
     "actions": ["Bio-Wings"]
   },
--- a/packages/salvageunion-reference/etc/salvageunion-reference.api.d.ts
+++ b/packages/salvageunion-reference/etc/salvageunion-reference.api.d.ts
@@ -1874,6 +1874,8 @@ export type { DamageKind, MechDamageInput, MechDamageEffect, PilotDamageInput, P
 export { PILOT_BASE_HP, PILOT_BASE_AP, PILOT_BASE_INVENTORY_SLOTS, injuryMaxHpPenalty, pilotMaxHP, pilotMaxAP, isPilotDead, clampPilotCurrentStats, installedStatBonus, mechMaxSP, mechMaxEP, mechMaxHeat, mechMaxCargo, clampMechCurrentStats, unifiedMechConditions, crawlerMaxSP, crawlerMaxSPParts, mechMaxSPParts, mechMaxEPParts, mechMaxHeatParts, mechMaxCargoParts, pilotMaxHPParts, pilotMaxAPParts, clampCrawlerCurrentStats, } from './derivedStats.js';
 export type { ChassisStats, CrawlerMaxSPParts, StatBreakdown } from './derivedStats.js';
 export { statesMechanicalChange } from './rulesBearing.js';
+export { mechTraits } from './mechTraits.js';
+export type { MechTrait } from './mechTraits.js';
 export type { RulesClaim } from './rulesBearing.js';
 export { abilityContributions, resolveAmount, sumContributions, } from './contributions.js';
 export type { ContributionStat, ContributionTarget, ContributionAmount, DeclaredContribution, ResolvedContribution, } from './contributions.js';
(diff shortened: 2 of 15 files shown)


alxjrvs (commented):
Reviewed the `parity-targeting` branch diff (1 commit, `8b256eac`, reference package only: `EffectTargetSchema` + `appliedEffects` on `SystemModuleSchema`, new `mechTraits` derivation, Bio-Wings data, parity-validator update, regenerated JSON schemas + api report).

Baseline health: `bun --filter salvageunion-reference test` (975 pass), `bun run typecheck` (all workspaces), `bun run knip`, `validate:all` and `validate:parity` all pass, and `bun run build:package` produces no schema/api drift. The bugs below are latent — none of them break the current suite.

Findings:

- `packages/salvageunion-reference/tools/validateParityLogic.ts:179` — `hasEffectData` counts *any* non-empty `appliedEffects` as "encoded", but the only consumer (`mechTraits`) handles just `addTrait`/`removeTrait` with `target: 'hostMech'`; `resolveChoiceView` never reads `appliedEffects` at all. An author can retire a parity finding with `setRange`/`addDamage`/self-targeted data that nothing applies — the exact inert-content failure the audit exists to catch.
- `packages/salvageunion-reference/lib/schemas/objects.ts:568` — `target` was added to `ChoiceEffectSchema`, which also types `choices[].choiceOptions[].effects`, but `resolveChoiceView` switches on `op` only and ignores `target`; a `hostMech`-targeted choice effect is applied to the declaring item ("the Bio-Wings system flies"), and `mechTraits` never sees it.
- `packages/salvageunion-reference/lib/rules/mechTraits.ts:68` — doc claims parity with `resolveChoiceView`, but the Map is keyed case-sensitively and keeps the max amount, while `resolveChoiceView` matches case-insensitively and last-amount-wins. Burn 2 then Burn 1 disagree; `'Fly'` + `'fly'` yields two entries; `removeTrait 'fly'` won't strip `'Fly'`; a non-numeric amount makes both sides NaN so first-grant-wins.
- `packages/salvageunion-reference/lib/rules/mechTraits.test.ts:42` — the "does NOT collect self-targeted effects" test uses Cargo Pod, which has no `appliedEffects` and no `choices`; deleting the `target !== 'hostMech'` guard leaves the test green.
- `packages/salvageunion-reference/tools/validateParityLogic.ts:123` — the `KNOWN_UNRESOLVED` JSDoc still says "Both need the SAME missing capability" and documents Bio-Wings as blocked, contradicting the one-entry list directly beneath it.
- `packages/salvageunion-reference/lib/schemas/objects.ts:545` — `hostMech` is a second target vocabulary alongside `ContributionTargetSchema`'s `self | pilot | pilotedMech | crawler` (the one ADR-029 §1 specifies); a pilot ability granting the piloted mech a trait is inexpressible, and future `target` consumers must handle two disjoint enums.

PR #641, opened 2026-07-28
feat(itun): crew visibility and owner chips (Phase 2)

alxjrvs:
Ninth in the stack. **Phase 2** of [ADR-030](https://github.com/SalvageUnion-io/SU-SRD/blob/main/docs/adrs/ADR-030-accounts-games-server-of-record.md).

Every member sees every crewmate's vitals live and can drill into a crewmate's sheet read-only (D12). The Mediator's prepared opposition is the one thing that stays hidden — it is simply not queried here.

## Why vitals are a separate, narrow query

The crew strip re-renders on **every point of damage anyone takes**. Serving it from `entities.listForGame` would push every crewmate's loadouts, abilities and cargo down the wire on each change, for a row that shows four numbers.

It also makes the privacy boundary legible: this query returns what a crewmate is *entitled* to see at a glance, and nothing more.

## `readEntity` has no write counterpart, anywhere

Membership is the whole check — which is what "lean over and look at their sheet" means at a physical table. A Mediator who wants to *change* what they're reading proposes instead (D7).

**A shelved entity is unreachable through it entirely** — not guarded, absent. A shelf belongs to one person and to no crew, so there is no membership that could grant access. Returning null is safer than a check somebody might later loosen.

## Bodies are opaque, so the projection doesn't trust them

A missing numeric field reads as `null`, not zero. **Zero would render as "dead" on a vitals strip**, which is worse than blank.

## Owner chips are data, not a component

`component-lib` is shared with `apps/srd`, which has no accounts and never will, so teaching the card about users would push an ITUN-only concept into a package whose contract is to stay backend-agnostic. The app passes a `badge` control instead — which renders at every card extent, so a dense listing row and an expanded card show the same chip.

**The load-bearing case is unclaimed vs. owned-but-unresolved.** Both have no name to show, and collapsing them would tell a Mediator a pilot is free to hand out when it already belongs to somebody — a data-integrity error dressed as a label.

## Tests

18 new: 11 on chip resolution (including that the label is *never* empty) and 7 on the crew queries, weighted toward what must stay out — the non-member and the shelved entity.

## Verification

typecheck 4/4 · knip clean · `format:check` clean · design-token gate clean · full suite green · lint at baseline. Re-verified after rebasing across 12 newly-landed commits on `main`.

### DIFF
--- a/apps/itun/CHANGELOG.md
+++ b/apps/itun/CHANGELOG.md
@@ -4,82 +4,71 @@ Maintained by release-please (see ADR-024).
 
 ## [0.7.0](https://github.com/SalvageUnion-io/SU-SRD/compare/itun-v0.6.0...itun-v0.7.0) (2026-07-28)
 
-
 ### Features
 
-* **component-lib:** make the roll-table title its own table picker ([#592](https://github.com/SalvageUnion-io/SU-SRD/issues/592)) ([18f315a](https://github.com/SalvageUnion-io/SU-SRD/commit/18f315ac0501f00f52db63181dbf383d66909245))
-* **itun:** render partners in place, delete the partner sheet ([#590](https://github.com/SalvageUnion-io/SU-SRD/issues/590)) ([33ddccd](https://github.com/SalvageUnion-io/SU-SRD/commit/33ddccd79c6b9db417dff0384f27188d3fe0b67f))
-* **itun:** render the SRD home page in the Dashboard SRD Explorer ([#593](https://github.com/SalvageUnion-io/SU-SRD/issues/593)) ([05906b6](https://github.com/SalvageUnion-io/SU-SRD/commit/05906b629fe97708f5c9faf8e638e29033e363e9))
+- **component-lib:** make the roll-table title its own table picker ([#592](https://github.com/SalvageUnion-io/SU-SRD/issues/592)) ([18f315a](https://github.com/SalvageUnion-io/SU-SRD/commit/18f315ac0501f00f52db63181dbf383d66909245))
+- **itun:** render partners in place, delete the partner sheet ([#590](https://github.com/SalvageUnion-io/SU-SRD/issues/590)) ([33ddccd](https://github.com/SalvageUnion-io/SU-SRD/commit/33ddccd79c6b9db417dff0384f27188d3fe0b67f))
+- **itun:** render the SRD home page in the Dashboard SRD Explorer ([#593](https://github.com/SalvageUnion-io/SU-SRD/issues/593)) ([05906b6](https://github.com/SalvageUnion-io/SU-SRD/commit/05906b629fe97708f5c9faf8e638e29033e363e9))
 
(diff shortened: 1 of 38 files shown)


alxjrvs (commented):
Reviewed the 12-commit ITUN accounts/Games branch against `origin/main` (75 files: new Convex backend, connection-mode layer, account/games UI, v13 migration, Eldridge Coast removal). 13 findings.

**Findings**

- `apps/itun/convex/crew.ts:96` — `readEntity` does `ctx.db.get(entityId as never)` on an unvalidated string, so any authenticated Game member can read `encounterNpcs` (the Mediator's hidden prep) and other users' unshared `mechPatterns` by passing their `_id`.
- `apps/itun/convex/crew.ts:66` — `vitals` reads `currentHp`/`currentAp`/`currentSp`; the schemas store `currentHP`/`currentAP`/`currentSP`, so all three are permanently `null`. `crew.test.ts:42` seeds the wrong casing, so it passes tautologically.
- `apps/itun/src/stores/entityStore.ts:264` — the mirror sends the *local* `gameId` (a workspace UUID) into `upsertByAppId`'s `v.id('games')` validator; Convex rejects it and the error is swallowed into a `console.warn`, so no non-shelf entity is ever mirrored.
- `apps/itun/convex/entities.ts:263` — `claimLocal` inserts rows without `appId`, so the first mirrored edit after a claim creates a duplicate row instead of updating.
- `apps/itun/convex/templates.ts:94` — template soft links carry `starter-*` local ids while the seeded pilots/mechs get fresh `_id`s and no `appId`: every relationship in a template-seeded Game is dangling.
- `apps/itun/src/stores/entityStore.ts:262` — `withActiveWorkspace` sets `workspaceId` but never `gameId`, so newly created entities always mirror as shelf entities.
- `apps/itun/convex/entities.ts:176` — `update`/`remove` never cross-check the `table` arg against the id's table; a mech body can be written into a `pilots` row, and `remove` can delete a `mechPatterns` row.
- `apps/itun/src/stores/entityStore.ts:510` — the remote delete is issued before the local IDB delete, so a failed local transaction leaves the entity locally present but gone server-side.
- `apps/itun/src/stores/entityBackend.ts:129` — `WritesBlockedOffline` is thrown but caught nowhere; a false `navigator.onLine === false` turns every edit (including purely local shelf/Starter-Set entities) into an unhandled rejection.
- `apps/itun/src/stores/entityBackend.ts:36` — `authState` defaults to signed-out until the provider effect flushes; writes in that window silently skip the mirror.
- `apps/itun/src/components/account/AccountScreen.tsx:30` — `revokeObjectURL` runs synchronously after `a.click()` on a detached anchor; the data export can produce an empty/failed download.
- `apps/itun/convex/games.ts:150` — `destroy` (and `deleteAccount`'s last-member path) leaves `changeLog` and `mechPatterns` rows pointing at a deleted game.
- `apps/itun/src/components/games/GamesScreen.tsx:209` — the template button hardcodes `'starter-set'` instead of `t.id`.

PR #642, opened 2026-07-28
feat(itun): Mediator server layer — NPC tray and presence (Phase 3)

alxjrvs:
Tenth in the stack. **Phase 3 server layer** of [ADR-030](https://github.com/SalvageUnion-io/SU-SRD/blob/main/docs/adrs/ADR-030-accounts-games-server-of-record.md).

## The NPC tray is the one genuinely secret thing

These gate on `requireMediator`, not `requireMember` — the only place in the app that does. Everything else is readable by every member, which is what makes a crew a crew. **A player who can read prepared opposition can read the encounter before it happens**: the one leak that changes how the game is *played*, rather than merely who can edit what.

A test covers the case easiest to get wrong: **an Organizer who does not mediate cannot read the tray either.** The Organizer flag is administrative and confers no content authority, and prepared opposition is very much content.

## Presence is a heartbeat, not a stream

It upserts one row per `(game, member)` so a long session cannot grow it without bound, and **`present` is computed at read time from `lastSeen` rather than stored.** A stored boolean needs something to clear it — a timer, a disconnect hook — and every such mechanism can fail in a way that leaves somebody permanently "present" long after they closed the tab.

Presence itself is not privileged: any member may heartbeat and read it. A Mediator needs to see players arrive exactly as much as players need to see each other.

`amMediator` returns `false` rather than throwing for a non-member — "can I mediate this" is a reasonable question for anyone to ask, and a surface gating itself on the answer shouldn't have to catch.

## Scope, stated plainly

**This is the server layer only.** The Mediator *screen* is net-new product design — crew roster, live vitals, NPC tray, absorbing `/encounter` — and the plan specifies its behaviour but not its form.

This repo treats UI as a design problem: the `design-token` gate has already caught me twice, `ttrpg-ux-designer` exists for exactly this, and the locked v52 Dashboard went through ~60 iterations. Inventing that surface unreviewed would be presenting guesswork as specification. **The data it needs is here and tested; the screen wants a design pass.**

## Tests

10 new, weighted toward what must stay hidden: player-cannot-read, player-cannot-write, non-mediating-Organizer-cannot-read, plus presence upsert-not-append and the computed-not-stored window.

## Verification

### DIFF
--- a/apps/itun/CHANGELOG.md
+++ b/apps/itun/CHANGELOG.md
@@ -4,82 +4,71 @@ Maintained by release-please (see ADR-024).
 
 ## [0.7.0](https://github.com/SalvageUnion-io/SU-SRD/compare/itun-v0.6.0...itun-v0.7.0) (2026-07-28)
 
-
 ### Features
 
-* **component-lib:** make the roll-table title its own table picker ([#592](https://github.com/SalvageUnion-io/SU-SRD/issues/592)) ([18f315a](https://github.com/SalvageUnion-io/SU-SRD/commit/18f315ac0501f00f52db63181dbf383d66909245))
-* **itun:** render partners in place, delete the partner sheet ([#590](https://github.com/SalvageUnion-io/SU-SRD/issues/590)) ([33ddccd](https://github.com/SalvageUnion-io/SU-SRD/commit/33ddccd79c6b9db417dff0384f27188d3fe0b67f))
-* **itun:** render the SRD home page in the Dashboard SRD Explorer ([#593](https://github.com/SalvageUnion-io/SU-SRD/issues/593)) ([05906b6](https://github.com/SalvageUnion-io/SU-SRD/commit/05906b629fe97708f5c9faf8e638e29033e363e9))
+- **component-lib:** make the roll-table title its own table picker ([#592](https://github.com/SalvageUnion-io/SU-SRD/issues/592)) ([18f315a](https://github.com/SalvageUnion-io/SU-SRD/commit/18f315ac0501f00f52db63181dbf383d66909245))
+- **itun:** render partners in place, delete the partner sheet ([#590](https://github.com/SalvageUnion-io/SU-SRD/issues/590)) ([33ddccd](https://github.com/SalvageUnion-io/SU-SRD/commit/33ddccd79c6b9db417dff0384f27188d3fe0b67f))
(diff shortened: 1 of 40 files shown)


alxjrvs (commented):
Reviewed PR #642 (`itun-accounts-phase3` → `main`). Scope: `origin/main...HEAD` — 77 files, ~6.3k insertions across the new Convex backend (`apps/itun/convex/`), the connection-mode layer, the write mirror, migration 13, and the account/games UI. (`git diff @{upstream}...HEAD` was empty — the branch is already pushed — and `main...HEAD` was misleading because local `main` is stale.)

**Findings (9):**

- `apps/itun/netlify.toml:111` — Production CSP `connect-src` has no Convex origin, so once `VITE_CONVEX_URL` is set the client's `wss://*.convex.cloud` socket is blocked and the app is permanently, silently Solo in production.
- `apps/itun/src/stores/entityStore.ts:264` — `mirrorEntityWrite` passes `record.gameId` (a *local* workspace UUID after migration 13) as a Convex `Id<'games'>`; argument validation rejects it and `mirrorWrite` swallows the error, so nothing in a non-default workspace ever mirrors. `gameIdOf()` exists for this and isn't used.
- `apps/itun/convex/entities.ts:264` — `claimLocal` inserts without `appId`, so the first mirrored edit after a claim takes `upsertByAppId`'s create branch and duplicates every claimed entity.
- `apps/itun/convex/entities.ts:284` — `byAppId` is a global, owner-unscoped lookup, but Starter Set ids are deterministic constants (`starter-pilot-bonesaw`); the second user to edit theirs gets a permanent silent `NotAuthorized` on every mirror.
- `apps/itun/convex/crew.ts:96` — `readEntity` ignores `args.table` and casts an arbitrary id, so any member can read an `encounterNpcs` body — the one thing the module says must stay Mediator-only.
- `apps/itun/convex/entities.ts:176` — `update`/`remove` never verify the loaded row's table matches `args.table`; a Mech body can be written into a pilots row you own, bricking it against `PilotSchema`.
- `apps/itun/src/stores/entityStore.ts:510` — the remote delete fires *before* the local delete, inverting the create/update ordering; a failed local delete leaves the entity locally present but server-side destroyed.
- `apps/itun/convex/games.ts:124` — `destroy` leaves `changeLog` and `mechPatterns` rows pointing at the deleted game.
- `apps/itun/convex/mediator.ts:141` — `amMediator` throws for signed-out callers despite documenting that it returns `false`.

Positives worth noting: the permission module is genuinely centralized and well tested (328-line authz suite), the migration is fixture-tested through the real opener, the Solo-build branch is structurally guarded by tests in both new screens, and the connection-mode table is exhaustively covered.

PR #643, opened 2026-07-28
feat(itun): proposals and alerts — the Mediator never writes your sheet (Phase 4)

alxjrvs:
Eleventh in the stack. **Phase 4** of [ADR-030](https://github.com/SalvageUnion-io/SU-SRD/blob/main/docs/adrs/ADR-030-accounts-games-server-of-record.md) (D7, D25).

The mechanism that lets a Mediator reach a player's sheet **without ever writing it**: they push a proposal, the player applies or declines.

## Alerts and cross-player writes are one system

Both are rows in the Change Log — a proposal is an entry in `proposed` state carrying `before`/`after` for one field; an alert is the same row with no entity target.

That's the payoff for ADR-022 having built the log append-only, ordered and replay-shaped *before any of this existed*. There was no second bus to design.

## Three rules the tests pin

Each is a way the promise could **quietly erode** rather than break loudly:

**Nothing is force-applied.** There is no mutation here that writes an entity on the player's behalf; `apply` refuses if the caller isn't the owner. A test has the Mediator try to apply their own proposal and expects rejection — that path appearing is exactly how propose-and-confirm collapses back into the direct write ADR-030 rejects.

**Nothing expires.** A test backdates a proposal a month and asserts it's still pending. A timer would silently drop damage a player was meant to take — the bookkeeping error this feature exists to prevent.

**Newer supersedes older, per field.** A second proposal against the same `(entityId, field)` retires the first, so a player never faces two contradictory pending values. Supersession is per *field*, not per entity — damage and heat are separate facts, and a test covers that a different field is left alone.

## Smaller decisions

- **`pending` is scoped to the asker** — a player isn't shown proposals against a crewmate's entity, which would leak an edit in flight.
- **An unclaimed entity can't be proposed to** — nobody would see it, so it'd sit pending forever.
- **Declining records rather than deletes.** The log is append-only; a declined proposal is history, not a gap.

## Tests

12 new, weighted toward erosion rather than the happy path.

## Verification

typecheck 4/4 · knip clean · `format:check` clean · design-token gate clean · **4034 tests pass** · lint at baseline.

### DIFF
--- a/apps/itun/CHANGELOG.md
+++ b/apps/itun/CHANGELOG.md
@@ -4,82 +4,71 @@ Maintained by release-please (see ADR-024).
 
 ## [0.7.0](https://github.com/SalvageUnion-io/SU-SRD/compare/itun-v0.6.0...itun-v0.7.0) (2026-07-28)
 
-
 ### Features
 
-* **component-lib:** make the roll-table title its own table picker ([#592](https://github.com/SalvageUnion-io/SU-SRD/issues/592)) ([18f315a](https://github.com/SalvageUnion-io/SU-SRD/commit/18f315ac0501f00f52db63181dbf383d66909245))
-* **itun:** render partners in place, delete the partner sheet ([#590](https://github.com/SalvageUnion-io/SU-SRD/issues/590)) ([33ddccd](https://github.com/SalvageUnion-io/SU-SRD/commit/33ddccd79c6b9db417dff0384f27188d3fe0b67f))
-* **itun:** render the SRD home page in the Dashboard SRD Explorer ([#593](https://github.com/SalvageUnion-io/SU-SRD/issues/593)) ([05906b6](https://github.com/SalvageUnion-io/SU-SRD/commit/05906b629fe97708f5c9faf8e638e29033e363e9))
+- **component-lib:** make the roll-table title its own table picker ([#592](https://github.com/SalvageUnion-io/SU-SRD/issues/592)) ([18f315a](https://github.com/SalvageUnion-io/SU-SRD/commit/18f315ac0501f00f52db63181dbf383d66909245))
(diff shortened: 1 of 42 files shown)


alxjrvs (commented):
Reviewed the Phase 4 diff (`origin/itun-accounts-phase3...HEAD`): `apps/itun/convex/proposals.ts` (new, 227 lines), its tests, the harness/api registration, and a one-word schema comment fix. Tests pass (12/12). Six findings:

- `apps/itun/convex/proposals.ts:160` — `apply()` persists the entity body with **no Zod parse**, unlike every other write in `entities.ts`. Since `MechSchema`/`PilotSchema` are `.strict()`, one typo'd field name (`currentSp` vs the real `currentSP`) permanently corrupts the row, and every later `upsertByAppId` fails in `parseBody` — swallowed by `mirrorWrite` as a `console.warn`, so the player's sheet silently stops syncing forever.
- `apps/itun/convex/proposals.ts:161` — `apply` writes only the Convex row, but the client is local-first and never reads entities from the server (`api.entities.listForGame` has zero callers; `entityStore` reads IndexedDB only). The applied change never reaches the sheet, and the player's next edit mirrors the stale local body back up, reverting it.
- `apps/itun/convex/proposals.ts:71` — `requireProposalTarget` blind-casts any `entityId` to a pilot/mech id with none of `ownership.ts`'s `'ownerId' in doc` guard; `ownerId === null` is false for `undefined`, so a `crawlers` id passes the unclaimed check and creates a proposal nobody can ever answer or expire. `mechPatterns` is worse — `apply` succeeds against it.
- `apps/itun/convex/proposals.ts:216` — `alerts` `.collect()`s the entire game change log to return 20 rows; `propose` (line 77) `.collect()`s an entity's entire history to check one field. changeLog is documented as an append-only spine, so both eventually cross Convex's 16,384-doc / 8 MiB transaction cap and start throwing.
- `apps/itun/convex/proposals.ts:84` — `entityType` is stored verbatim from the client, never checked against the table `entityId` resolves to, and is handed back by `pending`; a mismatch routes drill-in/rendering to the wrong table.
- `apps/itun/convex/proposals.ts:162` — `apply` leaves `actorId` as the proposing Mediator, so the log records the Mediator as actor of an applied write to a player's sheet — indistinguishable from the direct write the feature exists to disprove.

PR #644, opened 2026-07-28
feat(itun): crew-wide Downtime (Phase 5)

alxjrvs:
Twelfth in the stack. **Phase 5** of [ADR-030](https://github.com/SalvageUnion-io/SU-SRD/blob/main/docs/adrs/ADR-030-accounts-games-server-of-record.md).

Downtime is the one procedure the whole crew performs **in step**, so the phase is Game state advanced by the Mediator rather than something each player tracks alone. Before this, six players ran six private Downtimes and reconciled verbally.

## Upkeep is spent once, not six times

`upkeepSpent` is a flag on the **Game's** downtime, not on each member. The crawler is communal, so its upkeep is one cost the crew pays together — six members each spending it is the exact double-charging that made per-player Downtime unworkable.

The flag resets when a **new** Downtime begins and explicitly **not** when a step advances, which would charge the crew again mid-procedure. Both halves are tested.

`spendUpkeep` returns `false` rather than throwing when it was already paid. Two players pressing the button at the same moment is an ordinary race at a table, not a fault to show anybody; the caller uses the result to decide whether to also deduct the scrap.

## Completion is per step, not cumulative

`completedBy` clears on every advance. Keeping it would leave a player who finished step 1 **looking finished for the whole Downtime** — the opposite of what the Mediator needs to know.

A player marks only *themselves* done. A Mediator cannot tick somebody off, which would be the same overreach as writing their sheet.

## Small shape decision

`state` reports not-running as a *state* rather than returning null, so every caller doesn't invent the same default. That's also why `stepIndex` is nullable on the row instead of the row being absent.

## Tests

13 new, weighted toward the two properties above plus the Mediator-only phase controls.

## Verification

### DIFF
--- a/apps/itun/CHANGELOG.md
+++ b/apps/itun/CHANGELOG.md
@@ -4,82 +4,71 @@ Maintained by release-please (see ADR-024).
 
 ## [0.7.0](https://github.com/SalvageUnion-io/SU-SRD/compare/itun-v0.6.0...itun-v0.7.0) (2026-07-28)
 
-
 ### Features
 
-* **component-lib:** make the roll-table title its own table picker ([#592](https://github.com/SalvageUnion-io/SU-SRD/issues/592)) ([18f315a](https://github.com/SalvageUnion-io/SU-SRD/commit/18f315ac0501f00f52db63181dbf383d66909245))
-* **itun:** render partners in place, delete the partner sheet ([#590](https://github.com/SalvageUnion-io/SU-SRD/issues/590)) ([33ddccd](https://github.com/SalvageUnion-io/SU-SRD/commit/33ddccd79c6b9db417dff0384f27188d3fe0b67f))
-* **itun:** render the SRD home page in the Dashboard SRD Explorer ([#593](https://github.com/SalvageUnion-io/SU-SRD/issues/593)) ([05906b6](https://github.com/SalvageUnion-io/SU-SRD/commit/05906b629fe97708f5c9faf8e638e29033e363e9))
+- **component-lib:** make the roll-table title its own table picker ([#592](https://github.com/SalvageUnion-io/SU-SRD/issues/592)) ([18f315a](https://github.com/SalvageUnion-io/SU-SRD/commit/18f315ac0501f00f52db63181dbf383d66909245))
+- **itun:** render partners in place, delete the partner sheet ([#590](https://github.com/SalvageUnion-io/SU-SRD/issues/590)) ([33ddccd](https://github.com/SalvageUnion-io/SU-SRD/commit/33ddccd79c6b9db417dff0384f27188d3fe0b67f))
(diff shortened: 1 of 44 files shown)


alxjrvs (commented):
Reviewed PR #644 (`itun-accounts-phase5`, 16 commits, 81 files, ~7.2k insertions vs `origin/main`) — the ITUN accounts/Games/Convex server-of-record work. I read every hunk and empirically confirmed five of the findings by running throwaway `convex-test` cases against the real functions (scratch files removed afterwards). The existing suite passes (1448 tests, 0 fail).

Findings, most severe first:

- `apps/itun/convex/crew.ts:96` — **CONFIRMED**: `readEntity` never uses its `table` arg and gates only on `gameId`, so any player can read the Mediator-only `encounterNpcs` (returned `{"body":{"hp":99,"name":"Secret Boss"}}` in a test) — the one thing the module says must stay hidden.
- `apps/itun/convex/proposals.ts:160` — **CONFIRMED**: `apply()` patches the entity body with no Zod parse, violating the invariant every other mutation honours. A proposal of `currentHp: 'not a number'` persisted verbatim; the client can no longer parse that pilot.
- `apps/itun/convex/entities.ts:176` — **CONFIRMED**: `update`/`remove` trust `args.table` without checking the id's actual table; `update({table:'pilots', entityId:<mech id>})` overwrote a mech row with a pilot body.
- `apps/itun/convex/entities.ts:263` — **CONFIRMED**: `claimLocal` inserts without `appId`, so the first mirrored edit after claiming inserts a duplicate (2 rows observed).
- `apps/itun/src/stores/entityBackend.ts:110` — **CONFIRMED**: the mirror forwards a local workspace UUID as `v.id('games')`; Convex rejects it and the error is swallowed, so entities in any non-Default workspace never reach the server of record.
- `apps/itun/src/stores/entityStore.ts:396` — `transfer()` got neither `requireWritableBackend()` nor the mirror, so disconnected users still write locally and connected transfers never sync.
- `apps/itun/src/lib/connection/ConnectionProvider.tsx:77` — a signed-in user who loads offline resolves to `solo`, not `disconnected`: no banner, writes go to IndexedDB and fork against the server.
- `apps/itun/src/components/account/AccountScreen.tsx:30` — re-implements `downloadJson` and drops its Safari fix (synchronous `revokeObjectURL`, anchor never attached).
- `apps/itun/convex/proposals.ts:40` — a proposal against a crawler id is accepted and becomes permanently unanswerable (`ownerId` is `undefined`, not `null`).
- `apps/itun/convex/account.ts:174` — `presence` rows outlive `leaveGame` and non-last-member account deletion, leaving ghost "Crewmate" entries.

Also noted but not reported as defects: `games.destroy`/`deleteAccount` leave `downtime`, `changeLog` and `mechPatterns` rows orphaned (storage leak only, nothing exposed), and the Phase 2–5 server layers (`crew`, `mediator`, `proposals`, `downtime`, `ownership`, `ownerChip`, `container`) have no UI consumers yet — consistent with the phased plan, but it means `knip` is only green because the tests count as consumers.

PR #645, opened 2026-07-28
feat(itun): Discord bot as a Game participant (Phase 6)

alxjrvs:
Thirteenth and final in the stack. **Phase 6** of [ADR-030](https://github.com/SalvageUnion-io/SU-SRD/blob/main/docs/adrs/ADR-030-accounts-games-server-of-record.md).

Closes the intent of the old #165 — "connect the bot to live campaign state" — **without the mechanism that issue assumed.** It called for a Supabase service-role key and RLS policies: a bot that can act as anybody.

## The bot acts as a participant, not an admin

A **channel binding** says "rolls in this channel belong to this Game". The actor is resolved from the Discord user id against `users.discordId`, so the bot can only ever act as somebody who has **linked their own account and joined the Game**. A stranger in the channel is not a member and gets nothing.

This is why Discord was chosen as the sole identity provider back in D3: **the id the bot already holds *is* the account key.** No separate credential, no impersonation surface, no privileged token to leak.

## Every failure is silent

No binding, no linked account, not a member — all return `false`. Distinguishing them in a public channel would announce **who has an account and who is in which Game**.

## Two constraints worth naming

- **One Game per channel.** A channel meaning two Games makes every roll ambiguous, and the bot has no way to ask which was meant. Re-binding the same pair is a no-op rather than an error.
- **Two accounts cannot claim one Discord id**, or the bot would resolve rolls to whichever it found first.

## Rolls are Game history, not a separate store

They land as Change Log entries, so a roll made at the table and a roll made in the channel are the same kind of fact in the same history — for the same reason alerts needed no separate bus.

## Also in this PR

`docs/architecture/accounts-and-games.md` marks Phases 0–6 landed, and says plainly that the **Phase 3–5 surfaces are wired to this data but want a design pass** before being called finished.

## Tests

11 new, weighted toward proving the bot *cannot* act as anybody.

## Verification

### DIFF
--- a/apps/itun/CHANGELOG.md
+++ b/apps/itun/CHANGELOG.md
@@ -4,82 +4,71 @@ Maintained by release-please (see ADR-024).
 
 ## [0.7.0](https://github.com/SalvageUnion-io/SU-SRD/compare/itun-v0.6.0...itun-v0.7.0) (2026-07-28)
 
-
 ### Features
 
-* **component-lib:** make the roll-table title its own table picker ([#592](https://github.com/SalvageUnion-io/SU-SRD/issues/592)) ([18f315a](https://github.com/SalvageUnion-io/SU-SRD/commit/18f315ac0501f00f52db63181dbf383d66909245))
-* **itun:** render partners in place, delete the partner sheet ([#590](https://github.com/SalvageUnion-io/SU-SRD/issues/590)) ([33ddccd](https://github.com/SalvageUnion-io/SU-SRD/commit/33ddccd79c6b9db417dff0384f27188d3fe0b67f))
-* **itun:** render the SRD home page in the Dashboard SRD Explorer ([#593](https://github.com/SalvageUnion-io/SU-SRD/issues/593)) ([05906b6](https://github.com/SalvageUnion-io/SU-SRD/commit/05906b629fe97708f5c9faf8e638e29033e363e9))
+- **component-lib:** make the roll-table title its own table picker ([#592](https://github.com/SalvageUnion-io/SU-SRD/issues/592)) ([18f315a](https://github.com/SalvageUnion-io/SU-SRD/commit/18f315ac0501f00f52db63181dbf383d66909245))
+- **itun:** render partners in place, delete the partner sheet ([#590](https://github.com/SalvageUnion-io/SU-SRD/issues/590)) ([33ddccd](https://github.com/SalvageUnion-io/SU-SRD/commit/33ddccd79c6b9db417dff0384f27188d3fe0b67f))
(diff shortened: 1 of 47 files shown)


alxjrvs (commented):
Reviewed the Phase 6 commit (`8e7ab9bd`): new `convex/bot.ts`, the `channelBindings` table, harness/api registration, and doc status updates. Working tree was clean, so `git diff HEAD~1` was the whole scope. I verified the two most serious findings empirically against the repo's own convex-test harness (temp probe test, since removed — tree is clean).

Findings (6):

- `apps/itun/convex/bot.ts:132` — **`recordRoll` is a public mutation with zero authentication.** Confirmed by running it with no identity: it returns `true` and writes a `changeLog` row attributed to the victim. Any internet caller with the deployment URL (shipped in the client bundle) plus a public Discord channel id and user snowflake can forge roll history for any member of any bound Game, with an unvalidated `v.any()` payload. The module's "no impersonation surface, no privileged token to leak" claim holds only because the endpoint is unauthenticated. Every test calls it while signed in, which hides this.
- `apps/itun/convex/bot.ts:185` — **`linkDiscordId` trusts a self-asserted Discord id.** Nothing else in the repo writes `users.discordId` (the OAuth flow stores the provider id in `authAccounts`), so any signed-in user can claim a stranger's snowflake: their rolls get attributed to the impostor, and the real owner is permanently locked out by the "already linked to another user" throw, with no unlink path.
- `apps/itun/convex/bot.ts:56` — **Bindings orphaned by game deletion permanently kill a channel.** Confirmed: `games.destroy` (and `deleteAccount`'s last-member-out path) don't clean `channelBindings`. Afterwards rebinding fails with "already bound to another game" and `unbindChannel` fails with "Not a member of this game", since `requireOrganizer` looks up a membership in a deleted game. The unused `by_game` index on the new table suggests the cleanup was intended and dropped.
- `apps/itun/convex/bot.ts:71` — **`gameForChannel` only checks `requireUser`, not membership**, so any signed-in account can resolve an arbitrary channel id to a private Game's id and name — the exact leak `resolveActor`'s doc comment says the design avoids.
- `apps/itun/convex/_generated/api.d.ts:34` — commit adds the missing `downtime` entry but not `bot`. Harmless today only because `convex/` is excluded from `apps/itun/tsconfig.json`; the first `src/` component calling `api.bot.linkDiscordId` fails typecheck.
- `apps/itun/convex/bot.ts:174` — `rolls` `.collect()`s the entire game Change Log to return 20 rows; blows Convex's per-query read limit on a long campaign and then fails permanently. Same shape as the pre-existing `proposals.alerts`, so pattern-consistent rather than new.

PR #646, opened 2026-07-28
feat(itun): build the Phase 3–5 surfaces

alxjrvs:
Fourteenth in the stack. The screens the previous PRs deliberately left unbuilt: the **Mediator surface**, the player's **proposal inbox**, and the **crew-wide Downtime panel**.

## Mediator screen (`/mediator/$gameId`)

One screen, because the Mediator's problem is holding four things at once — the crew's numbers, the opposition, what they want to change, and where the table is in a Downtime. Splitting those across routes recreates the tab-flipping it exists to replace.

**Deliberately not the player Dashboard.** That's a locked 1280×800 canvas built around one pilot+mech+crawler (ADR-020). An N-player table view doesn't fit it, and stretching it would reopen a layout that took ~60 iterations to settle. A plain scrolling surface is the right shape for a list that grows with the crew.

The propose form only offers **claimed** entities — an unclaimed pre-gen has nobody to answer, so listing it would build a dead end into the UI.

## Proposal inbox

Shows **before *and* after**. The Mediator captured `before` from their own view at propose time and it's deliberately not re-read on apply — if it disagrees with what the player is looking at, that mismatch is information the table needs, not something the UI should reconcile away.

**Decline sits beside Apply as a peer**, not behind an X. Declining is a recorded answer; hiding it would imply accepting is the only real option.

## Downtime panel

The point is the **"who is still working"** list. Before this, six players ran six private Downtimes and the Mediator's only way to know whether the table could move on was to ask.

The list emptying on every advance is **correct, not a bug** — a name persisting would mean "finished at some point", the opposite of what's needed.

## Crew vitals

A null vital renders as `—`, **never `0`**. Bodies are opaque server-side, so a missing field means "unknown", and zero would read as *dead* on a vitals strip. Unclaimed renders as a state, not a blank — a pre-gen waiting to be handed out is available, not broken.

## Solo builds

Every surface sits behind the `isConvexConfigured` branch, so a Solo build renders an honest explanation rather than an inert console. Tested, because this screen has more Convex hooks than any other in the app and is the likeliest to leak one past the branch.

## Two gate catches worth recording

- The **design-token checker scans file text**, not just class attributes — naming the arbitrary-tracking pattern literally inside a doc comment tripped it.
- **knip** caught an unused chrome constant I'd defined but never consumed.

## Design status

Functional, using the app's existing chrome vocabulary: hard ink borders, stamped condensed caps, the shipped tracking ladder. **The visual arrangement has not had a design pass** and is the part most likely to change. The behaviour and the data are settled; the composition is a first draft.

## Verification

typecheck 4/4 · knip clean · `format:check` clean · design-token gate clean · full suite green · lint at baseline. Re-verified after rebasing across 16 newly-landed commits on `main`.

### DIFF
--- a/apps/itun/CHANGELOG.md
+++ b/apps/itun/CHANGELOG.md
@@ -4,82 +4,71 @@ Maintained by release-please (see ADR-024).
 
 ## [0.7.0](https://github.com/SalvageUnion-io/SU-SRD/compare/itun-v0.6.0...itun-v0.7.0) (2026-07-28)
 
-
 ### Features
 
-* **component-lib:** make the roll-table title its own table picker ([#592](https://github.com/SalvageUnion-io/SU-SRD/issues/592)) ([18f315a](https://github.com/SalvageUnion-io/SU-SRD/commit/18f315ac0501f00f52db63181dbf383d66909245))
-* **itun:** render partners in place, delete the partner sheet ([#590](https://github.com/SalvageUnion-io/SU-SRD/issues/590)) ([33ddccd](https://github.com/SalvageUnion-io/SU-SRD/commit/33ddccd79c6b9db417dff0384f27188d3fe0b67f))
-* **itun:** render the SRD home page in the Dashboard SRD Explorer ([#593](https://github.com/SalvageUnion-io/SU-SRD/issues/593)) ([05906b6](https://github.com/SalvageUnion-io/SU-SRD/commit/05906b629fe97708f5c9faf8e638e29033e363e9))
+- **component-lib:** make the roll-table title its own table picker ([#592](https://github.com/SalvageUnion-io/SU-SRD/issues/592)) ([18f315a](https://github.com/SalvageUnion-io/SU-SRD/commit/18f315ac0501f00f52db63181dbf383d66909245))
+- **itun:** render partners in place, delete the partner sheet ([#590](https://github.com/SalvageUnion-io/SU-SRD/issues/590)) ([33ddccd](https://github.com/SalvageUnion-io/SU-SRD/commit/33ddccd79c6b9db417dff0384f27188d3fe0b67f))
+- **itun:** render the SRD home page in the Dashboard SRD Explorer ([#593](https://github.com/SalvageUnion-io/SU-SRD/issues/593)) ([05906b6](https://github.com/SalvageUnion-io/SU-SRD/commit/05906b629fe97708f5c9faf8e638e29033e363e9))
 
 ## [0.6.0](https://github.com/SalvageUnion-io/SU-SRD/compare/itun-v0.5.0...itun-v0.6.0) (2026-07-28)
 
-
 ### Features
 
-* **itun:** show the Actions deck as one catalog-tile grid, mech + pilot ([#591](https://github.com/SalvageUnion-io/SU-SRD/issues/591)) ([f8979b0](https://github.com/SalvageUnion-io/SU-SRD/commit/f8979b04b8d804c4c464bd596c0b31074d8f4769))
-
+- **itun:** show the Actions deck as one catalog-tile grid, mech + pilot ([#591](https://github.com/SalvageUnion-io/SU-SRD/issues/591)) ([f8979b0](https://github.com/SalvageUnion-io/SU-SRD/commit/f8979b04b8d804c4c464bd596c0b31074d8f4769))
 
 ### Bug Fixes
 
-* **itun:** start Heat at zero, never at capacity ([#588](https://github.com/SalvageUnion-io/SU-SRD/issues/588)) ([67cebf2](https://github.com/SalvageUnion-io/SU-SRD/commit/67cebf2901aff7bfda198e55f49c10a3f7442f3f))
+- **itun:** start Heat at zero, never at capacity ([#588](https://github.com/SalvageUnion-io/SU-SRD/issues/588)) ([67cebf2](https://github.com/SalvageUnion-io/SU-SRD/commit/67cebf2901aff7bfda198e55f49c10a3f7442f3f))
 
 ## [0.5.0](https://github.com/SalvageUnion-io/SU-SRD/compare/itun-v0.4.1...itun-v0.5.0) (2026-07-25)
 
-
 ### Features
 
-* **itun:** live-sheet redesign — card/slab containers, folding, and a crew-shaped crawler ([#561](https://github.com/SalvageUnion-io/SU-SRD/issues/561)) ([b180c24](https://github.com/SalvageUnion-io/SU-SRD/commit/b180c244ab7b8d8df39512ed3805b80661efc22a))
-* Partner Sheets — statted drones/companions owned by a pilot or a mech ([#578](https://github.com/SalvageUnion-io/SU-SRD/issues/578)) ([7ee4d1b](https://github.com/SalvageUnion-io/SU-SRD/commit/7ee4d1b45e11d44f90da0c90dad82e08a1edca30))
+- **itun:** live-sheet redesign — card/slab containers, folding, and a crew-shaped crawler ([#561](https://github.com/SalvageUnion-io/SU-SRD/issues/561)) ([b180c24](https://github.com/SalvageUnion-io/SU-SRD/commit/b180c244ab7b8d8df39512ed3805b80661efc22a))
+- Partner Sheets — statted drones/companions owned by a pilot or a mech ([#578](https://github.com/SalvageUnion-io/SU-SRD/issues/578)) ([7ee4d1b](https://github.com/SalvageUnion-io/SU-SRD/commit/7ee4d1b45e11d44f90da0c90dad82e08a1edca30))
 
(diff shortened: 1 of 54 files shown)


alxjrvs (commented):
Reviewed `origin/main...HEAD` (17 commits, 90 files — the ITUN accounts/Games/Convex work). Nine of the thirteen findings were reproduced directly against the `convex-test` harness (scratch test written, run, and deleted; tree is clean).

Findings, most severe first:

- `apps/itun/convex/bot.ts:132` — `recordRoll` has no authentication and takes `discordId` as an argument; an anonymous caller forges Change Log rolls attributed to any linked user (verified: returns `true` with no identity).
- `apps/itun/convex/crew.ts:96` — `readEntity` ignores its `table` arg and `db.get`s any id; a plain player read the Mediator-only `encounterNpcs` body (`{name:'SECRET BOSS'}`) after being denied `mediator.npcs`.
- `apps/itun/convex/proposals.ts:160` — `apply` patches the entity body with no Zod parse; a proposal of `currentHp: 'not-a-number'` persists verbatim and poisons the row for every later write.
- `apps/itun/convex/entities.ts:176` — `update`/`remove` trust the client's `table` rather than the row's; a valid pilot body was written into a mech row without error.
- `apps/itun/convex/games.ts:141` — `destroy` doesn't clean `channelBindings` (nor `downtime`/`changeLog`/`mechPatterns`); the orphan binding makes the Discord channel permanently unbindable and un-unbindable.
- `apps/itun/convex/entities.ts:263` — `claimLocal` inserts without `appId`, so the first mirrored edit after sign-in creates a duplicate row (verified: 2 pilots).
- `apps/itun/src/stores/entityBackend.ts:110` — the mirror casts a local workspace UUID to `Id<'games'>`; Convex rejects it and the catch swallows it, so migrated workspace entities silently never sync.
- `apps/itun/src/stores/entityStore.ts:396` — `transfer()` skips both `requireWritableBackend()` and the mirror: writes land offline (read-only rule broken) and never reach the server.
- `apps/itun/convex/bot.ts:185` — `linkDiscordId` accepts any snowflake unverified; an attacker claims another player's Discord identity and locks them out.
- `apps/itun/convex/entities.ts:291` — `byAppId` calls `.unique()` on a non-unique column; two accounts holding the same exported entity id break each other's mirror permanently.
- `apps/itun/convex/account.ts:174` — `deleteAccount` leaves `presence` rows in surviving Games, producing a permanent ghost crewmate.
- `apps/itun/src/stores/entityStore.ts:510` — `delete()` mirrors the server delete before the local delete succeeds.
- `apps/itun/src/components/games/MediatorScreen.tsx:172` — propose always sends `before: null` (defeating the inbox's stated before/after comparison) and `Number(value)` → `NaN` on non-numeric input.

alxjrvs (commented):
Reviewed `origin/main...HEAD` (17 commits, 90 files, +8.3k) — the ITUN accounts/Games/Convex work. I read every Convex function and client surface, then **empirically confirmed 11 of 12 findings** by writing throwaway `convex-test` probes against the real harness (probe files removed; tree is clean, and the existing suite passes 1462/0).

Findings — `file:line — summary`:

1. `apps/itun/convex/bot.ts:132` — `recordRoll` is a public mutation with **no auth check at all**; confirmed an anonymous caller forges a Change Log roll attributed to another user in a Game they aren't in.
2. `apps/itun/convex/proposals.ts:149` — `apply` patches the entity body with **no Zod parse**; confirmed it writes `currentHp: "not-a-number"` into a `.strict()` pilot, after which every later server write of that pilot throws forever.
3. `apps/itun/src/components/games/MediatorScreen.tsx:37` — proposable fields are misspelled (`currentHp/Ap/Sp` vs schema `currentHP/AP/SP`) and not filtered by target type; a proposal never changes the real stat and corrupts the body.
4. `apps/itun/convex/crew.ts:90` — `readEntity` accepts `table` then ignores it; confirmed a non-Mediator member reads the hidden NPC tray (`{"name":"SECRET BOSS"}`) straight through the `requireMediator` gate.
5. `apps/itun/convex/bot.ts:185` — `linkDiscordId` lets any signed-in user claim any Discord snowflake unverified: permanent impersonation, and the real owner can never link.
6. `apps/itun/convex/crew.ts:66` — crew vitals read the same misspelled keys, so HP/AP/SP always render `—` for the whole crew.
7. `apps/itun/convex/entities.ts:263` — `claimLocal` (and `templates.createGame`) insert without `appId`; confirmed the next mirrored edit creates a **duplicate** row.
8. `apps/itun/convex/entities.ts:284` — `by_app_id` is unscoped, and Starter Set ids are static; confirmed the second user to sign in can never mirror their starter roster (silently).
9. `apps/itun/src/stores/entityStore.ts:259` — the mirror passes a local workspace UUID into a `v.id('games')` arg; confirmed validator rejection, swallowed into a console.warn. Also, new entities never set `gameId`, so everything mirrors to the shelf.
10. `apps/itun/convex/games.ts:124` — `destroy` leaves `channelBindings` (plus `downtime`/`changeLog`/`mechPatterns`) orphaned; confirmed the Discord channel then can be neither re-bound nor unbound by anyone.
11. `apps/itun/convex/entities.ts:176` — `update`/`remove` never check the doc is in `args.table`; a `mechPatterns` id deletes through the pilots API.
12. `apps/itun/src/components/account/AccountScreen.tsx:23` — re-implements `lib/export/downloadJson.ts` but revokes the blob URL synchronously (the existing helper defers 100ms explicitly "so Safari has time to start the download").

Themes worth the author's attention: the module headers make strong safety claims (`every mutation MUST Zod-parse`, `the bot has no impersonation surface`, `the NPC tray is the one thing that must stay hidden`) that three separate functions violate — and the tests only ever exercise the happy path with an identity attached, so none of it was caught. The `appId`/`gameId` mirror contract is the other cluster: it works in tests because they use fresh random ids and `null` game ids, which is exactly the case that never occurs in the migrated app.

alxjrvs (commented):
Reviewed `origin/main...itun-accounts-surfaces` (90 files, ~8.3k insertions — the whole Convex accounts/Games backend plus its ITUN surfaces). Working tree is clean, so the range diff was the full scope. `lint`, `format:check`, `knip`, `typecheck` and all 1,462 tests pass on the branch — several findings below are things those gates structurally cannot see.

**Findings (14)**

- `apps/itun/convex/bot.ts:139` — `recordRoll` never calls `requireUser`; any anonymous caller can forge a Change Log roll attributed to any linked user in a bound channel.
- `apps/itun/convex/bot.ts:185` — `linkDiscordId` accepts an unverified snowflake, so a user can claim someone else's Discord identity and permanently block the real owner from linking.
- `apps/itun/convex/crew.ts:96` — `readEntity` ignores its `table` argument entirely; a plain member holding an `encounterNpcs` id reads the Mediator-only NPC tray (and unshared `mechPatterns`).
- `apps/itun/convex/proposals.ts:160` — `apply` patches the entity body with no Zod parse, violating the invariant `entities.ts` states; a `NaN` or off-schema field corrupts a `.strict()` sheet.
- `apps/itun/src/lib/connection/ConnectionProvider.tsx:77` — `useConvexAuth().isLoading` is dropped, so signed-in users resolve to `solo` during auth load: the "You are playing solo" UI flashes and writes in that window skip the mirror.
- `apps/itun/src/stores/entityStore.ts:264` — after migration 13, `gameId` holds a *local workspace UUID*; it is sent to Convex as `v.id('games')`, so every mirror write for a non-Default-workspace entity fails validation into a swallowed `console.warn`.
- `apps/itun/src/stores/entityStore.ts:361` — `requireWritableBackend()` throws from every entity write when Disconnected, but no call site catches it and `canWrite` is never consumed; edits vanish as unhandled rejections, including on shelf-only entities.
- `apps/itun/tsconfig.json:19` — `convex/**/*` is not in `include`, so the entire new backend is excluded from `bun run typecheck`/CI; adding it surfaces 17 real errors today.
- `apps/itun/convex/_generated/api.d.ts:13` — stale codegen: the `bot` module is absent, so `api.bot.*` is untyped (only works via `anyApi` at runtime).
- `apps/itun/convex/account.ts:189` — `deleteAccount` and `games.destroy` orphan `changeLog` (carrying the deleted user's `actorId` and sheet values), `presence`, `downtime`, `channelBindings` and game-scoped `mechPatterns`.
- `apps/itun/src/components/account/AccountScreen.tsx:30` — re-implements `downloadJson` and revokes the blob URL synchronously; the data export silently fails in Safari. The existing helper already solves this.
- `apps/itun/convex/entities.ts:263` — `claimLocal` never sets `appId`, so the first post-claim mirror inserts a duplicate server row instead of updating the claimed one.
- `apps/itun/src/components/games/MediatorScreen.tsx:171` — `ProposeForm` always sends `before: null` (inbox always renders `— → N`) and coerces a free-text field with `Number()`.
- `apps/itun/src/stores/entityStore.ts:510` — `delete` mirrors the remote delete *before* the local transaction, inverting the ordering used by `create`/`update`; a local failure leaves the server row gone with nothing to repair it.

PR #647, opened 2026-07-28
feat(itun): accounts, Games, and a live Mediator surface (ADR-030)

alxjrvs:
The whole accounts feature plus the coverage work, consolidated into one PR.
This branch contains every commit from the former #631–#646 stack (each of those
branches was a patch-identical ancestor) and from #649. **All 14 of those PRs are
closed as superseded by this one.**

Implements [ADR-030](docs/adrs/ADR-030-accounts-games-server-of-record.md),
which supersedes ADR-001, amends ADR-022 and extends ADR-021 with an ownership
axis.

## The feature

- **Three storage modes.** Solo (not signed in — IndexedDB is the source of
  truth, nothing gated, and this stays the default forever), Connected, and
  Disconnected (signed in but offline = **read-only**, not a write queue).
  Resolved in one place, `src/lib/connection/`.
- **Two containers, not one.** Workspaces split into a shared **Game** and a
  personal **Shelf**, encoded as a nullable `gameId` — `null` means deliberately
  shelved, `undefined` means not yet decided.
- **Roles.** Base role Player | Mediator, plus an orthogonal Organizer flag that
  is administrative only.
- **The Mediator never writes your sheet.** Changes are *proposed*: a Change Log
  row the player applies or declines, showing before and after so a mismatch is
  visible rather than papered over. Proposals never expire and are never
  force-applied.
- **Discord sign-in**, account management, export and deletion — with the shared
  crawler and the table's game surviving an account deletion.
- **Discord bot as a Game participant** (Phase 6).
- **Your existing local data comes with you**: a v13 migration gives every
  pre-accounts row a container, v1 export bundles still import, and a one-time
  prompt copies a device's roster into an account — offered, not automatic,
  because uploading somebody's whole roster on sign-in is a decision made with
  their data.

## Going live

Two fixes here are what make the deployed app work at all:

- `connect-src` now allows Convex (`https://*.convex.cloud`,
  `wss://*.convex.cloud`, `https://*.convex.site`). `VITE_CONVEX_URL` is
  *already* set on the production site, so without this the app would have
  shipped Connected, failed every request at the CSP layer, and reported
  nothing — sign-in silently never completing. `wss:` does not inherit from
  `https:`, and the hosts are wildcarded because the deployment differs by
  context.
- The Netlify build runs `convex deploy` before the site build when
  `CONVEX_DEPLOY_KEY` is present, so backend and bundle deploy together.

Without that key the build is byte-for-byte the old one, so previews keep
building Solo. **Setting `CONVEX_DEPLOY_KEY` on the Netlify site is the single
remaining switch that turns production accounts on** — it cannot live in a PR.

## Coverage (was #649)

The ratchet was **measuring the wrong files**. It resolved lcov `SF:` paths
against the lcov's directory, but Bun writes them relative to the workspace
root. `apps/srd` and `apps/itun` reach siblings as `../../packages/…` and landed
outside either way; `packages/component-lib` is one level shallower, so its
`../salvageunion-reference/…` resolved to something that looked local and it was
charged for 4,600 lines of another package. The gate passed, printed a number,
and the number was of something else. `parseLcov` moves to `tools/lib/` so the
rule is testable directly — three of the seven new cases fail against the old
resolution.

| Workspace | Before | Now |
| --- | --- | --- |
| packages/salvageunion-reference | 92.13% | **96.08%** ✅ |
| apps/srd | 95.57% | 95.57% ✅ |
| apps/discord-bot | 92.40% | **95.47%** ✅ |
| packages/component-lib | 76.18% (mis-measured) | **88.57%** |
| apps/itun | 89.20% | **90.09%** |

New tests cover the connected surfaces (GamesScreen, MediatorScreen,
AccountScreen, ClaimLocalData, CrewVitals, DowntimePanel, ProposalInbox) — the
branch people actually use, previously untested because the test build has no
`VITE_CONVEX_URL` — plus `observability.ts` (the bot's only channel for learning
production broke), `helpers.ts`, and `InlineEditField`.

Three real defects fell out of writing them, all fixed here: the claim prompt
hid its own confirmation, `mock.module` is process-global in Bun and was
poisoning 219 tests, and Testing Library's auto-cleanup proved order-dependent.
The mocking lesson is written into `.claude/rules/testing-patterns.md`.

`component-lib` and `itun` remain short of 95% — 658 and ~920 lines, concentrated
in `RollTable` (859 lines, nondeterministic rolls), `ReferenceEntityCard` and the
itun sheet surfaces. That is its own piece of work.

### DIFF
--- a/.claude/rules/testing-patterns.md
+++ b/.claude/rules/testing-patterns.md
@@ -42,6 +42,46 @@ Testing patterns using Bun's test runner and Testing Library for component tests
 - Use `mock()` from `bun:test` for module/function mocking
 - Mock external dependencies appropriately
 
+### `mock.module` is process-global — restore what you replace
+
+`mock.module` rewrites the entry in the module registry for the whole test
+process, **not** for the file that called it. Every test file that runs after
+yours gets your replacement. This is not theoretical: mocking the entity store
+in one component test broke 219 tests across the suite, and the failures
+surfaced in files that had never heard of the mocked module.
+
+Capture the real exports first, and put them back in `afterAll`:
+
+```typescript
+// The spread is load-bearing. A module namespace is a LIVE view, and mocking
+// rewrites it in place — hold the namespace itself and by `afterAll` it
+// already reads as the mock, so you restore the mock over the mock.
+const realStore = { ...(await import('../../stores/entityStore')) }
+
+mock.module('../../stores/entityStore', () => ({ useEntityStore: fake }))
+
+afterAll(() => {
+  mock.module('../../stores/entityStore', () => realStore)
+})
+```
+
+Two corollaries:
+
+- **Partial mocks break importers you did not think about.** Replacing
+  `convex/react` with only the hooks under test made `@convex-dev/auth/react`
+  fail at import on a missing `ConvexProviderWithAuth`. Mock every export the
+  transitive importers reach, not just the ones your component calls.
+- **Beware module-scope environment reads.** `config.ts` reads `process.env`
+  once at import, so setting an env var to drive a test hands that value to
+  every later file too. Mock the config module instead of setting the variable.
+
+### Call `cleanup()` explicitly in component tests
+
+Testing Library's auto-cleanup has proven order-dependent here — a file passed
+alone and failed in the full suite. When several tests in a file render the
+same accessible name, one leaked render turns every later query into "found
+multiple elements". Add `afterEach(cleanup)` rather than depending on it.
+
 ## Examples
 
 **Package test:**
--- a/apps/discord-bot/src/__tests__/observability.test.ts
+++ b/apps/discord-bot/src/__tests__/observability.test.ts
@@ -1,0 +1,151 @@
+import { afterAll, beforeAll, describe, expect, mock, test } from 'bun:test'
+
+/**
+ * observability — the bot's only channel for finding out that production
+ * broke. It is entirely env-gated, and the gate is the interesting part: with
+ * no DSN every function must be a silent no-op (local dev and CI run that way),
+ * and with one it must actually report. A regression in either direction is
+ * invisible until the day you need it — a bot that has stopped reporting looks
+ * exactly like a bot with no errors.
+ *
+ * `config` is mocked rather than driven through `process.env`, deliberately.
+ * `config.ts` reads the environment once at module scope, so setting
+ * `SENTRY_DSN` here would hand a live DSN to every test file that runs after
+ * this one — `mock.module` is process-global in Bun, not file-scoped, and the
+ * env is worse still. Both the mocks below are captured and restored.
+ */
+
+const sentryCalls: Array<{ fn: string; args: unknown[] }> = []
+let flushShouldThrow = false
+let dsn: string | undefined
+
+// `config.ts` throws at module scope on a missing token, and it is evaluated by
+// the capture below. Same `??=` as the sibling tests, and deliberately no
+// SENTRY_DSN: the DSN is supplied through the mock, never the environment.
+process.env.DISCORD_TOKEN ??= 'test-token'
+process.env.DISCORD_CLIENT_ID ??= 'test-client-id'
+
+const realSentry = { ...(await import('@sentry/node')) }
+const realConfig = { ...(await import('../config.js')) }
+
+mock.module('@sentry/node', () => ({
+  init: (...args: unknown[]) => sentryCalls.push({ fn: 'init', args }),
+  flush: async (...args: unknown[]) => {
+    sentryCalls.push({ fn: 'flush', args })
+    if (flushShouldThrow) throw new Error('transport down')
+    return true
+  },
+  captureException: (...args: unknown[]) => sentryCalls.push({ fn: 'captureException', args }),
+  captureMessage: (...args: unknown[]) => sentryCalls.push({ fn: 'captureMessage', args }),
+}))
+
+mock.module('../config.js', () => ({
+  config: {
+    get sentryDsn() {
+      return dsn
+    },
+    nodeEnv: undefined,
+    releaseSha: 'abc123',
+  },
+}))
+
+let obs: typeof import('../observability.js')
+
+beforeAll(async () => {
+  obs = await import('../observability.js')
+})
+
+afterAll(() => {
+  mock.module('@sentry/node', () => realSentry)
+  mock.module('../config.js', () => realConfig)
+})
+
+function calls(fn: string): Array<{ fn: string; args: unknown[] }> {
+  return sentryCalls.filter((c) => c.fn === fn)
+}
+
+/**
+ * These run in order against one module instance, because `enabled` is
+ * module-level state and the transition from off to on is exactly what is
+ * being tested.
+ */
+describe('with no DSN configured', () => {
+  test('init reports nothing and does not enable Sentry', () => {
+    dsn = undefined
+    obs.initObservability()
+    expect(calls('init')).toHaveLength(0)
+  })
+
+  test('captureException is a no-op rather than an unconfigured send', () => {
+    obs.captureException(new Error('boom'))
+    expect(calls('captureException')).toHaveLength(0)
+  })
+
+  test('captureMessage is a no-op too', () => {
+    obs.captureMessage('ready')
+    expect(calls('captureMessage')).toHaveLength(0)
+  })
+
+  test('flush resolves immediately instead of waiting on a transport', async () => {
+    // A 2s flush timeout on every exit of a bot that never had a DSN would be
+    // two wasted seconds on every restart.
+    await obs.flushObservability()
+    expect(calls('flush')).toHaveLength(0)
+  })
+})
+
+describe('once a DSN is configured', () => {
+  test('init passes the DSN, environment and release through', () => {
+    dsn = 'https://key@example.ingest.sentry.io/1'
+    obs.initObservability()
+
+    const init = calls('init')[0]
+    expect(init).toBeDefined()
+    expect(init?.args[0]).toMatchObject({
+      dsn: 'https://key@example.ingest.sentry.io/1',
+      // Unset NODE_ENV must not leave the environment blank — an untagged event
+      // is indistinguishable from one raised in dev.
+      environment: 'production',
+      release: 'abc123',
+    })
+  })
+
+  test('a second call is a no-op, so startup can call it freely', () => {
+    obs.initObservability()
+    expect(calls('init')).toHaveLength(1)
+  })
+
+  test('exceptions are reported, with context as extra when given', () => {
+    const error = new Error('boom')
+    obs.captureException(error, { guildId: 'g1' })
+
+    const c = calls('captureException')[0]
+    expect(c?.args[0]).toBe(error)
+    expect(c?.args[1]).toEqual({ extra: { guildId: 'g1' } })
+  })
+
+  test('no context sends undefined rather than an empty extra bag', () => {
+    obs.captureException(new Error('bare'))
+    expect(calls('captureException')[1]?.args[1]).toBeUndefined()
+  })
+
+  test('messages are reported — the worker has no port to health-probe', () => {
+    obs.captureMessage('logged in', { shard: 0 })
+    const c = calls('captureMessage')[0]
+    expect(c?.args[0]).toBe('logged in')
+    expect(c?.args[1]).toEqual({ extra: { shard: 0 } })
+  })
+
+  test('flush is awaited before exit, with the timeout passed through', async () => {
+    // The transport is async; a synchronous exit right after captureException
+    // drops the one event the restart happened for.
+    await obs.flushObservability(500)
+    expect(calls('flush')[0]?.args[0]).toBe(500)
+  })
+
+  test('a failing flush never rejects — the process is exiting anyway', async () => {
+    flushShouldThrow = true
+    expect(await obs.flushObservability(10).then(() => 'resolved')).toBe('resolved')
+    flushShouldThrow = false
+  })
+})
--- a/apps/itun/CHANGELOG.md
+++ b/apps/itun/CHANGELOG.md
@@ -28,82 +28,71 @@ Maintained by release-please (see ADR-024).
 
 ## [0.7.0](https://github.com/SalvageUnion-io/SU-SRD/compare/itun-v0.6.0...itun-v0.7.0) (2026-07-28)
 
-
 ### Features
 
-* **component-lib:** make the roll-table title its own table picker ([#592](https://github.com/SalvageUnion-io/SU-SRD/issues/592)) ([18f315a](https://github.com/SalvageUnion-io/SU-SRD/commit/18f315ac0501f00f52db63181dbf383d66909245))
-* **itun:** render partners in place, delete the partner sheet ([#590](https://github.com/SalvageUnion-io/SU-SRD/issues/590)) ([33ddccd](https://github.com/SalvageUnion-io/SU-SRD/commit/33ddccd79c6b9db417dff0384f27188d3fe0b67f))
-* **itun:** render the SRD home page in the Dashboard SRD Explorer ([#593](https://github.com/SalvageUnion-io/SU-SRD/issues/593)) ([05906b6](https://github.com/SalvageUnion-io/SU-SRD/commit/05906b629fe97708f5c9faf8e638e29033e363e9))
+- **component-lib:** make the roll-table title its own table picker ([#592](https://github.com/SalvageUnion-io/SU-SRD/issues/592)) ([18f315a](https://github.com/SalvageUnion-io/SU-SRD/commit/18f315ac0501f00f52db63181dbf383d66909245))
+- **itun:** render partners in place, delete the partner sheet ([#590](https://github.com/SalvageUnion-io/SU-SRD/issues/590)) ([33ddccd](https://github.com/SalvageUnion-io/SU-SRD/commit/33ddccd79c6b9db417dff0384f27188d3fe0b67f))
+- **itun:** render the SRD home page in the Dashboard SRD Explorer ([#593](https://github.com/SalvageUnion-io/SU-SRD/issues/593)) ([05906b6](https://github.com/SalvageUnion-io/SU-SRD/commit/05906b629fe97708f5c9faf8e638e29033e363e9))
 
 ## [0.6.0](https://github.com/SalvageUnion-io/SU-SRD/compare/itun-v0.5.0...itun-v0.6.0) (2026-07-28)
 
-
 ### Features
 
-* **itun:** show the Actions deck as one catalog-tile grid, mech + pilot ([#591](https://github.com/SalvageUnion-io/SU-SRD/issues/591)) ([f8979b0](https://github.com/SalvageUnion-io/SU-SRD/commit/f8979b04b8d804c4c464bd596c0b31074d8f4769))
-
+- **itun:** show the Actions deck as one catalog-tile grid, mech + pilot ([#591](https://github.com/SalvageUnion-io/SU-SRD/issues/591)) ([f8979b0](https://github.com/SalvageUnion-io/SU-SRD/commit/f8979b04b8d804c4c464bd596c0b31074d8f4769))
 
 ### Bug Fixes
 
-* **itun:** start Heat at zero, never at capacity ([#588](https://github.com/SalvageUnion-io/SU-SRD/issues/588)) ([67cebf2](https://github.com/SalvageUnion-io/SU-SRD/commit/67cebf2901aff7bfda198e55f49c10a3f7442f3f))
+- **itun:** start Heat at zero, never at capacity ([#588](https://github.com/SalvageUnion-io/SU-SRD/issues/588)) ([67cebf2](https://github.com/SalvageUnion-io/SU-SRD/commit/67cebf2901aff7bfda198e55f49c10a3f7442f3f))
 
 ## [0.5.0](https://github.com/SalvageUnion-io/SU-SRD/compare/itun-v0.4.1...itun-v0.5.0) (2026-07-25)
 
-
 ### Features
 
-* **itun:** live-sheet redesign — card/slab containers, folding, and a crew-shaped crawler ([#561](https://github.com/SalvageUnion-io/SU-SRD/issues/561)) ([b180c24](https://github.com/SalvageUnion-io/SU-SRD/commit/b180c244ab7b8d8df39512ed3805b80661efc22a))
-* Partner Sheets — statted drones/companions owned by a pilot or a mech ([#578](https://github.com/SalvageUnion-io/SU-SRD/issues/578)) ([7ee4d1b](https://github.com/SalvageUnion-io/SU-SRD/commit/7ee4d1b45e11d44f90da0c90dad82e08a1edca30))
+- **itun:** live-sheet redesign — card/slab containers, folding, and a crew-shaped crawler ([#561](https://github.com/SalvageUnion-io/SU-SRD/issues/561)) ([b180c24](https://github.com/SalvageUnion-io/SU-SRD/commit/b180c244ab7b8d8df39512ed3805b80661efc22a))
+- Partner Sheets — statted drones/companions owned by a pilot or a mech ([#578](https://github.com/SalvageUnion-io/SU-SRD/issues/578)) ([7ee4d1b](https://github.com/SalvageUnion-io/SU-SRD/commit/7ee4d1b45e11d44f90da0c90dad82e08a1edca30))
 
 ## [0.4.1](https://github.com/SalvageUnion-io/SU-SRD/compare/itun-v0.4.0...itun-v0.4.1) (2026-07-24)
 
-
 ### Bug Fixes
 
-* **reference:** resolve data-inspection findings — dead flags, enum, null-marker, redundant fields ([#555](https://github.com/SalvageUnion-io/SU-SRD/issues/555)) ([2658095](https://github.com/SalvageUnion-io/SU-SRD/commit/2658095793f18341538cd977d328aca98a7861cb))
+- **reference:** resolve data-inspection findings — dead flags, enum, null-marker, redundant fields ([#555](https://github.com/SalvageUnion-io/SU-SRD/issues/555)) ([2658095](https://github.com/SalvageUnion-io/SU-SRD/commit/2658095793f18341538cd977d328aca98a7861cb))
 
 ## [0.4.0](https://github.com/SalvageUnion-io/SU-SRD/compare/itun-v0.3.1...itun-v0.4.0) (2026-07-24)
 
-
 ### Features
 
-* **dashboard:** lay the Actions deck out as a masonry grid ([#549](https://github.com/SalvageUnion-io/SU-SRD/issues/549)) ([49fc203](https://github.com/SalvageUnion-io/SU-SRD/commit/49fc203e54bf35c0e90423544d4184981bd06a68))
-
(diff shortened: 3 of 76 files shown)


alxjrvs (commented):
Reviewed PR #647 (`itun-accounts-legacy-data` → main, 60 files / ~6.1k lines: the Convex accounts+Games server layer, the Game/Shelf container split, migration 13, and the v1→v2 export-bundle path). 12 findings:

1. **apps/itun/src/stores/entityStore.ts:264** — `mirrorEntityWrite` sends the entity's `gameId` to Convex, but migration 13 sets that to a *local workspace UUID*; `v.id('games')` (or `getMembership`) rejects it and `mirrorWrite` swallows the error, so signed-in users' campaign writes silently never reach the server of record. Newly created entities have `gameId === undefined` and mirror to the shelf instead.
2. **apps/itun/src/lib/export/mergeImport.ts:142** — import remaps `workspaceId` but leaves the newly-stamped `gameId` in `...rest`; since `containerOf` prefers `gameId`, an imported roster lands in a phantom game that exists nowhere on the device (same at :164, :186).
3. **apps/itun/convex/bot.ts:132** — `recordRoll` is a public mutation with no `requireUser`; the client-supplied `discordId` *is* the actor identity, so anyone can forge Change Log rows attributed to any linked member.
4. **apps/itun/src/components/account/AccountScreen.tsx:23** — the local `download()` clicks a detached anchor and revokes the blob URL synchronously; "Download my data" no-ops in Firefox/Safari. The repo's `downloadJson` already handles both.
5. **apps/itun/convex/proposals.ts:160** — `apply` patches the entity body with no Zod parse and an unvalidated `field`; one bad applied proposal makes every later `update`/`upsertByAppId` throw `Invalid pilots payload`, locking the owner out of their own sheet.
6. **apps/itun/convex/entities.ts:382** — `byAppId` uses `.unique()` on a client-supplied, non-unique, un-owner-scoped key; two claims of the same shared bundle permanently break mirroring for both users, silently.
7. **apps/itun/convex/account.ts:168** — game deletion omits `channelBindings`/`downtime` from cleanup; an orphaned binding makes the Discord channel permanently un-bindable and un-unbindable.
8. **apps/itun/convex/crew.ts:96** — `readEntity` never uses its `table` arg; `db.get` on an `encounterNpcs` id passes the guards and returns the Mediator's hidden opposition to any member.
9. **apps/itun/convex/entities.ts:176** — `update`/`remove` don't check that `entityId` belongs to `table`; a pilot body can be validated then written into a mech row.
10. **apps/itun/convex/entities.ts:356** — `claimLocal` inserts `mechPatterns` completely unvalidated (contra the module's "everything is Zod-parsed or skipped"); `softLinks` also get `gameId: null` when no crawler was claimed, making them invisible to `listForGame`.
11. **apps/itun/src/components/games/MediatorScreen.tsx:171** — always sends `before: null` (so the inbox's before→after comparison can never work) and `Number(value)` → NaN on non-numeric input, which Convex can't serialize.
12. **apps/itun/src/stores/entityStore.ts:510** — the remote delete is mirrored *before* the local IDB delete runs, so a failed local delete leaves the server row gone and the local row present.

Worth noting for context: `entities.update/remove/create/claimLocal`, `crew.readEntity` and the whole `bot` module have no client callers yet, so 3, 8, 9 and 10 are exposed API surface rather than paths the app currently exercises. Findings 1 and 2 are the ones that break the PR's stated purpose — carrying legacy IndexedDB data forward.

alxjrvs (commented):
Reviewed PR #647 (`origin/main...HEAD`, 16 commits / ~6,200 lines: the whole ADR-030 accounts + Games stack — Convex server layer, client backend selection, v13 migration, Games/Mediator/Account surfaces).

**Findings (most severe first)**

- `apps/itun/src/lib/export/mergeImport.ts:142` — Import remaps `workspaceId` through the old→new id map but passes the new `gameId` through untouched in `...rest`. Imported entities land in a container id that doesn't exist locally, so they're invisible to any container-filtered surface — the exact divergence `legacyContainers.ts` says it prevents.
- `apps/itun/src/stores/entityStore.ts:264` — `mirrorEntityWrite` forwards the local `gameId` (a `crypto.randomUUID()` workspace id after migration 13) as Convex's `v.id('games')`, which always rejects; the error is swallowed by `mirrorWrite`'s `console.warn`. Every workspace-scoped entity silently never syncs. Newly created entities have no `gameId` at all and mirror to the shelf instead of their game.
- `apps/itun/convex/bot.ts:132` — `recordRoll` is a public mutation with no `requireUser`; caller supplies `channelId` **and** `discordId`. Anyone with the (public) deployment URL can forge Change Log rolls attributed to a real player, and it's an unauthenticated unbounded-insert endpoint.
- `apps/itun/convex/proposals.ts:160` — `apply` patches the entity body without Zod-parsing it, breaking the invariant `entities.ts` declares mandatory. The Mediator UI offers all four fields for both entity types, so proposing `currentSp` on a pilot permanently wedges that pilot's writes against the strict schema.
- `apps/itun/convex/crew.ts:96` — `readEntity` never uses its `table` argument; `ctx.db.get` fetches whatever table the id belongs to. A member holding an `encounterNpcs` id reads the Mediator-only NPC tray through it. No test covers a table mismatch.
- `apps/itun/convex/entities.ts:242` — `claimLocal` has no caller anywhere in `src/`. First sign-in never uploads a Solo user's roster; only entities they subsequently edit reach the server via the upsert path.
- `apps/itun/convex/__tests__/harness.ts:1` — Not a `.test.ts`, so Convex deploys it (it's in `_generated/api.d.ts` as `"__tests__/harness"`), pulling the `convex-test` devDependency into the production bundle.
- `apps/itun/convex/account.ts:168` — `deleteAccount` leaves the user's `presence` rows in surviving Games (permanent ghost "Crewmate" in `mediator.presence`) and never sweeps `downtime` / `channelBindings` / `changeLog` / game-scoped `mechPatterns` when the last member leaves.
- `apps/itun/src/stores/entityStore.ts:510` — Delete is mirrored to the server *before* the local IDB delete is awaited; a local failure leaves the entity visible locally and erased on the server. `create`/`update` already order this correctly.
- `apps/itun/convex/entities.ts:176` — `update`/`remove` never verify the id belongs to `args.table`; a pilot-shaped body can be written into a mech row you own.
- `apps/itun/src/components/games/MediatorScreen.tsx:172` — `after: Number(value)` unguarded; a non-numeric typo proposes and applies `NaN` HP.
- `apps/itun/src/components/account/AccountScreen.tsx:30` — `URL.revokeObjectURL` fires synchronously after `a.click()`; "Download my data" can silently produce no file on Firefox/Safari.

Notes on things I checked and did **not** flag: the v13 migration itself is idempotent and correct, `containerOf`'s null-vs-undefined handling is sound, `permissions.ts` gating is consistent across the new modules, and the export bundle v1→v2 normalization in `parseImportBundle` is right (the bug is downstream, in `mergeImport`).

alxjrvs (commented):
Reviewed the PR-scoped diff (`97534c03..HEAD`, PR #647 on top of the #646 stack): 13 files, the `claimLocal` expansion, the new `ClaimLocalData` prompt, and the v1→v2 export-bundle migration. `bun --filter itun test` passes (1493/0), so none of these are caught by the suite.

Findings, most severe first:

- `apps/itun/src/components/account/ClaimLocalData.tsx:76` — The success/skip summary is unreachable dead UI: `markClaimed()` runs before `setSummary()`, so the line-76 guard returns `null` on the next render. After clicking "Copy to my account" the card just vanishes; the user never sees the copied count nor the "N could not be read" warning that `claimLocal` returns.
- `apps/itun/src/components/account/ClaimLocalData.tsx:74` — `userKey` is keyed on roster counts, not the account. Adding one pilot after a claim changes the key, re-shows the prompt, and a second click re-inserts the entire roster (plus a second "Claimed crawler" game), since `claimLocal` only inserts. Conversely a second account on the same device with matching counts is never offered the claim. Both contradict the file's own doc.
- `apps/itun/src/components/account/ClaimLocalData.tsx:65` — `usePatternStore((s) => s.mechPatterns)` never triggers hydration (every other consumer pairs it with `.getState().list()`), so on a direct load of `/account` patterns are `[]`: the prompt says "0 saved patterns" and none are claimed.
- `apps/itun/src/lib/export/parseImportBundle.ts:100` — `assignContainers` stamps `gameId` from the bundle's *original* `workspaceId`, but `mergeImport` then remaps `workspaceId` to a newly-minted local id and leaves `gameId` alone (it's in `...rest`). `containerOf` prefers `gameId`, so a restored v1 backup lands in a container id that doesn't exist on the device — the exact invariant `legacyContainers.ts` claims to guarantee.
- `apps/itun/convex/entities.ts:348` — Claimed soft links get the crawler's placeholder `gameId` (or `null`) while their endpoints always land on the shelf. `listShelf` returns no soft links at all and `listForGame` returns links with no matching pilots/mechs, so every claimed association is unreadable despite `byKind.softLinks` reporting success.
- `apps/itun/src/components/account/ClaimLocalData.tsx:104` — No `.catch` on the mutation: a rejection is unhandled, `busy` resets, and the user gets no error feedback at all.
- `apps/itun/convex/entities.ts:357` — `mechPatterns` are inserted with no Zod parse, breaking the module's stated "every write Zod-parses before it persists" contract; `MechPatternSchema` already exists. Nothing is ever counted as `skipped` for patterns.
- `apps/itun/convex/entities.ts:349` — Soft-link endpoint `type` is coerced with `String(l.from.type ?? '')` rather than validated, persisting unresolvable `{type: '', id}` refs as "claimed".

One non-finding worth flagging: `exportBundle.ts`'s header comment still says "schemaVersion: 1 is the current and only supported version; parseImportBundle() rejects any other value", and `parseImportBundle.ts:51` still says "schemaVersion is not 1" — both now false. Also, the new `SUPPORTED_VERSIONS`/`CURRENT_VERSION` consts were inserted between `parseImportBundle`'s JSDoc block and the function, so that doc comment now attaches to the constant.

alxjrvs (commented):
Review complete. Two notes on scope first: local `main` was stale (behind by 20 commits), so the real base is **`origin/main` (813e6662)** — I reviewed `git diff origin/main...HEAD`, 68 files / ~7.5k lines: the Convex accounts + Games server layer, the Games/Mediator/Account UI, migration 13, and the export-bundle v2 path. I also created two throwaway probe tests under `apps/itun/convex/__tests__/` to verify findings empirically and deleted them; separately, `coverage-baseline.json`, `tools/coverage-report.ts` and an untracked `tools/lib/` appeared dirty during the session and are **not** mine — worth checking whether another session is writing into this worktree.

Five of the twelve findings were confirmed by running probes against the `convex-test` harness, not by reading alone.

**Findings (most severe first)**

- `apps/itun/convex/crew.ts:96` — `readEntity` declares `table: 'pilots'|'mechs'` and never uses it; `ctx.db.get(entityId as never)` lets any member read any doc in the Game. Probe: a plain player retrieved `{"body":{"hp":99,"name":"SECRET BOSS"}}` from the Mediator-only `encounterNpcs` tray — the one thing `mediator.ts` calls "genuinely secret".
- `apps/itun/convex/bot.ts:139` — `recordRoll` never calls `requireUser`. Probe: with **no identity at all**, an anonymous caller inserted a `changeLog` row with `actorId` set to a real member. Needs only the deployment URL, a bound channel id, and a public Discord snowflake.
- `apps/itun/convex/bot.ts:202` — `linkDiscordId` accepts a self-asserted Discord id (nothing else in the codebase ever writes `users.discordId`; the OAuth provider doesn't). Probe: linking `'999-not-mine'` succeeded. Chained with the above: claim a victim's snowflake, receive their rolls, lock them out permanently.
- `apps/itun/src/stores/entityBackend.ts:110` — the mirror casts the *local* `gameId` (a migration-13 workspace UUID) to `Id<'games'>`. Probe: `Validator error: Expected one of id, null`. Swallowed by the `catch` → every entity in a pre-existing local workspace silently never mirrors, which is precisely the failure the doc comment claims to have fixed.
- `apps/itun/convex/entities.ts:385` — `byAppId` uses `.unique()` on a non-unique index, and the Starter Set ships **fixed** ids identical across every user's IndexedDB. Probe: two rows sharing `starter-pilot-bonesaw` make `removeByAppId` reject permanently.
- `apps/itun/src/components/account/ClaimLocalData.tsx:74` — the "already claimed" marker is keyed on roster *counts*, not the account. Build one more pilot after claiming and the card re-offers, re-uploading the whole roster (duplicate `appId`s → previous finding) plus a second placeholder Game.
- `apps/itun/convex/entities.ts:186` — `update`/`remove` never check the doc belongs to `args.table`. Probe: `remove({table:'pilots', entityId:<mechPatterns id>})` deleted the pattern silently.
- `apps/itun/convex/account.ts:172` — `deleteAccount`'s last-member cleanup omits `channelBindings`; the orphan makes both `unbindChannel` and `bindChannel` throw forever, so that Discord channel is permanently unusable.
- `apps/itun/src/components/account/ClaimLocalData.tsx:118` — `.then().finally()` with no `.catch`: a failed claim is completely silent (contrast `GamesScreen`'s redeem, which does surface the error).
- `apps/itun/src/components/games/MediatorScreen.tsx:171` — always sends `before: null` (so the inbox's documented "before → after" is always `— → x`) and `after: Number(value)` unvalidated, so a non-numeric entry writes `NaN` onto a player's sheet.
- `apps/itun/convex/entities.ts:307` — the placeholder Game is inserted before any crawler is parsed, so an all-unparseable crawler leaves a permanent empty "Claimed crawler" Game.
- `apps/itun/src/stores/entityStore.ts:510` — `delete` mirrors *before* the local write, unlike `create`/`update` which mirror after.

The permissions model in `convex/model/permissions.ts` is sound and consistently applied elsewhere; nearly every issue above is a boundary that was designed correctly and then bypassed by an unvalidated `v.string()` id or a missing `requireUser`.

alxjrvs (commented):
Two review agents reported back after the summary was rendered. Their findings are mostly new — here's the supplement, deduped against the 14 already reported. I verified the highest-impact claims directly against the branch's git objects (the worktree has since been switched to `coverage-attribution`, so the files aren't on disk).

## Convex server layer — new

- **`apps/itun/convex/bot.ts:132` — `recordRoll` has no caller authentication at all.** ✅verified: `resolveActor` never calls `requireUser`/`getAuthUserId`; identity comes entirely from the `channelId` + `discordId` arguments. Anyone who knows a channel snowflake and a member's Discord id can insert forged `changeLog` rows attributed to that member with `source: 'discord-bot'`. Every other function on the branch gates on `requireUser`; this one doesn't, and no bot-side caller exists yet to constrain it.
- **`apps/itun/convex/bot.ts:202` — `linkDiscordId` writes a self-asserted `discordId` onto the caller's user row**, never checked against the Discord OAuth identity (`auth.ts` never populates the field, so this is the only writer). Claim a victim's snowflake and their rolls resolve to your account — and line 198's "already linked to another user" check then locks them out permanently with no self-service remedy.
- **`apps/itun/convex/crew.ts:96` — `readEntity` takes a `table` argument and ignores it.** ✅verified: `ctx.db.get(args.entityId as never)` on an arbitrary string, then only `gameId !== null` + `requireMember`. Any ordinary member can pass an `encounterNpcs` id and read the Mediator's prepared opposition — the thing `mediator.ts`'s own header calls "the one thing that must stay hidden." Also reaches `mechPatterns` with `sharedToGame: false`.
- **`apps/itun/convex/bot.ts:71` — `gameForChannel` gates on `requireUser` only, not membership.** ✅verified. Any signed-in user can enumerate channel snowflakes and learn the id + name of every private campaign bound to them — undoing the leak-avoidance `resolveActor`'s docstring explicitly reasons about, and handing an attacker the `gameId` the finding above needs.
- **`apps/itun/convex/proposals.ts:160` — `apply` splices `after` into the body and patches with no Zod parse**, bypassing the invariant `schema.ts:32` states as mandatory. `field` is an unvalidated `v.string()`, so a proposal can write any key with any value and leave the body unparseable.
- **`apps/itun/convex/proposals.ts:71` — the unclaimed guard is `ownerId === null`, which misses `undefined`.** Propose against a `crawlers` id (no `ownerId` column) and the row is written, then permanently unanswerable: `pending` skips it, `apply`/`decline` both throw `NotAuthorized`.

## Client surfaces — new

- **`apps/itun/src/stores/entityBackend.ts:110` — `mirrorWrite` casts a local workspace UUID to `Id<'games'>`.** ✅verified cast. Migration 13 copies `workspaceId` into `gameId` verbatim, but `upsertByAppId` validates `v.id('games')` — so for any user with a non-default workspace, every mirrored edit fails argument validation and is swallowed by that `console.warn`. Combined with the already-reported swallow, this is the concrete case where the server of record silently never receives a write.
- **`apps/itun/src/components/games/MediatorScreen.tsx:37` — `PROPOSABLE_FIELDS` uses the wrong casing.** ✅verified: `['currentHp', 'currentAp', 'currentSp', 'currentHeat']` vs the schemas' `currentHP`/`currentAP`/`currentSP`. Applying a proposal writes a junk key beside the real one — HP unchanged on the sheet, proposal marked applied, body now fails `PilotSchema.strict()`. Same casing in `convex/crew.ts:66-67`, so `CrewVitals` renders "HP — · AP —" for everyone, always.
- **`apps/itun/src/components/games/MediatorScreen.tsx:227` — presence is dead on arrival.** ✅verified: nothing in `src` calls `api.mediator.heartbeat`, so `presence` is always empty and the "At the table" card never renders — indistinguishable from "nobody's here."
- **`apps/itun/src/components/account/ClaimLocalData.tsx:65` — the pattern store is never hydrated** (raw slice read without the `.getState().list()` every other consumer pairs it with). The card reports "0 saved patterns" and sends `mechPatterns: []` — patterns are silently left behind by the only path off the device.
- **`MediatorScreen.tsx:171` — `before` is unconditionally `null`** even though `crew.vitals` has the value, so `ProposalInbox`'s documented mismatch-detection can never fire; and `after: Number(value)` on a free-text input proposes `NaN` for "ten", which the empty-string disabled check doesn't catch.
- **`convex/entities.ts:307` — `claimLocal` inserts the placeholder Game + membership *before* parsing the crawler**, so a user whose only crawler fails validation is left in a phantom empty "Claimed crawler" game.
- Broader than the already-reported `ClaimLocalData` case: **every mutation across `GamesScreen`/`ProposalInbox`/`DowntimePanel`/`AccountScreen` is `void mutate(...)` with no `.catch`** while the server throws on ordinary races (two players applying the same proposal, a demoted organizer). Only `redeem` surfaces its error.

Combined with the earlier 14, the branch has 25 distinct findings. The four unauthenticated/under-authorized Convex endpoints are the ones I'd fix before this deploys anywhere — `recordRoll` in particular accepts writes from anyone on the internet.

alxjrvs (commented):
Reviewed the `itun-accounts-legacy-data` branch (19 commits, 68 files, ~7.5k insertions) against fresh `origin/main` — note that local `main` was 35 commits stale, so the initial `main...HEAD` diff was 4x oversized; I re-fetched and scoped to the real PR.

12 findings, most severe first:

- `apps/itun/convex/crew.ts:66` — `crew.vitals` reads `currentHp`/`currentAp`/`currentSp`, but the Zod schemas define `currentHP`/`currentAP`/`currentSP`. The crew strip renders `—` for every vital in production; the convex tests pass only because their fixtures use the wrong casing too.
- `apps/itun/src/components/games/MediatorScreen.tsx:37` — `PROPOSABLE_FIELDS` uses the same non-existent keys. Applying such a proposal injects an unknown key into a `.strict()` body, after which every `entities.update` for that pilot throws `Invalid pilots payload` — the player can never save that entity again.
- `apps/itun/convex/proposals.ts:160` — `proposals.apply` patches the body with no Zod parse, breaking the contract that `v.any()` bodies are validated at the mutation. `field`/`after` are unconstrained, so `NaN`, `id`, or any junk persists unchecked.
- `apps/itun/convex/crew.ts:96` — `readEntity` validates a `table` arg it never uses; `db.get` on an arbitrary id lets any game member read an `encounterNpcs` body — the one table the module says must stay Mediator-only.
- `apps/itun/src/components/account/ClaimLocalData.tsx:74` — the one-time-claim marker is keyed on roster counts, not the account. Add one pilot and the prompt returns; re-claiming duplicates the whole roster, and duplicate `appId`s then make `byAppId().unique()` throw, silently killing write mirroring for those entities forever.
- `apps/itun/src/stores/entityStore.ts:264` — `mirrorEntityWrite` sends `record.gameId ?? null` instead of using `containerOf()`. New entities (which only get `workspaceId`) mirror as shelf; migrated ones carry a local UUID that Convex rejects as `v.id('games')`, so nothing in a workspace ever mirrors — swallowed into a `console.warn`.
- `apps/itun/convex/bot.ts:140` — `recordRoll` never calls `requireUser`. Any unauthenticated caller with a channel id + a Discord snowflake (both public) can forge Change Log rolls attributed to another player.
- `apps/itun/convex/account.ts:189` — deleting the last member's account drops the game but leaves `channelBindings`/`downtime`/`changeLog`. The stale binding makes that Discord channel permanently unbindable: `bindChannel` throws "already bound", `unbindChannel` throws "Not a member" on the deleted game.
- `apps/itun/src/components/account/ClaimLocalData.tsx:118` — no `.catch` on the claim chain; a failure is an unhandled rejection and the card silently reverts, inviting the duplicate re-click above.
- `apps/itun/convex/entities.ts:186` — `update`/`remove` ignore `table` when resolving the id; a mech id with `table: 'pilots'` writes a pilot body onto the mechs row.
- `apps/itun/convex/__tests__/harness.ts:25` — not a `*.test.ts` and no `convex.json` exclusion, so it's in the generated module map and gets pushed by the new `bunx convex deploy` step, pulling in the devDependency `convex-test`.
- `apps/itun/convex/entities.ts:277` — `claimLocal` inserts the placeholder "Claimed crawler" game before parsing; if every crawler fails, the user is left in a permanent empty game.

alxjrvs (commented):
Reviewed PR #647 (`itun-accounts-legacy-data` → `main`, 22 commits, ~8.5k added lines: Convex accounts/Games/Mediator backend, ITUN client surfaces, v13 migration, export/import container split, coverage tooling).

10 findings, most severe first:

- `apps/itun/convex/bot.ts:132` — `recordRoll` is a public mutation with **no auth check at all**; it trusts the caller-supplied `discordId` as the actor, so anyone can forge Change Log rolls into any bound Game.
- `apps/itun/src/stores/entityStore.ts:264` — the mirror passes the *local workspace UUID* as Convex's `v.id('games')`; every mirrored write for an entity in a non-Default workspace fails argument validation and is swallowed into a `console.warn`.
- `apps/itun/src/stores/entityStore.ts:259` — `withActiveWorkspace` sets only `workspaceId`, never `gameId`, so every newly created entity mirrors to the server **shelf** regardless of which workspace/Game it belongs to.
- `apps/itun/src/lib/export/parseImportBundle.ts:95` — imports now carry `gameId`, but `mergeImport` only remaps `workspaceId`; the source device's `gameId` rides through in `...rest` and `containerOf()` prefers it → imported entities land in a phantom container.
- `apps/itun/src/components/account/ClaimLocalData.tsx:74` — the "already claimed" marker is keyed on roster item counts, and `claimLocal` blind-inserts; a second claim duplicates every row, and duplicate `appId`s make `byAppId`'s `.unique()` throw, permanently breaking all later mirrored writes.
- `apps/itun/convex/proposals.ts:160` — `apply` patches the entity body with an arbitrary field/value and never Zod-parses, violating the schema header's stated mandatory invariant (concrete: `Number("a lot")` → `NaN` persisted into `currentHp`).
- `apps/itun/convex/account.ts:189` — deleting the last member removes the Game but leaves `downtime`, `channelBindings`, `changeLog` and `mechPatterns` rows dangling; a stale `channelBindings` row makes that Discord channel permanently un-bindable and un-unbindable.
- `apps/itun/src/components/games/MediatorScreen.tsx:226` — nothing ever calls `mediator.heartbeat`, so `presence` is never written and the "At the table" card never renders.
- `apps/itun/convex/__tests__/harness.ts:1` — the harness is not a `.test.ts` file, so Convex codegen treats it as a deployable module (it's in `_generated/api.d.ts`); `convex deploy` (now run by `netlify.toml`) will bundle the `convex-test` devDependency into production.
- `apps/itun/src/lib/connection/ConnectionProvider.tsx:59` — `isAuthenticated` is false during session restore and the mode is pushed from an effect, so writes made in that window resolve to `local` and are never mirrored.

Also worth noting (not filed as defects): `entities.listForGame`, `listShelf`, `create`, `update`, `remove`, `patchCrawler`, `crew.readEntity`, and the whole `bot.*` surface have no client callers yet — reads still come exclusively from IndexedDB, so "Convex is the source of truth" is currently write-mirror-only.

alxjrvs (commented):
Reviewed PR #647 (`itun-accounts-legacy-data` → `main`, ~8.5k added lines: the Convex accounts/Games backend, the Games/Mediator/Account client surfaces, the v13 container migration, and the coverage-ratchet fix). 15 findings.

**Findings**

- `apps/itun/convex/bot.ts:132` — `recordRoll` has no caller authentication at all; it trusts a client-supplied `discordId`, so anyone with the (bundled) deployment URL can forge Change Log rolls attributed to any member of any bound Game.
- `apps/itun/convex/proposals.ts:160` — `apply` patches an entity body with no Zod parse, breaking the module-wide "every write validates" invariant; the Mediator UI's unfiltered `PROPOSABLE_FIELDS` lets `currentHeat` be applied to a pilot, after which the `.strict()` `PilotSchema` rejects the row on every later write.
- `apps/itun/src/stores/entityBackend.ts:110` — the mirror passes a *local workspace UUID* as `v.id('games')`, so every mirrored write for a non-shelf entity fails and is swallowed into a `console.warn`; new entities never get `gameId` at all and mirror to the shelf.
- `apps/itun/convex/crew.ts:96` — `readEntity` ignores its `table` arg and gates only on membership, so any doc id in a Game you belong to is readable, including the Mediator-only `encounterNpcs` tray.
- `apps/itun/convex/entities.ts:385` — `by_app_id` is not owner-scoped and `byAppId` uses `.unique()`; two users importing the same bundle (or one double claim) permanently break mirroring for that entity.
- `apps/itun/src/components/account/ClaimLocalData.tsx:74` — the one-time claim marker is keyed on roster *counts*, not the account; adding one pilot re-offers the claim and a second claim duplicates the entire roster server-side.
- `apps/itun/convex/templates.ts:84` — template entities are inserted without `appId` while `STARTER_SOFT_LINKS` reference app ids, so every mech↔pilot and pilot↔crawler link in a Starter Set Game dangles.
- `apps/itun/convex/entities.ts:348` — `claimLocal` puts pilots/mechs on the shelf but stamps softLinks with the placeholder "Claimed crawler" game, so the links live in a container holding none of their referents; `mechPatterns` also skip Zod entirely.
- `apps/itun/src/components/games/MediatorScreen.tsx:171` — the propose form hardcodes `before: null`, defeating the documented before/after mismatch surface even though `crew.vitals` already has the real value.
- `apps/itun/convex/account.ts:189` — deleting a last-member Game leaves `downtime`, `channelBindings` and `changeLog` orphaned; the stale binding makes the channel permanently un-unbindable.
- `apps/itun/convex/proposals.ts:216` — `alerts`, `bot.rolls` and `propose` all `.collect()` the whole game/entity changeLog and filter in memory; on a never-pruned append-only table this will hit Convex read limits.
- `apps/itun/convex/bot.ts:185` — `linkDiscordId` is an unverified self-assertion, and nothing in the OAuth flow ever writes `users.discordId`, so the bot path is inert as shipped and spoofable when used.
- `apps/itun/convex/_generated/api.d.ts:34` — `__tests__/harness` is now registered as a deployed Convex module, pulling the `convex-test` devDependency into the newly-enabled `convex deploy` (PLAUSIBLE — worth verifying against the Convex bundler's ignore rules).
- `apps/itun/src/stores/entityStore.ts:510` — `delete` mirrors the remote delete *before* the local one, inverting the order used by `create`/`update`.
- `apps/itun/netlify.toml:28` — the "without the deploy key it's a plain Solo build" claim doesn't hold on production, where `VITE_CONVEX_URL` is already set: the `else` branch ships Connected against an undeployed backend.

PR #649, opened 2026-07-28
coverage: fix workspace attribution, and take three workspaces past 95%

alxjrvs:
Two commits: an accuracy fix, then the coverage it made worth chasing.

## 1. The ratchet was measuring the wrong files

`SF:` records were resolved against the lcov's own directory, but Bun writes
those paths relative to the cwd the run happened in — the workspace root, not
`<workspace>/coverage`. Every path was one level too deep, and whether that
mattered depended on how far the import reached.

`apps/srd` and `apps/itun` reach a sibling as `../../packages/…`, which lands
outside the workspace either way, so they read correctly and the bug stayed
invisible. `packages/component-lib` sits one level shallower: its
`../salvageunion-reference/…` resolved to
`packages/component-lib/salvageunion-reference/…`, which looks local — so
component-lib alone was charged for **4,600 lines of the reference package**,
already measured at 92% in its own run.

The failure mode worth naming: the gate still passed, still printed a number,
and the number was of something else.

`parseLcov` moves to `tools/lib/` so the attribution rule is testable directly
instead of only through a full report run. **Three of the seven new cases fail
against the old resolution.**

## 2. Coverage

| Workspace | Before | After | |
| --- | --- | --- | --- |
| packages/salvageunion-reference | 92.13% | **96.08%** | ✅ |
| apps/discord-bot | 92.40% | **95.47%** | ✅ |
| apps/srd | 95.57% | 95.57% | ✅ |
| packages/component-lib | 76.18% → 86.15% (fix) | **87.32%** | |
| apps/itun | 89.50% | 89.50% | |

- **`observability.ts`** (31%) — the bot's only channel for learning that
  production broke, untested in both directions. A bot that has silently
  stopped reporting looks exactly like a bot with no errors.
- **`helpers.ts`** (13.7%) — schema/alias resolution, grant walking, facet
  extraction, and `extractStaticEntitySummary`, which the SRD's no-JS and
  crawler rendering is built from.
- **`InlineEditField`** (3%) — the edit-in-place atom every sheet value goes
  through, untested on the paths that quietly lose a player's typing.

### What is still short of 95%, and why

`component-lib` needs 658 more covered lines and `itun` 920, concentrated in
`RollTable` (859 lines, nondeterministic rolls), `ReferenceEntityCard`, and the
itun sheet surfaces. Those are worth doing but are their own piece of work, not
a rider on this one.

### DIFF
--- a/.claude/rules/testing-patterns.md
+++ b/.claude/rules/testing-patterns.md
@@ -42,6 +42,46 @@ Testing patterns using Bun's test runner and Testing Library for component tests
 - Use `mock()` from `bun:test` for module/function mocking
 - Mock external dependencies appropriately
 
+### `mock.module` is process-global — restore what you replace
+
+`mock.module` rewrites the entry in the module registry for the whole test
+process, **not** for the file that called it. Every test file that runs after
+yours gets your replacement. This is not theoretical: mocking the entity store
+in one component test broke 219 tests across the suite, and the failures
+surfaced in files that had never heard of the mocked module.
+
+Capture the real exports first, and put them back in `afterAll`:
+
+```typescript
+// The spread is load-bearing. A module namespace is a LIVE view, and mocking
+// rewrites it in place — hold the namespace itself and by `afterAll` it
+// already reads as the mock, so you restore the mock over the mock.
+const realStore = { ...(await import('../../stores/entityStore')) }
+
+mock.module('../../stores/entityStore', () => ({ useEntityStore: fake }))
+
+afterAll(() => {
+  mock.module('../../stores/entityStore', () => realStore)
+})
+```
+
+Two corollaries:
+
+- **Partial mocks break importers you did not think about.** Replacing
+  `convex/react` with only the hooks under test made `@convex-dev/auth/react`
+  fail at import on a missing `ConvexProviderWithAuth`. Mock every export the
+  transitive importers reach, not just the ones your component calls.
+- **Beware module-scope environment reads.** `config.ts` reads `process.env`
+  once at import, so setting an env var to drive a test hands that value to
+  every later file too. Mock the config module instead of setting the variable.
+
+### Call `cleanup()` explicitly in component tests
+
+Testing Library's auto-cleanup has proven order-dependent here — a file passed
+alone and failed in the full suite. When several tests in a file render the
+same accessible name, one leaked render turns every later query into "found
+multiple elements". Add `afterEach(cleanup)` rather than depending on it.
+
 ## Examples
 
 **Package test:**
--- a/apps/discord-bot/src/__tests__/observability.test.ts
+++ b/apps/discord-bot/src/__tests__/observability.test.ts
@@ -1,0 +1,151 @@
+import { afterAll, beforeAll, describe, expect, mock, test } from 'bun:test'
+
+/**
+ * observability — the bot's only channel for finding out that production
+ * broke. It is entirely env-gated, and the gate is the interesting part: with
+ * no DSN every function must be a silent no-op (local dev and CI run that way),
+ * and with one it must actually report. A regression in either direction is
+ * invisible until the day you need it — a bot that has stopped reporting looks
+ * exactly like a bot with no errors.
+ *
+ * `config` is mocked rather than driven through `process.env`, deliberately.
+ * `config.ts` reads the environment once at module scope, so setting
+ * `SENTRY_DSN` here would hand a live DSN to every test file that runs after
+ * this one — `mock.module` is process-global in Bun, not file-scoped, and the
+ * env is worse still. Both the mocks below are captured and restored.
+ */
+
+const sentryCalls: Array<{ fn: string; args: unknown[] }> = []
+let flushShouldThrow = false
+let dsn: string | undefined
+
+// `config.ts` throws at module scope on a missing token, and it is evaluated by
+// the capture below. Same `??=` as the sibling tests, and deliberately no
+// SENTRY_DSN: the DSN is supplied through the mock, never the environment.
+process.env.DISCORD_TOKEN ??= 'test-token'
+process.env.DISCORD_CLIENT_ID ??= 'test-client-id'
+
+const realSentry = { ...(await import('@sentry/node')) }
+const realConfig = { ...(await import('../config.js')) }
+
+mock.module('@sentry/node', () => ({
+  init: (...args: unknown[]) => sentryCalls.push({ fn: 'init', args }),
+  flush: async (...args: unknown[]) => {
+    sentryCalls.push({ fn: 'flush', args })
+    if (flushShouldThrow) throw new Error('transport down')
+    return true
+  },
+  captureException: (...args: unknown[]) => sentryCalls.push({ fn: 'captureException', args }),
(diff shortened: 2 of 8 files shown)


alxjrvs (commented):
**Scope note:** the branch is a stacked PR (#647, base `itun-accounts-surfaces`), so I reviewed `origin/itun-accounts-surfaces...HEAD` — 18 files, mostly the `claimLocal` expansion, the `ClaimLocalData` card, and the v1→v2 export-bundle migration. (`@{upstream}...HEAD` and `git diff HEAD` were both empty; local `main` is stale behind `origin/main`.)

Findings, most severe first:

- `apps/itun/src/lib/export/parseImportBundle.ts:100` — `assignContainers` now writes `gameId` into imported entities, but `mergeImport` (mergeImport.ts:142/164/186) only remaps `workspaceId`. The stale source-device `gameId` survives in `...rest` and wins in `containerOf`, so a cross-device import files entities into a Game that doesn't exist locally. Regression: without `gameId` the remapped `workspaceId` fallback placed them correctly.
- `apps/itun/src/components/account/ClaimLocalData.tsx:74` — the "one-time" marker is keyed on `pilots.length-mechs.length-crawlers.length`, not the account. Adding one local pilot after claiming re-offers the card, and `claimLocal` only inserts (no `appId` dedupe), so a second click duplicates the entire roster server-side — the exact failure the component's docblock says the marker prevents.
- `apps/itun/src/components/account/ClaimLocalData.tsx:65` — patterns are read via `usePatternStore((s) => s.mechPatterns)`, which does not trigger lazy hydration (unlike the sibling `s.list(...)` calls). On a direct load of `/account` nothing hydrates that store, so 0 patterns are uploaded and `markClaimed` then burns the one-time offer. Fix: `s.list()`.
- `apps/itun/convex/entities.ts:348` — all claimed soft links get `gameId: shelfGameId` (the placeholder "Claimed crawler" Game, or `null`), while their endpoints go to the shelf. `listForGame` returns links with no endpoints; `listShelf` returns no links at all — every claimed relationship is unreachable.
- `apps/itun/src/components/account/ClaimLocalData.tsx:107` — no `.catch` on the `claimLocal` promise: a rejected mutation is an unhandled rejection and the card silently reverts with no error shown.
- `apps/itun/convex/entities.ts:307` — the placeholder Game + membership are inserted before any crawler is validated, leaving an empty orphan "Claimed crawler" Game if every crawler fails `CrawlerSchema`.
- `apps/itun/convex/entities.ts:353` — soft links are folded into `claimed` but excluded from the card's offered `total`, so the confirmation reports more items than were ever offered.

One non-finding worth mentioning: the docblock at `parseImportBundle.ts:51` still says "schemaVersion is not 1 (incompatible format)" and is now stale.

alxjrvs (commented):
## Review: `coverage-attribution` (2 commits, 7 files)

Scope resolved to `origin/main...HEAD` — `664cf537` (extract `parseLcov` into `tools/lib/`, resolve `SF:` against the workspace root instead of the lcov dir) and `20ecde88` (three new test files + baseline bumps).

**The core fix is correct.** I verified the premise directly against a real lcov: `packages/component-lib/coverage/lcov.info` contains `SF:src/catalog/catalogColors.ts` and `SF:../salvageunion-reference/lib/BaseModel.ts` — paths relative to the workspace root (the test-run cwd), not to `coverage/`. Resolving against `workspaceRoot` is right, and the `apps/srd` case that previously worked by accident still resolves outside its workspace either way, so the change is strictly better.

**The baseline bumps hold.** I ran the full coverage pipeline before the tree shifted: salvageunion-reference 96.08% (base 95), component-lib 88.57% (87), srd 95.57% (95), itun 89.48% (89), discord-bot 95.47% (95). Ratchet passes, exit 0. `bun run knip` is clean, full test suite is 0 fail.

### Findings

- `tools/lib/parse-lcov.test.ts:1` — the new regression suite never executes. `bun run test` is `bun --filter "*" test`, which I ran: it targets only the five workspaces, never the root package, so nothing under `tools/` runs in CI or pre-push. Reverting the attribution fix would keep every gate green. (Pre-existing for `tools/run-coverage.test.ts` too — the fix is one root-level test invocation.)
- `tools/lib/parse-lcov.test.ts:71` — the "path that merely starts with the workspace name" test builds `<tmp>/x-extra/...` against a root of `<tmp>/lcov-XXXXXX`. Those share no prefix, so a naive `abs.startsWith(workspaceRoot)` implementation passes this test and every other test in the file. The documented guard isn't tested.
- `packages/salvageunion-reference/lib/helpers-display.test.ts:94` — `expect(resolved.length).toBeLessThanOrEqual(nonChoice.length)` is satisfied by `[]`, and `resolved.every(...)` is vacuous on `[]`, so the test cannot fail if grant resolution breaks entirely. I measured the real data: 14 granters, 15 resolved, 15 non-`choice` grants — exact equality is available.
- `apps/discord-bot/src/__tests__/observability.test.ts:58` — `afterAll` restores both module mocks but leaves `observability.ts`'s module-level `enabled = true`. `ready.test.ts` sorts later and calls `captureMessage` through the real `@sentry/node`. Inert today only because real Sentry was never `init`-ed.

### One thing to flag outside the review

**The worktree switched branches under me mid-review.** `/Users/jarvis/Code/SU-SRD/.claude/worktrees/accounts-legacy` went from `coverage-attribution` @ `20ecde88` to `itun-accounts-legacy-data` @ `e0e5ea19` partway through — another session shares this worktree. All findings above were re-verified against the `20ecde88` blobs via `git show`, so they're sound, but the coverage numbers and `knip` result were captured before the switch and can't be re-run here.

alxjrvs (commented):
Reviewed the 3 commits on `coverage-attribution` vs `origin/main` (8 files: the lcov-attribution fix + extraction, three new test suites, a raised baseline, and a testing-patterns doc).

**Verification performed:** ran all three touched workspace suites in full (`salvageunion-reference` 1003 pass, `component-lib` 675 pass, `discord-bot` 80 pass), ran the new root test directly (7 pass), ran `knip`, biome and prettier on the changed files (all clean), and recomputed every workspace's coverage with the new parser against the branch's own lcov files. The attribution fix itself is correct — I confirmed against the real lcov output that Bun writes `SF:` relative to the workspace root (the run cwd), and that `packages/component-lib` is the only workspace whose numbers change, exactly as the code comment claims.

Findings:

- `tools/lib/parse-lcov.test.ts:1` — The new regression suite never executes. Root `test` is `bun --filter "*" test`, and the root package is matched by no filter (verified: `bun --filter 'SURef' run knip` → "No packages matched the filter"), so neither CI's `test` job, `pre-push`, nor `test:coverage` runs it. Reverting the fix would leave every gate green. The pre-existing `tools/run-coverage.test.ts` is dead for the same reason, which makes the guarantee asserted in `tools/run-coverage.ts`'s docstring unenforced.
- `coverage-baseline.json:6` — `apps/discord-bot` raised 92 → 95 while actual is 95.47%, leaving ~0.97pp (≈7 lines of 773) before the ratchet fails; a single small untested module fails the coverage job on an unrelated PR. The other two raises (`salvageunion-reference` 1.58pp, `component-lib` 2.07pp) have adequate slack.

PR #652, opened 2026-07-29
feat(itun): retire Workspaces in favour of Games and the Shelf

alxjrvs:
## Why

ADR-030 §2 replaced Workspaces with two containers — a shared **Game** and the owner's personal **Shelf** — but only the *data* layer had been cut over. Migration v13 and `lib/container.ts` landed in #609; the UI still rendered a Workspace switcher, a Workspace manager and a Workspace assign control, and still filtered the Roster by `workspaceId`. This finishes the client cutover.

## What

- `activeContainerStore` replaces `activeWorkspaceStore`, persisting one string (`shelf` | `game:<id>`) rather than a raw id.
- `ContainerSwitcher` / `MoveToContainerControl` replace `WorkspaceSwitcher`, `WorkspaceList` and `AssignToWorkspaceButton`. Both render **nothing** in Solo: with no account there are no Games, so a one-option select would be furniture that never does anything.
- Roster and Encounter resolve through `containerOf` + a new `sameContainer`.

### The one decision worth reviewing

**Solo surfaces render the whole pile unfiltered.** This is load-bearing rather than a shortcut: migration v13 mapped every non-Default workspace onto `gameId: <that workspace id>`, so a Solo user who once made custom Workspaces now has entities addressed by ids matching no real Game. Filtering their roster by container would render those **invisible** — they would open the app and find builds missing. Not filtering is both simpler and the only rendering that cannot lose one.

`sameContainer` is structural, not `===`: `containerOf` mints a fresh object per call, so reference equality would make every filter silently return nothing.

### Rehomed, because Workspaces were carrying them

| Thing | Was | Now |
| --- | --- | --- |
| Starter Set | its own Workspace | seeds onto the Shelf behind an opt-in **Load Starter Set** button; idempotence guards on its own rows, not a container record |
| Dashboard dial prefs | the Workspace record | `cockpitPrefsStore` — localStorage, keyed by container. Neither container is a right home: a Game is shared state and a dial layout is personal; the Shelf is not a record |
| Imported entities | `workspaceId` remapped | land on the importer's **Shelf**. Game membership comes from the server, not from a file |

`ExportBundle.workspaces` stays in the schema, permissive and always `[]`, so **existing backups keep parsing** — dropping the key would turn every saved export into a parse error.

The `workspaces` object store and migration v10 stay: v13 reads what v10 writes when an old database is opened. Constants are inlined where the deleted modules used to supply them (the same policy migration v13 already follows).

## Verification

- `bun run check:all` → exit 0 (lint, format, typecheck ×4, tests, validate, knip, audit, tokens, styling).
- 1503 itun tests pass, 0 fail. New coverage for `activeContainerStore` (encoding round-trip + tolerant parse + persistence) and `sameContainer`.
- Stale knip entry for the deleted `workspaceStore` replaced with the two new stores.

## Note for the reviewer

`MoveToContainerControl` **re-stamps the container in place**, exactly as the control it replaces did. ADR-030 says a cross-container move should be an explicit *fork*, and no fork mutation exists yet (`convex/entities.ts` has create/update/claimLocal). Turning it into a fork is a separate change with its own server surface — deliberately not smuggled in behind a rename.

https://claude.ai/code/session_01E3Uy8Pgtx2AcJJgHD9trbQ


alxjrvs (commented):
Reviewed `main...HEAD` (the "retire Workspaces in favour of Games and the Shelf" commit, ~3.5k lines across 67 files). Baseline checks are green: `typecheck:itun`, `bun --filter itun test` (1503 pass / 0 fail), and `bun run knip` all pass, so nothing here is a build break — the findings are behavioural.

Findings (most severe first):

- `apps/itun/src/lib/starterSet/seedStarterSet.ts:47` — Re-seeding the Starter Set overwrites user edits. The guard moved from "the Starter Set workspace exists" to "every seeded pilot id is present", so deleting one seeded pilot brings the "Load Starter Set" button back, and clicking it `put`s all seed records over the user's customized copies (and resurrects the deleted row) — the exact behaviour the deleted module promised never to do.
- `apps/itun/src/components/container/ContainerSwitcher.tsx:63` — The switcher renders only `Shelf` + `api.games.listMine`, so a Connected user whose entities carry a v13 phantom `gameId` (a retired custom workspace) or a left Game has no option to select that container; the Roster filters them out and they become unreachable from every surface. `MoveToContainerControl` handles this case with an "Unknown game" option; the switcher does not.
- `apps/itun/src/components/roster/Roster.tsx:186` — `mode !== 'connected'` lumps `disconnected` (signed in + offline) in with `solo`, so going offline merges every Game's builds into one unfiltered list with the switcher gone. Same branch in `EncounterScreen` also breaks NPC sibling numbering across containers.
- `apps/itun/src/stores/entityStore.ts:311` — `activeContainer` is never reset or reconciled on sign-out, and `withActiveContainer` stamps new builds with it regardless of mode; a signed-out user's new entities get the last Game's id, and a second account on the same browser lands on an empty filtered roster while mirroring upserts with a foreign `gameId`.
- `apps/itun/src/components/sheet/SheetMech.tsx:158` (and `SheetPilot.tsx:147`) — These pass `containerOf(entity)` unconditionally while `Roster` passes `undefined` in non-connected mode, so a Solo user with phantom containers gets an empty, unlaunchable pilot picker from the sheet even though the Roster shows those pilots.
- `apps/itun/src/lib/export/mergeImport.ts:125` — All imports are forced to the Shelf, but the Connected Roster filters by the active Game, so an import toasts "N entities created" with nothing visible on screen.
- `apps/itun/src/lib/export/parseImportBundle.ts:96` — The v1→v2 `assignContainers` step is now entirely discarded by `mergeImport`, making `legacyContainers.ts` dead and the "must not land somewhere different depending on how it arrived" comment false.

PR #653, opened 2026-07-29
feat(discord-bot): the bot as an authenticated ITUN Game client (#623)

alxjrvs:
Implements **Phases 0–4** of the plan added in the first commit. Phases 5 (alerts push) and 6 (proposal Apply/Decline) stay deferred, exactly as the plan drew them. Closes the intent of #623.

## Two defects closed first

- ⚠️ **`recordRoll` was a public Convex mutation with no authorization at all.** It trusted a `discordId` passed as an argument, and the deployment URL ships in the SPA bundle — so once a channel were bound, anyone could forge Change Log entries attributed to a real player. It is now `internal.botClient.recordRoll`; an **internal** Convex function is unreachable from any client, which is a stronger fix than a check inside a public mutation because it can't be undone by someone later forgetting the check.
- **`gameForChannel` opened with `requireUser`**, so the bot — which holds no Convex token — could never call the one query it most needed.

## A third, caught while rebasing onto #656

#656 found that `crew.vitals` read `currentHp` while the Zod schemas define `currentHP`, so every vital on the Mediator's crew strip was null and rendered as an em-dash — indistinguishable from an undamaged crew. **The bot's crew board was written from that same wrong spelling** and would have shipped the same lie, on the surface whose entire purpose is showing those four numbers. Reads are now salvage-tolerant (canonical first, historical second) and the spelling is pinned by a test asserting real derived numbers rather than the field names it was written with.

## There was never a linking step to build

Discord is the sole auth provider, so `@convex-dev/auth` has already written the snowflake to `authAccounts.providerAccountId` before anyone could be asked to paste it. `users.discordId` is now stamped in an `afterUserCreatedOrUpdated` callback and backfilled by an internal mutation; `linkDiscordId` is deleted. `/su me` is a confirmation, not a form.

## The credential

A `/bot/*` HTTP route with a bearer secret that authenticates the **bot**, never the **actor**. Every forwarded call re-resolves its actor from a linked Discord id and runs the same `model/permissions.ts` checks the web runs — so holding it lets you ask *"what may this Discord user see?"*, never *"show me everything"*. `model/permissions.ts` gained explicit-actor `…As` variants and the token-based functions now delegate to them, so there is one implementation of every rule rather than a second copy for the bot. The residual risk is stated plainly in `botHttp.ts` and in the plan's §3, along with the endgame (Discord-signed interactions).

## What shipped

`/su me`, `/su games`, `/su shelf`, `/su crew`, `/su sheet`, and the `/su game bind|unbind|info` group. Roll attribution is an upgrade in place — `/su roll` and `/su check` gain no option; in a bound channel their footer grows ` · recorded to <Game>`.

## Judgement calls worth reviewing

- **No `convex` dependency was added**, though the plan called for one. With the surface exposed as HTTP routes the client is `fetch` and nothing else; the package buys a websocket only Phase 5 needs.
- **Rolls reply before recording, never deferred.** A dice bot must feel instant, and the reference commands must behave identically with no ITUN deployment configured. The footer is edited afterwards, only once the recording lands.
- **Game commands defer ephemerally and send public results as a follow-up.** Discord fixes ephemerality at defer time, so deferring publicly would put *"you are not a member of this game"* — a fact about a person — into the channel.
- **The re-roll buttons record too**, or the Change Log would quietly disagree with the channel about what happened at the table.
- **Maxima are derived in the bot**, not fetched: Convex stores bodies opaquely and can't compute max HP/SP/Heat, so `gameEmbed.ts` uses `salvageunion-reference/rules` (ADR-006).
- **`itunReply.ts` exposes one named test seam.** `config.ts` reads env at module scope and the client is resolved once at import, so an env var does nothing and `mock.module` is process-global in Bun. It returns a restore function rather than exposing a setter.
- **`buttons.ts` now takes a narrow interaction type** like the command handlers, which is what removed the last cast from the tests.
- **The _Post to channel_ button was dropped.** `/su crew` is public already — the case it was meant to serve.
- `apps/itun/convex/_generated/api.d.ts` is **hand-edited**: `convex codegen` needs a configured deployment. It matches the generator's output exactly, but is worth re-running codegen over.

## Verification

**All 20 CI checks green**, including `coverage` and the `quality-checks` aggregate, rebased onto current main (#656). Locally `bun run check:all` exits 0 and `bun run build:bot` bundles clean.

Coverage: the ratchet caught the new code landing at **75.43%** against a 95% baseline; it is now **95.96%**. The added tests target invariants rather than the percentage — a denial on a public command staying ephemeral, an outage reading as an outage rather than a permissions problem, `/su sheet` refusing a hand-typed value before any network call, a thrown client never escaping past the user's roll, and every client method sending the arguments the server declares.

`soloMode.test.ts` pins that the reference bot is untouched with no ITUN deployment configured — **the guarantee that must not regress**.

## Still open for you

1. **Credential model** — shipped option A (shared secret). Option B (Discord-signed interactions on Convex) remains the documented endgame.
2. **Deployment** — needs `ITUN_BOT_SECRET` on the Convex deployment and `ITUN_CONVEX_SITE_URL` + `ITUN_BOT_SECRET` on Render (both now declared in `render.yaml`). Until then the bot stays in Solo mode, which is safe and deliberate.
3. **Glyph check** — the `█`/`░`/`▲` gauge glyphs want eyeballing on desktop + mobile, light + dark, before they are considered locked.

https://claude.ai/code/session_01BDQLC27L4KFJDN32FhE2ii


alxjrvs (commented):
Reviewed commit `8919d383` against `origin/main` — a docs-only diff (3 files, +501/-10): the new `docs/architecture/discord-bot-game-client.md` plan, a Phase 6 status correction in `accounts-and-games.md`, and a `docs/README.md` index row.

I verified the plan's factual claims against the code rather than only reading prose. **Everything it asserts about current state checks out**: `convex/bot.ts` is exactly 204 lines, `recordRoll` genuinely has zero auth, `gameForChannel` genuinely opens with `requireUser`, `grep` genuinely finds no `api.bot.*` callers in either app, the bot declares only `GatewayIntentBits.Guilds`, `lookupEmbed.ts` has the stated `LIMIT` budget and `0xb7410e`, `customId.ts` has the `su:<action>:<payload>`/100-char scheme, `--external` is in `apps/discord-bot/package.json`, prettier `format:check` passes, and I ran `bun run knip` — it is green with `bot.ts`'s uncalled exports, so the "not a risk: knip" claim holds.

The problems are all in the forward-looking design, and the first one is load-bearing.

Findings (7):

- `docs/architecture/discord-bot-game-client.md:94` — §3 Option A claims the bot route runs "the same `model/permissions.ts` checks"; every entry point there resolves the actor via `getAuthUserId(ctx)`, which is null under a shared-secret request, so all ten functions in §6's Convex column throw "Not signed in". Phase 0 doesn't budget the actor-injection refactor this requires.
- `docs/architecture/discord-bot-game-client.md:92` — `Authorization: Bearer $ITUN_BOT_SECRET` collides with the header Convex HTTP actions reserve for JWT auth; use a distinct header.
- `docs/architecture/discord-bot-game-client.md:239` — the `· recorded to Tenacity` footer forces a blocking binding lookup on `/su roll`'s reply path, which §9's `deferReply()` mitigation explicitly does not cover and §5's Degraded rule forbids breaking.
- `docs/architecture/discord-bot-game-client.md:160` — retiring `linkDiscordId` drops its duplicate-snowflake guard while `resolveActor` still reads `by_discord` with `.unique()`, which throws instead of returning null.
- `docs/architecture/discord-bot-game-client.md:74` — the "no new mutations" non-goal contradicts what Phase 0 and §6 actually require (and Phase 0 already breaches it).
- `docs/architecture/discord-bot-game-client.md:181` — the §5 mode matrix has no row for `CONVEX_URL` set with the secret unset, a state Phase 0 makes reachable.
- `docs/architecture/discord-bot-game-client.md:461` — "Watermark by `changeLog.ts`" names a module that doesn't exist; the real read path is `proposals.alerts`, which has no `since` cursor.

alxjrvs (commented):
Reviewed `origin/main...HEAD` (3 commits: the ITUN Game-client plan, the bot implementation, and the deployment-vars doc) — ~3.8k added lines across `apps/discord-bot`, `apps/itun/convex`, and docs. Bot tests pass (115); ITUN's one intermittent failure is a pre-existing `ShareSnapshotScreen` timeout unrelated to this diff.

The design (bot credential authenticates the bot, actor resolved from a linked Discord id, shared `model/permissions` `…As` variants, internal-only functions behind one HTTP door) is sound. The defects are in the wiring at both ends of the identity/data path — and notably the two most severe would each make the headline feature silently not work in production while every test stays green.

Findings, most severe first:

- `apps/itun/convex/auth.ts:44` — `profile.id` is always `undefined`: `@convex-dev/auth` destructures `id` out of the provider profile before calling `afterUserCreatedOrUpdated`, so `users.discordId` is never stamped and every bot command returns `unlinked` until an operator manually runs `backfillDiscordIds`.
- `apps/discord-bot/src/gameEmbed.ts:127` — reads `currentHp`/`currentAp`/`currentSp`, but the Zod schemas store `currentHP`/`currentAP`/`currentSP`, so HP/AP/SP always render `0/max` on `/su crew` and `/su sheet`. Test fixtures share the same typo. (`convex/crew.ts` on main has it too, outside this diff.)
- `apps/itun/convex/botClient.ts:283` — `sheet` never uses the `table` arg, so any member can read any same-Game document by id, including the Mediator-only `encounterNpcs` the module header says it can "never" reach. `ctx.db.normalizeId` fixes this and the malformed-id throw.
- `apps/discord-bot/src/gameEmbed.ts:100` — `gauge` treats an absent current value as 0, but absent means "at max" in the schemas; an undamaged crew renders as wiped out.
- `apps/discord-bot/src/gameEmbed.ts:173` — embed deep links are built from the Convex `_id`, while `/sheet/$kind/$id` resolves the local `appId` UUID, so every shelf/sheet link is dead. (`gameUrl`'s `?game=` param is also inert — `games.tsx` has no `validateSearch`.)
- `apps/itun/convex/botHttp.ts:131` — the catch-all returns every error as 400 with the raw internal message, contradicting its own "deliberately opaque" comment and conflating server faults with bad arguments.
- `apps/discord-bot/src/gameEmbed.ts:357` — the `' unclaimed'` bucket sorts *first* under `localeCompare`, not last as the comment states (verified empirically).
- `apps/discord-bot/src/commands/crew.ts:117` — `indexOf(':') === -1` mis-slices (`pilotss` → table `pilots`) instead of rejecting, turning a typo into an "ITUN could not be reached" outage message.

One process note: partway through the review another process began a rebase in this worktree (HEAD moved to `644d1336` with a conflict in `docs/architecture/accounts-and-games.md`). I did not touch it; all findings are against the original `00df4d58` tree, and line numbers refer to that state.

alxjrvs (commented):
Reviewed the 5-commit PR scope (`644d1336...HEAD`, 42 files) — the Discord bot as an authenticated ITUN Game client. 7 findings:

- **apps/itun/convex/botClient.ts:283** — `sheet` casts an arbitrary string to a pilot/mech id and only checks `gameId`, never the table. A member typing `pilots:<encounterNpcId>` into the free-text autocomplete gets the Mediator-only NPC body back. `botHttp.ts` and `render.yaml` both claim this can "never see `encounterNpcs`". Fix with `ctx.db.normalizeId(args.table, …)`.
- **apps/discord-bot/src/gameEmbed.ts:130** — `gauge` treats an absent `currentHP`/`currentAP`/`currentSP` as `0`, while all 18 app call sites treat it as full (`?? max`). An undamaged pilot/mech renders as `░░░░░░░░░░ 0/12` on `/su crew` — the same class of bug as the `currentHp` fix in commit 2b82f103.
- **apps/discord-bot/src/gameEmbed.ts:196** — `sheetUrl` uses the Convex `_id`, but `/sheet/$kind/$id` resolves against `entityStore`, keyed by `appId`. Every sheet deep link in `/su shelf` and `/su sheet` is dead; `botClient` doesn't even return `appId`.
- **apps/discord-bot/src/commands/crew.ts:85** — autocomplete does a blocking Convex round trip with an 8s client timeout inside Discord's 3s autocomplete deadline; on a slow deployment `respond()` throws `Unknown interaction` per keystroke.
- **apps/discord-bot/src/gameEmbed.ts:191** — `gameUrl` emits `/games?game=<id>`; `games.tsx` has no `validateSearch` and the param is ignored. The real route is `/games/$gameId` (added on this branch).
- **apps/itun/convex/botHttp.ts:131** — every error is returned verbatim as a 400, contradicting the "deliberately opaque over the wire" comment and misclassifying server faults as client errors.
- **apps/discord-bot/src/gameEmbed.ts:380** — `' unclaimed'` with a leading space sorts *first* under `localeCompare`, not last as commented; past 25 buckets it displaces a real crewmate.

Things I checked and found sound: the `…As` permission split (one implementation per rule), `secretsMatch` timing behaviour, the unconfigured-is-404 choice, Solo-mode degradation, `stampDiscordId`'s refuse-don't-throw contract, and the rules-derivation field names (`maxHpModifier`, `chassisRef`, `systems`/`modules`) which do match the ITUN Zod schemas.

alxjrvs (commented):
Reviewed the full `main...HEAD` diff (139 files, ~8.6k insertions: the Discord bot's ITUN client + Game commands, the Convex bot HTTP door, and the Workspace→Container migration). Working tree is clean, so the committed range was the whole scope.

**Findings (4):**

- `apps/itun/convex/botClient.ts:295` — `sheet()` calls `ctx.db.get(args.entityId as Id<'pilots'|'mechs'>)` and only verifies `gameId`, never that the doc belongs to `args.table`. Convex ids are table-tagged but `db.get` returns any table's doc (this repo already guards for it in `ownership.ts:loadOwnable`). A Game member passing an `encounterNpcs` / `crawlers` / `mechPatterns` id gets its `body` back — contradicting the bound `botHttp.ts` and `render.yaml` both state ("never `encounterNpcs`"). Fix with `ctx.db.normalizeId(args.table, args.entityId)`.
- `apps/itun/src/stores/entityStore.ts:298` — crawler container moves never mirror. An update always routes to `patchCrawlerByAppId` (which deliberately never inserts), and a move to the Shelf returns early on `gameId === null`. `MoveToContainerControl` renders on crawler sheets, so moving a crawler into a Game looks like it worked locally while the Game gets no crawler (`gameHasCrawler` stays false, blocking the crew), and moving one out orphans the server row permanently. This is the exact bug the new re-home branch in `upsertByAppId` was added to fix for pilots/mechs.
- `apps/discord-bot/src/gameEmbed.ts:165` — `mechStats` passes no `PilotingContext` to `mechMaxSP`/`mechMaxHeat`, so `pilotedMech` ability contributions and `perTechLevel` amounts are dropped. Real data has such an ability, so `/su crew` and `/su sheet` render a different Max SP than the app for the same mech.
- `apps/discord-bot/src/commands/crew.ts:118` — `raw.indexOf(':')` result is used unchecked; with no colon, `table = raw.slice(0,-1)` and `entityId = raw`. Input like `mechsX` slips past the "pick from the list" guard and surfaces as "In The Union Now could not be reached" (an outage message for a typo).

Things I checked and cleared: the bearer-token comparison and 404-when-unconfigured in `botHttp.ts`; the `…As` permission split (no rule duplicated, wrappers delegate); `ownership.claim`'s free-only + transaction race handling; `proposals.apply` now parsing before persisting; the `currentHp`→`currentHP` rename across `crew.ts`/`CrewVitals`/`MediatorScreen` (all consumers updated, `encounterNpc.currentHp` is a separate schema); `containerOf`'s null-vs-undefined discipline and the v13 phantom-id reasoning in `Roster`/`activeContainerStore`; `mergeImport` shelving everything and `ExportBundleSchema` staying permissive for old bundles; and the `getSubcommandGroup()` dispatch in `su.ts` (discord.js defaults `required` to false there, so plain subcommands still fall through correctly).

alxjrvs (commented):
Reviewed PR #653 (`worktree-discord-bot-game-client-plan` → `main`): 42 files, +4865/−319 — the Discord bot as an authenticated ITUN Game client (`/bot/*` Convex HTTP route + bearer credential, `botClient`/`model/bot` split, `/su me|games|shelf|crew|sheet|game …`, roll attribution). I read every hunk plus surrounding context, and ran `bun run typecheck` (clean), `bun --filter discord-bot test` (176 pass), `bun --filter itun test` (1635 pass).

The authorization core holds up: `internal*` functions behind one credential-checked door, actor always re-resolved from `authAccounts` + a real membership, `normalizeId` making `table` load-bearing in `sheet`, `gameId` re-checked against the channel's binding, denials always ephemeral. I verified the `authAccounts.providerAccountId` claim against the installed `@convex-dev/auth` source, and every canonical body field name (`currentHP`/`currentSP`/`currentHeat`/`max*Modifier`/`max*Override`) against the ITUN Zod schemas.

Findings:

- `apps/discord-bot/src/commands/game.ts:110` — `/su game bind` passes the `game` option straight through as a `v.id('games')`; text typed instead of picked fails server-side validation and surfaces as "In The Union Now could not be reached." (an outage message) rather than "pick from the list". `/su sheet` guards this exact case; `bind` doesn't.
- `apps/discord-bot/src/gameEmbed.ts:179` — `mechMaxSP(input)` is called without the `PilotingContext` that `MechSheet.tsx:185` and `dialItems.ts:89` pass, so a mech piloted by a `Beefcake` pilot (`+3 + techLevel` Max SP, real data in `abilities.json`) renders a lower max than the app; a damaged one renders a full bar labelled `22/20`.
- `apps/discord-bot/src/config.ts:37` — `ITUN_WEB_URL` falls back only on `undefined` and is never validated; a blank or scheme-less value yields a relative URL, and `EmbedBuilder.setURL` throws (verified against the pinned discord.js), breaking every Game command with a generic error instead of degrading.
- `apps/itun/convex/bot.ts:21` — the rewritten module header still claims the actor resolves "against `users.discordId`", the mechanism `model/bot.ts` documents as never having worked; `schema.ts:315` repeats it and the `by_discord` index is now dead.

Non-findings I checked and cleared: `followUp` ephemerality after an ephemeral defer (discord.js sends via webhook, no flag propagation — the public path is correct), `editReply` preserving components in `attributeRoll`, `shelf`'s `gameId === null` against the schema's `v.union(v.id, v.null())`, `downtime.stepIndex !== null` likewise, `opFromPath` traversal, `secretsMatch` timing, autocomplete's 2s budget wiring, and the removed `recordRoll`/`linkDiscordId` having no remaining callers.

alxjrvs (commented):
Reviewed the whole PR diff (`origin/main...HEAD`, 8 commits, ~4.9k added lines): the Convex bot half (`botClient.ts`, `botHttp.ts`, `model/bot.ts`, `model/permissions.ts` `…As` split, schema/http wiring), the bot half (`itun/client.ts`, `gameEmbed.ts`, the new `/su me|games|shelf|crew|sheet` + `/su game` group, roll attribution), and `render.yaml`. I also ran both suites: `discord-bot` 177 pass, `itun` 1636 pass.

The authorization story checks out — I specifically verified the paths a reviewer should distrust: `normalizeId(table, …)` plus the `gameId` equality check does block an `encounterNpcs` or cross-table id from `/su sheet`; `resolveActor` re-resolves the actor from `authAccounts` on every forwarded call; `bind`/`unbind` go through `requireOrganizerAs`; the old unauthenticated public `recordRoll` is genuinely gone; and `shelf`'s `gameId === null` filter matches the schema (`v.union(v.id('games'), v.null())`, not optional). The bearer-secret residual risk is documented rather than hidden.

Findings:

- `apps/discord-bot/src/commands/roll.ts:110` — Recorded rolls carry no outcome. `buildRollMessage` keeps `outcome` local and returns only `{embeds, components}`, so `attributeRoll` forwards `{ table: 'Core Mechanic' }` as the `result`. The Change Log entry says a roll happened but not what it produced; same at `check.ts:94` (`{ notation }`) and `buttons.ts:58` (`{ rerolled: … }`).
- `apps/discord-bot/src/commands/game.ts:73` — Autocomplete choice `name` is unbounded user text with no `truncate`. A >100-char game name (nothing caps it: `games.create` only trims, no input sets `maxLength`) makes Discord 400 the whole `respond()`, so the picker is empty for *every* game — and since `bind` also rejects hand-typed values, the channel can't be bound at all. Same at `crew.ts:94/98` for long callsigns.
- `apps/discord-bot/src/gameEmbed.ts:179` (plausible) — `mechStats` calls `mechMaxSP`/`mechMaxHeat` without the `PilotingContext` the app passes, so a real ability (`abilities.json:602`, +3+TL Max SP to a piloted mech) makes the bot's derived max lower than the stored `currentSP`, rendering a full bar labelled "20/17".

Things I checked and deliberately did not report: public `followUp` after an ephemeral `deferReply` (Discord sets ephemerality per follow-up message via `flags`, so the "post publicly, fail privately" design holds); non-Convex-function exports from `convex/botHttp.ts` (`mediator.ts` already exports a const, so it's established); and the 25-field crew-embed slice (would need 25+ distinct owners).

PR #654, opened 2026-07-29
feat(itun): move the account controls into the masthead

alxjrvs:
The sign-in strip sat above the brand header on its own light band, which read as a second bar competing with the masthead rather than as part of it. Its controls now render **inside** the dark header on a row of their own, under the nav.

Encounter leaves the nav in the same change — it is reached through the Mediator sheet now, not as a top-level destination. The `/encounter` route itself is untouched; only the nav entries (desktop + drawer) and their Alpha pills are gone.

## Why the utility row is a row of the header, not part of the nav cluster

First attempt nested it inside the right-hand nav cluster as a stacked column. Below `lg` the cluster's intrinsic width won the flex negotiation and squeezed the brand into a three-line column, with `GAMES ACCOUNT` overlapping the eyebrow text. As its own row it right-aligns under the nav on desktop and spans the width on mobile — and costs no row at all when an app passes no slot, so the SRD masthead (which shares `AppBar`) renders unchanged.

## Colour

The controls are paper-on-ink now, so `SignInControl` takes an `onDark` flag. Signed out, the Discord button drops `primary` for the same quiet outline — rust directly beneath "Buy the game" gave the bar two identical CTAs and no hierarchy.

## Verification

- `bun run check:all` green (lint, format, typecheck, 1505 itun tests, validate, knip, audit, tokens, styling).
- Measured in headless Chrome at 1440 and 390, signed-in and signed-out.
- SRD masthead re-measured through the shared component: 95px desktop / 75px mobile, unchanged.
- Tests: updated the drawer test that asserted on the Encounter link; added one case that Encounter is absent from both surfaces and one that the utility row renders inside `<header>` and outside `<nav>`.

https://claude.ai/code/session_01E3Uy8Pgtx2AcJJgHD9trbQ

### DIFF
--- a/apps/itun/src/components/account/AccountStrip.tsx
+++ b/apps/itun/src/components/account/AccountStrip.tsx
@@ -1,4 +1,5 @@
 import { Link } from '@tanstack/react-router'
+import { FOCUS_RING } from 'component-lib'
 
 import { useConnection } from '../../lib/connection/connectionContext'
 import { isConvexConfigured } from '../../lib/connection/convexClient'
@@ -29,8 +30,10 @@ import { SignInControl } from './SignInControl'
  * users see exactly the chrome they saw before.
  */
 
-const ACCOUNT_LINK =
-  'font-cond text-caption font-semibold uppercase tracking-caps-wide text-paper/60 no-underline transition-colors hover:text-paper'
+// The masthead nav-link treatment one size down — same paper-on-ink ramp and
+// the same rust focus ring its siblings in `AppBar` use, so the utility row
+// reads as part of the bar rather than as a transplanted light-mode strip.
+const ACCOUNT_LINK = `font-cond text-caption font-semibold uppercase tracking-caps-wide text-paper/60 no-underline transition-colors hover:text-paper ${FOCUS_RING}`
 
 export function AccountStrip() {
   const { mode } = useConnection()
--- a/apps/itun/src/components/account/AccountStrip.tsx
+++ b/apps/itun/src/components/account/AccountStrip.tsx
@@ -5,16 +5,22 @@ import { isConvexConfigured } from '../../lib/connection/convexClient'
 import { SignInControl } from './SignInControl'
 
 /**
- * A thin strip above the brand header carrying sign-in and a link to the
- * account page.
+ * The account cluster — sign-in plus links to Games and the account page.
  *
- * ## Why not a nav item in `AppBar`
+ * ## Where it renders
+ *
+ * It is handed to `AppHeader` as its `utilityRow` slot, so it sits on its own
+ * row *inside* the dark masthead, beneath the nav. It used to be a light strip
+ * stacked above the header, which read as a second bar competing with the
+ * brand. Living on the masthead means these are paper-on-ink controls, not the
+ * ink-on-paper ones the old strip used.
+ *
+ * ## Why not nav items in `AppBar`
  *
  * `AppBar` lives in `component-lib` and is shared with `apps/srd`, which has no
- * accounts and never will. Threading an auth slot through it would push an
- * ITUN-only concept into a package whose contract is to stay backend-agnostic,
- * and would touch a component every page of both apps renders. A local strip
- * keeps the blast radius inside ITUN.
+ * accounts and never will. Threading auth through `navItems` would push an
+ * ITUN-only concept into a package whose contract is to stay backend-agnostic.
+ * A generic slot keeps the blast radius inside ITUN.
  *
  * ## It renders nothing in a Solo build
  *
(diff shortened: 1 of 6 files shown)


alxjrvs (commented):
**Scope note:** local `main` was stale — `origin/main` is at `ab9ca745`, so the real diff is the single commit `652b0e9c` (move account controls into the masthead) plus one uncommitted hunk in `AccountStrip.tsx` that adds `FOCUS_RING`.

**Verification run:** `bun --filter itun test` (1505 pass, 0 fail), `bun run typecheck` (clean across all workspaces), `biome check` on the changed files (clean). The structural refactor of `AppBar` into two rows is behavior-preserving for `apps/srd` — `SiteHeader` passes no `utilityRow`, and `flex flex-col gap-2` with one child renders identically to the old single-row header. `DARK_BUTTON`'s overrides do win the `cn()` merge: `border-chrome` is registered as a `border-w` group in `cn.ts`, so `border-paper/40` recolors without stripping the width.

**Findings (2):**

- `packages/component-lib/src/components/shared/AppHeader.tsx:29` — Removing Encounter from `DESKTOP_NAV`/`DRAWER_NAV` orphans the still-registered `/encounter` route: no `to=`/`href=` reference to it survives anywhere in `apps/itun/src`, and the Mediator sheet cited as its replacement is a separate Convex-backed surface (`api.mediator.npcs`) that never links to or renders `EncounterScreen`. Solo builds have no Mediator surface at all, so a Solo user's saved `encounterNpcs` become reachable only by hand-typing the URL.
- `apps/itun/src/components/account/AccountStrip.tsx:36` — The uncommitted hunk adds `FOCUS_RING` (25%-alpha rust) to links that now sit on `bg-ink-deep`; `interaction.ts:38` documents that this exact ring "simply disappears" on dark grounds. Composited contrast is ~1.27:1, versus the 3:1 WCAG 2.4.11 requires, so keyboard focus on Games/Account is imperceptible. `AppBar`'s `NAV_LINK` has the same defect already, so this propagates rather than introduces it — but `FOCUS_RING_ON_TONE` is no fix either (`ring-ink` on ink-deep), so the masthead genuinely lacks a correct rung.

alxjrvs (commented):
Reviewed `main...HEAD` (3 commits: retire Workspaces → Games/Shelf, plus the masthead account row). `bun run typecheck:itun` and `bun --filter itun test` (1505 pass / 0 fail) both green, so everything below is behavioural rather than compile-time.

**Findings (most severe first)**

1. `packages/component-lib/src/components/shared/AppHeader.tsx:29` — Encounter was removed from both `DESKTOP_NAV` and `DRAWER_NAV`, and nothing else links to `/encounter`. The docblock/test claim it "lives on the Mediator sheet"; `MediatorScreen.tsx` has no such link. The whole encounter tray is now reachable only by typing the URL.
2. `apps/itun/src/components/container/ContainerSwitcher.tsx:73` — the switcher only offers Shelf + live Games. Migration v13 maps every non-Default workspace to `gameId: <workspace id>`, so a legacy user who signs in has entities in containers the switcher can't list; `Roster.tsx:185` filters them out and there is no in-app way back to them. (`MoveToContainerControl` has the "Unknown game" escape hatch; this one doesn't.) Also renders a blank selection when `activeContainer` isn't in `games`.
3. `apps/itun/src/stores/activeContainerStore.ts:35` — the localStorage key changed from `itun.activeWorkspaceId` to `itun.activeContainer` with no migration, so every existing user resets to Shelf and new builds get stamped into a different container from their existing ones.
4. `apps/itun/src/lib/starterSet/seedStarterSet.ts:38` — `isStarterSetSeeded` requires *every* starter pilot, so deleting one makes "Load Starter Set" permanent chrome and re-clicking resurrects the deleted rows — the exact behaviour the old workspace-existence guard was written to prevent.
5. `apps/itun/src/lib/starterSet/seedStarterSet.ts:55` — seeding bypasses `requireWritableBackend()` and `mirrorEntityWrite`: Connected-mode seeds are local-only (never reach Convex) and Disconnected-mode seeds write despite ADR-030's read-only rule.
6. `apps/itun/src/stores/encounterStore.ts:49` — `listForContainer` now filters through `containerOf`, but migration 13 only touched pilots/mechs/crawlers, so pre-existing encounter NPCs with a custom `workspaceId` resolve to a phantom Game and vanish from the tray when connected.
7. `apps/itun/src/lib/export/mergeImport.ts:125` — `parseImportBundle` → `assignContainers` maps legacy `workspaceId` → `gameId`, then `mergeImport` overwrites it all with `gameId: null`. `legacyContainers.ts` (and its tests) do no observable work, and its stated "restores to the same place the migration would have" invariant no longer holds.
8. `apps/itun/src/components/sheet/SheetMech.tsx:158` (and `SheetPilot.tsx:147`) — these pass `activeContainer={containerOf(entity)}` unconditionally, so `DashboardChooser` scopes by container even in Solo, contradicting the "Solo renders unfiltered" rule the Roster (`:266`) and `apps/itun/CLAUDE.md` both state.

Non-blocking observations: the `AppBar` flex-col restructure is behaviourally neutral for `apps/srd` (no `utilityRow` ⇒ single child, no extra gap); `cockpitPrefs` previously stored on Workspace records are silently dropped rather than migrated to the new localStorage store (documented as a deliberate narrowing, but existing users lose their dial layout).

PR #655, opened 2026-07-29
feat(itun): a Game invite scheme that can be managed, carried, and gated

alxjrvs:
## Why

The invite scheme was three mutations and one button: mint a code, redeem it as a bare Player, revoke it. Except **nothing in the app ever called revoke**, and with no `invites.list` a minted code could not be seen again, audited, or killed for its whole 14-day life. A code pasted into the wrong Discord channel was both undetectable and unstoppable — and joining hands over read access to every crewmate's sheet (ADR-030 §5).

## What landed

**Invites are manageable.** `invites.list` derives status from the row and the clock rather than a stored flag that would need a cron to stay true. `revoke` becomes a soft delete, so the new `inviteRedemptions` trail keeps resolving — "used by Sam" instead of "2 uses left". Revoking closes a code; it never evicts anyone already seated.

**Invites carry what the Organizer already decided** — a seat (`role: 'mediator'` hands over the GM chair) and grants (unclaimed entities handed over on join). This is `assign` *scheduled*, not self-claim: the Organizer names the recipient in advance, so the Change Log records **them** as `actorId`. A grant gone stale is skipped and the join still succeeds — arriving without the promised pilot is a notice, whereas failing over a stale pointer would strand someone outside a game they were genuinely invited to.

**Invites can be knocks.** `requiresApproval` means the code identifies the game and grants nothing until the Organizer approves. A knock is not a membership and **consumes no use**, so it cannot burn a code before anyone joins. Both doors seat people through one shared `seat()` path, so role and grants cannot drift apart.

**A Game is a place you can be sent.** `/join/$code` describes the invitation *before* asking anyone to sign in — signed-out is the expected state for a link you hand a stranger, and being asked to authenticate only to learn the code expired is the worst version of that page. The lobby lists each Game as an `EntityRow` like the pilots and mechs do — a fourth **blue** ontology whose badges say which crawler the crew rides and how much is built.

## Notes for review

- **Rebuilt on top of #656**, which landed while this was in flight. That PR had already made `/games/$gameId` the place per-game work happens, so the invite panel went *there* rather than into the duplicate detail screen this branch originally had — that screen and its route are gone. `GameScreen` now reads `games.get` instead of `listMine`-and-find (one game, not all, and it distinguishes loading from not-a-member).
- **The ADR amendment is reconciled with #656's self-claim amendment**, not written over it. Self-claim made an unclaimed entity an offer to the crew; `assign` survived as the power to place an entity with a *particular* person. An invite grant is that power, scheduled.
- `--color-sheet-game-deep` (#1e5364) is a **literal, not `var(--color-tl-5)`** — that token holds the same value but means *tech level*, and aliasing it would tie the Game ontology to tech-level semantics. It was picked for contrast, not by eye: `EntityRow`'s empty variant paints `tone.rail` as **text**, and #1e5364 measures 7.7:1 on the game wash, mid-family between mech (9.1) and crawler (7.5).
- `summarize()` now counts pilots and mechs, and **Convex has no count API** — counting means collecting. That is stated in the code: fine at a table's scale, and if it ever bites the fix is a denormalised counter, not a cleverer query.
- **Deliberately not done:** rate limiting on `redeem`/`preview` (40 bits needs ~10¹² online attempts; a limiter would be theatre and real friction for a table mistyping a code), and any change to the code generator.

## Verification

`bun run check:all` green on top of current main: **4335 tests** (itun 1587, reference 1003, srd 984, component-lib 681, discord-bot 80), plus lint, format, typecheck, knip, validate, audit, design-tokens, styling-ownership.

The design doc (`docs/design/game-invites-and-membership-plan.md`) is kept as the reasoning behind the shape; its §1 describes the state this replaced.

https://claude.ai/code/session_01DxsNxRvv4nDHZuZ9SqyNYG

### DIFF
--- a/apps/itun/convex/__tests__/invites.test.ts
+++ b/apps/itun/convex/__tests__/invites.test.ts
@@ -1,0 +1,425 @@
+import { describe, expect, test } from 'bun:test'
+
+import { api } from '../_generated/api'
+import type { Id } from '../_generated/dataModel'
+import { testConvex } from './harness'
+
+/**
+ * The invite scheme past a bare code (ADR-030 §2–4).
+ *
+ * The three things worth defending here are the ones a UI cannot enforce:
+ * an invite grants exactly the seat it was minted for, a grant that has gone
+ * stale must not strand the joiner outside the Game, and a knock must not be
+ * a membership until someone says so.
+ */
+
+type Ctx = ReturnType<typeof testConvex>
+
+/**
+ * The first row, or a failure that says so.
+ *
+ * Destructuring and then asserting non-null hides the interesting case: when a
+ * query unexpectedly returns nothing, the test should fail on "expected a row"
+ * rather than on a property access against `undefined` several lines later.
+ */
+function firstRow<T>(rows: T[]): T {
+  const row = rows[0]
+  if (row === undefined) throw new Error('expected at least one row, got none')
+  return row
+}
+
+async function makeUser(t: Ctx, name: string) {
+  const userId = await t.run(async (ctx) => {
+    return await ctx.db.insert('users', { name, displayName: name })
+  })
+  return { userId, as: t.withIdentity({ subject: userId }) }
+}
+
+async function seedGame(t: Ctx) {
+  const organizer = await makeUser(t, 'Organizer')
+  const gameId = await organizer.as.mutation(api.games.create, { name: 'Tenacity' })
+  return { organizer, gameId }
+}
+
+async function seedPilot(t: Ctx, gameId: Id<'games'>, ownerId: Id<'users'> | null) {
+  return await t.run(async (ctx) =>
+    ctx.db.insert('pilots', {
+      gameId,
+      ownerId,
+      body: { callsign: 'Roach-Boy' },
+      updatedAt: Date.now(),
+    })
+  )
+}
+
+describe('invite management', () => {
+  test('list reports derived status and who redeemed', async () => {
+    const t = testConvex()
+    const { organizer, gameId } = await seedGame(t)
+    const joiner = await makeUser(t, 'Sam')
+
+    const code = await organizer.as.mutation(api.invites.create, {
+      gameId,
+      label: 'for Sam',
+      usesRemaining: 1,
+    })
+    await joiner.as.mutation(api.invites.redeem, { code })
+
+    const invite = firstRow(await organizer.as.query(api.invites.list, { gameId }))
+    expect(invite.label).toBe('for Sam')
+    // One use, spent — so the code is exhausted rather than merely used.
+    expect(invite.status).toBe('exhausted')
+    expect(invite.redeemers).toEqual(['Sam'])
+  })
+
+  test('revoke is a soft delete that keeps the audit trail', async () => {
+    const t = testConvex()
+    const { organizer, gameId } = await seedGame(t)
+    const joiner = await makeUser(t, 'Sam')
+
+    const code = await organizer.as.mutation(api.invites.create, { gameId })
+    await joiner.as.mutation(api.invites.redeem, { code })
+
+    const before = firstRow(await organizer.as.query(api.invites.list, { gameId }))
+    await organizer.as.mutation(api.invites.revoke, { inviteId: before._id })
+
+    const [after] = await organizer.as.query(api.invites.list, { gameId })
+    expect(after?.status).toBe('revoked')
+    // The row survived, so "who joined via which code" is still answerable.
+    expect(after?.redeemers).toEqual(['Sam'])
+  })
+
+  test('a revoked code is refused, but does not evict whoever already joined', async () => {
+    const t = testConvex()
+    const { organizer, gameId } = await seedGame(t)
+    const early = await makeUser(t, 'Early')
+    const late = await makeUser(t, 'Late')
+
+    const code = await organizer.as.mutation(api.invites.create, { gameId })
+    await early.as.mutation(api.invites.redeem, { code })
+
+    const invite = firstRow(await organizer.as.query(api.invites.list, { gameId }))
+    await organizer.as.mutation(api.invites.revoke, { inviteId: invite._id })
+
+    await expect(late.as.mutation(api.invites.redeem, { code })).rejects.toThrow(/revoked/i)
+    // Early is still seated — revocation closes the door, it does not sweep the room.
+    const games = await early.as.query(api.games.listMine, {})
+    expect(games).toHaveLength(1)
+  })
+
+  test('a non-organizer can neither list nor revoke', async () => {
+    const t = testConvex()
+    const { organizer, gameId } = await seedGame(t)
+    const player = await makeUser(t, 'Player')
+    const code = await organizer.as.mutation(api.invites.create, { gameId })
+    await player.as.mutation(api.invites.redeem, { code })
+
+    await expect(player.as.query(api.invites.list, { gameId })).rejects.toThrow()
+
+    const invite = firstRow(await organizer.as.query(api.invites.list, { gameId }))
+    await expect(player.as.mutation(api.invites.revoke, { inviteId: invite._id })).rejects.toThrow()
+  })
+})
+
+describe('invites carry a seat and a hand-out', () => {
+  test('a mediator invite seats the joiner as Mediator', async () => {
+    const t = testConvex()
+    const { organizer, gameId } = await seedGame(t)
+    const gm = await makeUser(t, 'Vex')
+
+    const code = await organizer.as.mutation(api.invites.create, { gameId, role: 'mediator' })
+    await gm.as.mutation(api.invites.redeem, { code })
+
+    const [game] = await gm.as.query(api.games.listMine, {})
+    expect(game?.mediator).toBe(true)
+  })
+
+  test('a mediator invite defaults to single use', async () => {
+    const t = testConvex()
+    const { organizer, gameId } = await seedGame(t)
+    const first = await makeUser(t, 'First')
+    const second = await makeUser(t, 'Second')
+
+    const code = await organizer.as.mutation(api.invites.create, { gameId, role: 'mediator' })
+    await first.as.mutation(api.invites.redeem, { code })
+
+    // The schema permits several Mediators; the UI is built for one, so an
+    // unlimited mediator code would quietly seat a second.
+    await expect(second.as.mutation(api.invites.redeem, { code })).rejects.toThrow(/used up/i)
+  })
+
+  test('a grant assigns the pilot and logs the ORGANIZER as the actor', async () => {
+    const t = testConvex()
+    const { organizer, gameId } = await seedGame(t)
+    const joiner = await makeUser(t, 'Sam')
+    const pilotId = await seedPilot(t, gameId, null)
+
+    const code = await organizer.as.mutation(api.invites.create, {
+      gameId,
+      grants: [{ table: 'pilots', entityId: pilotId }],
+    })
+    const result = await joiner.as.mutation(api.invites.redeem, { code })
+
+    expect(result).toMatchObject({ kind: 'joined', granted: 1 })
+
+    const pilot = await t.run(async (ctx) => ctx.db.get(pilotId))
+    expect(pilot?.ownerId).toBe(joiner.userId)
+
+    // ADR-030 §4: ownership is ASSIGNED, never self-claimed. The Organizer made
+    // this call when they minted the invite, so the log must name them — not the
+    // person who happened to walk through the door.
+    const entry = await t.run(async (ctx) =>
+      (await ctx.db.query('changeLog').collect()).find(
+        (row) => row.entityId === pilotId && row.field === 'ownerId'
+      )
+    )
+    expect(entry?.actorId).toBe(organizer.userId)
+    expect(entry?.after).toBe(joiner.userId)
+  })
+
+  test('a stale grant is skipped and the joiner still gets in', async () => {
+    const t = testConvex()
+    const { organizer, gameId } = await seedGame(t)
+    const claimer = await makeUser(t, 'Claimer')
+    const joiner = await makeUser(t, 'Sam')
+    const pilotId = await seedPilot(t, gameId, null)
+
+    const code = await organizer.as.mutation(api.invites.create, {
+      gameId,
+      grants: [{ table: 'pilots', entityId: pilotId }],
+      usesRemaining: 5,
+    })
+
+    // Somebody else takes the pilot between minting and redeeming.
+    await t.run(async (ctx) => ctx.db.patch(pilotId, { ownerId: claimer.userId }))
+
+    const result = await joiner.as.mutation(api.invites.redeem, { code })
+
+    // Arriving without the promised pilot is a notice. Failing the join over a
+    // stale pointer would strand someone outside a Game they were invited to.
+    expect(result).toMatchObject({ kind: 'joined', granted: 0 })
+    const games = await joiner.as.query(api.games.listMine, {})
+    expect(games).toHaveLength(1)
+    const pilot = await t.run(async (ctx) => ctx.db.get(pilotId))
+    expect(pilot?.ownerId).toBe(claimer.userId)
+  })
+
+  test('redeeming twice is idempotent and consumes no second use', async () => {
+    const t = testConvex()
+    const { organizer, gameId } = await seedGame(t)
+    const joiner = await makeUser(t, 'Sam')
+    const code = await organizer.as.mutation(api.invites.create, { gameId, usesRemaining: 2 })
+
+    await joiner.as.mutation(api.invites.redeem, { code })
+    const second = await joiner.as.mutation(api.invites.redeem, { code })
+
+    expect(second).toMatchObject({ kind: 'already' })
+    const invite = firstRow(await organizer.as.query(api.invites.list, { gameId }))
+    expect(invite.usesRemaining).toBe(1)
+  })
+})
+
+describe('preview', () => {
+  test('describes the invite without spending it or leaking the crew', async () => {
+    const t = testConvex()
+    const { organizer, gameId } = await seedGame(t)
+    const stranger = await makeUser(t, 'Stranger')
+    const code = await organizer.as.mutation(api.invites.create, { gameId, role: 'mediator' })
+
+    const preview = await stranger.as.query(api.invites.preview, { code })
+    expect(preview).toMatchObject({
+      gameName: 'Tenacity',
+      invitedBy: 'Organizer',
+      role: 'mediator',
+      status: 'active',
+    })
+    // Looking is not joining.
+    expect(await stranger.as.query(api.games.listMine, {})).toHaveLength(0)
+    // Nothing about who is already at the table.
+    expect(JSON.stringify(preview)).not.toContain('memberCount')
+  })
+
+  test('an unknown code is null rather than an error', async () => {
+    const t = testConvex()
+    const stranger = await makeUser(t, 'Stranger')
+    expect(await stranger.as.query(api.invites.preview, { code: 'ZZZZZZZZ' })).toBeNull()
+  })
+
+  test('a revoked code previews as revoked, so the landing page can say so', async () => {
+    const t = testConvex()
+    const { organizer, gameId } = await seedGame(t)
+    const stranger = await makeUser(t, 'Stranger')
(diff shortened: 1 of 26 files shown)


alxjrvs (commented):
**Scope:** `origin/main...HEAD` is a single docs commit — `docs/design/game-invites-and-membership-plan.md` (+433 lines, new file). The large workspace-retirement diff in `main...HEAD` is stale local `main`; that work is already merged upstream as #652. Working tree is clean, so the plan doc is the whole review scope. I reviewed it as a spec: every factual claim checked against `apps/itun/convex/*`, `component-lib`'s `Card`, and ADR-030.

**Verified-correct claims** (worth stating, since much of the doc is assertions about existing code): the invite generator's entropy/bias/alphabet analysis, `revoke` having no app caller, no `invites.list`, no `games.get`, the `memberships` insert shape, the six unclaimed Starter Set pilots/mechs in `templates.ts`, the `bot.ts` function set, the `changeLog` `source: 'ownership'` / `field: 'ownerId'` write, ADR-030 §1/§3/§5/Consequences citations, `EntityRow`'s three tones and `--color-sheet-*`, and that `Card` really does carry `headerContent` / `label` / `stats` / `controls.cardClick` / `extent`.

**Findings (11):**

- `docs/design/game-invites-and-membership-plan.md:231` — Shared seat path re-enters the knock step, so `joinRequests.decide(approve)` returns `pending` again and never seats the member.
- `docs/design/game-invites-and-membership-plan.md:96` — `createdAt`/`role`/`requiresApproval` declared required on the already-deployed `invites` table; Convex deploy fails on existing rows, no backfill step in §8.
- `docs/design/game-invites-and-membership-plan.md:376` — Slice 3 makes revoke soft but the `revokedAt` check on redeem is slice 4, so revoked codes keep working between slices.
- `docs/design/game-invites-and-membership-plan.md:235` — Grants assign ownership with `actorId = Organizer`, bypassing `requireOwnershipAssigner`'s Mediator-first rule (ADR-030 §3's deliberate bend).
- `docs/design/game-invites-and-membership-plan.md:179` — §4.1 marks `invites.preview` "Any signed-in"; §4.2/§5.2 require it to answer signed-out visitors.
- `docs/design/game-invites-and-membership-plan.md:332` — `footMeta` (organizer · joined date) never renders because the grid card uses `extent: 'head'`, which suppresses the footer band.
- `docs/design/game-invites-and-membership-plan.md:229` — Approval re-runs step 1's expiry/uses validation, so pending knocks become un-approvable once the invite expires or another approval exhausts it.
- `docs/design/game-invites-and-membership-plan.md:391` — Nothing checks `state: 'declined'` in step 3, contradicting slice 5's "decline is terminal" test.
- `docs/design/game-invites-and-membership-plan.md:404` — Draft amendment says "§2 describes an invite code as sufficient to join"; ADR-030 §2 is Containers and states no such thing (the relevant authority is in §3).
- `docs/design/game-invites-and-membership-plan.md:214` — `preview`'s syntax check is uppercase-only (`SNAPSHOT_ID_PATTERN`); a lowercased `/join/<code>` link reads as invalid while the typed form works, and no normalization is specified.
- `docs/design/game-invites-and-membership-plan.md:127` — `ownableTable` is module-private in `convex/ownership.ts:31`, so `schema.ts` cannot import it as written.

alxjrvs (commented):
## Review scope

Local `main` was stale, so `git diff main...HEAD` looked like a 68-file monster. The real scope against `origin/main` is **two commits adding one new file**: `docs/design/game-invites-and-membership-plan.md` (479 lines, design-only, nothing implemented). No code changed, no working-tree changes.

I reviewed it as a spec — verifying every claim it makes about the codebase, and hunting for places where an implementer following it would build something broken.

**The plan's descriptive half is accurate.** Every factual claim I checked holds: `invites.ts` shape and defaults, `revoke` having no app caller (only `permissions.test.ts:440`), `templates.ts` seeding starter entities `ownerId: null`, `routes/games.tsx` being one leaf, `bot.ts`'s three mutations, the entire `EntityRow` prop/`TONE`/`EMPTY_GLYPH` analysis, `BadgeTone` being exactly `pilot|mech|crawler|ok|warn|bad`, `Roster.tsx`'s `joinMeta(' · ')`, and `--color-wk-accent: rgb(125, 206, 235)` down to the comment text. The 40-bit / no-modulo-bias argument is correct (`b % 32`, 256/32 = 8).

The problems are all in the prescriptive half.

## Findings (11)

- `docs/design/game-invites-and-membership-plan.md:96` — §3.1 adds `createdAt`, `role`, `requiresApproval` as **required** fields to the live `invites` table; `schemaValidation` defaults on, so PR 3's `convex deploy` is rejected by every pre-existing row.
- `:235` — Grant invites bypass `requireOwnershipAssigner`. It permits an Organizer to assign ownership *only when the Game has no Mediator*, but `invites.create` is plain Organizer-only, so a grant code hands out a pilot the minter cannot hand out directly.
- `:179` — §4.1 lists `invites.preview` as "Any signed-in"; §4.2 and §5.2 require it to answer signed-out. `/join/$code`'s whole purpose fails if the table is followed.
- `:182` — A knock outlives its code: the shared path re-validates not-expired/not-revoked, so a day-13 knock approved on day 15 throws and the request is stuck `pending` with no other door.
- `:219` — The new discriminated `redeem` has no failure arm, but `GamesScreen.tsx` shows errors via `.catch(err => setJoinError(err.message))`. Return failures instead of throwing and a dead code becomes a silent no-op.
- `:231` — §8 asserts "decline is terminal"; §4.3 step 3 checks only membership, so a declined user re-redeems and re-knocks forever. §3.3's indexes don't even support the dedupe (`by_game_user` missing).
- `:213` — §4.2's "uniform 'not valid' for every failure mode" directly contradicts §5.2 step 2's "say which" and its own returned `status`.
- `:230` — §4.3 step 2 returns `already` before step 5, so a grant invite sent to someone already seated transfers nothing, logs nothing, and reports success — and that's the plan's own flagship Starter Set flow.
- `:442` — The drafted ADR amendment opens "§2 describes an invite code as sufficient to join". ADR-030 §2 says no such thing (its only invite mention is "a shelf has no ... invites"). The mis-citation is inherited from a wrong header comment in `convex/invites.ts`.
- `:330` — `BADGE_TONES` resolves `--color-pilot`/`-mech`/`-crawler`, not `--color-sheet-*`. The plan mints only `--color-sheet-game`, so a `game` badge has no backing token. The four-point change is really five.
- `:127` — `ownableTable` is a private `const` in `ownership.ts` (only `OWNABLE`/`OwnableTable` are exported), and importing it into `schema.ts` would cycle through `_generated/server`.

The top two are the ones that would actually cost a day: #1 wedges the PR 3 deploy, and #2 is an authorization hole that the plan's own §2 constraints list walks right past by citing ADR-030 §4 without §3.

alxjrvs (commented):
## Review complete

**Scope:** the PR against `origin/main` is one new file — `docs/design/game-invites-and-membership-plan.md` (543 lines, no code). The large `main...HEAD` diff was a stale-ref artifact: `origin/main` is already at `ab9ca745`, so the Workspaces→Games retirement is *not* under review. Working tree is clean.

Since the doc specifies concrete Convex schema, mutation logic, and `component-lib` changes, I reviewed it as executable spec and checked every load-bearing claim against `apps/itun/convex/*` and `packages/component-lib/src`. Most of its factual claims hold up — I verified the code generator (8 chars ×5 bits = 40 bits, no modulo bias), `revoke` genuinely having no caller, `summarize()`'s exact field list, `ownableTable` excluding communal crawlers, `crawlers.body.name` being a real top-level field, the `body as Record<string, unknown> | null` precedent in `crew.ts`, and **all six contrast ratios in §6.2, which I recomputed and found accurate to the stated decimal** (7.68, 5.37, 12.07, 4.75).

### Findings

- `docs/design/game-invites-and-membership-plan.md:237` — §4.3 step 5's grant assignment throws `NotAuthorized` (the check reads the *caller*, a plain Player, not `actorId`) and rolls back the membership insert, so the whole join fails.
- `docs/design/game-invites-and-membership-plan.md:229` — step 1 validates `usesRemaining` before step 2's membership check, so `already` is unreachable for exactly the mediator/grant invites §4.4 defaults to single-use.
- `docs/design/game-invites-and-membership-plan.md:96` — `createdAt`/`role`/`requiresApproval` are non-optional on an existing table with no backfill; `convex deploy` rejects the push on the first legacy invite row.
- `docs/design/game-invites-and-membership-plan.md:179` — §4.1 makes `preview` signed-in-only, contradicting §4.2/§5.2 where the signed-out visitor is the primary path.
- `docs/design/game-invites-and-membership-plan.md:134` — `games.destroy` is never extended for `inviteRedemptions`/`joinRequests`, and `joinRequests` lacks the `by_game` index its loop requires.
- `docs/design/game-invites-and-membership-plan.md:232` — step 3 ignores an existing request's `state`, so a declined knocker re-knocks forever, contradicting §8's "decline is terminal".
- `docs/design/game-invites-and-membership-plan.md:279` — the route split moves `ProposalInbox` behind per-Game navigation with no row indicator, removing the only discovery surface for pending proposals.
- `docs/design/game-invites-and-membership-plan.md:214` — §4.2's "uniform not valid" contradicts §4.1's derived `status` and §5.2's "say which".
- `docs/design/game-invites-and-membership-plan.md:507` — the ADR amendment is filed against ADR-030 §2, which never states the invariant it claims to change (that's §3).
- `docs/design/game-invites-and-membership-plan.md:332` — `ReactNode | ReactNode[]` collapses to `ReactNode`; the real change is an untyped runtime `Array.isArray` branch the proposed test can't catch.
- `docs/design/game-invites-and-membership-plan.md:407` — `BADGE_TONES` fills key off `--color-pilot`-style base tokens, not the `--color-sheet-*` pair §6.2 adds, so a `bg-game` entry would resolve to nothing.
- `docs/design/game-invites-and-membership-plan.md:105` — `ownableTable` is a module-private `const` in `ownership.ts`; only the type is exported, so the §3.1 import won't compile.

alxjrvs (commented):
Reviewed the full `main...HEAD` diff (117 files, ~6.4k insertions) — the ADR-030 invite/container work across the Convex backend, ITUN stores/components, and component-lib.

1. `apps/itun/src/components/games/GameRoster.tsx:283` — `ensureLocal` adopts the raw server `body`, whose `gameId` is the body's own copy, not the row's column. Template rows carry `gameId: null`, so a claimed pre-gen lands on the local Shelf and its next edit mirrors back `gameId: null`, pulling it out of the Game.
2. `apps/itun/convex/templates.ts:83` — template-created entities get no `appId`, so `upsertByAppId` can't find them and inserts a **duplicate** row instead of updating the Game's.
3. `apps/itun/src/stores/entityStore.ts:301` — `mirrorEntityWrite` skips any crawler whose local `gameId` is null (which is every template crawler after adoption), so crew edits to the communal crawler are silently local-only, with no warning logged.
4. `apps/itun/convex/entities.ts:602` — moving a crawler between Shelf and Game never reaches the server: `patchCrawlerByAppId` never inserts, and it merges `gameId` into the *body* rather than the row's column, so the column and body can diverge.
5. `apps/itun/src/components/roster/Roster.tsx:329` — only the crawler column got the new crawler-gate handling; the Pilots/Mechs CTAs still open the wizard, so a player in a crawler-less Game creates a local-only ghost whose mirror is refused into a `console.warn`.
6. `apps/itun/src/components/games/GamesScreen.tsx:26` — the deleted crew-management block was never re-created on `GameScreen`; `games.rename`, `setMediator`, `transferOrganizer` and `destroy` now have zero callers, so an Organizer cannot promote an existing member to Mediator, rename, or end a Game.
7. `packages/component-lib/src/components/shared/AppHeader.tsx:29` — `/encounter` nav entries removed with no replacement link anywhere; the encounter tray (reworked in this same PR) is now reachable only by typing the URL.
8. `apps/itun/convex/invites.ts:153` — `list` sorts by `_id.localeCompare`, which carries no ordering guarantee; the comment itself names `_creationTime` as the right field. Newly minted codes don't sort to the top.
9. `apps/itun/src/components/container/ContainerSwitcher.tsx:68` — no "Unknown game" fallback option (unlike `MoveToContainerControl`), so a stale persisted container makes the select read "Shelf" while the Roster filters to a phantom Game and renders empty.
10. `apps/itun/convex/proposals.ts:174` — the new strict re-parse of the whole merged body permanently blocks proposals on any entity whose server body already carries a stray key written by the old unparsed `apply` path; no migration cleans them.
11. `apps/itun/convex/invites.ts:47` — `grants[].entityId` is unvalidated `v.string()`; `seat`'s `ctx.db.get` throws on a malformed id (bricking the code), and `table` is never checked against the id's real table.

Note: I did not run typecheck/tests (no `node_modules` verification performed in this worktree); findings are from reading the diff plus surrounding files.

alxjrvs (commented):
Reviewed PR #655 (`origin/main...HEAD`: the invite-scheme commit plus its test follow-up) — ~3,000 lines across the Convex backend (`invites.ts`, `schema.ts`, `games.ts`), the new Games/Join/Invite React surfaces, and `EntityRow`/`Badge`/theme in component-lib. I ran the new `convex/__tests__/invites.test.ts` (20 pass) and traced the auth, permission, and destroy paths through their existing implementations.

Findings, most severe first:

- `apps/itun/src/components/games/JoinScreen.tsx:168` — Signing in from `/join/$code` loses the invite. `SignInControl` calls `signIn('discord')` with no `redirectTo`, and `@convex-dev/auth`'s `redirectAbsoluteUrl` falls back to `siteUrl()` when it's undefined, so OAuth returns the user to the app root, not back to the invite. The file's own header claims the code survives the round-trip; nothing carries it.
- `apps/itun/src/components/games/GamesScreen.tsx:156` — Deleting the inline `GameRow` removed the only call sites for `games.setMediator`, `games.transferOrganizer` and `games.rename`, and nothing in the new `GameScreen`/`GameRoster`/`GameRow` replaces them. An Organizer can no longer promote an existing member to Mediator, stand one down, hand over Organizer, or rename a Game — and because `isTableRunner` only falls back to the Organizer while a Game has no Mediator, a mis-seated Mediator is unrecoverable.
- `apps/itun/convex/invites.ts:153` — `invites.list`'s "newest first" sort uses `_id.localeCompare`; Convex ids aren't creation-ordered, so a just-minted code lands in arbitrary position. The projection also drops `_creationTime`/`createdAt`, so the client can't fix it.
- `apps/itun/src/components/games/InvitePanel.tsx:59` — `createInvite`, `revoke` and `decideRequest` are all `void fn(...)` with no `.catch`. A rejected mutation (e.g. Organizer role lost in another tab) is an unhandled rejection with zero UI feedback; there's no global rejection handler in the app.
- `apps/itun/convex/games.ts:204` — `games.destroy` doesn't sweep the two game-scoped tables this PR adds (`inviteRedemptions`, `joinRequests`), orphaning rows with dangling `gameId`/`inviteId`. `joinRequests` has no `by_game` index, only `by_game_state`.
- `apps/itun/convex/invites.ts:258` — `seat()` trusts `grant.table` rather than the fetched doc; `'ownerId' in doc` is true for pilots *and* mechs, so a mismatched grant hands over the wrong entity and logs the wrong `entityType`. A malformed `entityId` makes `ctx.db.get` throw at redeem time — surfaced to the joiner, not the Organizer who mis-minted it. `create` validates grants not at all.
- `apps/itun/convex/invites.ts:349` — A declined join request is reset to `pending` on the next `redeem`, with no cooldown or block-list, so declining a stranger on a publicly-posted approval code is unenforceable; the only remedy is revoking the code for everyone.

alxjrvs (commented):
Reviewed `origin/main...HEAD` (2 commits: the invite scheme + its test follow-up), covering the Convex backend (`invites.ts`, `games.ts`, `schema.ts`, `ownership.ts`), the four new/rewritten ITUN Game surfaces, and the `EntityRow`/`Badge`/theme changes in `component-lib`. I ran the new convex test suite (20 pass) and wrote throwaway probes against the same harness to confirm four of the behaviours below, then deleted them (tree is clean).

Findings, most severe first:

- `apps/itun/convex/invites.ts:349` — An approval-gated code re-redeemed by someone whose request is already `approved` but who has left the Game returns `pending` while creating no pending request, so the Organizer sees nothing and the user is stuck at "waiting for the organizer" permanently. Only `declined` is revived. Confirmed in the harness.
- `apps/itun/src/components/games/GamesScreen.tsx:26` — The rewrite deleted the Organizer's rename / Make Mediator / Hand over controls without re-adding them anywhere; `api.games.rename`, `setMediator` and `transferOrganizer` now have zero callers. Since `invites.seat` is the only other writer of `mediator: true` and it fires only for a non-member, a Game can no longer gain a Mediator at all except via a mediator invite to a brand-new account — leaving `/mediator/$gameId` unreachable for a solo Organizer.
- `apps/itun/src/components/account/SignInControl.tsx:67` — `signIn('discord')` passes no `redirectTo`, so a signed-out visitor to `/join/$code` is returned to the app root after OAuth and the invite is lost. The plan's §5.2(3) explicitly requires carrying the code through the round-trip.
- `apps/itun/src/components/games/InvitePanel.tsx:136` — `createInvite` / `revoke` / `decide` are all `void promise` with no `.catch`. Approving a knock whose invite expired throws (their own server test covers that throw) and the UI silently does nothing.
- `apps/itun/convex/invites.ts:153` — "Newest first" sorts by `_id.localeCompare`, but Convex `_id` is opaque and unordered; the adjacent comment names `_creationTime`, which is never used. Passes only because convex-test issues sequential ids.
- `apps/itun/src/components/games/InvitePanel.tsx:167` — Nothing in the app ever renders a `/join/<code>` link, so the new route, `JoinScreen` and the unauthenticated `preview` query have no product entry point.
- `apps/itun/convex/invites.ts:331` — A Mediator invite redeemed by an existing member returns `already` and silently grants no seat upgrade; combined with the finding above, this is the workaround an Organizer would naturally reach for.
- `apps/itun/convex/games.ts:203` — `destroy` cleans `invites` but not the two tables added in the same diff (`inviteRedemptions`, `joinRequests`), leaving permanently unreachable orphan rows.
- `apps/itun/convex/invites.ts:252` — A grant's declared `table` is never checked against the row; a `pilots`-labelled grant pointing at a mech assigns it and writes `entityType: 'pilot'` into the Change Log. Confirmed in the harness.

alxjrvs (commented):
Reviewed the 3-commit invite-scheme branch (`origin/main...HEAD`): Convex `invites.ts`/`games.ts` backend, the new `InvitePanel`/`JoinScreen`/`GameRow` surfaces, the `/join/$code` route, and the `EntityRow`/`Badge` `game`-ontology widening. Working tree is clean, so the range diff was the whole scope.

Findings (5):

- `apps/itun/src/components/games/GamesScreen.tsx:158` — the GamesScreen rewrite deleted the only UI callers of `games.rename`, `games.setMediator`, and `games.transferOrganizer`; nothing on the new Game screen replaces them, so an Organizer can no longer rename a Game, promote/demote a Mediator, or hand over the Organizer flag — and since `ownership.leaveGame` refuses while `organizer` is true, they can never leave their own Game either.
- `apps/itun/convex/invites.ts:351` — `redeem` returns `{kind:'pending'}` unconditionally but only writes a joinRequest when the prior row is absent or `declined`; an approved-then-departed user gets a permanent "waiting for approval" with nothing in the Organizer's queue.
- `apps/itun/convex/invites.ts:155` — `invites.list` sorts by `_id.localeCompare`, which is not creation-ordered in Convex, so the "newest first" guarantee (the just-minted code) does not hold; `_creationTime` is named in the comment but unused.
- `apps/itun/src/components/games/InvitePanel.tsx:136` — Approve/Decline/Revoke/mint are all `void mutation(...)` with no `.catch`, and `pendingRequests` doesn't filter knocks against expired/revoked invites, so approving one throws `NotAuthorized` and the UI shows nothing at all.
- `apps/itun/convex/invites.ts:61` — `grants[].entityId` is an unvalidated `v.string()`; a `table`/`entityId` mismatch writes a changeLog row with the wrong `entityType`, and a malformed id makes `ctx.db.get` throw, failing the whole redeem against the documented "skip the stale grant, seat them anyway" contract.

Verified-clean areas worth noting: the `games.destroy` invite/redemption/joinRequest sweep matches the declared `by_game` indexes; permission gating on every new query/mutation (`requireOrganizer`/`requireMember`) is consistent; `preview` is deliberately unauthenticated and leaks only what the invite already tells its bearer; and the `EntityRow` `meta` array normalisation preserves single-node behaviour for all pre-existing callers.

PR #656, opened 2026-07-29
feat(itun): a Game's crew as a roster, and the rules for setting a table up

alxjrvs:
## Why

The Mediator surface shipped as a stack of bordered cards listing names and numbers: no way into a sheet, no way to launch anything, no way to make anything. A Game asks the same question the home Roster answers — *what have we got and what can I do with it* — so it now answers it in the same vocabulary.

Two shapes for one question is how an app stops feeling like one app.

## What

**`GameRoster`** — a Game's crew in the home Roster's shape: three ontology-toned `EntityRow` columns, a create CTA per column, sheet and Dashboard launches, an owner chip per row.

- `/games/:id` — the crew, for any member (new route)
- `/mediator/:id` — opens with the roster, keeps the private instruments below
- `/games` goes back to being a lobby. It used to render a proposal inbox and a Downtime panel *inside every row*, so a member of three campaigns met three stacked Downtime panels with nothing saying whose was whose.

**The table-setup rules** (ADR-030 §5a, new), enforced in Convex and mirrored — never re-decided — for the UI in `lib/games/gameRoster.ts`:

| Rule | Who |
| --- | --- |
| Raise / scrap a crawler | the table runner (Mediator, or Organizer while there is none) |
| Edit a crawler's fields | every member — that is what communal means |
| Add pilots & mechs to a Game | any player, **once the Game has a crawler**; the table runner is exempt |
| Create a character unclaimed | the table runner |
| Pick up an unclaimed character | any member, if nobody holds it |

A Game may hold **several** crawlers — losing one and rebuilding is ordinary play, not a state to reject.

**Self-claim amends ADR-030 §4**, which said players never self-claim. An unclaimed entity is an *offer to the crew* — that is why a Mediator pre-builds characters and a template seeds six. Requiring the Mediator to hand each one over added a round trip to every table's first ten minutes and bought no safety. The half that mattered is intact: claiming touches only what is **free**; what a crewmate holds must be released first.

Unclaimed reads as an **UNCLAIMED stamp seal** on the row, which opens a pick-up confirm. One control, because it is one fact — a muted "Unclaimed" chip beside a separate "Pick up" said the same thing twice and buried the invitation in the quieter half.

**Adoption** (`entityStore.adopt` / `forget`) caches a Game row into IndexedDB under its own id and drops it again. That is what makes a sheet or the Dashboard openable at all for a character built at somebody else's table. Local crawler edits now mirror up as a **field merge**, so "players may edit its fields" is true end to end rather than only locally.

## Two live defects found on the way

- **`crew.vitals` read `currentHp` / `currentSp`** where the Zod schemas define `currentHP` / `currentSP`. Every vital on the Mediator's crew strip came back null and rendered as an em-dash — indistinguishable from an undamaged crew, which is why it survived review. The existing test hand-wrote the same wrong key and agreed with the bug.
- **`proposals.apply` wrote the merged body without parsing it.** A proposal against a misspelled field added a key nothing reads: the player pressed Apply, no number moved, nothing said why. It now Zod-parses and refuses. (The Mediator surface was offering exactly those misspelled fields.)

Both are pinned by tests that build fixtures by **parsing the real schema** rather than hand-writing field names.

## Testing

- `convex/__tests__/tableSetup.test.ts` — 22 cases over the new rules, server-side
- `lib/games/__tests__/gameRoster.test.ts` — the UI's copy of those rules, pinned to the same table of cases
- `components/games/__tests__/GameRoster.connected.test.tsx` — the surface tells the truth about what you may do
- `bun run check:all` green (lint, format, typecheck, 1500+ tests, validate, knip, audit, tokens, styling)

## Known gaps, deliberately left

- **A refused mirror is only a console warning.** The surfaces avoid offering actions the server would refuse, so this is reachable mainly by going around them; a toast is the next step.
- **No read-only drill-in.** ADR-030 §5 permits reading a crewmate's sheet, but ITUN's sheet is a live *editing* surface — opening one would hand over an editor whose saves are refused. Rows you do not own show vitals and an owner, no link.
- **Adopted copies re-sync on the way in**, not while a sheet sits open.

https://claude.ai/code/session_016osniJWrGgSrfq4VaCbHoq

### DIFF
--- a/apps/itun/convex/__tests__/crew.test.ts
+++ b/apps/itun/convex/__tests__/crew.test.ts
@@ -1,5 +1,7 @@
 import { describe, expect, test } from 'bun:test'
 
+import { MechSchema } from '../../src/lib/schemas/mech'
+import { PilotSchema } from '../../src/lib/schemas/pilot'
 import { api } from '../_generated/api'
 import { testConvex } from './harness'
 
@@ -39,7 +41,7 @@ describe('vitals', () => {
       await ctx.db.insert('pilots', {
         gameId,
         ownerId: player.userId,
-        body: { callsign: 'Roach-Boy', currentHp: 7, currentAp: 3 },
+        body: { callsign: 'Roach-Boy', currentHP: 7, currentAP: 3 },
         updatedAt: 1,
       })
     })
@@ -47,9 +49,63 @@ describe('vitals', () => {
     const crew = await organizer.as.query(api.crew.vitals, { gameId })
     expect(crew.pilots).toHaveLength(1)
     expect(crew.pilots[0]?.name).toBe('Roach-Boy')
-    expect(crew.pilots[0]?.currentHp).toBe(7)
+    expect(crew.pilots[0]?.currentHP).toBe(7)
     // The chip needs a name, not just an id.
     expect(crew.pilots[0]?.ownerName).toBe('Beefcake')
+  })
+
+  test('reads the field names the Zod schema actually defines', async () => {
+    const t = testConvex()
+    const { organizer, player, gameId } = await seedGame(t)
+
+    // The body is built by PARSING a real record rather than hand-written, so
+    // the key names come from the schema instead of from this test's memory of
+    // them. That is the whole point: this query used to read `currentHp` while
+    // the schema defines `currentHP`, every vital came back null, and the crew
+    // strip rendered em-dashes that were indistinguishable from an undamaged
+    // crew. A hand-written fixture agreed with the bug and kept it green.
+    const pilot = PilotSchema.parse({
+      id: 'p1',
+      schemaVersion: 1,
+      name: 'Roach-Boy',
+      callsign: 'Roach-Boy',
+      classRef: 'salvager',
+      abilities: [],
+      equipment: [],
+      motto: '',
+      keepsake: '',
+      appearance: '',
+      conditions: [],
+      currentHP: 9,
+      currentAP: 4,
+      createdAt: '2026-01-01T00:00:00.000Z',
+      updatedAt: '2026-01-01T00:00:00.000Z',
+    })
+    const mech = MechSchema.parse({
+      id: 'm1',
+      schemaVersion: 1,
+      name: 'Iron Mongrel',
+      chassisRef: 'iron-mongrel',
+      systems: [],
+      modules: [],
+      cargoLots: [],
+      conditions: [],
+      currentSP: 12,
+      currentHeat: 2,
+      createdAt: '2026-01-01T00:00:00.000Z',
+      updatedAt: '2026-01-01T00:00:00.000Z',
+    })
+
+    await t.run(async (ctx) => {
+      await ctx.db.insert('pilots', { gameId, ownerId: player.userId, body: pilot, updatedAt: 1 })
+      await ctx.db.insert('mechs', { gameId, ownerId: player.userId, body: mech, updatedAt: 1 })
+    })
+
+    const crew = await organizer.as.query(api.crew.vitals, { gameId })
+    expect(crew.pilots[0]?.currentHP).toBe(9)
+    expect(crew.pilots[0]?.currentAP).toBe(4)
+    expect(crew.mechs[0]?.currentSP).toBe(12)
+    expect(crew.mechs[0]?.currentHeat).toBe(2)
   })
 
   test('an unclaimed pilot has a null owner name rather than a missing row', async () => {
@@ -82,7 +138,7 @@ describe('vitals', () => {
     const crew = await organizer.as.query(api.crew.vitals, { gameId })
     // Bodies are opaque, so the projection must not trust their shape. Zero
     // would render as "dead" on a vitals strip, which is worse than blank.
-    expect(crew.pilots[0]?.currentHp).toBeNull()
+    expect(crew.pilots[0]?.currentHP).toBeNull()
   })
 
   test('a non-member gets nothing', async () => {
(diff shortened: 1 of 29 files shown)


alxjrvs (commented):
Reviewed `origin/main...HEAD` (2 commits: the Game-roster feature + the offer-back/CTA fix). Typecheck passes; the diff is large but coherent. Seven findings, most-severe first:

- `apps/itun/convex/entities.ts:613` — **`patchCrawlerByAppId` never re-homes the row.** It merges the client patch into `existing.body`, so a `{gameId}` patch is written into the crawler *body* (CrawlerSchema has that field) while the row's `gameId` column stays put. Moving a crawler between Games corrupts server state; moving a shelf crawler into a Game silently no-ops (no server row → `.unique()` null → return), leaving a Game-scoped crawler that only exists locally. Contrast `upsertByAppId`, which was explicitly given a re-home branch.
- `apps/itun/src/stores/entityStore.ts:511` — **`transfer()` never mirrors.** Crawler `scrapPool`/`cargoLots` writes (`useCargo.ts:135/198`, `usePartnerCargo.ts:123`) — the exact contended fields `mirrorCrawlerWrite`'s field-merge doc claims to serialize — never reach Convex at all.
- `apps/itun/src/components/games/GameRoster.tsx:283` — **`ensureLocal` unconditionally overwrites the local copy** on the premise "every local write mirrors up immediately". That premise is false (see above), so opening a crawler from the Game roster silently discards unmirrored cargo/scrap edits.
- `apps/itun/src/components/games/MediatorScreen.tsx:44` — **`PROPOSABLE_FIELDS` isn't filtered by target type.** `currentSP`/`currentHeat` proposed against a pilot now hard-fails at apply time (strict `PilotSchema` + the new parse in `proposals.apply`), leaving a proposal the player can never apply.
- `apps/itun/src/components/sheet/SheetMech.tsx:140` (and `SheetPilot.tsx:122`) — **sheet crawler `+ Create` CTAs bypass the Roster reroute** this PR added, reproducing the "local crawler the Game refuses" failure the Roster comment calls "the worst of both".
- `apps/itun/src/components/games/GameRoster.tsx:349` — "This game has no Union Crawler yet" renders *above* the loading guard, so every viewer (Mediator included) sees a false statement while queries are in flight.
- `apps/itun/src/components/games/GameRoster.tsx:343` — the shared `error` line renders behind `ModalShell`, so a failed claim (e.g. lost race) gives the user no feedback at all.

Things I checked and cleared: the `currentHp`→`currentHP` casing fix is correct against both schemas; `proposals.apply`'s new strict parse does genuinely reject typo'd keys (all three schemas are `.strict()`); `assertMayAddToContainer` correctly covers both the `create` and `upsertByAppId` insert/re-home paths for pilots/mechs; `claim`'s race is properly serialized; the `games_.$gameId` route escape is right. The `upsertByAppId` update path did drop its unconditional membership check, but `leaveGame` nulls `ownerId` before deleting the membership, so `assertMayWrite` still closes it.

alxjrvs (commented):
Reviewed the full `main...HEAD` diff (99 files: Workspaces → Games/Shelf containers, the new `GameRoster`, crawler authorship rules on the Convex side, and the masthead/`EntityRow` changes in `component-lib`). Verification runs: `bun run typecheck` clean, `bun run knip` clean, `bun --filter itun test` green (1559 tests) after several re-runs.

Findings (most severe first):

- `apps/itun/src/stores/entityStore.ts:306` — Crawler container moves never reach the server. `mirrorEntityWrite`'s crawler branch sends `patchCrawlerByAppId`, which never inserts and never touches the row's `gameId` column; Shelf→Game silently no-ops (no server row to patch), Game→Shelf early-returns at line 300, Game→Game leaves the crawler in the old Game. The Game still reports "no crawler", which keeps every player blocked from adding crew. Pilots/mechs got exactly this re-home fix in `upsertByAppId`; crawlers have no equivalent.
- `packages/component-lib/src/components/shared/AppHeader.tsx:29` — `/encounter` is now orphaned. The nav item was removed from `DESKTOP_NAV` and `DRAWER_NAV`, and nothing anywhere in `apps/itun/src` links to it. The file comment and the new AppHeader test both claim it "lives on the Mediator sheet", but `MediatorScreen`'s `NpcTray` is the server-backed `api.mediator.npcs`, a different feature from the local `EncounterScreen`/`encounterStore` this PR just migrated to containers.
- `apps/itun/src/components/roster/Roster.tsx:229` — Re-loading the Starter Set overwrites played characters. `isStarterSetSeeded()` requires *every* starter pilot, so deleting one un-hides the "Load Starter Set" button; clicking it `put`s all seed records back, resetting the five the user has been playing. No confirmation, no error handling.
- `apps/itun/src/components/games/GameRoster.tsx:283` — `ensureLocal`'s unconditional `adopt` clobbers local writes that were never mirrored. Its premise ("a local copy is never legitimately ahead") is false for `entityStore.transfer`, which cargo moves use and which mirrors nothing — opening the crawler from the roster silently discards them.
- `apps/itun/src/components/container/MoveToContainerControl.tsx:66` — The control's `FieldError` can never fire for a refused move: `mirrorWrite` swallows `NotAuthorized` into a `console.warn`, so a server-rejected move renders as success and the local record diverges.
- `apps/itun/src/components/games/GameRoster.tsx:349` — The "no Union Crawler yet" banner renders outside the `loading` guard (`crawlerCount` defaults to 0 while `listing` is undefined), so a properly set-up Game briefly tells every player it isn't set up.

One non-finding worth noting: `bun --filter itun test` reported `1 fail` on two early runs but passed on five subsequent runs (direct and filtered, including with the junit reporter), and the filter TUI elides the failing test name. The logs are full of `An update to GameRoster inside a test was not wrapped in act(...)` warnings from the new `GameRoster.connected.test.tsx`, which is a plausible source of an order/timing-dependent flake worth watching in CI.

PR #657, opened 2026-07-29
feat(srd): render each entity's og:image from its Catalog tile

alxjrvs:
Entity pages all shared one static `/og-image.png`, so every link preview looked the same. They now get a per-entity preview that is a **screenshot of the entity's Catalog tile**.

**Coverage is exact: 1004 pages, 1004 images.** Chassis patterns are included as entities in their own right — a pattern has its own page, its own card view and its own provenance, so it gets its own tile at `…/item/{itemId}/pattern/{patternId}.og.png` rather than inheriting the chassis banner.

## Fidelity comes from sharing the component

Rather than matching props by hand, the schema index grid and the og:image render surface both render a new `CatalogTile`. The preview *cannot* drift from the Catalog view, because they are the same component.

- The tile lays out at **432px** — its measured width in the index's 3-up masonry, which is fluid (`flex-1`) and so has no intrinsic width to inherit.
- The frame is rasterised at exactly **1200 device pixels**, so a card that fits the canvas is written with **no resampling at all**; taller cards are cleanly downsampled by `contain`.
- The canvas is matted in `--color-wk-bg`, the surface the index actually puts behind its tiles, read from the live page rather than hardcoded.

**Verified against the live index tile**, not asserted: the tile lands at exactly the predicted position (dx=0) and differs by a mean of **1.02/255** across all bytes. The diff map is glyph *outlines* only — single-width and centred, with all fills and text interiors identical — i.e. antialiasing phase, because the live tile sits at a fractional `y` on the index page while the render surface starts at 0.

## A clipping bug the pattern work exposed

The catalog card is `overflow-visible`, so decorations render **outside** its box — a pattern's chassis name-tab sits 8px above it — and an element-scoped screenshot of the tile sliced them off. Captures now target a **padded frame** around the tile.

The padding is *guarded, not assumed*: every capture measures the card's real overhang and warns when it exceeds `CATALOG_TILE_PADDING`. That guard promptly found `roll-tables/callsign-table` overhanging by **17px** — 1px past the original 16 — so the padding is 24 with headroom. A silent clip is exactly the failure it exists to prevent.

## Why this is safe to run in the deploy this time

A per-entity og pass previously **broke every srd deploy** (#482): it rendered the *full* entity card for ~1.5k entities with its PNG cache in `node_modules`, so a cold builder re-rendered everything and blew the build limit. Catalog tiles are far cheaper — **~50s cold for 1004 entities, 36MB** palette-quantised (visually identical, ~3× smaller) — and three guards make that failure mode impossible:

1. **Astro always emits the DEFAULT og:image.** The script rewrites the meta only for pages whose PNG actually landed, so a skipped, capped, or failed run leaves a working default rather than an og:image that 404s.
2. **`OG_SCREENSHOTS_BUDGET_MS` caps the pass** (300s on Netlify); whatever doesn't render keeps the default.
3. **The script never fails the build** — errors are logged, not thrown.

Worst case is "some previews are generic", never a failed deploy. Generation is also opt-in (`OG_SCREENSHOTS=1` via `og:generate`), so a plain `build` — and therefore CI — renders nothing.

## Correctness details

- Pattern hashes key off the **pattern**, not its chassis, so two patterns of one chassis can't collide and editing one invalidates only its own tile.
- A pattern target that fails to resolve reports `MISSING` rather than silently falling back to the chassis tile, which would ship the wrong card.
- `SCRIPT_VERSION` invalidates the cache when the *rendering* changes, which entity data alone can't see.

## Verification

- `bun run check:all` passes (lint, format, typecheck, tests, validate, knip, audit, tokens, styling).
- Full cold run: 1004/1004 rendered, 1004 pages relinked, ~50s, no overhang warnings.
- Spot-checked an item tile, a tall chassis tile, a pattern tile, and the 17px overhang case.

https://claude.ai/code/session_01Jf5eQbe1AVATgR5qTsBE35


alxjrvs (commented):
Reviewed commit `9324f32b` (the srd og:image feature) plus the uncommitted pattern-support work in progress. Verified against the live component tree (`ReferenceEntityCard`, `CardImage`), the Astro config, `netlify.toml`, and by running knip (green), lint (green), and srd typecheck (green). I also disproved two hypotheses I'd formed: `bun install --no-optional` does *not* strip sharp's transitive platform binaries, and there are no duplicate entity slugs (850 targets, 850 unique keys).

Findings, most severe first:

- `apps/srd/scripts/og-screenshots.ts:318` — `__ogSetEntity` is called with only `(schemaId, itemId)` while the wait key from `entityKey()` includes `/pattern/<id>`; every pattern target times out 30s + 30s retry and exhausts the 300s budget almost immediately. (Uncommitted WIP.)
- `apps/srd/scripts/og-screenshots.ts:352` — capture happens mid-`transition-opacity duration-300` fade-in of `CardImage`; only two rAFs are awaited and `animations: 'disabled'` isn't set, so artwork is baked in at partial opacity.
- `apps/srd/scripts/og-screenshots.ts:385` — the 432px viewport disables all `sm:`/`md:` utilities, so `CardImage`'s `md:float-left md:mr-4` never applies and the OG tile is the mobile layout, not the desktop Catalog tile the PR claims 1:1 parity with.
- `apps/srd/scripts/og-screenshots.ts:361` — `fit: 'contain'` letterboxes a ~400-600px-tall portrait tile into a 1200x630 landscape canvas, downscaling it to ~45% with ~328px background bars each side; the `SCALE = OG_WIDTH/CATALOG_TILE_WIDTH` "no resampling" rationale never holds.
- `apps/srd/scripts/og-screenshots.ts:336` — a swallowed `img.decode()` rejection still writes the PNG, mirrors it to cache, and records it in the manifest, so a transient asset-host failure permanently ships an artwork-less preview with zero reported failures.
- `apps/srd/scripts/og-screenshots.ts:41` — top-level `import sharp` and `import ../src/lib/staticPaths` (which top-level-awaits `preload('all')`) run outside `run()`'s catch, so a failure there exits non-zero and fails the Netlify deploy, contradicting the "never fails the build" guarantee in `netlify.toml`.
- `apps/srd/scripts/og-screenshots.ts:362` — `palette: true` quantises painted artwork to 256 colours; the justifying comment ("flat brand colours, text and line art") contradicts the script's own note that catalog tiles lead with artwork.
- `apps/srd/scripts/og-screenshots.ts:411` — if `freshPage()` throws inside the recovery block, `page` stays closed and every remaining target in that worker fails instantly, which is the cascade the comment claims to prevent.
- `apps/srd/src/components/islands/CatalogTile.tsx:27` — name collides with `component-lib`'s already-exported `CatalogTile` (`packages/component-lib/src/index.ts:48`).

alxjrvs (commented):
## Review complete

Scope: commit `9324f32b` (`origin/main` is at `b72ea0c7`) **plus** the uncommitted working-tree changes that add chassis-pattern coverage (`og-screenshots.ts`, `CatalogTile.tsx`, `OgCardIsland.tsx`, `ogCard.ts`, `apps/srd/CLAUDE.md`).

### Findings

- `apps/srd/netlify.toml:29` — `og:install-browser` is `&&`-chained into the build command, so a flaky chromium download fails the entire deploy — exactly the "never fails the build" property the design claims (and the #482 regression class).
- `apps/srd/scripts/og-screenshots.ts:46` — the never-fail guarantee is only `run().catch(...)`. Static imports run first: `sharp` (whose native binary is an *optional* dependency, and this same netlify.toml sets `BUN_FLAGS = "--no-optional"`) and `../src/lib/staticPaths` → `lib/gameData.ts`, which top-level-awaits `preload('all')`. Either failing exits non-zero, before the `OG_SCREENSHOTS` opt-out is even reached. Playwright is correctly dynamic-imported; these are not.
- `apps/srd/scripts/og-screenshots.ts:341` — the island's `data-og-current="MISSING:<key>"` marker is never read. An unresolvable or wedged target burns 30s + a page reload + another 30s instead of failing fast; 4-5 of them exhaust the 300s Netlify budget and drop every remaining entity to the default banner. `OgCardIsland.tsx:92` claims such targets are "skipped" — nothing skips them.
- `apps/srd/scripts/og-screenshots.ts:219` — pattern tiles are cache-keyed on the pattern object only, but the card renders `data={chassis} pattern={pattern}`. Editing chassis name/artwork/tech level/source leaves every one of its pattern previews stale until `SCRIPT_VERSION` is bumped (which then invalidates all ~1000).
- `apps/srd/netlify.toml:71` — the 300s budget is sized off a *local* ~70s measurement while running 5 concurrent chromium pages on a 2-vCPU builder with a cache that (per the file's own #482 note) doesn't survive a cold builder. Likely outcome is a partial, nondeterministic-per-deploy set of previews.
- `apps/srd/scripts/og-screenshots.ts:398` — `palette: true` justified as "flat brand colours, text and line art", but line 388 of the same function notes tiles lead with licensed illustrated artwork; expect banding in the upscaled link card.

### Verified as *not* problems (checked, so they don't get re-litigated)

- `define:vars` + `is:global` works — the dev server emits `<html style="--tileWidth: 432px;">`, so `#og-card { width: var(--tileWidth) }` resolves.
- `--color-wk-bg` is `rgb(230, 240, 245)` (not `oklch`), so sharp's `background` parser accepts the value read from `getComputedStyle`.
- Slug derivation matches on both sides: `getItemStaticPaths` uses `getReferenceEntityData().slug`, which is `getEntitySlug()` — the same function `OgCardIsland` matches on. Patterns likewise agree on `visiblePatterns` + `nameToSlug`.
- `itemHtmlPath` matches Astro's output (`build.format` defaults to `directory`), including the nested `pattern/<id>/index.html`; the pattern route page uses the default og:image, so `pointPageAtOgImage`'s `html.includes(defaultUrl)` guard holds. The `split/join` only hits `og:image` and `twitter:image` — structured data doesn't carry the URL.
- `page.evaluate(() => document.fonts.ready)` does not throw on a non-serializable return (verified against real Playwright 1.62 — returns `{}`).
- `bun run knip` is green and `bun --filter srd typecheck` passes (0 errors) with the working-tree pattern changes applied.

alxjrvs (commented):
Reviewed the two OG commits (`5cc73c1a`, `495fc5b5`) against `origin/main` — 11 files, +895/−27 in `apps/srd`. Verified along the way that lint and `knip` pass, that `getReferenceEntityData().slug` really is `getEntitySlug` (so the island's entity lookup matches the route paths), that `gameData.ts`'s top-level `await preload('all')` makes the standalone script's `getItemStaticPaths()` work (it enumerates 1004 targets: 850 items + 154 patterns, so the netlify.toml "~850 entities / ~70s" measurement predates the pattern commit), and that neither the item nor the pattern page passes `ogImage`, so the post-build rewrite is well-founded.

Findings:

- `apps/srd/netlify.toml:29` — `og:install-browser` is `&&`-chained ahead of the build, so a chromium download failure aborts the whole deploy; contradicts the PR's "never fails the build" guarantee. Wrap in `|| true`.
- `apps/srd/scripts/og-screenshots.ts:48` — `sharp` is statically imported, outside the `run().catch()` net (unlike playwright, which is dynamic-imported for exactly this reason); its platform binaries are `optionalDependencies` and Netlify sets `BUN_FLAGS = "--no-optional"`, so a load failure fails the deploy instead of degrading.
- `apps/srd/scripts/og-screenshots.ts:491` — `writeManifest()`/`linkAllRendered()` aren't in a `finally`; one worker's un-retried `freshPage` (lines 435, 472) rejects `Promise.all` and orphans every PNG already written to dist, including the cache-restored ones.
- `apps/srd/scripts/og-screenshots.ts:120` — cache hash omits `CATALOG_TILE_PADDING`, so following the script's own "raise CATALOG_TILE_PADDING" warning changes nothing until `SCRIPT_VERSION` is also bumped.
- `apps/srd/scripts/og-screenshots.ts:221` — pattern hash covers only the pattern, but the tile renders chassis-derived content; a chassis rename leaves every one of its pattern previews stale forever.
- `apps/srd/scripts/og-screenshots.ts:67` — unguarded `Number(env)`: a bad `OG_SCREENSHOTS_BUDGET_MS` silently disables the budget cap (the #482 failure mode), a bad `OG_SCREENSHOTS_CONCURRENCY` yields zero workers and a "successful" 0-image run.
- `apps/srd/scripts/og-screenshots.ts:342` — the island's `MISSING:` marker is never detected; each unresolvable or wedged entity burns 2 × 30s of a 300s budget.
- `apps/srd/netlify.toml:124` — the existing `/*.png` `immutable, max-age=31536000` header now applies to ~1000 non-content-hashed `.og.png` files, so regenerated previews can't bust caches.
- `apps/srd/scripts/og-screenshots.ts:270` — a missing HTML file is counted as `mismatched` and reported as "BaseLayout's meta block may have changed", misdirecting diagnosis.

alxjrvs (commented):
I reviewed the 3 commits on `worktree-og-catalog-previews` (`origin/main...HEAD`): the build-time og:image generator, the shared `CatalogTile`, the `/og-card` render surface, and the Netlify wiring.

I verified the feature end-to-end rather than only reading it: `bun --filter srd og:generate` ran clean (1004 PNGs, all 1200×630, 1004 pages relinked, 0 failures, 0 overhang warnings, 34 MB of images), `bun run knip` is green, `bun --filter srd typecheck` is green, the sitemap correctly excludes `/og-card` and the PNGs, and `define:vars` does reach `#og-card` (checked via the Astro compiler). The tree is left clean.

Findings (10):

- `apps/srd/scripts/og-screenshots.ts:417` — `fit: 'contain'` shrinks tall tiles to an unreadable thumbnail. Measured on real output: `crawler-bays/item/pilot-bay.og.png` renders the card 178px wide in a 1200px canvas; 19% of a 168-image sample is under 600px. Worse than the default banner for those entities.
- `apps/srd/scripts/og-screenshots.ts:233` — pattern cache key hashes only the pattern, but the tile's name tab, stats, artwork and lead prose all come from the chassis, so chassis edits never invalidate pattern PNGs.
- `apps/srd/scripts/og-screenshots.ts:354` — the island's deliberate `MISSING:<key>` sentinel is never checked; an unresolvable or crashing card costs 2×30s, so five of them exhaust the 300s Netlify budget.
- `apps/srd/netlify.toml:29` — the "never fails the build" guarantee doesn't hold: the unguarded `og:install-browser` step and og-screenshots.ts's module-level init (`import sharp`, the top-level-await ORM preload) both bypass `run().catch`.
- `apps/srd/scripts/og-screenshots.ts:367` — a swallowed `img.decode()` failure produces an artwork-less card that the manifest then caches permanently.
- `apps/srd/scripts/og-screenshots.ts:503` — the manifest is only persisted at the very end, so a mid-run abort voids the entire incremental cache.
- `apps/srd/netlify.toml:124` — `/*.png` gives the content-derived og:images a 1-year `immutable` cache at a stable URL, so corrections never reach social scrapers.
- `apps/srd/scripts/og-screenshots.ts:67` — `Number(env ?? default)` turns an empty-string env var into 0, silently rendering zero images.
- `apps/srd/scripts/og-screenshots.ts:441` — viewport is the tile width (432), not the frame width (480); media queries evaluate mobile, latent until a `sm:`/`md:` class enters the catalog path (PLAUSIBLE, not currently triggered).
- `apps/srd/src/components/islands/CatalogTile.tsx:29` — name collides with the already-exported `CatalogTile` in component-lib.

PR #660, opened 2026-07-29
fix(itun): sweep invite rows on game deletion, and correct the ADR citation

alxjrvs:
Follow-up to #655. Two findings from that PR's recorded review that were real against the merged code — the merge queue landed #655 before these could be folded in.

## The leak

`games.destroy` deletes `invites` but was never extended for the two tables added beside it. Deleting a Game left:

- `inviteRedemptions` rows pointing at a Game that no longer exists, and
- **a pending `joinRequests` knock at a door that is gone** — unanswerable forever, since the only surface that could approve or decline it went with the Game.

Both now go with the Game. `joinRequests` gains a plain `by_game` index for the sweep: a prefix scan of `by_game_state` would have worked, but it reads as "pending only" to whoever touches it next, which is exactly the kind of thing that quietly stops sweeping declined rows later.

Covered by a test that fails without the fix — it seeds a redemption and a pending knock, destroys the Game, and asserts all three tables are empty.

## The citation

The amendment opened by attributing to ADR-030 §2 a claim §2 does not make. §2 is *Containers*, and its only mention of invites is that a shelf has none; the authority is §3, which grants the Organizer "invites, membership" and stops there.

The mis-citation was inherited from the module header in `convex/invites.ts`, which said the same thing — fixed at the source too, since that comment is what an implementer reads first.

Also corrects the design doc's "decline is terminal", which the built behaviour deliberately contradicts: a declined knocker may ask again, so an Organizer who declined by mistake need not mint a fresh code.

## Verification

`bun --filter itun typecheck` clean; all 174 Convex tests pass, including the new destroy case.

https://claude.ai/code/session_01DxsNxRvv4nDHZuZ9SqyNYG

### DIFF
--- a/apps/itun/convex/__tests__/invites.test.ts
+++ b/apps/itun/convex/__tests__/invites.test.ts
@@ -379,6 +379,35 @@ describe('knock and approve', () => {
   })
 })
 
+describe('deleting a game', () => {
+  test('sweeps redemptions and pending knocks along with the invites', async () => {
+    const t = testConvex()
+    const { organizer, gameId } = await seedGame(t)
+    const joiner = await makeUser(t, 'Sam')
+    const knocker = await makeUser(t, 'Knocker')
+
+    const open = await organizer.as.mutation(api.invites.create, { gameId })
+    await joiner.as.mutation(api.invites.redeem, { code: open })
+
+    const gated = await organizer.as.mutation(api.invites.create, {
+      gameId,
+      requiresApproval: true,
+    })
+    await knocker.as.mutation(api.invites.redeem, { code: gated })
+
+    await organizer.as.mutation(api.games.destroy, { gameId })
+
+    // A knock at a door that no longer exists would otherwise sit pending
+    // forever, and a redemption row would point at a deleted Game.
+    const leftovers = await t.run(async (ctx) => ({
+      invites: (await ctx.db.query('invites').collect()).length,
+      redemptions: (await ctx.db.query('inviteRedemptions').collect()).length,
+      requests: (await ctx.db.query('joinRequests').collect()).length,
+    }))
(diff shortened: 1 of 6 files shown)


alxjrvs (commented):
Reviewed commit `684ac742` (the only commit not on the upstream/pushed branch; working tree clean). The change itself is sound — the added `inviteRedemptions`/`joinRequests` sweep, the new `joinRequests.by_game` index, the ADR citation fix, and the new test all check out, and `bun --filter itun test` passes (1619 tests). I verified each finding empirically with throwaway convex-test files (since removed; tree is clean).

Findings:

- `apps/itun/convex/games.ts:206` — the sweep still omits `channelBindings` (plus `downtime`, `changeLog`, `mechPatterns`). Confirmed: after `games.destroy`, the binding row survives; `bot.bindChannel` for that channel throws "already bound to another game" and `bot.unbindChannel` throws "Not a member of this game" (its `requireOrganizer` needs a membership the same sweep deleted). The Discord channel is permanently unusable.
- `apps/itun/convex/invites.ts:351` — the re-knock branch handles only `state === 'declined'`. Confirmed: after `ownership.leaveGame` (which deletes the membership but leaves the request `approved`), re-redeeming the same gated code returns `{kind:'pending'}` while `pendingRequests` shows 0 — the user waits on a knock the Organizer can never see.
- `apps/itun/convex/account.ts:168` — `deleteAccount`'s last-member-out teardown is a second hand-copied table list that wasn't updated. Confirmed: after that path deletes a Game, `inviteRedemptions: 1, joinRequests: 1` survive — exactly the leak this commit declares impossible. The two lists should be one shared helper.
- `apps/itun/convex/__tests__/invites.test.ts:8` — header still cites "ADR-030 §2–4", the citation this commit just corrected elsewhere (§2 is Containers). Docs nit.

PR #662, opened 2026-07-29
feat(srd): fit each entity's og:image card to the preview canvas

alxjrvs:
The per-entity `og:image` rendered as a tall, narrow card floating in wide empty margins — the shape in the Discord preview that prompted this. Two causes, both fixed here.

## 1. It was capturing the *mobile* card, not the Catalog tile

The capture ran at a **432px viewport** — the tile's own width — so every `sm:`/`md:`/`lg:` variant on the card was off. `ReferenceEntityCard` keys its shape off viewport media queries (`md:flex-row`, `md:w-2/5`, `sm:columns-2`), so what got photographed was the **mobile stack** (artwork above the prose), not the desktop tile the module claims to be 1:1 with.

Layout now happens at `CATALOG_VIEWPORT_WIDTH` (1440), the breakpoint the 3-up masonry grid is measured at. The capture stays element-scoped, so the extra viewport width never lands in the image.

## 2. One fixed tile width can't fit every card

432px was pinned for every entity, so any card whose content made it taller than the canvas letterboxed into pillars.

The width is now **fitted per entity**: the generator re-flows the real card at each width in `CATALOG_TILE_WIDTHS` (432→1008, step 32) and keeps whichever covers the most of the 1200×630 canvas. The grid tile is fluid (`flex-1`), so every candidate is a genuine catalog layout — this picks which one to photograph, it does not invent an OG-only layout.

Near-ties go to the **narrowest** width, so a card only widens when widening actually buys canvas.

## Measured, not asserted

Over the full **1,004-entity** corpus: 639 stay at the 432px grid width (content-light tiles are already at full canvas width — a one-line card cannot fill a 1.9:1 canvas at any width), 365 widen, 2 hit the ceiling.

On a 107-entity sample, before vs. after:

| | mean | median | best |
|---|---|---|---|
| before (vw432, fixed 432) | 50.9% | 49.3% | 74.7% |
| viewport fix only | 51.6% | 50.7% | 74.7% |
| **after (fitted)** | **55.3%** | **50.7%** | **83.2%** |

Content-rich tiles — the ones that were badly pillarboxed — gain 35–40pp: `classes/hauler` +40.3, `equipment/survey-drone` +39.7, `chassis/atlas` +35.4. **Zero entities regressed by more than 1pp.**

Full run: 1,004 rendered in ~51s, no failures, no overhang warnings.

## Notes

- `pickTileWidth` is a **pure, unit-tested** function over real measurements — card height is a step function of width, so there is no closed form to derive. 13 new tests in `apps/srd/src/lib/__tests__/ogCard.test.ts` (the OG pipeline had none).
- The run now **logs the chosen width distribution** and flags a pile-up at the ceiling. A heuristic that silently collapses onto one width looks exactly like a working one.
- `SCRIPT_VERSION` and the geometry cache key both move, so existing PNGs self-invalidate rather than being served at the old geometry.
- Rasterisation scale is derived from the *narrowest* candidate, so no chosen width is ever upscaled onto the canvas.
- All of `check:all` passes: lint, format, typecheck ×4, tests, validate, knip, audit, tokens, styling.

https://claude.ai/code/session_01XAhCRvuZY6USFBYzqLjUoT


alxjrvs (commented):
**Scope:** local `main` is stale, so the real diff is `origin/main...HEAD` — one commit, `de4fe52f feat(srd): fit each entity's og:image card to the preview canvas`. 5 files: `apps/srd/src/lib/ogCard.ts`, `apps/srd/scripts/og-screenshots.ts`, `apps/srd/src/pages/og-card.astro`, `apps/srd/src/lib/__tests__/ogCard.test.ts`, `apps/srd/CLAUDE.md`.

**Verified clean:** `bun --filter srd typecheck` (0 errors), `bun --filter srd test` (996 pass), `bun run knip` (clean). I also confirmed the two load-bearing mechanics hold: the inline `--tileWidth` on `#og-card` does beat the `define:vars` default (inline declarations win), and Playwright 1.62 passes `captureBeyondViewport: !fitsViewport`, so the reduced 1000px viewport height still captures taller cards in full. The card stack uses only `sm:`/`md:`/`lg:` variants (no `xl:`/`2xl:`), so a widened tile at a 1440 viewport really is layout-identical to a real 3-up tile on a wider monitor — the "1:1 with the Catalog view" claim survives the widening.

**Findings (6):**

- `apps/srd/src/lib/ogCard.ts:132` — `pickTileWidth` scores purely on canvas area with no floor on render scale, so the prose-heavy cards it widens most are downsampled to ~1.14–1.29× CSS instead of 2.5×, roughly halving preview text size for exactly the entities that most need readable text.
- `apps/srd/scripts/og-screenshots.ts:400` — the pass runs on every Netlify deploy (`netlify.toml` → `og:generate`, 300s budget); SCRIPT_VERSION 6 invalidates the whole cache while per-entity cost rises (19 forced re-layouts + 2.4× larger page surface), eroding the headroom over the measured ~70s cold run. Overrun silently leaves pages on the default og:image.
- `apps/srd/src/components/islands/OgCardIsland.tsx:22` — header comment still names `CATALOG_TILE_WIDTH` (deleted by this PR) and asserts a fixed tile width that no longer exists.
- `apps/srd/scripts/og-screenshots.ts:416` — `widthPicks` is incremented before the screenshot, so retried entities double-count; with `generated === 0` the `atMax > generated / 10` ceiling NOTE fires on a run that rendered nothing.
- `apps/srd/scripts/og-screenshots.ts:141` — `TILE_FILL_EPSILON` changes the chosen width but is absent from `entityHash`, contradicting that comment's own "every geometry input that changes the OUTPUT is in the key"; tuning it serves stale cached PNGs.
- `apps/srd/src/lib/__tests__/ogCard.test.ts:27` — the no-upscale test asserts `width + 2*padding <= OG_WIDTH`, comparing CSS px to canvas device px; it would block the script's own recommended `CATALOG_TILE_MAX_WIDTH` raise past 1152 for a reason that doesn't hold, and wouldn't catch a real upscale if `SCALE` were decoupled from `CATALOG_MIN_FRAME_WIDTH`.

PR #663, opened 2026-07-29
fix(component-lib): render one card per installed copy in a pattern loadout

alxjrvs:
## What

A chassis pattern's systems and modules were de-duplicated before rendering, so **every multiple collapsed into a single card**.

Atlas's Thunder Storm prints as `.50 Cal Machine Gun x6` in the Core Book (Digital Edition 2.0a, p.125 — identical in 1.2). The SRD pattern view rendered one machine gun.

## Scope

**60 of the 209 patterns** in the dataset carry a multiple that was being dropped. The heaviest case, Leviathan / Destroyer, showed one `.50 Cal` instead of six, one `Red Laser` instead of three, and one `30mm Autocannon` instead of two — reading as a far lighter build than the book describes.

## Cause

`resolvePatternGroups` (`resolveNestedEntities.ts`) looked each system up **by name only**, never read `count`, and de-duplicated by `label:schema:id`.

The data spells multiplicity two ways, and both were swallowed:

- a `count` on the entry — the common case (`{ name: '.50 Cal Machine Gun', count: 6 }`)
- the **same name repeated as separate entries** — Trooper / DronTek lists `Articulated Rigging Arm` and `Chaff Launcher` twice each with no `count`

A fix that only read `count` would still have dropped the second. Removing the de-dupe and expanding `count` covers both.

## Why this shape

ITUN's wizard path (`useChassisPatternConfig`) **already** expanded counts into one card per copy. This brings the SRD pattern view in line with it rather than introducing a third behaviour. Rendering choice confirmed with @alxjrvs — one card per installed copy, not a `×6` badge.

Duplicate cards were already safe to render: the card's `cardKey` disambiguates by index, not by entity id.

## Tests

One test pinned the old behaviour (`'duplicate entries within a group are de-duplicated'`) and is replaced by its inverse. Added:

- `resolveNestedEntities.test.ts` — Atlas / Thunder Storm's full 11-entry system list; Leviathan / Destroyer's three independent counts; Trooper / DronTek's repeated-entry spelling; count-of-1 and module-count cases
- `PatternView.test.tsx` — end-to-end render assertion that six `.50 Cal Machine Gun` cards appear and that uncounted systems stay single

`bun run check:all` exits 0.

## Not in this PR

Two book-side findings surfaced while verifying, left alone deliberately:

1. The book annotates `Weapon Link` with the weapons it links (`.50 Cal Machine Gun × 6`, `FM-3 Flamethrower // Chainsaw Arm`). Our pattern-module shape is `{ name, count }`, so that designation isn't captured for any of the 19 patterns carrying one.
2. Thunder Storm needs **22 system slots on a 21-slot chassis**. Every slot value matches the book (verified pp. 164/165/171 against the `Slots Required / Salvage Value` legend), so this is an error in the published pattern, not in our data — but it would fail any future slot-budget validation.

https://claude.ai/code/session_01AuHFE7oDVDau2EZjy1Nmk3

### DIFF
--- a/packages/component-lib/src/components/referenceEntity/card/__tests__/PatternView.test.tsx
+++ b/packages/component-lib/src/components/referenceEntity/card/__tests__/PatternView.test.tsx
@@ -104,6 +104,27 @@ describe('full pattern view reading order', () => {
     expect(patternProse).toBeLessThan(systems)
     expect(systems).toBeLessThan(modules)
   })
+
+  test('a system installed six times renders six cards, not one', () => {
+    // The end-to-end guard for the multiplicity fix in `resolvePatternGroups`.
+    // Atlas's Thunder Storm prints as ".50 Cal Machine Gun x6" in the book; the
+    // card used to de-duplicate the loadout by entity id, so the whole pattern
+    // rendered a single machine gun and read as a far lighter build than it is.
+    const chassis = SalvageUnionReference.Chassis.all().find((c) => c.name === 'Atlas')
+    if (!chassis) throw new Error('Atlas fixture missing')
+    const pattern = chassis.patterns?.find((p) => p.name === 'Thunder Storm')
+    if (!pattern) throw new Error('Thunder Storm pattern fixture missing')
+    // Guard the fixture: the count is what drives the expansion.
+    expect(pattern.systems?.find((s) => s.name === '.50 Cal Machine Gun')?.count).toBe(6)
+
+    render(<ReferenceEntityCard data={chassis} pattern={pattern} size="large" />)
+
+    // One name node per rendered card.
+    expect(screen.getAllByText('.50 Cal Machine Gun')).toHaveLength(6)
+    // The uncounted systems stay single — the fix must not multiply everything.
+    expect(screen.getAllByText('Shotgun Pit')).toHaveLength(1)
+    expect(screen.getAllByText('Armour Plating')).toHaveLength(1)
+  })
 })
 
 /**
--- a/packages/component-lib/src/components/referenceEntity/card/__tests__/resolveNestedEntities.test.ts
+++ b/packages/component-lib/src/components/referenceEntity/card/__tests__/resolveNestedEntities.test.ts
@@ -130,12 +130,64 @@ describe('resolvePatternGroups', () => {
     expect(names(group(groups, 'Systems')?.entities ?? [])).toEqual(['Green Laser'])
   })
 
-  test('duplicate entries within a group are de-duplicated', () => {
+  test('a `count` renders one card per installed copy', () => {
+    // Atlas's Thunder Storm is the printed case: the book's pattern block reads
+    // ".50 Cal Machine Gun x6". This resolver used to de-duplicate by entity id,
+    // so all six collapsed into one card and the pattern silently under-reported
+    // its armament.
+    const groups = resolvePatternGroups(pattern('Atlas', 'Thunder Storm'))
+    expect(names(group(groups, 'Systems')?.entities ?? [])).toEqual([
+      '.50 Cal Machine Gun',
+      '.50 Cal Machine Gun',
+      '.50 Cal Machine Gun',
+      '.50 Cal Machine Gun',
+      '.50 Cal Machine Gun',
+      '.50 Cal Machine Gun',
+      'Armour Plating',
+      'Escape Hatch',
+      'Locomotion System',
+      'Personnel Transport Pod',
+      'Shotgun Pit',
+    ])
+  })
+
+  test('several counted systems in one pattern each keep their own multiplicity', () => {
+    // Leviathan's Destroyer is the heaviest case in the dataset: three separate
+    // counted weapons, which the old de-dupe flattened to one card each.
+    const systems = names(
+      group(resolvePatternGroups(pattern('Leviathan', 'Destroyer')), 'Systems')?.entities ?? []
+    )
+    expect(systems.filter((n) => n === '.50 Cal Machine Gun')).toHaveLength(6)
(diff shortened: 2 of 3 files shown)


alxjrvs (commented):
## Review: `fix(component-lib): render one card per installed copy in a pattern loadout`

**Scope** — the local `main` ref is stale (a `main...HEAD` diff shows 179 files); the real PR against `origin/main` is 3 files / 1 commit: `resolveNestedEntities.ts` plus two test files. Working tree is clean.

### Findings (1)

- `packages/component-lib/src/components/referenceEntity/card/resolveNestedEntities.ts:125` — `Math.max(1, count)` promotes a schema-legal `count: 0` to one rendered card, while `useChassisPatternConfig` (the expansion the docstring cites as the shared contract) uses `count ?? 1` and renders zero. The default parameter `count = 1` already handles the only case needing a fallback, so the clamp only fires on `0`. Latent — no pattern in the dataset carries `count: 0`.

### What I verified (and found clean)

- **The fix is semantically right.** Expanding `count` and preserving repeated entries makes 133 of 209 patterns sum *exactly* to their chassis system/module slot budgets. Trooper/DronTek — the one pattern in the dataset spelling multiplicity as repeated entries with no `count` — hits 16/16 systems and 3/3 modules only when both duplicates are counted, confirming the old de-dupe was dropping real data rather than masking a data error.
- **No `count`/repeat double-spelling.** Scanned all patterns: zero entries both repeat *and* carry a `count`, so nothing gets multiplied twice. Also zero `count: 0`, no non-integers, none above 10.
- **React keys hold.** `cardKey` is `${id}-${index}`, and all three `renderNested` layout paths (flat, compact-column, masonry-plus-orphan) index uniquely — including the orphan, keyed `entities.length - 1`. Ran the srd suite with this file reverted to `origin/main`: the 3 pre-existing "unique key" warnings are identical before and after, and the Atlas pattern render emits none.
- **Single call site, no leakage.** `resolvePatternGroups` is used only at `ReferenceEntityCard.tsx:1255`, is not in the `component-lib` barrel, and its now-un-deduped `Drones` group is filtered out there. ITUN's `MechChassisStep` doesn't pass `pattern`, so no double-render of the loadout alongside `useChassisPatternConfig`'s `afterExtraContent`.
- **Aggregations unaffected.** `computeSvOverride` reads `patternOverride` directly with its own `count ?? 1`, not this resolver, so salvage value isn't double-counted.
- **Gates green.** `bun --filter component-lib test` 686 pass, `bun --filter srd test` 996 pass, `bun run typecheck` clean across all workspaces, `bun run lint` clean on the changed files, `knip` clean.

### Non-blocking notes

- The SEO/no-JS fallback (`apps/srd/src/lib/patternSummary.ts`) still expresses multiplicity as `"Name ×6"` while the hydrated island now renders six cards. Intentional given the PR's stated goal, but the two representations of the same pattern now differ in shape.
- `resolveNestedEntities.test.ts` asserts the complete ordered 11-name Systems list for Atlas/Thunder Storm; any future addition to that pattern breaks a test that isn't about the addition. The companion `PatternView.test.tsx` guards its fixture with an explicit `count` assertion, which is the more robust shape.

alxjrvs (commented):
**Scope reviewed:** `git diff origin/main...HEAD` — 2 commits (`85e57bed`, `5295adb8`), 3 files, all in `packages/component-lib/src/components/referenceEntity/card/` (the `resolvePatternGroups` de-dupe removal plus its tests). Working tree is clean.

**Verification performed:** `bun run test` (1636 + 1003 + 688 tests, 0 fail), `bun run typecheck` (all workspaces clean), plus a data scan of `chassis.json` (209 patterns; 61 carry a multiple; max `count` is 6, on Atlas/Thunder Storm; the only repeated-entry spelling is Trooper/DronTek — both cases the tests cover).

**Things I checked and cleared:**

- **React key collisions from duplicate entities** — `cardKey` at `packages/component-lib/src/components/referenceEntity/card/ReferenceEntityCard.tsx:1544` appends the array index, and the masonry orphan path uses `entities.length - 1`, so the 10 column keys and the orphan key can't collide. The doc comment's claim is accurate.
- **Duplicate cards sharing DOM state** — no `id=`, `useId`, `htmlFor`, or `type="radio"` anywhere in the entity-display path, so six identical cards can't cross-wire form state.
- **Blast radius of the `Drones` group losing its de-dupe** — `ReferenceEntityCard.tsx:1255` filters `label !== 'Drones'`, and `resolvePatternGroups` has no other caller (it isn't in the `component-lib` barrel), so Big Brother's four same-stat-block drones don't double-render.
- **Chassis pages and og:images** — pattern listing rows use `extent="head"`, which returns at `ReferenceEntityCard.tsx:1061`, before `patternGroups` is computed; catalog tiles force `hide.patterns = true` at line 500. Neither surface expands, so the chassis index and the generated tile screenshots are untouched.
- **Consistency with the wizard** — `useChassisPatternConfig.tsx:73/86`, `patternOverrideUtils.ts:36`, and `MechWizard.tsx:255` all use `?? 1`, which agrees with the new `count = 1` default parameter (they differ only on `null`, which the Zod `.optional()` rejects).
- **Untrusted `count` values** — ITUN's user-saved patterns never reach the `pattern` prop (`PatternList.tsx` renders them as plain rows), so `Array.from({ length: count })` only ever sees schema-validated `NonNegativeIntegerSchema` data.

**Finding (1, low severity, latent):**

- `apps/srd/src/lib/patternSummary.ts:7` — `loadoutLine` tests `i.count && i.count > 1`, so an explicit `count: 0` is falsy and falls through to the bare-name branch. The card now renders zero copies for `count: 0` while the same page's static/no-JS summary would still print the system's name. Not reachable with current data (no pattern uses `count: 0`), but it is the one place that doesn't honour the semantic commit `5295adb8` deliberately introduced.

PR #664, opened 2026-07-30
feat(itun): the Games lobby as a controls band over its list

alxjrvs:
## What

Reshapes the Games index to the Roster's rhythm: **what you can DO is one ink-ruled band across the top, and what you HAVE lists directly beneath it.**

Before, three creation Cards (start / from-a-template / join) stacked above the list, so the tables you came to open sat below a screenful of forms you had already used. A lobby's subject is the list; creating is a control on it.

```
BEFORE                              AFTER
+------------------------+          [Start a game ___] [Create] [From a template]
| Start a game  [Create] |          [Join with code ___] [Join]
+------------------------+          ------------------------------------ <- border-b-2
| Or start from a        |
| template ...           |          THE ELDRIDGE COAST      Organizer - Mediator
+------------------------+          Hamlet - 4 Pilots - 3 Mechs    [View]
| Join with a code [Join]|
+------------------------+
  ...games below the fold
```

Start-a-game and join-by-code stay one click, inline. The template list moves behind a dialog: it is a list to *read* — a name plus a paragraph each — and inlining it was exactly what wedged prose between the controls and the games.

## Two fixes on the way through

**`templates.list` types its `id` as the `TemplateId` union, not `string`.** With a plain `string` the only way to satisfy `createGame`'s `v.literal('starter-set')` validator was to hardcode the id at the call site — so the lobby rendered one template and would have created a *different* one the moment a second existed. Typed this way, adding a template widens the union and the validator becomes the type error, which is where you want to notice.

**The template query stays in the parent and is handed down.** `open` is local to the picker, so toggling it re-renders only that child. A query living there would make the screen's read sequence depend on which child last changed state — and the connected tests answer queries by call order. Found the hard way: mid-implementation the picker rendered the *games array* as its template list, showing "Union Crawler #430" as a template name.

## Tests

- `offers a template to start from` rewritten for the new contract: the template is **absent** until the control is clicked, then present.
- New `the creating controls come BEFORE the games they act on` — asserts document order, so the band-over-list arrangement is pinned rather than assumed.
- 88 games tests pass; `bun run check:all` exits 0.

## Not in this PR

The per-game dashboard at `/games/$gameId` (`GameScreen`) **already exists on main** — shipped by #656/#655 — so the second half of the original ask needed no work. This PR is only the lobby layout.

Separately: prod's Convex backend is missing `games.get`, `ownership.claim`, and the `invites.*` functions that the deployed frontend calls, which is why the live per-game page currently errors. That is a deploy-skew issue, not a code issue, and is not addressed here.

https://claude.ai/code/session_01AuHFE7oDVDau2EZjy1Nmk3

### DIFF
--- a/apps/itun/convex/templates.ts
+++ b/apps/itun/convex/templates.ts
@@ -40,8 +40,15 @@ export type TemplateId = keyof typeof TEMPLATES
 export const list = query({
   args: {},
   handler: async () => {
+    // `id` is typed as the TemplateId union rather than `string` so a caller can
+    // pass the row straight to `createGame`. With a plain `string` the only way
+    // to satisfy that mutation's `v.literal` validator was to hardcode the id at
+    // the call site, which meant the lobby rendered one template and created
+    // another the moment a second one existed. Typed this way, adding a template
+    // widens the union and `createGame`'s validator becomes the type error —
+    // which is the right place to notice.
     return Object.entries(TEMPLATES).map(([id, t]) => ({
-      id,
+      id: id as TemplateId,
       name: t.name,
       description: t.description,
     }))
--- a/apps/itun/src/components/games/GamesScreen.tsx
+++ b/apps/itun/src/components/games/GamesScreen.tsx
@@ -1,6 +1,7 @@
 import { useState } from 'react'
-import { Button, Card, EntityRow, Text } from 'component-lib'
+import { Button, Card, EntityRow, ModalShell, Text } from 'component-lib'
 import { useMutation, useQuery } from 'convex/react'
+import type { FunctionReturnType } from 'convex/server'
 import { useNavigate } from '@tanstack/react-router'
 
 import { api } from '../../../convex/_generated/api'
@@ -8,7 +9,7 @@ import { useConnection } from '../../lib/connection/connectionContext'
 import { isConvexConfigured } from '../../lib/connection/convexClient'
 import { SignInControl } from '../account/SignInControl'
 import { GameRow } from './GameRow'
-import { STAMP } from './gameChrome'
+import { INPUT, STAMP } from './gameChrome'
 
 /**
  * Games — the shelf of tables you belong to.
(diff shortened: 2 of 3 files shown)


alxjrvs (commented):
Review scope: `origin/main...HEAD` is a single commit — `64ad2996 feat(itun): the Games lobby as a controls band over its list` — touching 3 files (`apps/itun/convex/templates.ts`, `apps/itun/src/components/games/GamesScreen.tsx`, and its connected test). No uncommitted changes; the branch is fully pushed.

Verification I ran: `bun run typecheck` in `apps/itun` (clean), the 13 tests in `GamesScreen.connected.test.tsx` (all pass), `biome check` on the three files (clean), and both design-system gates (`check:tokens`, `check:styling` — no new violations; the removed `border-[var(--color-ink)]` literals only lower the count).

Findings:

- `apps/itun/src/components/games/GamesScreen.tsx:83` — The name typed into "Start a game" is no longer forwarded to template creation. The old call was `createFromTemplate({ templateId: 'starter-set', name: newName || undefined })`; the new one passes only `templateId`. Since "From a template" now sits in the same control band immediately right of that input, a user who types a name and then starts a template game gets the template's default name, with their input silently discarded and still sitting in the field. The server explicitly supports this (`convex/__tests__/templates.test.ts:91` — "a custom name overrides the template name"), and `createGame`'s optional `name` arg now has no production caller.
- `apps/itun/src/components/games/GamesScreen.tsx:83` — `void createFromTemplate(...).then(() => setOpen(false))` has no `.catch`. On rejection (expired session hitting `requireUser`, or a network error) the dialog stays open with no error text and the rejection is unhandled, so the user's only signal is nothing happening; each retry that succeeds creates another Game. The sibling `join` handler in the same file deliberately catches and surfaces the server's wording, and the band already has a `joinError`/`joinNotice` slot.

Things I checked and cleared: the `id as TemplateId` cast in `templates.list` is sound (`TemplateId` is currently exactly `'starter-set'`, and adding a template does surface a compile error at this call site against `v.literal`, as the comment claims); `ModalShell` is exported from `component-lib`, renders nothing when closed, and portals outside `container` (so the new ordering test's `container.textContent` assertions are unaffected by the dialog); `Button variant="default"` exists; `INPUT` exists in `gameChrome.ts` and its concatenation with `tracking-caps-wide uppercase` has no conflicting utilities; the `Card` import is still used by `GamesBody`/`GamesScreen`; and the comment's claim about keeping the templates query in the parent is correct — `setOpen` re-renders only `TemplatePicker`, so the call-order-keyed query mock is not disturbed.

PR #665, opened 2026-07-30
feat(component-lib): render a pattern loadout as linked shortform badges

alxjrvs:
## What

A pattern's **Systems** and **Modules** now render as the entity card's own **shortform badge** (`size="small" extent="head"`), each linking to that entity's page — instead of a full inline card per entry.

Atlas's *Thunder Storm* was the worst case: six identical, full-length `.50 Cal Machine Gun` cards, a **5462px** page. It's now **~1300px** of card content, and the pattern's own flavour is no longer buried under reprinted rules.

| Before | After |
| --- | --- |
| Six full cards, each repeating the same paragraph | Six linked badges, tone band + TL tail intact |

## Decisions

- **Reuses the card, not a new chip.** The badge *is* `ReferenceEntityCard` at its smallest rung — the "SHORTFORM token" the card already implements — so the tone band and tech-level tail come for free. An earlier pass hand-rolled a `Badge` with a swatch; that was wrong and was replaced.
- **Multiples repeat, they don't collapse.** Six machine guns are six badges, not one reading `×6`. That matches what the pattern installs and what `resolvePatternGroups` already emits; a count compresses the shape of the build out of the one row whose job is showing it.
- **Real `<a>` links**, like `PatternListRow` — middle-click, hover preview, crawlable. Inert with no `EntityHrefProvider` (ITUN), rather than linking nowhere.

## Layout (all scoped to the pattern view)

- **Systems / Modules side by side** in one grid so the columns line up (a pattern installs far more systems than modules); single column below `sm`, with a **vertical rule** carried on the second column so it spans whichever column is taller.
- **Pattern title tightened** — the solid `Slab`'s `flex-1` leader rule ran clear across the card and read as a page divider; `w-fit` collapses it to a stub.
- **Below the fold.** A pattern card now takes the **aside lead**: artwork + chassis prose as a centred lead row, then chassis ability, pattern prose and loadout at full width beneath — the way a class page's ability trees read.

### Reversal, stated plainly

The last point reverses an earlier ruling (pinned by `AsideLead.test.tsx`) that patterns must keep the float. That ruling was about not being caught *accidentally* by the generic `afterExtraContent` gate. The loadout now renders inline and reads as a section, so the layout is **chosen, not inherited**. The gate is still never inferred from the generic slot. The **basic chassis card is untouched** and still wraps its pattern list around the illustration — there's a new test pinning that.

## Verification

Verified locally in the dev server at 1280px and 390px, plus the Big Brother *DronTek* pattern (drones still render as full cards with their own loadouts) and the basic Atlas chassis page (unchanged).

- `typecheck` — 4/4 workspaces clean
- `test` — 0 fail (3 new tests for link resolution, no inline rules, inert-without-provider; 2 rewritten for the reversal)
- `lint`, `format:check`, `knip`, `validate:all` — clean
- Link resolution confirmed against the hydrated page: all six copies → `/schema/systems/item/50-cal-machine-gun/`

https://claude.ai/code/session_01XmNCbFJgv4mcbRFKiXJabX

### DIFF
--- a/packages/component-lib/src/components/referenceEntity/card/ReferenceEntityCard.tsx
+++ b/packages/component-lib/src/components/referenceEntity/card/ReferenceEntityCard.tsx
@@ -1641,19 +1641,24 @@ function ReferenceEntityCardInner({
   // on a line under a mostly-empty one. Single column below `sm` — two columns
   // of badges on a phone would each be too narrow to hold a system name.
   //
-  // FLOWS AROUND the floated anchor (artwork): a grid container establishes its
-  // own independent formatting context, so it is placed BESIDE the float in the
-  // width that is left rather than overlapping it — no `clear` and no
-  // `flow-root` needed, and `flow-root` would in fact fight the `grid` display
-  // for the same property. Both columns narrow together, so the loadout never
-  // splits into one indented column and one full-width one.
-  const renderPatternLoadout = (groups: NestedGroup[], flat = false): ReactNode => (
-    <div className={cn('grid grid-cols-1 gap-x-3 gap-y-1.5 sm:grid-cols-2', flat && 'mb-1.5')}>
+  // NO float handling here, deliberately: a pattern card takes the ASIDE LEAD,
+  // which drops the float entirely, so this always renders full width beneath
+  // the lead row. (An earlier revision took a `flat` argument and documented how
+  // the grid would sit beside the floated artwork. Once patterns moved below the
+  // fold that state became unreachable — `flat` is `hasAnchor && !asideLead`,
+  // and a pattern's `asideLead` is gated on the same artwork — so the parameter
+  // was dead and the comment described a layout that can no longer happen.)
+  const renderPatternLoadout = (groups: NestedGroup[]): ReactNode => (
+    <div className="grid grid-cols-1 gap-x-3 gap-y-1.5 sm:grid-cols-2">
       {groups.map((group, index) => (
         <div
           key={group.label}
           className={cn(
-            'flex flex-col gap-1.5',
+            // `min-w-0` is load-bearing: a grid item defaults to `min-width:
+            // auto`, so a column refuses to shrink below its widest badge and
+            // the card's `overflow-hidden` hard-clips it. "Electro-Magnetic
+            // Shield Projector" plus its TL tail overruns a two-column track.
+            'flex min-w-0 flex-col gap-1.5',
             // A VERTICAL RULE between the columns. On the column itself rather
             // than a third grid cell, so it spans whichever column is taller
             // (Systems usually far outruns Modules) instead of standing at the
@@ -1664,11 +1669,19 @@ function ReferenceEntityCardInner({
           )}
         >
           <Slab variant="dashed" label={group.label} />
-          <div className="flex flex-wrap items-start gap-1.5">
+          {/* A real LIST: the loadout is a countable set of installed parts, so
+              a screen reader should announce "list, 6 items" rather than a run
+              of anonymous links. `list-none` keeps the markers off. */}
+          <ul
+            aria-label={`${group.label} installed`}
+            className="flex list-none flex-wrap items-start gap-1.5"
+          >
             {group.entities.map((item, itemIndex) => (
-              <LoadoutBadge key={cardKey(item, itemIndex)} data={item} hostDown={isDown} />
+              <li key={cardKey(item, itemIndex)} className="min-w-0">
+                <LoadoutBadge data={item} hostDown={isDown} />
+              </li>
             ))}
-          </div>
+          </ul>
         </div>
       ))}
     </div>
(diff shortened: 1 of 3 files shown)


alxjrvs (commented):
Reviewed `main...HEAD` (commit 5d83b420, "render a pattern loadout as linked shortform badges"); working tree clean, no unpushed changes. I read the full card component around every hunk, checked all `pattern`-passing consumers (srd `ReferenceEntityIsland`/`CatalogTile`, itun `MechChassisStep`), and verified claims empirically with throwaway probe tests (removed afterward). All 692 component-lib, 996 srd and 1636 itun tests pass.

Things I checked and cleared: the catalog tile does **not** nest the new badge anchors inside its own outer `<a>` (`isCatalog` forces `hide.patterns: true`); itun's wizard pattern cards are unaffected because `useChassisPatternConfig` never passes `pattern`, so `isPattern` stays false and `asideLeadRequested` still gates them; `border-l-chrome` *is* registered in `cn()`'s `border-w-l` group, so it survives the merge with `border-ink/20`; `w-fit` on the solid `Slab` conflicts with nothing; `patternGroups` filters out `Drones`, so the grid never exceeds two cells and the `index > 0` divider can't land in column one; and the new `not.toContain('This simple ballistic weapon')` assertion is non-vacuous (that prose comes from the `.50 Cal Machine Gun` *action*, which a full nested system card does fold in).

Findings:

- `packages/component-lib/src/components/referenceEntity/card/ReferenceEntityCard.tsx:2112` — Drone list children lose their key on pattern cards. `asideLead` now makes `flat` permanently false for pattern cards, and `wrapFlat(false, key, node)` discards the key it's handed; line 2112 is the only `wrapFlat` call site that doesn't also set `key` on the inner card. Reproduced the React warning on Little Sestra / Surveyor (all three Little Sestra patterns have both artwork and a drone, so all three regress).
- `packages/component-lib/src/components/referenceEntity/card/ReferenceEntityCard.tsx:1644` — `renderPatternLoadout`'s `flat` parameter is unreachable and its comment is false. For a pattern card `flat = showImage && !showImage` = always false, so the `flat && 'mb-1.5'` branch is dead and the eight-line comment about the grid flowing around the floated artwork describes a state that can't occur (confirmed: no `md:float-left` in the rendered output).
- `packages/component-lib/src/components/referenceEntity/card/ReferenceEntityCard.tsx:1667` — Long badge names overflow and get hard-clipped rather than truncating. Each badge is a flex item with default `min-width: auto` over `whitespace-nowrap` text, so it can't shrink; "Electro-Magnetic Shield Projector" plus the TL tail is ~280px against a ~270px column at `sm:grid-cols-2`, and the card frame's `overflow-hidden` cuts it off with no ellipsis. `min-w-0` on the wrapper would fix it. (PLAUSIBLE — layout claim not browser-verified.)
- `packages/component-lib/src/components/referenceEntity/card/ReferenceEntityCard.tsx:2312` — `LoadoutBadge`'s anchor has no `aria-label`, so its accessible name is ".50 Cal Machine Gun TL 1" repeated six times; the sibling `PatternListRow` supplies one deliberately for exactly this reason, and the wrap is a plain `div` rather than a list, so no item count is announced.

alxjrvs (commented):
Reviewed the 3-file diff (`ReferenceEntityCard.tsx` + two test files) against `origin/main`, opened the surrounding card internals, the srd/ITUN consumers, and verified claims empirically (render probes on both HEAD and main, plus reintroducing the drone-key bug to confirm the new regression test actually fails).

Finding:

- `packages/component-lib/src/components/referenceEntity/card/ReferenceEntityCard.tsx:1904` — the new `asideLead = showImage && (isPattern || …)` gate is not scoped to the full pattern view, so **catalog-extent** pattern tiles also lose the artwork float and switch to the centered aside row. Catalog mode forces `hide.patterns = true` (line 499), so the loadout that justifies the layout is never rendered there. Verified: Atlas / "Thunder Storm" at `extent="catalog"` emits `md:float-left` on main and the aside row on this branch. That changes every chassis-pattern tile in the srd schema-index masonry grid (now laid out unlike the chassis tiles beside it) and every pattern og:image, which `OgCardIsland` → `CatalogTile` produces and `og-screenshots.ts` fits by measuring tile height at widths 432–1008 on a 1440 viewport. Gating on `extent === 'full'` (or on `inFlowGroups.length > 0`) confines the reversal to the surface that asked for it.

Things I checked that are **not** issues:

- The drone `key` fix is real and the new `PatternView.test.tsx` guard genuinely catches its regression, both in isolation and in the full suite (React's per-owner key-warning dedup does not make it vacuous — confirmed by removing the key).
- `LoadoutBadge`'s links bypass `useEntityDetailLink` (the gate `useDetailModal` uses to keep ITUN from linking off-site), but no ITUN surface renders a `pattern`-prop card today — ITUN's wizard goes through `useChassisPatternConfig`/`afterExtraContent`, so `isPattern` is false there and there is no reachable failure.
- No nested-anchor hazard: `CatalogTile` wraps the card in an `<a>`, but the loadout is suppressed at catalog extent.
- No chassis pattern has an empty systems+modules loadout, so the unguarded empty grid `<div>` in `renderPatternLoadout` is unreachable.
- `typecheck`, `lint`, `knip`, `component-lib` (694) and `srd` (996) tests all pass; the only lint warnings are pre-existing.

PR #666, opened 2026-08-03
fix(component-lib): fall back to badge stats when a card header is too narrow

alxjrvs:
## The bug

On a phone, a chassis' header stats ran off the side of the card, over the breadcrumb and out of the viewport (reported against `/schema/chassis/item/mule/`).

The cluster was `shrink-0`, so the `flex-wrap` beside it could never engage — a flex item that cannot shrink never reaches a width its children must wrap at. Eight stats want a ~420px line, which a 320px band does not have.

## The fix — two layers, so it does not depend on JS

1. **The cluster shrinks**, so the value boxes wrap instead of overflowing. This is the no-JS floor, and what a crawler and the pre-hydration paint see.
2. **The header measures its own band** and, when the boxes don't fit, swaps to the compact `[label | value]` cells — the badge anatomy a listing card already uses — with the short-form labels (TL / SV / SYS / MODS) that anatomy is written for. The card builds both label sets and hands the header both; an override cluster (editable sheet stats) is already short-form and serves both.

Measured rather than breakpointed: "enough space" is a function of the **card's** width and its stat **count**, not the viewport — a 1-stat card is fine at 320px, an 8-stat one is cramped in a 500px column on a desktop. A width of 0 (happy-dom, a `display:none` ancestor) is explicitly *not* a verdict.

At that width the **title also stacks** above the cluster: a full card's title is `text-5xl`, so its min-content (one word) is already most of a phone-width band and it painted straight over the stats.

## Verification

Measured in headless Chrome on `/schema/chassis/item/mule/` (8 stats):

| viewport | band | anatomy | overflow | page scroll |
|---|---|---|---|---|
| 390 | 320 | badge cells | none | none |
| 500 | 430 | badge cells | none | none |
| 768 | 698 | value boxes (unchanged) | none | none |
| 1440 | 1114 | value boxes (unchanged) | none | none |

A 1-stat equipment card keeps its value box at 390. The pattern page (`/pattern/hauler/`) behaves identically.

`bun run check:all` green (lint, format, typecheck, 4330 tests, validate, knip, audit, tokens, styling). New `narrowStats.test.tsx` pins both layers by stubbing `clientWidth` + `ResizeObserver`, since happy-dom performs no layout.

https://claude.ai/code/session_013QzYAhR1MjVmeKU8zttL7V

### DIFF
--- a/packages/component-lib/src/components/referenceEntity/card/EntityCardHeader.tsx
+++ b/packages/component-lib/src/components/referenceEntity/card/EntityCardHeader.tsx
@@ -65,11 +65,18 @@ const HYSTERESIS = 24
  * before the name has any room at all, which a phone (or a narrow grid column)
  * does not have — the cluster used to be `shrink-0`, so it neither shrank nor
  * wrapped and simply ran off the card, over the header and out of the viewport.
- * The cluster now wraps (that alone keeps it on the card with no JS, which is
- * what a crawler and the pre-hydration paint see), and once mounted the header
- * measures itself and swaps to the compact `[label | value]` cells — the same
- * badge anatomy a listing card already uses — which fit a phone at 2 rows
- * instead of 3 stacks of boxes.
+ * The cluster now wraps, and on top of that the header measures itself and
+ * swaps to the compact `[label | value]` cells — the same badge anatomy a
+ * listing card already uses — which fit a phone at 2 rows instead of 3 stacks
+ * of boxes.
+ *
+ * The wrap is the floor for every case this measurement declines to judge: an
+ * unmeasurable band (below), no `ResizeObserver`, or a host that renders the
+ * card without ever running an effect. It is NOT what a crawler sees — srd
+ * gates this card behind `GameDataGate` and serves `StaticEntityContent` to
+ * no-JS readers, and ITUN is client-only, so the card is never
+ * server-rendered. Nor is it a visible frame: `useLayoutEffect` measures
+ * before paint, so the boxes never flash on a narrow card.
  *
  * Measured, not breakpointed, because "enough space" is a function of the CARD's
  * width and its stat COUNT, not the viewport: a 2-stat card is fine at 320px and
@@ -152,9 +159,13 @@ export function EntityCardHeader({
       {title}
     </TitleTag>
   )
-  // Too narrow for the value boxes → the compact cells, with the short-form
-  // labels that anatomy is written for.
-  const clusterStats = narrow ? (narrowStats ?? stats) : stats
+  // On the compact cells → the short-form labels that anatomy is written for.
+  // `listing` is included because it renders the cells too: a `size="large"`
+  // + `extent="head"` card is NOT `compact`, so its `stats` carry the two-line
+  // long form, and without this it would put "Structure / Points" inside a
+  // shortform cell. (`compact` needs no clause — such a card's two lists are
+  // the same list.)
+  const clusterStats = narrow || listing ? (narrowStats ?? stats) : stats
   const statsNode =
     stats.length > 0 ? (
       <EntityCardStatBox stats={clusterStats} compact={compactStats || narrow} />
@@ -255,7 +266,14 @@ export function EntityCardHeader({
         // word) is already most of a phone-width band and it painted straight
         // over the stats, which is the same collision the compact cells fix one
         // level down. Wrapping gives each side a whole row instead.
-        narrow && 'flex-wrap gap-y-2',
+        //
+        // Only WITHOUT prose. The prose rule owns both columns' widths (a 75%
+        // ceiling on the title, a 55% basis on the description), and those beat
+        // the stacking classes below — `max-w-[75%]` clamps `basis-full`, and a
+        // non-auto basis beats `w-full`. So it would read as a stack rule that
+        // silently does nothing; a card carrying BOTH prose and stats keeps the
+        // prose allocation, which is what it did before this change.
+        narrow && !hasProse && 'flex-wrap gap-y-2',
         accent.className
       )}
       style={accent.style}
@@ -269,7 +287,7 @@ export function EntityCardHeader({
         className={cn(
           hasProse ? TITLE_VS_PROSE : 'min-w-0 flex-1',
           // Stacked: the title takes the whole first row.
-          narrow && 'basis-full'
+          narrow && !hasProse && 'basis-full'
         )}
       >
         {titleNode}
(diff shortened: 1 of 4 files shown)


alxjrvs (commented):
**Findings (3, all low severity — no correctness bug found in the shipped path):**

- `packages/component-lib/src/components/referenceEntity/card/EntityCardHeader.tsx:272` — Narrow-mode stacking is inert when the header also has flavour prose: `basis-full` is clamped by `max-w-[75%]` and `w-full` is overridden by the non-auto `flex-[1_1_55%]` basis, so the "title row, then cluster row" stack the comment promises doesn't happen for that (test-pinned, but currently unreachable) prop combination.
- `packages/component-lib/src/components/referenceEntity/card/EntityCardHeader.tsx:157` — `narrowStats` is consulted only on the measured `narrow` verdict, not on `listing`, so a `size="large" extent="head"` card renders the compact cells with the two-line long-form labels (`Structure`/`Points`) that `narrowHeaderStats` exists to replace. Fix is `narrow || listing`; no in-repo call site hits it today.
- `packages/component-lib/src/components/referenceEntity/card/EntityCardHeader.tsx:69` (also `EntityCardStatBox.tsx:107-112` and the new test docblock) — the "no-JS floor / what a crawler and the pre-hydration paint see" rationale is false: srd gates the card behind `GameDataGate` (server snapshot hardcoded `false`, renders a `Skeleton`; crawlers get `StaticEntityContent`) and ITUN is client-only, so the card is never server-rendered.

**What I verified and found sound:** the two stat lists are guaranteed same-length and same-keyed (`buildReferenceEntityStats` filters identically for `compact` true/false), so the swap can't drop or re-key a stat; the hysteresis direction is right and cannot oscillate (the narrow verdict changes header height, not band width, and `MasonryColumns` columns are `flex-1 min-w-0`, i.e. content-independent); width 0 correctly abstains; `enabled` toggling resets state; hooks are unconditional; `shrink-0` removal doesn't change the desktop title-yields rule (the title's `basis-0` contributes zero scaled shrink, so the cluster still absorbs all shrinkage only in the overflow case that was previously the bug); `useLayoutEffect` runs pre-paint so there's no anatomy flash. Editable-stat measurement drift (a stepper column makes a box ~70–94px, not 52) exists in theory but every production `statsOverride` with `onChange` is on a `size="medium"` card, which is never measured — so I did not report it. `bun --filter component-lib test` (699 pass), `bun run typecheck` and `bun run lint` are all clean; the new test's `clientWidth` prototype stub does restore correctly (happy-dom defines it on `HTMLElement.prototype`, which I checked), so it doesn't leak into other test files.

alxjrvs (commented):
**Scope note:** `git diff @{upstream}...HEAD` was empty; the change under review is commit `c0b9b6d9` plus the uncommitted edits that appeared in the worktree mid-review (another session is actively iterating on these same three files — it has already landed fixes for the `listing`-card label mismatch and the no-op prose stacking classes, both of which I'd otherwise have flagged).

**Verification done:** read every hunk plus `EntityCardStatBox`, `Stat`, `buildReferenceEntityStats`, and all `statsOverride` / `EntityCardHeader` consumers; ran `bun --filter component-lib test` (699 pass); and measured the real rendered page in headless Chrome across 390/500/560/600/640/700/768/900/1024/1440.

Things I checked and cleared: `compact:true/false` only changes stat *labels*, never the stat count, so `stats.length` is a valid threshold input; the anatomy swap remounts `HorizontalValue` vs `ValueBox` (no hook-order violation); no ResizeObserver feedback loop (the compact anatomy is always narrower, so `narrow` can't oscillate); hysteresis direction is correct; `hide.stats` / `statsOverride` keep both lists consistent; the effect deps use `statCount`, not the freshly-built array, so no effect loop; no SSR exposure (srd gates the card behind `GameDataGate`, ITUN is client-only); the 1024px horizontal page overflow is the site nav, unrelated to this diff.

**Finding (1):**

- `packages/component-lib/src/components/referenceEntity/card/EntityCardHeader.tsx:56` — `TITLE_RESERVE = 140` under-reserves for a `text-5xl` title, so the fallback stops short of the widths where the collision it exists to fix still happens: on `/schema/chassis/item/stormterror/` the title paints over the first stat box by 173px at viewport 640, 113px at 700, and 45px at 768 (clear only at 900+). The h1 is `w-fit` + `break-words`, so its min-content is the entire word (307px) and it overflows its `flex-1` (basis-0) column rather than wrapping; the short-named "Mule" on the same page never collides. Measuring the title's actual width instead of a fixed constant would close the 557–860px gap. This overlap predates the change (the old `shrink-0` cluster occupied the same 416px), but it is the exact failure mode the new measurement is introduced to eliminate.

alxjrvs (commented):
## Findings

- **`packages/component-lib/src/components/referenceEntity/card/EntityCardHeader.tsx:56`** — `TITLE_RESERVE = 140` makes the narrow fallback fire ~300px too late. Browser-measured at the real geometry: at a 557px band (just above the 556px threshold for an 8-stat chassis) the title box is 105px, "IRON MONGREL" breaks mid-word across **5 lines**, header height **264px**; at 555px the fallback fires and the same title is 1 line in a **128px** header. Titles stay broken up to ~866px. Reserve the title's min-content at its own type scale rather than a flat 140px.
- **`packages/component-lib/src/components/referenceEntity/card/EntityCardStatBox.tsx:66`** — the read-only vertical `Stat` branch never forwards `stat.outOfMax`, but the compact-cell branch does. Now that the anatomy is width-dependent, ITUN's Mech Sheet chassis card shows System Slots as bare "4" on desktop and "4/6" on a phone.

## What I checked and cleared

- `headerStats` is byte-identical to pre-change; the second `buildHeaderStats(true)` build is pure and same-length (compact never changes the stat count in `referenceEntityStatsConfig.ts`), so `stats.length` is a valid measurement basis.
- `clusterStats`/`compact` truth table is correct across all four `compact` × `listing` × `narrow` combinations; `statsOverride` and `hide.stats` feed both lists consistently.
- No ResizeObserver feedback loop — narrow toggling changes band height, not width, and both states are stable attractors even in shrink-to-fit containers. Hysteresis direction is correct.
- The `basis-full` / `flex-1` CSS-order conflict on the stacked title is harmless: `w-full` on the cluster forces the wrap regardless, and the title (grow 1) fills line 1. Verified in-browser (1-line title, no overflow, at 555/500/390px).
- The `shrink-0` → `min-w-0` change genuinely fixes the overflow: `scrollWidth === clientWidth` on the cluster at every width from 1112 down to 390.
- SSR/`useLayoutEffect` concern is real but not reachable — `GameDataGate` returns its fallback until a client-only `ready` state flips, so srd never server-renders the card.
- The `VALUE_BOX_WIDTH = 52` constant ignores the editable stepper column (~94px on mobile), but every editable-`statsOverride` call site (`PartnerCard`, `size="medium"`) resolves to `compact`, which disables measurement — not live.
- Test-stub hygiene in `narrowStats.test.tsx` is sound: `clientWidth` is genuinely an own property of `HTMLElement.prototype` in happy-dom (probed), so the restore path can't leak.
- `bun --filter component-lib test` (700 pass), `bun run typecheck`, `bun run lint`, `bun run knip` all green.

PR #671, opened 2026-08-04
fix(discord-bot): fail fast on a silent login hang, and unwedge the bot build

alxjrvs:
Two production failures, both found while investigating "the discord bot is throwing errors". They compound: the first took the bot down, the second blocked the recovery.

---

## 1. The bot was up and deaf for 12.5 hours

On **2026-08-03** the `suref-discord-bot` worker logged:

```
11:28:41  [observability] Sentry tracking enabled
11:28:41  Starting Salvage Union Discord Bot...
```

…and then nothing. No `Logged in as`. The process sat alive at ~90MB and ~0% CPU, holding a gateway connection that never completed, while every slash command answered **"The application did not respond"** — Discord's wording for an interaction nobody acknowledged.

### Why nothing caught it

- **Render** saw a live process — a background worker has no HTTP port to health-check.
- **Sentry** saw no exception — `client.login()` neither resolved nor rejected, and no `error` event fired.
- The **`discord-bot ready`** event `handleReady` emits is only ever a *presence*. Nothing alerts on its absence, so the gap read as silence rather than as a signal.

The startup code was byte-identical to the last healthy boot (2026-07-28, `fe0a5cee`), which went `Starting` → `Logged in as` in **1.34s**. discord.js was unchanged at 14.27.0 (also the latest published), and `bun install --frozen-lockfile` is clean — a stuck connect, not a code regression. It will happen again.

### Fix

The process asserts its own liveness: reach ready within **60s**, or exit non-zero and let Render restart the worker.

- `src/startupWatchdog.ts` — `armReadyWatchdog()` returns a disarm function; idempotent, and `unref`'d so the watchdog is never itself the reason the process is still running.
- `src/index.ts` — armed **before** preload (a hang anywhere before a connected gateway looks identical from outside), disarmed by `ClientReady`. On expiry: log, report to Sentry, flush, `exit(1)`.

60s is ~45x a healthy boot, so it can only fire on a genuine hang. **A crash loop is a better failure than a bot that is up and deaf, because a crash loop is visible.**

---

## 2. `bun --filter` wedges the bot build on Render

Recovering needed a redeploy — and the redeploy hung for **17 minutes** on this line:

```
$ bun --filter discord-bot build
```

The bundle itself takes ~20ms. This has now happened twice, both times with no output after that line:

- **2026-07-28** — ran ~2h until Render SIGTERMed it. That is what drained the workspace's build-pipeline minutes, which cancelled the next 7 deploys and left the bot on a stale build for six days.
- **2026-08-04** — wedged the recovery deploy. Superseding it produced the identical signature, which is what confirmed the diagnosis:

```
error: script "build:bot" was terminated by signal SIGTERM (Polite quit request)
```

### Fix

`build:bot` now runs `bun run --cwd apps/discord-bot build` — the identical script with no filter runner in front of it. Bundles in 21ms and exits cleanly. `render.yaml` carries the reasoning so nobody "tidies" it back.

---

## Verification

- `bun --filter discord-bot test` — **182 pass, 0 fail** (5 new: fires undisarmed, silent when disarmed, disarm idempotent, error names the budget, default budget).
- `bun --filter discord-bot typecheck` clean; `biome lint` on changed files clean.
- Built the real bundle and ran `node dist/index.js`: an invalid token still fails fast and exits, so the watchdog does not interfere with the existing error path.
- Confirmed the new `build:bot` produces the same two entry points.

## Not in this PR

- **Production was restored by redeploying** (`dep-d9oito1t0dsc73b5ukp0`), not by this branch. This PR stops both failures recurring.
- The absent-`ready`-event alert is still unarmed — a Sentry-side config change, not code.
- The new `/su game`, `/crew` and `/account` commands from #653 still need a manual `deploy-commands:global`, unrelated to this outage.

https://claude.ai/code/session_019U1DfcdEKS1SqJ5T4sHAWH

### DIFF
--- a/apps/discord-bot/src/__tests__/startupWatchdog.test.ts
+++ b/apps/discord-bot/src/__tests__/startupWatchdog.test.ts
@@ -1,0 +1,64 @@
+import { describe, expect, test } from 'bun:test'
+import { armReadyWatchdog, READY_TIMEOUT_MS } from '../startupWatchdog.js'
+
+/**
+ * startupWatchdog — the guard against the bot being up and deaf.
+ *
+ * The failure it exists for is a silent one (see the module docs), so these
+ * tests care about exactly two things: that an undisarmed deadline actually
+ * fires, and that a disarmed one stays quiet forever. A watchdog that cried
+ * wolf would crash-loop a perfectly healthy bot, which is the worse of the two
+ * regressions — hence the disarm cases.
+ *
+ * Timers run for real at a few milliseconds rather than through a fake clock:
+ * the unit under test is a `setTimeout` and there is nothing left to assert
+ * once the clock is mocked away.
+ */
+
+const tick = (ms: number): Promise<void> => new Promise((resolve) => setTimeout(resolve, ms))
+
+describe('armReadyWatchdog', () => {
+  test('calls onExpire with an Error once the deadline passes', async () => {
+    const expirations: Error[] = []
+    armReadyWatchdog({ timeoutMs: 5, onExpire: (error) => expirations.push(error) })
+
+    await tick(30)
+
+    expect(expirations).toHaveLength(1)
+    expect(expirations[0]).toBeInstanceOf(Error)
+  })
+
+  test('names the elapsed budget in the error, so a log line explains itself', async () => {
+    const expirations: Error[] = []
+    armReadyWatchdog({ timeoutMs: 5, onExpire: (error) => expirations.push(error) })
+
+    await tick(30)
+
+    expect(expirations[0]?.message).toBe('Discord login did not reach ready within 5ms')
+  })
+
+  test('stays silent when disarmed before the deadline', async () => {
+    const expirations: Error[] = []
+    const disarm = armReadyWatchdog({ timeoutMs: 20, onExpire: (error) => expirations.push(error) })
+
+    disarm()
+    await tick(50)
+
+    expect(expirations).toEqual([])
+  })
+
+  test('disarming twice is safe — a reconnect re-firing ready must not throw', async () => {
+    const expirations: Error[] = []
+    const disarm = armReadyWatchdog({ timeoutMs: 20, onExpire: (error) => expirations.push(error) })
+
+    disarm()
+    expect(() => disarm()).not.toThrow()
+    await tick(50)
+
+    expect(expirations).toEqual([])
+  })
+
+  test('defaults to a budget far above a healthy 1.34s boot', () => {
+    expect(READY_TIMEOUT_MS).toBe(60_000)
+  })
+})
--- a/apps/discord-bot/src/index.ts
+++ b/apps/discord-bot/src/index.ts
@@ -5,6 +5,7 @@ import { commands } from './commands/index.js'
 import { handleReady } from './events/ready.js'
 import { handleInteractionCreate } from './events/interactionCreate.js'
 import { initObservability, captureException, flushObservability } from './observability.js'
+import { armReadyWatchdog } from './startupWatchdog.js'
 
 // Initialize error tracking as early as possible (no-op without SENTRY_DSN).
 initObservability()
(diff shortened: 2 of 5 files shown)


alxjrvs (commented):
Reviewed the diff (`origin/main...HEAD`, one commit, 3 files — local `main` was stale, so `main...HEAD` gave a misleading 17-file diff). Scope: a new `startupWatchdog.ts`, its wiring into `index.ts`, and a unit test.

Verified green locally: `bun test` in `apps/discord-bot` (182 pass / 0 fail), `bun run typecheck`, `bun run lint`, and root `bun run knip` (no unused-export complaint about `READY_TIMEOUT_MS`).

Findings:

- `apps/discord-bot/src/index.ts:37` — The `disarmReadyWatchdog()` call inside the ClientReady handler has zero test coverage; if it's ever dropped (reverting to the previous `client.once(Events.ClientReady, handleReady)` looks like a harmless simplification) a healthy bot self-kills after 60s in a permanent restart loop, with the whole suite still green. `index.ts` isn't importable in a test because it ends in a top-level `await client.login(...)`; extracting the handler pair into a small exported factory would make the disarm assertable.
- `apps/discord-bot/src/startupWatchdog.ts:53` — Expiry message is hard-coded to "Discord login did not reach ready", but the watchdog is intentionally armed before `SalvageUnionReference.preload('all')`. A preload hang therefore reports as a Discord problem, sending the operator to the token/gateway instead of the `await` one line earlier — and this message is the entire diagnostic, since the module docs note Sentry sees no exception and Render sees a live process.

Things I checked and deliberately did **not** report:
- `timer.unref?.()` initially looked like it could defeat the watchdog, but it's correct. In the motivating incident the process was alive at ~90MB holding the gateway socket, so the loop is non-empty and the timer fires. In the only case `unref` matters (empty loop with a pending top-level await), Node exits 13 on its own — still non-zero, still restarted by Render. It also correctly avoids holding the process for 60s when `client.login()` rejects and the registered `unhandledRejection` handler suppresses the default crash.
- Real-timer tests (`timeoutMs: 5` / `await tick(30)`) are not flaky: timers fire in expiry order, so the 5ms callback always precedes the 30ms tick regardless of load.
- The crash-loop tradeoff (self-imposed deadline could burn Render restarts / Discord session starts) is explicitly reasoned about in the module docs, so it's a documented decision rather than a defect.

alxjrvs (commented):
Reviewed the two-commit diff (`origin/main...HEAD`): the startup watchdog (`apps/discord-bot/src/startupWatchdog.ts` + wiring in `index.ts` + tests) and the `build:bot` `--filter` → `--cwd` change plus its `render.yaml` comment.

**Verification I ran:** `bun run typecheck` (clean), `bun test` in `apps/discord-bot` (182 pass), `bun run knip` (clean), `bun run lint` (only pre-existing warnings elsewhere), `bun run validate:doc-drift` / `validate:observability` / `validate:architecture` / `check-bun-version` (all pass), and `bun run --cwd apps/discord-bot build` itself (bundles in 21ms, exits cleanly on bun 1.3.14).

**The watchdog logic itself holds up.** Armed at module scope before both top-level awaits so it covers preload and login; disarm is idempotent and is called first in the `ClientReady` handler; `unref()` is safe because the observed hang keeps a ref'd gateway socket alive, and the one sub-case it declines to cover (empty event loop) already exits non-zero via Node's unsettled-TLA code 13. Deleting the disarm call would be caught by `noUnusedLocals`, so the untested wiring is compiler-guarded. Timer tests are deterministic (expiry ordering), not flaky.

Findings:

- `.github/workflows/ci.yml:319` — CI still runs `bun --filter discord-bot build`, the exact invocation this PR declares wedges. The job has no `timeout-minutes` and gates the aggregate check, so a wedge blocks the merge queue for GitHub's 360-minute default; and since CI never runs `build:bot`, a break in the new `--cwd` script passes CI and fails only on Render (the `shared` filter includes `package.json`, so the job *does* run on such PRs — it just tests the wrong command).
- `render.yaml:15` — the added comment says the SIGTERMed ~2h hung build "is what drained the workspace's build-pipeline minutes" on 2026-07-28, contradicting the pre-existing comment ten lines below, which attributes the same 2026-07-28 exhaustion to 13 cancelled builds from the missing `buildFilter`. Whoever later re-evaluates the `buildFilter` reads incompatible evidence for its existence.

PR #672, opened 2026-08-04
feat(itun): read a crewmate's sheet, and stamp ownership on every crew row

alxjrvs:
The Game crew view listed rows that mostly went nowhere, said ownership two ways at once, and ran in a narrower column than the Roster it is a twin of.

## What changed

**One ownership seal, three states.** `UNCLAIMED` / `YOU` / the holder's name, stamped in the row's top-right corner. It replaces an owner chip in the caption *plus* a separate UNCLAIMED stamp among the buttons — one field said twice, in two registers, in two places. `UNCLAIMED` is the pressable one (an unclaimed pre-gen is an offer); the other two are inert facts. A crawler carries no seal — it is communal, so "who owns it" is not a question the game asks.

**Every row is a door.** New read-only crew sheet at `/games/$gameId/view/$kind/$entityId`, rendering a crewmate's build from the server body through a frozen store extracted to `sheet/frozenSheet.ts` and shared with the snapshot surface. **Nothing is adopted into IndexedDB** — reading a crewmate's pilot leaves no copy among your own builds. What was never safe was handing a non-owner ITUN's *live* sheet, whose writes the server refuses; that stays owner-only, now labelled **Edit**.

**Rows speak the card's language.** `EntityRow` drops the 6px left rail and the tinted body. A card states its ontology in a band across the top over a paper body, so a row does too, at the card's `--bw-entity` frame weight. Every piece of text on a row is now a badge: strings in `metaLine` are chipped automatically, nodes (cross-links) pass through un-nested.

**Game surfaces run full width** on the Roster's own page shell (`PAGE` in `gameChrome`). Join keeps a capped form inside the wide shell — a six-character code does not want 1440px.

Also resolves the mech chassis slug to its real name on the crew roster, the way the Roster already did ("Iron Mongrel · TL 1", not "iron-mongrel").

## Blast radius

`EntityRow` is shared, so the home Roster and the three live sheets change with it — that is the point (the two surfaces were drifting), but it is the thing to review hardest.

## Verification

Full `check:all` green. Both surfaces were rendered from the **real** components against the **built** CSS and measured at 1440 / 768 / 390: no horizontal overflow at any width, no truncated names. Two truncation regressions were caught and fixed that way ("IR…" on desktop, "IRON MONG…" on mobile).

Draft until the screenshots are reviewed.

### DIFF
--- a/apps/itun/src/components/games/GameEntitySheet.tsx
+++ b/apps/itun/src/components/games/GameEntitySheet.tsx
@@ -1,0 +1,159 @@
+/**
+ * GameEntitySheet — a crewmate's build, read-only (ADR-030 §5).
+ *
+ * ## Why this surface exists
+ *
+ * A Game's roster used to be a list of rows most of which went nowhere: only
+ * what you owned offered a sheet, so a player looking at their crew could see
+ * four names and open one of them. That is the wrong answer to "who am I
+ * playing with" — the whole point of a shared table is that the table is
+ * shared. What was actually unsafe was not *reading* a crewmate's pilot but
+ * being handed ITUN's live sheet for it, which is an editing surface whose
+ * writes the server refuses (`assertMayWrite`).
+ *
+ * So the sheet opens, and it is frozen. `readOnly` suppresses every edit
+ * affordance, and the store behind it is the one from `frozenSheet.ts`, which
+ * holds this single entity and throws on any write. There is no path from here
+ * to a mutation, which is what makes the surface honest rather than merely
+ * discouraging.
+ *
+ * ## Nothing is cached locally
+ *
+ * The row is read straight from `entities.listForGame` and parsed. It is
+ * deliberately NOT adopted into IndexedDB: a crewmate's pilot in the viewer's
+ * own stores would show up among their builds, under a container they do not
+ * control, and would go stale the moment its owner touched it. Reading is not
+ * owning.
+ *
+ * The trade is a round trip — this re-queries the game listing rather than
+ * reading a local copy — which is the correct cost for a surface you visit to
+ * look something up.
+ */
+
+import { useMemo } from 'react'
+import { useQuery } from 'convex/react'
+import { Card, Text } from 'component-lib'
+
+import { api } from '../../../convex/_generated/api'
+import type { Id } from '../../../convex/_generated/dataModel'
+import { useConnection } from '../../lib/connection/connectionContext'
+import { isConvexConfigured } from '../../lib/connection/convexClient'
+import type { RosterKind } from '../../lib/games/gameRoster'
+import { makeFrozenStore, parseFrozenEntity } from '../sheet/frozenSheet'
+import { Sheet } from '../sheet/Sheet'
+import { AppLink } from '../shared/AppLink'
+import { PAGE, TITLE } from './gameChrome'
+
+type GameEntitySheetProps = {
+  gameId: string
+  kind: RosterKind
+  /** The Convex row id — what the crew roster addresses every row by. */
+  entityId: string
+}
+
+/** A framed message with the way back to the crew, for every state but success. */
+export function GameSheetNotice({
+  gameId,
+  children,
+}: {
+  gameId: string
+  children: React.ReactNode
+}) {
+  return (
+    <main className={PAGE}>
+      <Text as="h1" className={TITLE}>
+        Crew sheet
+      </Text>
+      <Card>
+        <div className="flex flex-col items-start gap-3 p-4">
+          <Text>{children}</Text>
+          <AppLink href={`/games/${gameId}`} className="font-cond text-sm font-bold uppercase">
+            ← Back to the crew
+          </AppLink>
+        </div>
+      </Card>
+    </main>
+  )
+}
+
+function GameEntityBody({ gameId, kind, entityId }: GameEntitySheetProps) {
+  // The same query the roster uses, so this surface is governed by exactly the
+  // membership check that decides whether the row was listable in the first
+  // place — a non-member gets `null` here for the same reason they get an empty
+  // roster, with no second permission rule to drift from the first.
+  const listing = useQuery(api.entities.listForGame, { gameId: gameId as Id<'games'> })
+
+  const row = useMemo(() => {
+    if (listing === undefined) return undefined
+    const rows =
+      kind === 'pilot' ? listing.pilots : kind === 'mech' ? listing.mechs : listing.crawlers
+    return rows.find((r) => r._id === entityId) ?? null
+  }, [listing, kind, entityId])
+
+  const parsed = useMemo(
+    () => (row === undefined || row === null ? null : parseFrozenEntity(kind, row.body)),
+    [row, kind]
+  )
+  const store = useMemo(() => (parsed?.ok ? makeFrozenStore(parsed) : null), [parsed])
+
+  if (listing === undefined) {
+    return <GameSheetNotice gameId={gameId}>Loading this sheet…</GameSheetNotice>
+  }
+
+  if (row === null) {
+    // Covers "not in this game", "no such entity", and "it was scrapped"
+    // identically — a non-member must not be able to tell them apart.
+    return (
+      <GameSheetNotice gameId={gameId}>
+        That {kind} is not in this game. It may have been removed, or you may no longer be a member.
+      </GameSheetNotice>
+    )
+  }
+
+  if (parsed === null || !parsed.ok || store === null) {
+    return (
+      <GameSheetNotice gameId={gameId}>
+        This {kind}&rsquo;s data doesn&rsquo;t match anything this app knows how to show. It may
+        have been built in a newer version.
+      </GameSheetNotice>
+    )
+  }
+
+  return (
+    <div>
+      {/* The read-only banner, in the same shape the snapshot surface uses.
+          Saying WHY it is read-only matters more than saying that it is: a
+          player who cannot edit a crewmate's pilot should understand that as
+          the rule of the table, not as something broken. */}
+      <div
+        role="note"
+        aria-label="Read-only crew sheet"
+        className="border-b-2 border-ink bg-caution px-4 py-2 font-body text-sm font-semibold text-ink sm:px-[30px]"
+      >
+        You are reading a crewmate&rsquo;s sheet. Only whoever holds it can make changes.
+      </div>
+
+      <Sheet
+        kind={parsed.kind}
+        id={parsed.entity.id}
+        store={store}
+        back={{ href: `/games/${gameId}`, label: 'the crew' }}
+        readOnly
+      />
+    </div>
+  )
+}
+
+export function GameEntitySheet(props: GameEntitySheetProps) {
+  const { mode } = useConnection()
+
+  if (!isConvexConfigured || mode !== 'connected') {
+    return (
+      <GameSheetNotice gameId={props.gameId}>
+        A Game is shared state, so reading a crewmate&rsquo;s sheet needs a connected account.
+      </GameSheetNotice>
+    )
+  }
+
+  return <GameEntityBody {...props} />
+}
--- a/apps/itun/src/components/games/GameRoster.tsx
+++ b/apps/itun/src/components/games/GameRoster.tsx
@@ -492,7 +492,7 @@ export function GameRoster({ gameId, gameName }: GameRosterProps) {
                               {row.can.openSheet && (
                                 <Button
                                   variant="default"
-                                  size="compact"
+                                  size="mini"
                                   disabled={busy !== null}
                                   onClick={() =>
                                     void run(`open-${row.serverId}`, () => openSheet(row))
@@ -504,7 +504,7 @@ export function GameRoster({ gameId, gameName }: GameRosterProps) {
                               {row.kind === 'mech' && row.can.openSheet && (
                                 <Button
                                   variant="primary"
-                                  size="compact"
+                                  size="mini"
                                   disabled={busy !== null}
                                   onClick={() =>
                                     void run(`dash-${row.serverId}`, () => launchDashboard(row))
@@ -521,7 +521,7 @@ export function GameRoster({ gameId, gameName }: GameRosterProps) {
                               {row.can.release && (
                                 <Button
                                   variant="ghost"
-                                  size="compact"
+                                  size="mini"
                                   disabled={busy !== null}
                                   onClick={() =>
                                     void run(`offer-${row.serverId}`, async () => {
@@ -546,7 +546,7 @@ export function GameRoster({ gameId, gameName }: GameRosterProps) {
                               {row.can.scrap && (
                                 <Button
                                   variant="ghost"
-                                  size="compact"
+                                  size="mini"
                                   disabled={busy !== null}
                                   onClick={() =>
                                     void run(`scrap-${row.serverId}`, async () => {
--- a/apps/itun/src/components/games/GameRoster.tsx
+++ b/apps/itun/src/components/games/GameRoster.tsx
@@ -51,6 +51,7 @@ import {
   ModalShell,
   Text,
 } from 'component-lib'
+import type { EntityRowDetail } from 'component-lib'
 
 import { api } from '../../../convex/_generated/api'
 import type { Id } from '../../../convex/_generated/dataModel'
@@ -161,24 +162,24 @@ function statsFor(row: RosterRow): Array<{ label: string; value: string | number
 }
 
 /**
- * The row's caption parts — callsign, class, chassis — each rendered by
- * `EntityRow` as its own quiet chip.
+ * The row's body details — `CALLSIGN | Ghost`, `CLASS | Salvager` — each
+ * rendered by `EntityRow` as a horizontal `Stat` with its label plate tinted by
+ * ontology.
  *
- * Ownership is NOT among them any more: it is the row's seal, stamped in the
- * top-right corner (see `OwnerSeal`). It used to be a toned chip down here
- * beside the callsign, which put the single most consequential fact about a
- * row — whose character is this — in the same visual register as their
- * nickname.
+ * Ownership is NOT among them: it is the row's seal, stamped on the top border
+ * (see `OwnerSeal`). It used to be a toned chip down here beside the callsign,
+ * which put the single most consequential fact about a row — whose character is
+ * this — in the same visual register as their nickname.
  */
-function captionFor(row: RosterRow): string[] {
-  const parts: string[] = []
+function captionFor(row: RosterRow): EntityRowDetail[] {
+  const parts: EntityRowDetail[] = []
   if (row.kind === 'pilot') {
     const callsign = row.body.callsign
     if (typeof callsign === 'string' && callsign.length > 0 && callsign !== row.name) {
-      parts.push(`"${callsign}"`)
+      parts.push({ label: 'Callsign', value: callsign })
     }
     const className = resolveClassName(String(row.body.classRef ?? ''))
-    if (className) parts.push(className)
+    if (className) parts.push({ label: 'Class', value: className, tone: 'pilot' })
   }
   // A mech's chassis is NOT here: it is a `CHASSIS | Iron Mongrel` stat in the
   // header band. It is a named property of the mech, which is what a Stat is
--- a/apps/itun/src/components/games/GameRoster.tsx
+++ b/apps/itun/src/components/games/GameRoster.tsx
@@ -135,6 +135,17 @@ function statsFor(row: RosterRow): Array<{ label: string; value: string | number
     if (num('currentAP') !== undefined) out.push({ label: 'AP', value: num('currentAP') as number })
   }
   if (row.kind === 'mech') {
+    // The chassis leads: it is what the mech IS, where SP and Heat are how it
+    // is doing. Rendered as `CHASSIS | Iron Mongrel` rather than a bare chip,
+    // so the name arrives labelled.
+    const chassis = row.body.chassisRef
+    if (typeof chassis === 'string' && chassis.length > 0) {
+      const resolved = chassisOf(chassis)
+      out.push({ label: 'Chassis', value: resolved.name })
+      // TL is its own stat rather than a suffix on the chassis value: two facts
+      // crammed into one value box is the thing `Stat` exists to stop.
+      if (resolved.techLevel != null) out.push({ label: 'TL', value: resolved.techLevel })
+    }
     if (num('currentSP') !== undefined) out.push({ label: 'SP', value: num('currentSP') as number })
     if (num('currentHeat') !== undefined) {
       out.push({ label: 'Heat', value: num('currentHeat') as number })
@@ -169,27 +180,28 @@ function captionFor(row: RosterRow): string[] {
     const className = resolveClassName(String(row.body.classRef ?? ''))
     if (className) parts.push(className)
   }
-  if (row.kind === 'mech') {
-    const chassis = row.body.chassisRef
-    // Resolved, not printed raw. The home Roster reads "Iron Mongrel · TL 1"
-    // where this said "iron-mongrel" — the stored value is a slug, and showing
-    // the slug is the surface admitting it never looked the chassis up.
-    if (typeof chassis === 'string' && chassis.length > 0) parts.push(chassisLabel(chassis))
-  }
+  // A mech's chassis is NOT here: it is a `CHASSIS | Iron Mongrel` stat in the
+  // header band. It is a named property of the mech, which is what a Stat is
+  // for — a bare chip saying "Iron Mongrel" left the reader to infer what the
+  // word was doing there.
   return parts
 }
 
-/** A mech's chassis as "Name · TL n", falling back to the raw ref. */
-function chassisLabel(chassisRef: string): string {
+/**
+ * A mech's chassis, resolved. The stored value is a SLUG, and printing the slug
+ * is the surface admitting it never looked the chassis up — the home Roster has
+ * always resolved it, and this said "iron-mongrel" where that said "Iron
+ * Mongrel".
+ */
+function chassisOf(chassisRef: string): { name: string; techLevel?: number } {
   // `resolveChassisRef` throws when the Chassis model is not preloaded (test
   // and snapshot contexts), so this falls back rather than taking the screen
   // down with it — the same guard the Roster's `mechChassisMeta` uses.
   try {
     const chassis = resolveChassisRef(chassisRef) as { name: string; techLevel?: number } | null
-    if (!chassis) return chassisRef
-    return chassis.techLevel != null ? `${chassis.name} · TL ${chassis.techLevel}` : chassis.name
+    return chassis ?? { name: chassisRef }
   } catch {
-    return chassisRef
+    return { name: chassisRef }
   }
 }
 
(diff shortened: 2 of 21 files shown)


alxjrvs (commented):
Reviewed commit `965d59f9` against `origin/main` (the crew read-only sheet + `EntityRow` band/seal rework). `bun --filter itun test`, `bun --filter component-lib test`, and `bun run typecheck` all pass on the committed tree.

Findings:

- `apps/itun/src/components/games/GameEntitySheet.tsx:136` — Crew mech sheet drops pilot ability bonuses. The frozen store has no soft-links and no pilot record, and no `pilotAbilities` is passed, so `pilotingContext(mech, undefined)` strips Beefcake-style Max SP / Max Cargo bonuses. Worse, `sp = Math.min(mech.currentSP ?? maxSP, maxSP)` then clamps the *stored current* SP down, so a crewmate reads a mech as 15/15 that its owner reads as 18/18. This is precisely the failure the `pilotAbilities` prop's own docstring was added to prevent for snapshots. `entities.listForGame` already returns `softLinks` and pilot bodies, so the fix data is available.
- `apps/itun/src/components/games/GameEntitySheet.tsx:84` — Non-member hits the root error page, not the crafted notice. The comment claims `useQuery` returns `null` for a non-member, but `listForGame` calls `requireMember`, which throws `NotAuthorized`; Convex re-throws in render, so `__root.tsx`'s `errorComponent` replaces the shell. The `row === null` branch at line 103 that says "you may no longer be a member" is unreachable for that exact case.
- `packages/component-lib/src/components/shared/EntityRow.tsx:299` — `pr-24` (96px) does not bound the absolutely-positioned seal. `OwnerSeal`'s stamp has no `max-w`/`truncate`, so a long crewmate display name overflows the reserved corner and overlaps the name stamp, contradicting the docstring's guarantee.
- `apps/itun/src/components/games/GameRoster.tsx:29` — Module header still asserts the rule this commit reverses ("Only what you own offers a sheet", "adopted on the way in", "an owner chip on every row"). Per this repo's CLAUDE.md, stale prose here gets followed rather than checked.

One caveat: the worktree is being edited concurrently while I reviewed — `EntityRow.tsx` and `GameRoster.tsx` picked up uncommitted follow-up changes mid-pass (stats moved into the header band, `STAMP_SEAM` seal placement, chassis moved to a Stat). A transient `tsc` TS1005 in `EntityRow.tsx` during that window was a mid-write artifact and cleared on re-run, not a defect. I scoped the findings to the committed diff; all four still apply to the in-flight tree, except that the follow-up already removed the no-op `rotate-[-4deg]` on the inert seal's inline `<span>` (a cosmetic issue I therefore did not report).

alxjrvs (commented):
Reviewed `git diff main...HEAD` (7 commits: Games crew view, EntityRow reseal, card narrow-stats fallback, pattern loadout badges) plus the uncommitted working-tree changes. Note: the working tree was being actively edited *during* the review — `git status` was clean at the start and picked up 5 modified files mid-pass, so findings 1, 2 and 6 are against an in-progress edit that may still be moving.

Verified: `bun --filter component-lib test` (700 pass), `bun --filter itun test` (1638 pass), `bun run typecheck` (passes at HEAD, **fails on the working tree**), `bun run lint` (warnings only, pre-existing), `bun run format:check` (fails on the working tree).

Findings:

- `packages/component-lib/src/components/shared/EntityRow.tsx:418` — `toNodes` was widened to `Array<ReactNode | EntityRowDetail>` for both `meta` and `metaLine`, so `metaBadges.map` feeds a possible `EntityRowDetail` into `<Badge>`; confirmed `TS2322` from `tsc --noEmit`.
- `apps/itun/src/components/roster/Roster.tsx:23` — `Badge` import left behind after `linkSegment` switched to `Stat`; confirmed `TS6133`.
- `apps/itun/src/components/games/GameEntitySheet.tsx:84` — the comment claims a non-member "gets `null` here", but `listForGame` → `requireMember` throws and Convex's `useQuery` re-throws during render, so the crafted "not in this game / no longer a member" notice is unreachable and the root error page shows instead.
- `apps/itun/src/components/games/GamesScreen.tsx:83` — template creation dropped the `name: newName || undefined` argument, so the "Start a game" name field is silently ignored when starting from a template.
- `apps/itun/src/components/games/GamesScreen.tsx:83` — `createFromTemplate(...).then(...)` has no `.catch`; a rejected mutation leaves the dialog open with no message and an unhandled rejection (the sibling `join` handler does surface errors).
- `packages/component-lib/src/components/shared/EntityRow.tsx:229` — fails `bun run format:check` (reformatted `toNodes` signature + misindented `metaLine` JSDoc).

Things I checked and cleared: the `useNarrowStats` measure/hysteresis loop (no oscillation, `bandRef` attached on both return paths, card is never SSR'd since `GameDataGate` gates it); the `asideLead` widening for patterns (pattern *listing* cards return early at `extent === 'head'`, so they can't be dragged in); the `wrapFlat` drone-key fix (correct — `wrapFlat` does discard the key when `flat` is false); `frozenSheet.ts` extraction from `SnapshotSheet` (behaviour-preserving, `parseContext` still only applied to mechs); and `ownableRows`' `openSheet` narrowing to owner-only while every row gets a read-only `View`.

alxjrvs (commented):
Reviewed the 3-commit PR (`origin/main...HEAD`): the read-only Game crew-sheet surface (`GameEntitySheet` + `frozenSheet.ts` + new route), the `EntityRow` card-anatomy rewrite (band/body/seal), and the Roster/GameRoster caption→`Stat` migration. Verified with `bun run typecheck` (clean), `bun --filter itun test` (1638 pass), `bun --filter component-lib test` (700 pass), `bun run lint` and `bun run knip` (clean) — so every finding below is behavioural, not a broken build.

Findings:

- `apps/itun/src/components/games/GameEntitySheet.tsx:84` — Non-member (or malformed `gameId`) crashes the whole app instead of showing the notice. `listForGame`'s `requireMember` **throws**, and convex/react's `useQuery` rethrows in render (`if (result instanceof Error) throw result`), so the `row === null` "you may no longer be a member" branch is unreachable and `RootErrorComponent` takes over. `GameScreen` guards this by checking `api.games.get` (which returns `null`) first; the new route has no such gate.
- `apps/itun/src/components/games/GameEntitySheet.tsx:136` — Mech crew sheets omit `pilotAbilities`, so a crewmate's mech reads lower Max SP / Cargo than on the owner's sheet (frozen store ⇒ `pilot: null` ⇒ `pilotingContext(mech, undefined)`). This is the exact failure `Sheet`'s own JSDoc warns about and `SnapshotSheet` guards against; the needed data is already in the same query result (`listing.softLinks` + `listing.pilots`).
- `apps/itun/src/components/games/GamesScreen.tsx:249` — The shared `PAGE` shell drops the lobby's `max-w-2xl` cap, so the single-column Games list and its inputs go full-bleed at desktop widths. `PAGE`'s rationale ("three ontology columns") doesn't apply here, and `JoinScreen` got an explicit inner cap in the same commit for exactly this reason.
- `apps/itun/src/components/roster/Roster.tsx:230` — `linkSegment`'s Badge→`Stat` swap inherits `Stat`'s hardcoded `cursor-default` (Stat.tsx:354), so Roster cross-links no longer show a pointer cursor.
- `packages/component-lib/src/components/shared/EntityRow.tsx:141` — The `seal` prop doc claims "the band reserves the width for it", directly contradicting the implementation comment 230 lines down ("no right reserve"). No visual break today (the `pt-3` clears the stamp), but it's a false guarantee in a doc-driven codebase.
- `apps/itun/src/components/games/GameEntitySheet.tsx:85` — No tests for the new surface: none of the four render states, the shared `parseFrozenEntity`/`makeFrozenStore`, or the route's `isRosterKind` guard are exercised. `GameRoster.connected.test.tsx` only asserts the `href`, never the destination.

One thing I checked and cleared: `EntityRow`'s filled variant silently drops its `className` prop (used by `SheetPilot`/`SheetMech` rails) — but that's pre-existing on `origin/main`, not introduced here.

alxjrvs (commented):
Reviewed `origin/main...HEAD` (4 commits: frozen crew sheet + EntityRow band redesign) plus the uncommitted working-tree changes on top, which appeared mid-review and materially change the diff (they drop `EntityRowDetail`, move class/callsign into the band, and delete `captionFor`). Typecheck, lint, and all 4,514 tests are green.

Findings:

- `packages/component-lib/src/components/shared/EntityRow.tsx:362` — Row title hardcodes `text-paper` on all four ontology bands, contradicting the `TONE[...].ink` map three lines up (pilot/mech/game are declared ink) and `Badge.tsx`'s `BADGE_TONES` ("game: Light enough for dark ink"). Game rows (`--color-sheet-game` #7dceeb) render the game name at ~1.7:1 contrast — effectively invisible in the Games list; pilot ~2.4:1, mech ~3.0:1, crawler ~3.7:1, all below AA. `ReferenceEntityCard` solves this by resolving `titleTextClass` per tone rather than hardcoding paper.
- `apps/itun/src/components/roster/Roster.tsx:237` — `linkSegment`'s cross-link swapped a `Badge` with `max-w-full truncate` for a `Stat`, whose cell is `shrink-0 grow-0 w-fit whitespace-nowrap` with no `className`. A long linked pilot/mech name now can't shrink or ellipsise; it pushes the controls to their own line and gets clipped mid-word by the new inner `overflow-hidden` wrapper.
- `apps/itun/src/components/games/GameEntitySheet.tsx:82` — The comment asserts a non-member gets `null` from `listForGame`, but `convex/entities.ts:133` calls `requireMember`, which throws `NotAuthorized`; `useQuery` re-throws in render, so a non-member following a shared `/games/:id/view/...` link hits the root `errorComponent` instead of the notice — which also defeats the stated "must not be able to tell them apart" property, since members see the card and non-members see a crash screen.

Two notes that aren't findings: the new route's double-underscore opt-out, the `frozenSheet.ts` extraction, and the `SnapshotSheet` refactor are all behaviour-preserving as far as I can tell (`pilotAbilities` still only reaches mechs). Separately, an **untracked** scratch file `apps/itun/src/components/games/__tests__/captureGames.tsx` currently fails `bun run knip` locally — delete it or wire it up before pushing, or the pre-push gate will trip.

alxjrvs (commented):
Reviewed the full `origin/main...HEAD` diff (19 files: the `EntityRow` band/seal redesign, the new frozen crew-sheet surface + route, and the Roster/GameRoster stat migration). Working tree is clean, so the range diff was the whole scope.

Verification run: `bun run typecheck` (all 4 workspaces green), `bun --filter itun test` (1638 pass), `bun --filter component-lib test` (700 pass), `bun run lint`, `bun run knip` — all clean. I also rendered `EntityRow` in isolation to confirm the emitted markup for finding 1.

Findings:

- `packages/component-lib/src/components/shared/EntityRow.tsx:362` — the row title hardcodes `text-paper` instead of using `TONE[entityType].ink`, which the band itself already sets to `var(--color-ink)` for pilot/mech/game. Rendered output confirms `background: var(--color-sheet-game); color: var(--color-ink)` with a `text-paper` title span, i.e. #fbfaf7 on #7dceeb = 1.69:1 (pilot 2.40:1, mech 3.04:1). The file's own TONE docstring and the `EntityCardHeader` rule it claims to follow (`titleTextClass = 'text-ink'` by default) both say the opposite. Hits every row on Roster, GameRoster, Mediator and the sheet rails.
- `apps/itun/src/components/games/GameEntitySheet.tsx:136` — `Sheet` is rendered without `pilotAbilities`, so a crewmate's mech derives with `pilot: null` and loses pilot-sourced Max SP / Max Cargo (ADR-029). A Beefcake-piloted mech at 13/13 shows 10/10 (the `Math.min` clamp drags the current value down too). This is precisely what `Sheet.pilotAbilities` exists to prevent per its own docstring, and `entities.listForGame` — already fetched here — returns both `pilots` and `softLinks`.
- `packages/component-lib/src/components/shared/EntityRow.tsx:253` — the empty-row glyph switched from the deep tone to the light band tone while its `wash` backdrop was removed, dropping it from ~5.9:1 to 2.4:1 (pilot) / 1.7:1 (game) on paper. Low severity — the glyph is `aria-hidden` decorative — but it is a visible regression on all three live-sheet rails.

No issues found in the route wiring (`/games/$gameId/view/$kind/$entityId` generates correctly, the double `_` opt-out is right, `kind` is properly narrowed before it selects a Zod schema), the `frozenSheet.ts` extraction (behaviour-preserving; `parseContext` still supplies the snapshot's mech abilities), the store threading (`store` reaches `PartnerCard`/`MechSheet`/`PilotSheet`, so no global-store leak into a crewmate's sheet), or the `openSheet`/`claim` capability semantics.

alxjrvs (commented):
Reviewed `origin/main...HEAD` (6 commits: EntityRow band/seal redesign, Roster + GameRoster stat migration, new frozen crew-sheet route) plus the uncommitted working-tree changes (`bodyStats` split on `GameRow`, `mini` button rung). Baseline is green: `bun run typecheck`, `bun --filter itun test` (1638 pass), `bun --filter component-lib test` (700 pass), `bun run knip`, `bun run lint` all clean.

Findings:

- `packages/component-lib/src/components/shared/EntityRow.tsx:380` — The row title hardcodes `text-paper` on the new solid tone band while the same commit's `TONE` map defines `ink: var(--color-ink)` for pilot/mech/game (paper only for crawler) and applies it via `style={{ color: tone.ink }}`. The span overrides it, so a Game row's name is #fbfaf7 on #7dceeb (~1.7:1), pilot ~2.4:1, mech ~3.1:1 — all below AA. It read fine before only because the name sat on an ink stamp plate, which this change removed. `EntityCardHeader`, the rule cited in the comment, defaults `titleTextClass='text-ink'` and resolves per band.
- `apps/itun/src/components/games/GameEntitySheet.tsx:84` — `entities.listForGame` calls `requireMember`, which *throws*; `useQuery` rethrows in render. So the "not in this game / you may no longer be a member" notice is unreachable for a non-member, a stale link, or a malformed `gameId` — they get the root error page instead. The comment claims `null`. `GameScreen` sidesteps this by querying `games.get`, which returns `null` for non-members.
- `apps/itun/src/components/games/GameEntitySheet.tsx:136` — The frozen mech sheet is rendered without `pilotAbilities` and with `softLinks: []`, so `pilot` resolves null and pilot-sourced contributions drop out. This is verbatim the failure `Sheet`'s `pilotAbilities` prop was added to prevent (Beefcake's +3+X Max SP / +6 Cargo); `listForGame` already returns the soft-links and pilot bodies needed to supply them, and `SnapshotSheet` does pass them.
- `apps/itun/src/components/roster/Roster.tsx:239` — `value={detail ?? '—'}` misses the empty string. `resolveClassName('')` returns `''` (blank pilots get `classRef: ''` from `blankCreate.ts:72`), so the cross-link Stat renders `NAME | ` with an empty value cell. The chassis and crawler-TL paths return `undefined` and are fine.

alxjrvs (commented):
Reviewed `origin/main...HEAD` (7 commits, 21 files: `EntityRow` redesign in component-lib, the new read-only Game crew sheet + route in itun, roster/game row restructuring). Verified with `bun run typecheck` (clean), `bun --filter itun test` (1638 pass), `bun --filter component-lib test` (700 pass), `bun run lint` and `bun run knip` (clean) — so nothing here is caught by CI.

Findings, most severe first:

- `packages/component-lib/src/components/shared/EntityRow.tsx:380` — Row title hardcodes `text-paper` while the component's own `TONE` table says `ink` for pilot/mech/game and applies it to the band's `color`. Measured contrast against paper `#fbfaf7`: game `1.69:1`, pilot `2.41:1`, mech `3.05:1`, crawler `3.68:1` — all fail AA's 4.5:1 for 16px bold. `EntityCardHeader`, cited as the source of the rule, defaults `titleTextClass` to `text-ink`; only the reference card's deeper tone ramp uses paper.
- `apps/itun/src/components/games/GameEntitySheet.tsx:84` — The "non-member gets `null`" comment is wrong: `listForGame` calls `requireMember`, which throws `NotAuthorized`, and `useQuery` re-throws during render. The "you may no longer be a member" notice at line 103 is unreachable for exactly that case; the route errors out instead. Same for a `$gameId` that fails the `v.id('games')` validator.
- `apps/itun/src/components/games/GameEntitySheet.tsx:136` — A crewmate's mech renders with no piloting context (no `pilotAbilities`, `softLinks: []`), so ADR-029 contributions (e.g. Beefcake → Max SP/Cargo) are lost and the crew view shows lower maxima than the owner's sheet. `SnapshotSheet` threads this deliberately; `listing.softLinks`/`listing.pilots` are already available here.
- `packages/component-lib/src/components/shared/EntityRow.tsx:313` — Filled variant drops `className` (empty variant applies it). Pre-existing, but this diff rewrote that `cn()`; `SheetPilot`/`SheetMech`/`SheetCrawler` pass `flex-[1_1_0%]` on every rail slot, so filled and empty slots size differently.
- `apps/itun/src/components/roster/Roster.tsx:237` — Cross-links now wrap a `Stat`, whose cell sets `cursor-default` and `shrink-0 whitespace-nowrap`; the link loses its pointer cursor, and long names are hard-clipped instead of truncating as the old `Badge` did.
- `apps/itun/src/components/games/GameRow.tsx:60` — `Role` collapsed to a single value drops "Organizer" for a member who is both Mediator and Organizer (the ordinary path for whoever created the game); the old caption showed `Mediator · Organizer`.

alxjrvs (commented):
Reviewed the 8-commit range `main...HEAD` (EntityRow re-anatomy + ITUN Game crew read-only sheet). Typecheck, `bun --filter itun test` (1638 pass), `bun --filter component-lib test` (700 pass) and `knip` are all green, so nothing here is caught by the existing suite.

Findings:

- `packages/component-lib/src/components/shared/EntityRow.tsx:380` — The row title hardcodes `text-paper` while `TONE[entityType].ink` (computed and applied as the band's inline `color`) is `var(--color-ink)` for pilot/mech/game. Result: white on #7dceeb (≈1.8:1), #ef894f (≈2.5:1), #7a978a (≈2.8:1) — all below WCAG AA, and contrary to `EntityCardHeader`'s `titleTextClass` rule the doc comment cites. `tone.ink` is now effectively dead.
- `apps/itun/src/components/games/GameEntitySheet.tsx:136` — `Sheet` is rendered without `pilotAbilities`, so a crewmate's mech loses ADR-029 pilot contributions: Max SP short by 3+X and Cargo short by 6 for a Beefcake pilot, versus what the owner sees. `SnapshotSheet` threads exactly this to avoid the hazard Sheet.tsx:100-109 documents; `listForGame` already returns the pilot bodies.
- `apps/itun/src/components/games/GameEntitySheet.tsx:136` — With `softLinks: []`, the crew sheet's Linked Units rail states "No mech assigned" / "No crawler linked" / "No pilot assigned" even when the game listing (which returns `softLinks`) says otherwise. A read-only surface asserting a positive falsehood about somebody else's build.
- `apps/itun/src/components/games/GameEntitySheet.tsx:94` — `listing` is guarded only for `undefined` on the comment's claim that a non-member gets `null`; `requireMember` throws, so `useQuery` rethrows and a removed member hits an error boundary rather than the "not in this game" notice written for that case.
- `apps/itun/src/components/roster/Roster.tsx:237` — `linkSegment` swapped a `max-w-full truncate` Badge for a `Stat` cell that is `shrink-0 grow-0 whitespace-nowrap` with no max width; inside EntityRow's `overflow-hidden` wrapper a long cross-link is clipped mid-word with no ellipsis. The cell's `cursor-default` also removes the link affordance.
- `apps/itun/src/components/roster/Roster.tsx:107` — `pilotStats` emits `CALLSIGN | x` unconditionally; `GameRoster.statsFor` suppresses it when it equals the name, so a pilot named "Ghost" with callsign "Ghost" shows the string twice on the Roster but once on the crew roster.
- `packages/component-lib/src/components/shared/EntityRow.tsx:269` — The empty variant's glyph moved from `tone.rail` (deep) to `tone.band` (base), so it is now a pale tone drawn on paper (≈2.1:1) rather than a band fill.

alxjrvs (commented):
Reviewed the 8-commit branch (`main...HEAD`, ~2k diff lines across `component-lib/EntityRow`, the ITUN Roster/Games surfaces, the new frozen crew-sheet route, plus a dep bump). Verified against the repo's own gates first: `bun run typecheck`, `bun --filter itun test` (1638 pass), `bun --filter component-lib test` (700 pass), `lint` and `knip` are all green, so nothing here is a test-visible break — the findings are behaviours no test asserts. I also rendered `EntityRow` directly to confirm the emitted markup rather than reasoning about the class merge.

Findings:

- `packages/component-lib/src/components/shared/EntityRow.tsx:380` — The row title hardcodes `text-paper` on all four ontology bands while the component's own `TONE[*].ink` map resolves to `--color-ink` for pilot/mech/game. Confirmed by rendering: the band emits `color: var(--color-ink)` and the title span overrides it. Contrast is ~1.69:1 for game (`#7dceeb`), ~2.41:1 for pilot, ~3.04:1 for mech; only crawler (~3.67:1) is the case paper text suits. Every game name on the Games index is effectively invisible. The cited precedent, `EntityCardHeader`, actually defaults `titleTextClass` to `text-ink` and has callers resolve it per tone — so this is not the card rule the comment claims to follow.
- `apps/itun/src/components/games/GameEntitySheet.tsx:90` — The "not in this game / no longer a member" notice is unreachable. `entities.listForGame` calls `requireMember`, which throws `NotAuthorized`, and Convex `useQuery` re-throws during render, so `listing` is never `null`. A removed member or a stale shared link gets the root `errorComponent` page instead of the notice with its back-to-crew link.
- `apps/itun/src/components/games/GameEntitySheet.tsx:142` — `Sheet` is rendered without `pilotAbilities`, so a crewmate's mech loses pilot-sourced contributions (ADR-029, e.g. Beefcake's Max SP/Cargo) and reads lower than on the owner's sheet. This is the exact bug `SnapshotSheet` passes `pilotAbilities` to avoid, and `listForGame` already returns the `softLinks` needed to resolve it.
- `apps/itun/src/components/roster/Roster.tsx:966` — The cross-link chips now render a `Stat`, whose horizontal cell hardcodes `cursor-default`, overriding the wrapping `AppLink`'s pointer. The old `Badge` set no cursor class, so it inherited the link cursor.

One thing I considered and did not report: `GameRow.standingStats` collapses `Mediator · Organizer` to a single `Role` cell, so an organizer who is also mediator loses the organizer fact — that reads as deliberate given the one-cell framing in the comment, not a defect.

alxjrvs (commented):
Review complete. Scope: `git diff main...HEAD` — the EntityRow band/body re-composition in `component-lib`, the new read-only crew sheet (`GameEntitySheet` + `frozenSheet.ts` extracted from `SnapshotSheet`), the GameRoster/Roster/GameRow row-content rework, the shared `PAGE` shell, and the dependency-override commits.

Verification run: `bun --filter component-lib test` (700 pass), `bun --filter itun test` (1638 pass), `bun run typecheck` (all workspaces clean). Nothing below is a test failure — these are things the tests don't assert.

Findings, most severe first:

- `packages/component-lib/src/components/shared/EntityRow.tsx:380` — Row title hard-codes `text-paper` on the light `--color-sheet-*` band: ~1.77:1 contrast for game, 2.5:1 pilot, 3.2:1 mech, and it overrides the `TONE[].ink` value the same file declares three lines up. Affects every row in the Roster, Games index, crew roster and live-sheet rails.
- `apps/itun/src/components/games/GameEntitySheet.tsx:136` — The crew sheet doesn't pass `pilotAbilities`, so a crewmate's mech loses its ADR-029 pilot contributions; `Math.min(currentSP, maxSP)` then clamps the current value too, so a 15/15 Beefcake mech reads 12/12. This is verbatim the failure `SnapshotSheet` passes that prop to prevent.
- `packages/component-lib/src/components/shared/EntityRow.tsx:386` — Band stats render through `Stat`, whose cell is `shrink-0 whitespace-nowrap` with no max-width, so a long crawler name / chassis / callsign is hard-clipped by the inner `overflow-hidden` with no ellipsis. The Badges it replaced carried `max-w-full truncate`.
- `apps/itun/src/components/games/GameEntitySheet.tsx:84` — Comment claims a non-member "gets `null` here", but `requireMember` throws and Convex's `useQuery` re-throws in render → the root error component, not the tailored "not in this game" notice. That notice's stated purpose is unreachable for the case it was written for.
- `apps/itun/src/components/games/GameRoster.tsx:495` — All row actions moved `compact` → `mini` (~23px vs ~31px tall), putting Edit/Dashboard/Offer/Scrap and EntityRow's View/Delete below the 24×24 minimum touch target.
- `packages/component-lib/src/components/shared/EntityRow.tsx:269` — Empty-row glyph repainted from the deep tone to the light band tone on paper (~1.8:1 for game); theme.css:229-234 still cites the old `tone.rail`-on-wash measurement as the reason `--color-sheet-game-deep` holds its literal value.
- `packages/component-lib/src/components/shared/EntityRow.tsx:321` — Uses `border-[length:var(--bw-entity)]` instead of the canonical `border-entity`, which theme.css declares the one border vocabulary and `cn()` registers specifically to merge safely.
- `packages/component-lib/src/components/shared/EntityRow.tsx:137` — The `seal` prop doc ("the band reserves the width for it") directly contradicts the implementation comment at line 343 ("no right reserve") and the code.
- `apps/itun/src/components/games/GamesScreen.tsx:249` — Games index went full-bleed with no inner cap; its content is a single column plus forms, and `JoinScreen` in the same commit added `max-w-xl` for exactly this reason.

alxjrvs (commented):
Reviewed the 7-commit branch (`main...HEAD`): EntityRow band/seal rewrite in component-lib, the Game crew read-only sheet + new route, roster/GameRoster stat restructuring, and the shared `frozenSheet.ts` extraction. `bun run typecheck`, `bun run test` (all workspaces), `bun run lint` and `bun run knip` all pass on the branch, so everything below was found by reading, plus one live render of `EntityRow` to confirm the emitted markup.

Findings (most severe first):

- `apps/itun/src/components/games/GameEntitySheet.tsx:84` — `api.entities.listForGame` calls `requireMember`, which **throws** `NotAuthorized`; it never returns `null`. `convex/react`'s `useQuery` re-throws query errors during render, so a non-member (or a viewer who left the game) opening a shared `/games/<id>/view/<kind>/<id>` link hits the root `errorComponent` ("Something went wrong / Reload app") instead of the `GameSheetNotice` with the way back to the crew. The file's own comment (lines 86-89) and the `row === null` branch (line 104) both assume the `null` path. `GameScreen` guards this correctly by gating on `api.games.get` before mounting `GameRoster`; the new route has no equivalent guard.
- `packages/component-lib/src/components/shared/EntityRow.tsx:380` — the title hardcodes `text-paper` while the band sets `color: tone.ink` (ink for pilot/mech/game). Rendered output confirms white text on `--color-sheet-game` `#7dceeb` ≈ 1.69:1 contrast (pilot 2.41, mech 3.04, crawler 3.68) at `text-base` bold. The name becomes the only low-contrast element on a row where every Stat carries its own ink/paper plate, and `TONE[*].ink` ends up with no consumer.
- `packages/component-lib/src/components/shared/EntityRow.tsx:318` — the filled variant never applies `className` (the empty variant does, and the union type lets callers pass it). `SheetPilot`/`SheetMech`/`SheetCrawler` pass `flex-[1_1_0%]` to balance the Linked Units rail; it reaches the DOM only for empty slots, so a filled+empty rail is unbalanced. Pre-existing, but this hunk is exactly where it was rewritten.
- `packages/component-lib/src/components/shared/EntityRow.tsx:269` — removing `TONE.rail` re-tinted the empty-row ontology glyph from the deep tone to the light band tone while it still sits on paper (game ≈1.7:1, pilot ≈2.4:1). The message text in the same hunk was correctly moved to `text-ink`; the glyph was not.

PR #673, opened 2026-08-04
fix(deps): bump fast-uri and brace-expansion past the new high advisories

alxjrvs:
## Why

Every open PR's `audit` job was failing, which blocks the aggregate CI gate:

```
fast-uri  >=4.0.0 <4.1.2
  workspace:itun › vite-plugin-pwa
  workspace:srd › @astrojs/check
  high: fast-uri vulnerable to host confusion via backslash authority introducer
        GHSA-7p8r-x3mc-p8w7

brace-expansion  >=4.0.0 <5.0.9
  workspace:itun › @sentry/vite-plugin
  workspace:itun › vite-plugin-pwa
  high: brace-expansion: DoS via unbounded intermediate arrays, bypassing the
        CVE-2026-14257 mitigation — GHSA-rgw5-rvv9-x895
```

Both packages are transitive dev dependencies that the root `overrides` block
already pins to *exact* versions (added in #466 to clear earlier advisories).
Because the pins are exact, the new advisories — which land one patch above each
pin — could not resolve on their own: the override was actively holding both
packages on the vulnerable version.

## What

Bump the two pins to the patched releases:

| package | was | now | fixes |
| --- | --- | --- | --- |
| `fast-uri` | `4.1.1` | `4.1.2` | GHSA-7p8r-x3mc-p8w7 |
| `brace-expansion` | `5.0.8` | `5.0.9` | GHSA-rgw5-rvv9-x895 |

`bun.lock` is regenerated accordingly. No other dependency moves — the diff is
2 lines of `package.json` and the matching lockfile resolutions.

## Verification

- `bun audit --audit-level=high` — clean, 0 vulnerabilities (was 2 high)
- `bun run typecheck` — 0 errors across all workspaces
- `bun run test` — 3340 tests pass, 0 fail
- `bun run build` (package + srd + itun + bot) and `bun run build:itun` — all
  succeed; both bumps sit under build tooling, so the builds are the real check

## Note (not actioned)

This is the second time exact pins in `overrides` have themselves become the
thing blocking CI. A `>=`/caret range (e.g. `"fast-uri": "^4.1.2"`) would let
future patch fixes resolve without a manual bump, at the cost of the exact
determinism the current pins give. Left as-is to match the existing convention —
happy to change it if you'd prefer the self-healing shape.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -38,7 +38,7 @@
     },
     "apps/itun": {
       "name": "itun",
-      "version": "0.7.0",
+      "version": "0.10.0",
       "dependencies": {
         "@auth/core": "0.41.3",
         "@base-ui/react": "1.6.0",
@@ -76,7 +76,7 @@
     },
     "apps/srd": {
       "name": "srd",
-      "version": "1.3.0",
+      "version": "1.4.0",
       "dependencies": {
         "@astrojs/check": "0.9.10",
         "@astrojs/react": "^6.0.1",
@@ -142,7 +142,7 @@
     },
     "packages/salvageunion-reference": {
       "name": "salvageunion-reference",
-      "version": "2.7.0",
+      "version": "2.9.0",
       "dependencies": {
         "jsonc-parser": "^3.3.1",
         "zod": "^4.4.3",
@@ -153,8 +153,8 @@
     "@ladle/react@5.1.1": "patches/@ladle%2Freact@5.1.1.patch",
   },
   "overrides": {
-    "brace-expansion": "5.0.8",
-    "fast-uri": "4.1.1",
+    "brace-expansion": "5.0.9",
+    "fast-uri": "4.1.2",
     "filelist": "2.0.2",
     "js-yaml": "4.3.0",
     "postcss": "8.5.22",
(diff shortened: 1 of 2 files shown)


alxjrvs (commented):
## Review result

The bump itself is correct and lands cleanly — no defect in the two changed version strings. All four findings are about the `overrides` block the PR edits; findings 1, 2 and 4 are pre-existing conditions on adjacent lines that this PR's mechanism directly implicates.

**Findings**

- `package.json:12` — `undici` override pins `6.27.0`, inside three open advisories' `<6.28.0` range; every declarer asks `^6.27.0`, so the pin is the only reason patched `6.28.0` isn't resolved. All moderate, so CI's `--audit-level=high` gate stays green.
- `package.json:20` — `postcss` override pins `8.5.22`, the exact top of GHSA-fxqj-rqcc-2cmp's `<=8.5.22` range; declarers (`^8.5.3`…`^8.5.17`) would otherwise resolve the unaffected `8.5.25`. The pin now does the opposite of what a2752e46 added it for.
- `package.json:19` — `fast-uri` is declared only by `ajv@8.20.0` as `^3.0.1`; the override forces `4.1.2` across a major with no semver gate, though `3.1.5` patches the same GHSA and satisfies the range. Verified concretely: `resolve()` gained a `throw` in 4.1.2 (`resolve('https://ex.com/a/b', '/\evil.com/x')` returns a string on 4.1.1, throws on 4.1.2) and ajv's `resolveUrl` has no try/catch. (3.1.5 throws identically, so this specific change isn't avoidable — the point is the missing semver gate for the next one.)
- `package.json:16` — `brace-expansion` override is redundant: the sole declarer is `minimatch@10.2.5` at `^5.0.5`, which resolves to the patched `5.0.9` unaided. The pin buys nothing and guarantees a manual bump PR per advisory.

**Suggested action:** merge as-is (it clears the high gate), then separately consider dropping the `fast-uri` and `brace-expansion` pins and re-bumping `undici` → `6.28.0` and `postcss` → `8.5.25`.

PR #674, opened 2026-08-04
fix(deps): make the audit overrides security floors, not frozen pins

alxjrvs:
Follow-up to #673. Two items, both flagged there and both about the same thing:
the `audit` gate is fixed, but the *shape* that let it break is still in place.

## 1. Exact overrides are the bug, not the fix

The `overrides` entries that exist to hold a transitive dep above a vulnerable
range were written as exact versions. An exact override pins the package **down
as well as up**. So when the next advisory is fixed one patch above the pin, the
override itself becomes the thing holding the vulnerable version in place — and
no `bun update` can clear it, because the override wins. The only escape is a
human editing `package.json`.

That is not hypothetical. It is why `audit` has blocked every open PR twice:

| | package | bump | why |
| --- | --- | --- | --- |
| #565 | `brace-expansion` | 5.0.7 → 5.0.8 | GHSA-mh99-v99m-4gvg; *"blocking every PR"* |
| #673 | `brace-expansion` | 5.0.8 → 5.0.9 | GHSA-rgw5-rvv9-x895; blocked 5 of 6 open PRs |
| #673 | `fast-uri` | 4.1.1 → 4.1.2 | GHSA-7p8r-x3mc-p8w7 |

Same package, same failure, two hand-bumps.

**Measured rather than assumed** — a scratch project depending on `ajv`, with only
the override shape varied:

```
override "4.1.1"  (exact)  -> installs 4.1.1;  bun update fast-uri -> still 4.1.1
override "^4.1.1" (caret)  -> installs 4.1.2   (the patched release)
```

With carets in place, the advisory that prompted #673 would never have reached a PR.

Dependabot does not rescue this: it bumps *manifest ranges*, not the `overrides`
block, so an exact floor is invisible to it.

## 2. The `audit` job comment was actively wrong

It asserted the lockfile clears everything and *"NO package.json `overrides` are
required"*. False, and load-bearing in the wrong direction: several high
advisories are cleared **only** by these overrides, so a reader trusting the
comment would not think to look there — exactly the wrong place to send whoever
is next unblocking CI at 11pm. Rewritten to document the floor mechanism, the
`sharp` exception, and why a floor must never be frozen.

## What changed

Eight transitive security floors → caret ranges. **`sharp` deliberately stays
exact**: unlike the other eight it is a real pinned dependency mirrored in the
root and `apps/srd` devDependencies, so it is a compatibility pin, not a
security floor — and a range there risks splitting it from its direct pin.

## Risk: none measurable

This moves **no resolutions**. `bun install` reports `no changes`, all nine
packages resolve byte-identically, and the entire `bun.lock` delta is the
overrides block echoing the new ranges. Reproducibility is unaffected —
the committed lockfile still pins exact versions, and CI runs
`--frozen-lockfile`.

## Verification

`bun install --frozen-lockfile`, `bun audit --audit-level=high`, `typecheck`,
`test`, `build` (package + srd + itun + bot), `lint`, `format:check`, `knip`,
`validate:all` — all exit 0. `ci.yml` re-parsed and the `audit` job's `needs`
and `run` confirmed unchanged.


alxjrvs (commented):
Reviewed `main...HEAD` (2 commits: the fast-uri/brace-expansion bump + the exact-pin → caret-floor conversion). Diff is 3 files: `package.json` overrides, the regenerated `bun.lock`, and a rewritten explanatory comment on the `audit` job in `.github/workflows/ci.yml`. No application code.

What I verified as healthy: `bun install --frozen-lockfile --dry-run` passes (lockfile and package.json are consistent — every resolved version satisfies its new caret range), and `bun audit --audit-level=high` exits 0, so the PR does not break CI. The `sharp` exception the comment describes is accurate (it really is mirrored in root and `apps/srd` devDependencies).

The problem is that the PR's central premise — the thing it writes into CI as the standing maintenance procedure — does not hold under bun 1.3.14. I tested it against isolated scratch copies of the manifest + lockfile.

Findings (4):

- `.github/workflows/ci.yml:255` — The documented fix "`bun update <pkg>` plus the regenerated bun.lock. The floors are caret ranges so that works" does not work. bun does not re-resolve `overrides` entries: `bun update --lockfile-only` gave a zero-line lockfile diff and even `bun update --latest --lockfile-only` left `postcss@8.5.22` and `undici@6.27.0` in place, though 8.5.25 and 6.28.0 exist and satisfy the new floors (direct devDeps did bump — biome 2.5.5 → 2.5.6 — so the resolver was working). The per-package form is worse: `bun update postcss undici` left the resolutions unchanged and appended a new root `"dependencies"` block, in this repo's own tree with the *exact* versions `"postcss": "8.5.22", "undici": "6.27.0"` — recreating the frozen pin the PR exists to abolish, in a second location. The only move that actually raises a floor is still hand-editing the override and re-running `bun install`.
- `package.json:20` — The comment claims the floors are "held above the vulnerable range", but two of eight sit *inside* it: `bun audit` flags `postcss <=8.5.22` (GHSA-fxqj-rqcc-2cmp) and `undici <6.28.0` (three GHSAs). Moderate today so the high gate passes, but postcss's predecessor advisory was the `high` that motivated the 8.5.22 pin originally; combined with the finding above, these will not self-heal.
- `package.json:12` — Caret floors still freeze the major. `undici ^6.27.0` (latest 8.10.0) and `js-yaml ^4.3.0` (latest 5.2.3) cannot reach a fix that lands in the next major, reproducing exactly the brace-expansion stall the comment declares designed out; the unconditional "Raise the floor, never freeze it" gives no signal that case needs a hand edit.
- `.github/workflows/ci.yml:251` — Retained claim that the residual esbuild advisory comes from "the vite-7-pinned srd subtree" is misleading: `apps/srd/package.json` declares no vite at all, and `bun audit` traces esbuild `>=0.27.3 <0.28.1` to `srd › astro`, `itun › convex`, and `itun › vite` (in-range copies are convex's 0.27.0 and the 0.27.7 under `@vitejs/plugin-react-swc/vite@7.3.5` and `@vite-pwa/astro/astro@6.4.6/vite@7.3.5`).

The worktree is clean — I restored `package.json`/`bun.lock` from HEAD and removed all scratch dirs after the probes.

alxjrvs (commented):
**Scope note:** local `main` was stale — `a9064583` (#673) is already in `origin/main`, so the real diff is the two commits `d74aac1c` + `fda2c148`: eight `overrides` entries exact → caret (sharp deliberately left exact), the matching `bun.lock` overrides block, and a rewritten audit-job comment in `.github/workflows/ci.yml`.

**Mechanism verified as sound:** I confirmed by scratch install that a caret override does resolve to the highest matching version (override `^2.0.0` on `brace-expansion` → 2.1.4, not 2.0.0), `bun install --frozen-lockfile` still passes, `bun audit --audit-level=high` is green, and the `sharp` override is applied everywhere with no escaped nested copy (`undici@6.24.1` under `node_modules/.bun` is a stale store artifact; the live symlink points at 6.27.0). All findings are in the comment's factual claims and the shipped lockfile, not the override mechanism itself.

Findings:

- `.github/workflows/ci.yml:258` — Documented remedy is wrong: `bun update <pkg>` is a no-op for an override-only floor (measured: lock at 2.0.1 under `^2.0.0` stays at 2.0.1; only `bun update --latest <pkg>` moves it to 2.1.4).
- `.github/workflows/ci.yml:248` — sharp advisory path misattributed to `workspace:srd -> astro`; measured, `astro@7.1.4` resolves sharp 0.35.3 and audits clean, while the vulnerable `astro@6.4.6` (sharp 0.34.5, GHSA-f88m-g3jw-g9cj) comes in via `@vite-pwa/astro`'s peer range. Checking the stated path makes the override look droppable.
- `package.json:12` — Caret floors still freeze inside a major: `undici ^6.27.0` caps at the legacy 6.x line (registry latest 8.10.0, `seven` 7.29.0), and `bun update --latest` honours the cap — so a 7.x/8.x-only fix would leave the override holding the vulnerable version, the exact failure the comment says carets eliminate. Same for `js-yaml ^4.3.0` vs latest 5.2.3.
- `package.json:12` — Floors ship below current fixed versions: lockfile still holds `undici 6.27.0` (3 moderates fixed in 6.28.0) and `postcss 8.5.22` (GHSA-fxqj-rqcc-2cmp, latest 8.5.25). Both fixes satisfy the new carets, but a plain `bun install` won't float an already-satisfying entry, so the floating-floor mechanism isn't realised in the lockfile this PR ships.
- `.github/workflows/ci.yml:254` — esbuild advisory scoped to "the vite-7-pinned srd subtree" but measured under three paths (`srd › astro`, `itun › convex`, `itun › vite`); the low-severity conclusion holds, the attribution doesn't.

PR #676, opened 2026-08-04
docs(ci): correct the audit comment with measured advisory facts

alxjrvs:
Comment-only follow-up to #674. No workflow behaviour changes.

Reviewing #674's own comment against reality turned up two inaccuracies — the
same failure mode #674 exists to prevent: a durable comment asserting something
nobody had checked. Both are now measured rather than assumed, by deleting the
`overrides` block, re-resolving, and re-running `bun audit --audit-level=high`:

**1. "several high advisories currently depend on them"** — vague, and high.
Exactly **two** come back:

| package | advisory | reached via |
| --- | --- | --- |
| `sharp` | GHSA-f88m-g3jw-g9cj | `workspace:srd` → `astro` |
| `undici` | GHSA-vxpw-j846-p89q | `discord.js` |

The remaining floors do not bind on a fresh resolve today. They're kept as cheap
insurance against a parent range drifting back down — which is now what the
comment says, instead of implying they're all load-bearing.

**2. `sharp` was described as a compatibility pin and "the one exception"** —
implying it does no security work. It does: its override is exactly what clears
GHSA-f88m-g3jw-g9cj on the transitive `astro → sharp` path. A reader trusting
the old wording could have dropped it as redundant and quietly reintroduced a
high advisory. It's now documented as **both** — kept exact because it mirrors
the root and `apps/srd` devDependency pins, and load-bearing for security
regardless — along with the cost of that choice: a future sharp advisory needs
all three pins bumped by hand.

## Verification

`ci.yml` re-parsed (17 jobs; the `audit` job's `needs` and `run` unchanged),
`bun install --frozen-lockfile`, `bun audit --audit-level=high`, and
`format:check` all exit 0. The scratch experiment's repo state was restored and
confirmed clean before committing.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -244,12 +244,15 @@ jobs:
     # HOW ADVISORIES GET CLEARED HERE. Most resolve on their own when the
     # lockfile moves. The rest are held above the vulnerable range by the
     # `overrides` block in the root package.json — those entries are SECURITY
-    # FLOORS, not compatibility pins, and several high advisories currently
-    # depend on them. (`sharp` is the one exception: it is a real pinned
-    # dependency mirrored in the root and apps/srd devDependencies, so it stays
-    # exact.) No `--ignore` is needed: the formerly ignored GHSA-gv7w-rqvm-qjhr
-    # left the resolved lockfile with the vite 8 upgrade, and the residual
-    # esbuild advisory from the vite-7-pinned srd subtree is `low`, below this
+    # FLOORS, and the block is load-bearing: delete it, re-resolve, and two high
+    # advisories come straight back — `sharp` (GHSA-f88m-g3jw-g9cj, reached via
+    # workspace:srd -> astro) and `undici` (GHSA-vxpw-j846-p89q, via discord.js).
+    # The other floors do not bind on a fresh resolve today; they are kept as
+    # cheap insurance against a parent range drifting back down.
+    #
+    # No `--ignore` is needed: the formerly ignored GHSA-gv7w-rqvm-qjhr left the
+    # resolved lockfile with the vite 8 upgrade, and the residual esbuild
+    # advisory from the vite-7-pinned srd subtree is `low`, below this
     # `--audit-level=high` gate.
     #
     # WHEN THIS JOB FAILS ON A TRANSITIVE DEP, the fix is `bun update <pkg>`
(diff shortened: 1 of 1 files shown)


alxjrvs (commented):
Reviewed `98b227b0` (comment-only change to the `audit` job in `.github/workflows/ci.yml`). Since the diff is a durable factual comment, I verified every new claim empirically: fetched both GHSA advisories from the GitHub API, and re-ran `bun install` + `bun audit` in three throwaway copies of the manifests (overrides removed with the lockfile kept, overrides removed with the lockfile deleted, and sharp floated to `^0.35.0`). The working tree was never touched.

What holds up: both GHSA IDs are real and `high`; the undici attribution to discord.js is correct (`@discordjs/ws > @discordjs/rest` pins `undici: 6.24.1` exactly, and the `^6.27.0` override is what hoists it past the `<6.27.0` range); "the other floors do not bind" is accurate at the `high` gate; "all three pins" is the right count; baseline audit is 0 high / 8 moderate / 2 low, so the gate genuinely passes today.

Findings:

- `.github/workflows/ci.yml:248` — sharp advisory blamed on the wrong astro subtree. srd's own `astro ^7.1.4` resolves the safe sharp 0.35.3; the vulnerable copy is `@vite-pwa/astro > astro@6.4.6 > sharp@0.34.5`. The comment reproduces `bun audit`'s collapsed `workspace:srd > astro` label, so the comment's own `bun update <pkg>` remedy points at a dependency that cannot fix it.
- `.github/workflows/ci.yml:270` — the exact sharp override is justified by a false necessity. `"sharp": "^0.35.0"` clears the advisory identically (verified: single sharp@0.35.3, audit exits 0), since bun overrides replace the declared range wholesale. Keeping it exact re-creates the brace-expansion pin-down failure the same comment documents 10 lines earlier, on every transitive path.
- `.github/workflows/ci.yml:247` — the stated reproduction recipe only reproduces one way. Keeping bun.lock gives the 2 high advisories; deleting bun.lock gives "No vulnerabilities found" (undici floats to 6.28.0, sharp stays 0.35.3). Line 250 then says "on a fresh resolve", steering readers to the procedure that disproves line 247.
- `.github/workflows/ci.yml:255` — "the vite-7-pinned srd subtree" is wrong on both counts: nothing in the repo pins vite 7, and the vulnerable esbuild 0.27.7 copies sit under `@vite-pwa/astro > astro@6.4.6 > vite@7.3.5` *and* under itun's `@vitejs/plugin-react-swc > vite@7.3.5`.

Findings 1 and 4 are the same failure mode this commit exists to fix — a dependency path asserted rather than traced through the lockfile. Fixing 1 and 4 is a wording change; finding 2 is worth a decision (float the sharp override to `^0.35.0`, which is a one-line package.json change that keeps the audit green).

PR #678, opened 2026-08-04
fix(itun): give the Game panels a proper card header

alxjrvs:
The Game surface spoke two dialects: banded, toned entity rows above — and Invite / Downtime / Awaiting-your-answer stating their names as a small grey stamp inside an untoned slab below. Those three now wear a proper card header.

The band is **ink**, not the game ontology's blue. That's the convention ITUN's other panels already follow (`StorageManifest`, the encounter cards), and it keeps a panel visually distinct from an entity row rather than competing with one.

## What this PR no longer contains

It originally moved every card and row header onto the `-deep` tone so their paper-white titles would clear AA. **That is reverted in full** — cards and rows are exactly as they ship today. Reverted after seeing it rendered: the measured case was sound, but four ontologies of dark bands across 1039 SRD pages and every ITUN row is a different proposition from a contrast table, and it was growing a tail of consequences rather than settling (the frame border stayed on the base tone while the band moved; the tech-level ramp cards didn't darken at all).

## Contrast is still unfixed — the measurements, for whoever picks it up

| | pilot | mech | crawler | game |
|---|---|---|---|---|
| paper on tone (today) | 2.41 ✗ | 3.04 | 3.68 | 1.69 ✗ |
| ink on tone | 6.38 AA | 5.05 AA | 4.18 | 9.08 AA |
| paper on `-deep` | 5.17 AA | 10.16 AA | 8.49 AA | 8.12 AA |

**No white fixes this.** Contrast against a light ground improves as the text *darkens*, so a "safer white" is a contradiction in terms — white on game blue caps near 1.7:1 whatever shade you pick. Either the band darkens or the title goes ink.

Ink-on-light is the smaller change and clears AA everywhere except crawler (4.18, short for a 16px bold title). Deep-band clears everything but is the largest visual change in the codebase. That's a design call worth making deliberately, not as a side effect of a panel fix.

Local suite green.


alxjrvs (commented):
Reviewed `main...HEAD` (2 commits: the ITUN crew-sheet/row work plus the header-band tone change). Verified against the repo: `bun run typecheck`, `bun run lint`, `bun run knip`, `bun --filter component-lib test` (703 pass), `bun --filter itun test` (1638 pass) are all green, so every finding below is something the suite does not catch.

The dominant theme is the last commit (`bandSurface`): darkening every header band to the deep tone was applied to `Card` and `EntityCardHeader`, but three things that were tuned against the *base* tone were not moved with it.

Findings:

- `packages/component-lib/src/components/referenceEntity/card/ReferenceEntityCard.tsx:519` — `onBandText` is still `text-ink` for damaged/destroyed and ghosted cards on the premise that those bands are light; `bandSurface` now darkens them 65% toward black. A damaged card's title goes from ~4.8:1 to ~2.4:1 (grey band #918D88 → #5E5C58 under #282019 ink) — below the 4.5:1 floor this change cites as its justification. Ghosted cards also end up with a header band darker than their sub-header, inverting the ramp.
- `packages/component-lib/src/components/referenceEntity/card/ReferenceEntityCard.tsx:1996` — `Card` stepped its sub-header down to `accentDeeperColor` to stay distinct from the deep header, but `ReferenceEntityCard`'s `darkTone` (line 520) is still `accentDeepColor(tone.bg, tone.bgColor)` — the identical string `bandSurface` now paints on the header. Header, sub-header and footer collapse into one flat block on every SRD entity card: exactly the merge the Card.tsx comment says it fixed.
- `apps/itun/src/components/games/GameEntitySheet.tsx:169` — the new crew view renders a crewmate's mech without `pilotAbilities`, so pilot-sourced Max SP / Cargo contributions vanish (the failure `Sheet`'s own `pilotAbilities` docstring was written to prevent). The mech reads lower than on its owner's sheet; `listing.pilots` is already in hand, so the refs are recoverable.
- `packages/component-lib/src/components/shared/ModalShell.tsx:108` — the non-danger close (×) uses `text-ink/60`, picked against base pilot orange; on the darkened band its contrast drops to ~1.9:1, under the 3:1 non-text floor. `isDanger` modals are fine (already `text-paper/60`).
- `packages/component-lib/src/components/referenceEntity/card/ReferenceEntityCard.tsx:996` — the `small`+`head` badge shortform still fills with `accentSurface` (base tone) under `onBandText` paper-white: the exact 2.41:1 pairing `bandSurface` exists to remove, and it is commented as matching a header it no longer matches.
- `packages/component-lib/src/components/shared/__tests__/Card.test.tsx:209` — the only assertion that `Card` paints its header band became `expect(headerRowAround('Disabled')).toBeTruthy()`. Deleting the `band.style` spread from Card would leave every test green.

Things I checked and cleared: the new `/games/$gameId/view/$kind/$entityId` route (double-underscore escaping and `isRosterKind` narrowing are correct); `frozenSheet.ts` extraction is behaviour-preserving vs the old `parseSnapshot`/`makeSnapshotStore` (including the `pilotAbilities` move to `parseContext` at the call site); the frozen store genuinely blocks `adopt`; `EntityRow`'s `toNodes`/`hasBody` gating; `border-[length:var(--bw-entity)]` does compile under Tailwind 4.3.3 (I compiled it) and merges correctly through `cn()` — though `border-entity` is the registered canonical utility and theme.css says "never `border-[1.5px]`", so that is a convention deviation, not a bug. `SheetSectionCard` is safe from the `bg-[var(--tone)]` → `var(--color-[var(--tone)])` rewrite trap only because it passes `headerBgColor` explicitly; any future caller passing an arbitrary-value `headerBg` without it will now get a transparent band rather than the class fallback.

alxjrvs (commented):
Reviewed `main...HEAD` (3 commits: crew read-only sheet + EntityRow/Card band redesign). Verification run locally: `bun run typecheck`, `bun --filter component-lib test` (705 pass), `bun --filter itun test` (1638 pass), `bun run lint`, `bun run knip` — all green, so every finding below is reasoned from the code rather than a failing check.

Findings, most severe first:

- `packages/component-lib/src/components/shared/ModalShell.tsx:108` — Card's band now paints the DEEP fill (`color-mix(--color-pilot 65%, black)` ≈ #9b5933), but the non-danger modal close ✕ is still `text-ink/60` (#282019 @60%), chosen for the light `bg-pilot` band. That drops from ~4:1 to ~1.6:1 — the only mouse dismiss affordance becomes effectively invisible in every non-danger ModalShell.
- `apps/itun/src/components/games/GameEntitySheet.tsx:136` — the new crew sheet renders a crewmate's mech through the frozen store without `pilotAbilities`. `composition.pilot` is null (frozen store has no pilot, `softLinks: []`), so `SheetMech.tsx:55` resolves to `undefined` and pilot-sourced contributions (ADR-029 Beefcake +3+X Max SP, +6 Cargo) vanish — the mech reads lower than on its owner's sheet. `SnapshotSheet` passes `pilotAbilities` for exactly this reason; the new caller doesn't.
- `packages/component-lib/src/components/shared/Card.tsx:219` — `bandSurface` is applied unconditionally with no `bandDeep` opt-out, so `Callout`'s documented "always a pale tint" 12% band (`Callout.tsx:68`) is darkened to ≈ #939290. Every labelled Callout in the SRD renders a mid-dark slab. `ReferenceEntityCard` got the `bandDeep={false}` escape hatch; `Card` has none.
- `packages/component-lib/src/components/shared/EntityRow.tsx:401` (PLAUSIBLE) — band stats now carry free text (chassis/crawler/class names) through `Stat`, whose cell is `shrink-0 grow-0 whitespace-nowrap`. A value wider than a 390px row can neither wrap nor truncate and is clipped mid-word by the new inner `overflow-hidden` wrapper; the badges it replaced had `max-w-full truncate`.
- `apps/itun/src/components/roster/Roster.tsx:236` — the cross-link swapped `Badge` for `Stat` inside the `AppLink`; `Stat.tsx:354` hard-codes `cursor-default`, so the link now shows an arrow cursor. With `no-underline` on the anchor, nothing else signals it's clickable.
- `packages/component-lib/src/components/shared/EntityRow.tsx:327` — the filled row's root was rewritten but still never applies `className` (declared only on `EmptyEntityRowProps`, so the union's excess-property rule hides it). `SheetPilot`/`SheetMech`/`SheetCrawler` all pass `className="flex-[1_1_0%]"` for equal-width rail rows and it is discarded. Pre-existing, but this diff rewrote the exact element.
- `packages/component-lib/src/components/shared/__tests__/Card.test.tsx:216` — the band-tone assertion became `expect(...).toBeTruthy()`. Deleting `style={{...band.style}}` from `Card.tsx:339` now keeps the whole suite green; `bandSurface` is unit-tested but Card's use of it is not.

alxjrvs (commented):
Reviewed `git diff main...HEAD` on `itun-row-followup`. Note: **HEAD moved during the review** — a concurrent session landed `d3a24f3f "Revert the deep-band contrast change; keep only the Game panels"`, which backed out the three band commits (`949663a1`, `d3f69d82`, `faedf282`). Findings are against the post-revert HEAD. Typecheck, lint, and all tests (705 component-lib, 1638 itun) pass.

Findings:

- `packages/component-lib/src/components/shared/EntityRow.tsx:380` — The row title span hardcodes `text-paper`, but the revert put `TONE` back on the light base fills (`--color-sheet-pilot` #ef894f, `--color-sheet-mech` #7a978a, `--color-sheet-game` #7dceeb) and sets `color: tone.ink` = ink on the band wrapper. The span's own class beats the inherited colour, so every Roster / Games-lobby / crew row name paints white-on-light at 2.42:1, 3.05:1 and 1.70:1 against a 4.5:1 floor — the exact numbers the reverted commits cited. `tone.ink` is now dead for the title on every ontology. Either drop the `text-paper` class (let it inherit `tone.ink`) or restore the deep band.

- `apps/itun/src/components/games/GameEntitySheet.tsx:136` — The crew sheet renders a mech through `Sheet` without `pilotAbilities`. The frozen store carries `softLinks: []`, so the composition resolves `pilot: null` and pilot-sourced max-SP/cargo contributions (ADR-029) silently vanish — precisely the failure `SnapshotSheet` passes that prop to avoid, per Sheet.tsx's own docstring. A crewmate's mech reads a lower Max SP here than on its owner's sheet. `api.entities.listForGame`, already queried by this component, returns both `softLinks` and every pilot `body`.

- `apps/itun/src/components/games/GameRow.tsx:60` — The new `Role` stat is a short-circuit chain where the old caption was additive: `mediator ? 'Mediator' : organizer ? 'Organizer' : 'Player'`. A game's creator is normally both, so the common case drops "Organizer" entirely — the same flag that gates the Invite panel on the Game screen.

Things I checked and cleared: the `/games/$gameId/view/$kind/$entityId` route + generated tree, `isRosterKind` narrowing, `serverId` ↔ Convex `_id` addressing, the `frozenSheet.ts` extraction (parity with the old `parseSnapshot`/`makeSnapshotStore`, including `parseContext` still reaching mech snapshots), no `adopt`/write path out of the frozen store, no soft-link leakage from the viewer's own data, `Sheet`'s new `back` prop, and `Stat`'s self-coloured plates (so band stats stay legible regardless of band tone).

alxjrvs (commented):
## Review complete

**Scope:** The branch's committed range vs `origin/main` is 5 files / 36 lines (the earlier 24-file work is already merged as #672, and the `bandSurface` deep-band experiment was fully reverted in `d3a24f3f`). Working tree is clean, so no uncommitted changes to add.

**What the change does:** replaces the "small grey STAMP inside an untoned slab" title pattern with a real `Card` header band (`PANEL_TONE = 'bg-ink'` + `PANEL_TITLE`) on three Game panels — Invite, Downtime, Awaiting your answer — and adds those two constants to `gameChrome.ts`.

**Verification run:**
- `bun test src/components/games/` — 89 pass, 0 fail
- `bun run typecheck:itun` — clean
- `bunx biome check apps/itun/src/components/games/` — clean (no orphaned `STAMP`/`Text` imports; `STAMP` is still genuinely used in all three files' remaining call sites)
- `grep -rn bandSurface packages apps` — zero hits, so the revert is complete with no half-state left in `component-lib`
- Theme tokens `--color-ink` / `--color-paper` and the utilities `bg-ink`, `text-paper`, `font-cond`, `tracking-caps-tight` all exist and are already emitted from `@source`-scanned files, so no Tailwind JIT gap
- The `cn()` twMerge hazard does not apply here — `PANEL_TITLE` goes onto a raw `<span className={...}>`, never through `cn()`

I found no correctness bugs — no inverted conditions, null derefs, missing awaits, or broken callers. One finding:

### Findings

- `apps/itun/src/components/games/MediatorScreen.tsx:295` — The ink band is applied to only three panels, so `DowntimePanel` now renders framed and ink-banded directly between `AlertBar` (line 207) and `NpcTray` (line 63), both of which still title themselves with a `SECTION` span inside an untoned, borderless `Card` (as do `PresenceList` at 248, `ProposeForm` at 124, and the `CrewVitals` card at 272). Because `Card` only paints a frame when `headerBg`/`headerBgColor`/`borderColor` is set (`Card.tsx:261`), Downtime alone also picks up a 3px ink border and a 6px doubled bottom border (`Card.tsx:271`, since it has no footer). Net effect: the "surface speaking two dialects" problem the commit message sets out to fix is now reproduced *within a single screen*, with two title scales side by side (`PANEL_TITLE` `text-base`/`tracking-caps-tight` vs `SECTION` `text-lg`/`tracking-caps-snug`).

This is a UI-consistency issue rather than a functional defect, and it may well be a deliberate scope boundary — the commit explicitly says "keep only the Game panels." Worth an explicit decision either way: either extend `PANEL_TONE`/`PANEL_TITLE` to the four Mediator-local panels, or leave Downtime as-is on the Mediator surface until that pass happens.

alxjrvs (commented):
Reviewed PR #678 (`itun-row-followup` → `main`, 6 commits, 25 files): the `EntityRow` rail/wash → header-band rewrite, the new read-only crew-sheet route, and the Game panel header bands.

Verification run: `bun run typecheck` (clean), `bun --filter itun test` (1639 pass / 0 fail), `bun run lint` (only pre-existing warnings), `bun run knip` (clean). So nothing here is caught by CI — all four findings are runtime/visual.

Findings:

- `packages/component-lib/src/components/shared/EntityRow.tsx:380` — The band title hardcodes `text-paper` while the same container applies `style={{color: tone.ink}}`, whose `ink` is `var(--color-ink)` for pilot/mech/game. Result: white 16px bold on game `#7dceeb` = 1.76:1, pilot `#ef894f` = 2.51:1, mech `#7a978a` = 3.18:1, crawler `#ce5898` = 3.83:1 — all under the 4.5:1 AA floor. This PR's own `gameChrome.ts` cites the game-blue figure as the reason its panels use `bg-ink`, and the `EntityCardHeader` pattern the comment claims to follow resolves `titleTextClass` per band rather than fixing it to paper.
- `apps/itun/src/components/games/GameEntitySheet.tsx:136` — `Sheet` is rendered without `pilotAbilities`. The frozen store has no pilot and no soft-links, so `composition.pilot` is null and `SheetMech.tsx:55` falls through to `undefined`; a crewmate's mech renders with pilot-sourced contributions missing (Beefcake +3+X Max SP, +6 Cargo). `Sheet`'s own prop doc records this exact failure mode as why `SnapshotSheet` passes it.
- `apps/itun/src/components/games/GameEntitySheet.tsx:103` — The "not in this game" branch is unreachable for a non-member: `entities.listForGame` calls `requireMember`, which throws, and Convex's `useQuery` rethrows during render, so the viewer hits `RootErrorComponent` instead of the notice with its back-to-crew link. `GameScreen` sidesteps this by gating on `games.get` (returns `null`); the new route queries unguarded.
- `packages/component-lib/src/components/shared/EntityRow.tsx:357` — Band stats are `whitespace-nowrap shrink-0` cells inside the new `overflow-hidden` wrapper, so a long `CALLSIGN`/`CHASSIS`/`FROM` value is clipped mid-word rather than ellipsised as the `truncate` captions it replaced were.

alxjrvs (commented):
Reviewed `main...HEAD` (7 commits, 26 files) — the EntityRow band redesign, the new read-only Game crew sheet route, and the Game panel header/heading pass. Repo gates are all green on the branch (`typecheck`, `lint`, `bun --filter itun test`, `bun --filter component-lib test` 700 pass, `knip`), so everything below is behaviour the suite doesn't cover.

Findings, most severe first:

- `packages/component-lib/src/components/shared/EntityRow.tsx:380` — Row title hard-codes `text-paper` instead of the `TONE[entityType].ink` the band container is already styled with. White lands on the three light tones: game #7dceeb ≈ 1.7:1, pilot #ef894f ≈ 2.5:1, mech #7a978a ≈ 3.2:1, all under AA 4.5:1. Hits every row on Roster, Games index, Game/Mediator roster and the sheet rails. Notably, this same PR's `gameChrome.ts` `PANEL_TONE` comment cites the identical 1.69:1 measurement as the reason the Game *panels* use `bg-ink` — the row does the thing the panel deliberately avoided.
- `apps/itun/src/components/games/GameEntitySheet.tsx:84` — `useQuery(api.entities.listForGame)` runs with no membership gate, but that query begins with `requireMember`, which throws; convex/react's `useQuery` rethrows during render. A non-member (or a player removed from the game) following `/games/$gameId/view/...` gets the root boundary's generic "Something went wrong / Reload app" page — reload loops, no way back to the crew — instead of the "not in this game" notice the code comments promise. `GameScreen` avoids this by gating on `api.games.get` (returns `null`); this route skips that. A malformed `$gameId` fails `v.id('games')` the same way.
- `apps/itun/src/components/games/GameEntitySheet.tsx:136` — The crew view renders a mech through the frozen store (`softLinks: []`, `composition.pilot === null`) but never passes `pilotAbilities`, so `pilotingContext` gets `undefined` and pilot-sourced contributions (Beefcake +3+X Max SP, +6 Cargo, ADR-029) vanish. A crewmate's mech reads lower than the same mech on its owner's sheet. `Sheet.tsx`'s own `pilotAbilities` doc describes this exact failure; `SnapshotSheet` avoids it via `parseContext`, and the crew's pilots are already in the same `listing` payload.
- `packages/component-lib/src/components/shared/EntityRow.tsx:386` — Name-valued stats promoted into the band render via `Stat`, whose cell is `shrink-0 grow-0 whitespace-nowrap`, inside an `overflow-hidden` wrapper (line 328). A long `CHASSIS | …` or `CRAWLER | …` in a ~320px roster column is clipped mid-word with no ellipsis, where the old caption truncated.
- `packages/component-lib/src/components/shared/EntityRow.tsx:313` — `className` is declared only on `EmptyEntityRowProps` and applied only in the empty branch; the filled branch's `cn(...)` drops it. TS accepts it (union excess-property check), so `SheetPilot`/`SheetMech`/`SheetCrawler` passing `className="flex-[1_1_0%]"` to filled rails silently get `flex: 0 1 auto`. Pre-existing on `main`, but this hunk rewrote the exact element.

PR #679, opened 2026-08-04
audit: land all six waves — correct the map, make CI see it, finish the migrations, close the Convex seam

alxjrvs:
Lands **all six waves** of the deep audit ([full report](https://claude.ai/code/artifact/73d41d56-049d-4931-b1df-f1a8e258f528)) as seven reviewable commits on one branch — 80 verified findings, 340 files, +20,963/−11,752. `check:all` exits 0 at every commit.

> Currently rebased on `itun-row-followup` (#678) at its request, so its commits appear in this diff until #678 merges.

## The thesis: the map had rotted, not the machine

When ADR-030 replaced local-first with a Convex backend, the ADR was correctly marked superseded and **nothing propagated outward**. `CLAUDE.md` loads into every agent session and had zero mentions of Convex or Sentry while asserting itun has "no auth, no backend" — against 17 Convex modules and Sentry in all three apps. Every guard in the repo was written to catch the exact incident that burned someone, in the exact syntax it wore, and each had since been routed around.

So the fixes come in dependency order, not severity order: correct the map, make CI able to see, then change code.

## Wave 1 — `docs:` correct the agent-facing map
Rewrote the auto-loading instruction layer (`CLAUDE.md`, `.claude/rules/*`, `.claude/agents/*`, `data-flow.md`) around the real ADR-030 seam. Purged 11 dead component-lib symbol names — the hand-written postmortem about *that exact failure mode* is preserved; it had recurred 25 lines below itself. ADR inventory 26→30 with the real supersession chains. Astro 5→7 across 10 files. `.mcp.json` now declares all four servers (Netlify, Sentry, Render, GitHub).

## Wave 2 — `fix(ci):` make CI see what it believed it saw
**Fixed a live production defect:** `cn()` was silently deleting `border-l-entity`, so the 3px left accent spine was missing from every `SheetSectionCard`. Fixed via a `BORDER_WEIGHTS` cross-product over all five side groups — closing the bug class, not the instance — plus a `cn.test.ts` that parses `theme.css` and asserts every rung survives a conflicting class.

`check:tokens`/`check:styling` now run in CI (were bypassable pre-push only). `tools/` tests wired in — **11 tests there had never executed**. `apps/su-assets` (a live internet-facing handler) got a tsconfig, tests, and a pinned path-traversal guard. **`check-doc-drift` extended 3→6 checks** so wave 1 cannot silently re-rot.

## Wave 3 — `refactor:` reclaim the confirmed-dead code
Deleted three dead rules modules, 6 unread Zod fields, 10 never-passed props, the vestigial `heroRef`/`renderHero`/`rail`/`pill` plumbing, 3 uncalled Convex functions. Root scripts 60→40.

## Wave 4 — `refactor:` one encoding, and a guard for the half-migration
The choice model was half-migrated: 21 of 88 choices carried both encodings while `resolveChoiceView` read only the legacy one — so the obvious cleanup silently drops the inferred trait while the card still renders fine.

Migrated **readers before data**, then proved it: a harness compared 8 derived verdicts across all 88 choices. First run 86/88 — the 2 mismatches were a **latent bug**, where the old reader mistook `constraints: {field:'techLevel'}`, a *filter*, for a required-pick count. Encoded the real intent, re-proved **88/88 identical**. Added a coexistence guard so a half-migration now fails loudly.

## Wave 5 — `fix(itun):` name the auth handshake, refuse blocked writes out loud
`canWrite` had zero consumers and 45 store writes were voided into unhandled rejections. Landed `authSettled` **first** (an unsettled session resolved to Solo and silently stopped mirroring), then gated writes behind one `write(fields, meta)` helper with a real refusal state. `ChangeMeta` is now required, not an optional defaulting to `manual/unknown`. Replaced 7 `ctx.db.get(<string> as Id<…>)` casts with `normalizeId`.

## Wave 6 — `refactor:` split the grab-bags, enforce the conventions
`utilities.ts` 1233→32-line barrel; `objects.ts` split; PilotSheet 848→522 and MechSheet 815→537 with derived-model/actions hooks extracted; shared wizard step machine. **`additionalSources` wired into the card footer** — 383 researched provenance records that were read by nothing are now visible. `--color-wk-muted` retoned (was failing AA at 4.39:1 on paper and 3.96:1 on ITUN's own background; now 5.78 / 5.22). Dashboard buttons got the canonical focus ring — 1306 lines of CSS styled `:hover` five times and `:focus` zero.

## Deliberately not done
- **31 Tailwind geometry-string assertions** — the audit proposed replacing them with computed-style checks, but happy-dom cannot compute Tailwind styles, so that replacement does not exist. Left in place.
- A handful of items the agents correctly refused, e.g. deleting `constraints` from two Modification choices, which `ReferenceEntityCard:575` reads for `scalesWithField` — removing it would have silently stopped the header showing scaled Tech Level.

## Verification
`typecheck`, `lint`, `format:check`, `test`, `knip`, `validate:all`, `check:audit`, `check:tokens`, `check:styling` — all exit 0 at every commit. Test totals ~4,600 pass / 0 fail, no skips. Generated artifacts confirmed drift-free against a fresh `build:package`.

https://claude.ai/code/session_01L3TXYa34UqxoWH3xMMtJaH


alxjrvs (commented):
Reviewed the 3-commit diff against `origin/main` (91 files: docs rewrites, CI wiring, dead-code removal). Baseline verification first: `bun run typecheck`, `bun run knip`, `bun run test`, `bun run build:package` (clean tree, so `check:schemas` is safe), `bun run validate:doc-drift`, and the new su-assets tests all pass. Every deleted export/prop I checked (`EntityGrid`, `singleRoll`, `optionPane`, `statusSubject`, `vitalsTitle`, `activeAbilityIds`, `linkClassName`, `heroRef`, `pill`, `listShelf`, `remove`, `patchCrawler`, `computeCargoCapacity`, `effect`/`requiredTraits`/`dependsOn`/`purchaseLink`/`scrapTechLevel`) is genuinely unreferenced, and the itun/tanstack doc rewrite's line-number citations (`AccountStrip.tsx:40`, `SignInControl.tsx:75`, `api.games.members`, `api.ownership.claim`, `DB_VERSION = 13`, the migration versions) all check out.

Findings (`file:line — summary`):

- `CLAUDE.md:180` — component-lib roster still names `EntityGrid`, whose barrel export this PR deletes; `docs/architecture/package-contracts.md:213` (edited in this PR) repeats it. Third occurrence of the drift the same bullet warns about.
- `tools/check-doc-drift.ts:594` — `HISTORICAL_MENTION` skips the whole block, so one "the former X is gone" aside disables symbol checking for every other name in that bullet. Proven with a fixture; it's why the `EntityGrid` drift above passes CI.
- `tools/check-doc-drift.ts:552` — `componentLibSourceNames()` accepts a file basename as proof the symbol exists, so an export deleted from a still-present file (exactly `EntityGrid.tsx`) is invisible to check 5.
- `packages/component-lib/src/styles/theme.css:246` — the new "EVERY animation token here is applied `motion-safe:`" law is false: both surviving `--animate-*` tokens are used bare at `NavDrawer.tsx:108` and `MobileSearchDialog.tsx:66`.
- `.github/workflows/ci.yml:80` — root `CLAUDE.md`, `README.md` and `.claude/**` are in no path filter, so a PR editing only the doc-drift checks' own inputs skips `validate` and reports green; only the merge queue's force-all catches it.
- `tools/run-coverage.ts:52` — `apps/su-assets` gained tests but isn't in the coverage `WORKSPACES` lists, so the new internet-facing endpoint is unratcheted.
- `packages/salvageunion-reference/lib/rules/types.ts:118` — cargo types still document the deleted `computeCargoCapacity`, and the stale comments are republished in the committed `etc/*.api.d.ts`.
- `docs/architecture/data-flow.md:135` — "schema + 17 function modules" overcounts (17 `.ts` files total incl. `schema.ts`/`auth.config.ts`), and contradicts CLAUDE.md's "17 modules" for the same directory.
- `package.json:81` — `check:printed-names` is wired into nothing (not `check:all`, `validate:all`, CI, or lefthook), and soft-skips with exit 0 when run by hand.

alxjrvs (commented):
## Review complete

Scope: `git diff origin/main...HEAD` — 5 commits (two component-lib band fixes + the three audit waves), 102 files. I verified mechanically: `typecheck` (all 6 workspaces incl. the new `su-assets`), full `test` (3350+ tests), `knip`, `lint`, `format:check`, `check:tokens`, `check:styling`, `validate:doc-drift`, `test:tools`, and `build:package` (generated tree stays clean, so the widened `check:schemas` glob is honest). All green.

I also confirmed the removals are safe: no caller remains for `statusSubject`, `vitalsTitle`, `defaultCollapsed`, `optionPane`, `singleRoll`, `linkClassName`, `EntityGrid`, `activeAbilityIds`, `animate-heat-pulse`, the three deleted Convex functions, the removed data fields (`requiredTraits`, `scrapTechLevel`, `contextFrom`/`dependsOn`, `purchaseLink`, `verifiedAgainst`), or any of the ~20 deleted package.json scripts.

### Findings (4)

- `packages/component-lib/src/components/referenceEntity/card/ReferenceEntityCard.tsx:520` — the header band now resolves to `accentDeepColor(tone.bg, tone.bgColor)`, which is the *same call* as `darkTone`, the colour already fed to `EntityCardSubHeader` (:2000) and `EntityCardIdentityFooter` (:2245). On every ordinary entity card the three bands paint identical colour and collapse into one flat block. `Card.tsx` got the corresponding fix in this same PR (new `accentDeeperColor`, sub-header stepped to 45% "to stay a distinct band rather than merging with it"); the other card shell did not.
- `packages/component-lib/src/components/chrome/Callout.tsx:68` — Callout passes a deliberately pale 12% tint as `headerBgColor`; Card now darkens whatever it is handed to 65% black, so every callout band (incl. every "When Damaged" block) goes from a whisper of accent over paper to a mid-dark plate. Callout has no opt-out — `bandDeep` exists only on `EntityCardHeader`, not on `Card`.
- `package.json:81` — the new `check:printed-names` script is invoked by nothing (not `check:all`, not ci.yml, not lefthook), so the guard stays manual — the same gap this PR closed for `check:tokens`/`check:styling`.
- `packages/component-lib/src/components/sheet/SheetHero.tsx:71` — `heroRef` was unpicked from LiveSheet and all three sheet bodies but left in place here with zero callers; knip can't see it because it's a prop. (Also: SheetSectionCard.tsx:70's comment "`headerBg` stays for the header's own fill" is now false.)

alxjrvs (commented):
I reviewed `origin/main...HEAD` (7 commits: the Game-panel band work plus the three audit waves — 96 files, ~9.7k insertions). I verified the branch green locally: `bun run typecheck`, `bun run knip`, `bun run test` (all workspaces + the new `test:tools`), `bun run validate:doc-drift`, and the new `apps/su-assets` suite all pass. The `EntityGrid`/`RollTable.singleRoll`/`heroRef`/`pill`/`statusSubject`/`vitalsTitle`/`defaultCollapsed`/`optionPane` removals and the removed Convex functions (`listShelf`, `remove`, `patchCrawler`) have no remaining callers; `salvageunion-reference` is `"private": true`, so the schema-field removals are not an external API break.

Findings:

- `.github/workflows/ci.yml:33` — the `code` path filter (line 92) omits `CLAUDE.md`, `README.md`, and `.claude/**`, so a docs-only PR touching exactly the files the new checks 4–6 read skips the `validate` job entirely and `quality-checks` still reports green. Same gap wave 2 closed for `test/**`.
- `docs/architecture/display-system.md:54` — still instructs "A grid of entity cards → `EntityGrid` / `EntityGridRow`" after this PR deleted `EntityGrid`; `CLAUDE.md:180` and `docs/architecture/package-contracts.md:213` also still list it as a barrel export.
- `tools/check-doc-drift.ts:600` — check 5 accepts a symbol if it's a barrel export *or* any file/dir basename under `src/`, so a removed export whose file survives is undetectable; `EntityGrid` in this very PR is the proof.
- `packages/salvageunion-reference/data/crawler-bays.json:1236` — the knip "delete by default" rule was applied to rules *data*: the False Flag bay's cost drops from "5 Scrap at TL6" to "5 Scrap", and `requiredTraits` (Salvaging/Rigging gating) is gone from 7 actions. These facts now exist only in the gitignored PDFs.
- `docs/architecture/data-flow.md:135` — "schema + 17 function modules" (18) contradicts `CLAUDE.md`'s "17 modules"; the directory has 17 top-level `.ts` files, 15 of them function modules.
- `packages/salvageunion-reference/lib/rules/types.ts:118` — `CargoParent`/`CargoCapacityResult` still documented as the parameter/return of the deleted `computeCargoCapacity`.
- `README.md:33` — still describes itun as "local-first", the claim wave 1 corrected in CLAUDE.md.

alxjrvs (commented):
All three sweeps are in. They add four new defects and sharpen two of mine. I verified the load-bearing claims directly.

## New — ranked

**1. Offline-start pins a signed-in user in `connecting` forever.** `apps/itun/src/lib/connection/connectionMode.ts:80` tests `authSettled` before `online`:

```
if (!inputs.authSettled) return 'connecting'
if (!inputs.signedIn) return 'solo'
return inputs.online ? 'connected' : 'disconnected'
```

I confirmed that ordering. The agent traced the other half through `node_modules`: `@convex-dev/auth` reports `isAuthenticated: true` straight off the localStorage JWT, but `ConvexProviderWithAuth` holds `isLoading: true` until *the server* confirms the token over the websocket. Offline that confirmation never arrives, so `authSettled` never flips and `online` is never consulted. Installed PWA, in a Game, opens with no network: no NOT CONNECTED banner (`shouldWarnDisconnected` is `disconnected`-only), `Sheet.tsx:149` computes `writesBlocked = !canWrite && !settling` = `false`, so the full sheet renders with every edit affordance live — and each tap throws `WritesBlockedOffline('settling')` and toasts "Still signing in… Try again in a moment," indefinitely. ADR-030's Disconnected contract is unreachable in exactly the case it was written for. Pre-PR this resolved to `solo` and writes landed locally. **This is the most serious finding of the review** — I didn't re-verify the `node_modules` trace, so confirm that before fixing, but the ordering bug is real on its face.

**2. Convex schema narrowing may block deploy.** `apps/itun/convex/schema.ts:314-316` narrows `softLinks.from.type`/`to.type`/`type` from `v.string()` to closed literal unions. The PR's own new hunk at `convex/entities.ts:437-455` documents that old `claimLocal` wrote `String(l.from.type ?? '')`. Convex validates existing documents on push, so any deployment holding a legacy `''` row rejects the deploy. No backfill exists on the branch. Unverified by me — check the prod deployment for such rows before pushing.

**3. The `writeAwait` silence is wider than I reported, and not a sub-second window.** It's ~9 handlers across both `pilotSheetActions.ts:114` and `mechSheetActions.ts:101`, all rejecting into a `void` at the JSX call site with no `unhandledrejection` handler anywhere in `apps/itun/src`. Combined with finding 1, a Disconnected user sits in this state for the whole session, not a handshake window.

**4. The rewritten e2e round-trip can only fail.** `apps/itun/e2e/wired-build-and-snapshot.e2e.ts:95-105`. Fully confirmed: `openSheetFor` (`e2e/_helpers.ts:404`) clicks an `a[href*="/sheet/"]` and waits for `/sheet/`, but `ShareSnapshotScreen` — the only thing rendering "Publish snapshot" — is mounted solely at `routes/sheet/$kind/$id_.share.tsx`. The old soft-skip always returned, hiding it; the new hard `expect(...).toBeVisible()` will fail the moment `E2E_BASE_URL` is set, and blame the backend for a missing navigation.

## Sharpened

**CI filter (my #3)** — also omits root `README.md`, and the agent traced the full green path: no filter matches → all outputs `false` → `validate`/`test`/`typecheck`/`knip`/`coverage` all skip → `quality-checks` (`if: always()`, ci.yml:479) treats skipped as pass and the PR merges. The PR added `test/**`, `bunfig.toml`, `patches/**` to `shared` for exactly this reason but missed the dirs its own new checks read.

**check 5 (my #5)** — worse than scope-limited. `tools/check-doc-drift.ts:539-566` unions barrel exports with *every file basename* under `component-lib/src/`, so dropping an export while leaving the file passes. This PR does that: `index.ts:65` went from `export { EntityGrid, EntityGridRow }` to `export { EntityGridRow }` while `shared/EntityGrid.tsx` remains — both confirmed. `CLAUDE.md:180`, `package-contracts.md:213` and `display-system.md:54` still tell agents to use `EntityGrid`, and the check prints green. The basename fallback blinds it to the commonest form of the drift it exists to catch.

**Minor:** `theme.css:221-224`'s new comment claims `--color-wk-faint` no longer carries text. `apps/srd/src/pages/discord.astro:191` still renders `<span class="px-1.5 text-wk-faint">·</span>` — confirmed, the only survivor. A false invariant with nothing mechanical enforcing it.

The reference package came back clean, as I relayed. Want me to fix any of these on this branch?

alxjrvs (commented):
17 findings reported, from five area reviewers plus my own pass.

**Shape of it.** Every gate on this branch is green — typecheck, all six workspaces' tests, `validate:all`, `knip`, `lint`, `format:check` — and a large fraction of the diff verified as genuinely equivalent under scrutiny: the `resolveRefs` three-index rewrite (0 key mismatches across every schema), the `statBonus` → `contributions` migration (all 9 records 1:1, every total unchanged), the `cn()` tailwind-merge registration (complete against every `theme.css` token, with a new test closing that bug class), all 14 `capsLabel` call sites class-identical, and the su-assets DI extraction byte-identical in guard, allowlist, headers and status codes. The 45 removed `afterEach(cleanup)` lines are covered by the new bunfig preload.

**The two that matter most are both self-concealing.** `connectionMode.ts:80` doesn't just mis-order a branch — the branch it shadows is the one this wave existed to build, and its exhaustive test suite can't see the bug because it injects `authSettled` directly rather than deriving it. `entities.ts:584` drops Starter Set crawler writes silently, and the PR's own test rewrite added `appId` to the seed fixtures, so the fixtures now describe a world where the bug can't happen. Findings 7–9 (three separate write paths that refuse without a toast) are the blast radius of the first one.

**The tooling cluster (3–6) is one story:** wave 2 built a doc-drift checker, then gated it behind a path filter that excludes the docs it checks, and gave two of its six checks block-wide keyword exemptions that disable them on the real repo's content. The reviewer confirmed each by injecting a fake symbol and watching the checker exit 0 — so this isn't inference, it's three reproductions.

Nothing here is a merge blocker I'd impose unilaterally, but I'd fix 1, 2, and 3 before this lands — 1 and 2 are user-visible silent failures in the features the wave shipped, and 3 means the next doc regression won't be caught either.

alxjrvs (commented):
All four area agents returned. The review's four findings hold; the agents add eight more, and I verified the three most consequential myself.

## New findings — must fix before merge

**1. `apps/itun/src/stores/entityStore.ts:488` — `transfer()` neither gates nor mirrors writes.** Verified directly: `create` (419), `update` (472) and `delete` (601) all call `requireWritableBackend()` and `mirrorEntityWrite()`; `transfer()` calls neither. Signed into a Game and Connected, stowing cargo to the crawler or hitting Scrap Mech (`components/dashboard/ActiveItemBand.tsx:226`) changes `scrapPool`/`cargoLots` in IndexedDB only and never reaches Convex — and since the crawler mirror is a field patch that is never resent, the divergence is permanent after reconnecting. In Disconnected it writes anyway: the exact fork `requireWritableBackend` exists to prevent. It predates this branch, but this branch newly routes cargo through `transfer(…, meta)` and claims to complete the write-blocking work.

**2. Three data fields deleted with no replacement encoding** (`packages/salvageunion-reference`). I confirmed each is zero-occurrence at HEAD, data files included:
- `scrapTechLevel` — `crawler-bays.json` now prices the Nanite Processing Bay at `{scrap: 5}` while its own prose four lines down says "5 **Tech 6** Scrap". Bio-Mech and Bio-Crafting Bay lost their Tech 1 qualifier. Scrap tech level is the whole economy; the number is unrecoverable without re-reading the PDF.
- `requiredTraits` — the Salvaging/Rigging gate on seven actions is gone. `traits.json` only partly covers it: the `rigging` text names Load and Mount/Dismount but not Patch Up or Repair. "Can this mech perform Repair?" is no longer answerable from data.
- `purchaseLink` / `verifiedAgainst` — ten curated publisher URLs and the `2026-06-10` reconciliation date, deleted. No runtime failure (nothing read them), but `version: "1.5"` survives while the date recording what it was verified against does not.

## New findings — worth fixing

**3. `apps/itun/src/components/shared/useEntityChoices.ts:90`** — verified: a bare `void storeState.update(…, LIVE_SHEET_MANUAL)`, missed by the ~45-site `runWrite` sweep. Choice picked on an equipment/system card while writes are blocked → unhandled rejection, selection silently reverts.

**4. `apps/itun/src/components/sheet/pilotSheetActions.ts:127`** (also `:262`, `mechSheetActions.ts:161,272`) — `void softWarnings.saveAnyway()` bypasses `runWrite` too. Same class as the review's finding 3, one more path: the ability chip un-checks on the next render with no explanation.

**5. `.github/workflows/ci.yml:92`** — the `code` path filter covers `apps/**`, `packages/**`, `tools/**`, `docs/**`, `shared` but not root `CLAUDE.md`, root `README.md`, or `.claude/**` — exactly the docs the new `check-doc-drift.ts` checks read. A PR editing only `.claude/rules/monorepo-patterns.md` matches no filter → `validate` skipped → `quality-checks` counts `skipped` as pass → the drift merges green. Fix: add those three to the filter.

**6. `lib/resolveChoiceView.ts:107`** (same at `lib/rules/choiceCatalog.ts:75`) — the resolver dropped its legacy-field readers, but `ChoiceSchema` still accepts `choiceOptions`/`schemaEntities`/`schema` with `source` optional, `validateParityLogic` only fails on *coexistence*, and component-lib's `choiceSelectionHelpers.ts:89` kept its fallback. A newly authored legacy-only choice renders and is selectable but applies no effect. All 88 current choices resolve identically, so it's an authoring hazard, not a live regression.

**7. `lib/rules/derivedStats.ts:296` → `packages/component-lib/src/components/stat/provenanceLines.ts:104`** — `installed` is now hard-coded `0` (contributions arrive via `sources`), but `summarizeBreakdown` still prints `base + installed + adjustment = derived`, which I confirmed ignores `sources` entirely. Any mech stat through that tooltip renders arithmetic that doesn't add up. Latent — its only caller (`PartnerCard.tsx:114`) feeds partner stats that populate `installed` themselves. Sibling `linesFromBreakdown` handles `sources` correctly.

## Resolved and cleared

- **The e2e job-rename risk is closed.** The ruleset check confirms `quality-checks` is the only required status check, so merging `smoke-web`/`bundle-budget-*` into `e2e-web`/`e2e-itun` strands nothing.
- **component-lib: no findings.** The `cn()` border cross-product was checked against every `@utility` in `theme.css`, `capsLabel` was compared class-for-class at all 11 call sites, `useChipDraft` is a faithful extraction, and all ten removed props have zero consumers in either app.

The two areas the agents cleared in depth — the `statBonus` → `contributions` migration and the `BaseModel` lazy indexes — were independently confirmed by a second pass. Findings 1, 3 and 4 plus the review's finding 3 are one fix: route those paths through `runWrite`, or give `writeAwait` a `.catch(reportWriteFailure)` where nothing sequences after the write.

alxjrvs (commented):
All five deep-dive reviews have now reported. They found substantially more than my own pass did — here's the consolidated supplement to the 4 findings already reported.

## High — silent, user-facing failure

**`apps/itun/src/lib/connection/connectionMode.ts:80`** — a signed-in offline user gets a permanently, silently read-only app. `resolveConnectionMode` checks `authSettled` before `online`, and `authSettled` is `!isLoading && !isRefreshing`. Convex only clears `isLoading` on a *server* auth confirmation, while `@convex-dev/auth` restores the JWT from localStorage with no network — so offline, the mode is `connecting` forever. In that state `writesAllowed` is false, but `shouldWarnDisconnected` is also false and `Sheet.tsx:149`'s read-only lozenge is suppressed: every edit affordance stays live and every tap fails with no toast. This is exactly ADR-030's `disconnected` state rendering as a healthy app. New tests cover `authSettled:false` but never together with `online:false`. I confirmed the ordering at that line.

**`packages/component-lib/src/components/shared/EntityRow.tsx:380`** — the redesigned row title hardcodes `text-paper` while the same component computes the correct on-band ink at line 215. Measured contrast: **game 1.69:1**, pilot 2.41, mech 3.04, crawler 3.68 — all fail AA for 16px bold. On `main` this was fine because the name was a white-on-ink stamp plate; the plate was removed in this PR and the class stayed. Affects `GameRow`, `GamesScreen`, `MediatorScreen`. I confirmed the unconditional `text-paper`.

**`apps/itun/src/stores/entityStore.ts:488`** — `transfer()` has neither `requireWritableBackend()` nor `mirrorEntityWrite()`. I verified this directly: the guard appears at lines 419, 472, 601 but nowhere in transfer's 488–577 body. `useCargo`/`usePartnerCargo` commit exclusively through it, so a Connected player's cargo move never reaches Convex and a Disconnected player's is accepted locally while every other write is refused. Pre-existing on `main`, but this PR newly documents it as covered — `lib/cargo/reportCargo.ts`'s header lists "a write the connection mode blocked" among the refusals it surfaces, which `transfer` can never produce.

## Medium — guards that don't guard

- **`.github/workflows/ci.yml:57`** — path filters gained `docs/**`, `tools/**` etc. but nothing matches root `CLAUDE.md`, root `README.md`, or `.claude/**`, which are precisely what `check-doc-drift.ts` checks 3/5/6 read. Verified concretely: a PR editing only root CLAUDE.md sets every filter false, skips `validate`, and reports green — while running the checker by hand fails.
- **`tools/check-doc-drift.ts:213`** — the `HISTORICAL_MENTION` escape hatch skips any *whole block* containing "no longer"/"former"/"previously". Root `CLAUDE.md:180` contains "names that no longer exist", so that entire bullet is never inspected. **Already live:** this PR deletes the `EntityGrid` export, and both `CLAUDE.md:180` and `package-contracts.md:213` still list it, unflagged.
- **`packages/salvageunion-reference/tools/validateReferencesLogic.ts:40`** (+ `validateTraitsLogic.ts:89`) — the validators dropped legacy `schemaEntities`/`choiceOptions` reading, but `schemas/objects/choices.ts:127` still accepts those fields and `choiceSelectionHelpers.ts:88` still renders from them. A legacy-only choice pointing at a nonexistent entity now validates clean and still renders as a clickable card; `findDoubleEncodings` misses it because it only fires on *coexistence*. Latent, not live — all current data is migrated.
- **`apps/itun/tsconfig.json:19`** — `convex/` is typechecked by nothing in CI, which is where this PR's entire `normalizeId` correctness argument lives. Pre-existing; one latent error exists at `convex/__tests__/botHttp.test.ts:72`. Adding `"convex/**/*"` is a one-line fix.
- **`apps/itun/e2e/wired-build-and-snapshot.e2e.ts:101`** — the de-soft-skipped test asserts a "Publish snapshot" button on the sheet, but that button only exists on `ShareSnapshotScreen`. A phantom pass swapped for a guaranteed fail the moment `E2E_BASE_URL` is set.

## Lower

- **`apps/itun/convex/schema.ts:311`** — tightening `softLinks.*.type` from `v.string()` to closed unions is deploy-blocking if any existing row violates it; the *old* `claimLocal` wrote `String(l.from.type ?? '')`, i.e. `""`. Worth checking dev/prod before merge. (`changeLog` fields are safe.)
- **`apps/su-assets/netlify/functions/asset.ts:46`** — `decodeURIComponent` throws on a malformed escape, so `GET /%zz.png` is a 500 on the one internet-facing endpoint. Pre-existing, but this PR added the key-guard suite that misses it.
- **`packages/salvageunion-reference/lib/search.ts:28`** — the now-public `isSchemaName` declares `id is SURefEnumSchemaName` but tests a 24-member subset of the 25-member enum; `'ability-tree-requirements'` returns false. No caller hits it today.
- **`apps/srd/src/lib/searchIndexBuild.ts:26`** — `TEXT_FIELDS` still lists `'effect'` after the package index dropped it, directly under the new comment claiming byte-identical parity. No data has that key.
- **`packages/salvageunion-reference/lib/schemas/entities.ts:503`** — `verifiedAgainst` and its one data value were deleted. Same category as `purchaseLink`, which commit `2ec8bc31` deliberately reverted one commit ago. Owner call.
- **Stale references** (I verified all three): `docs/architecture/discord-bot-game-client.md:230` still maps `/su shelf` → the deleted `entities.listShelf`; `lib/search.ts:22-23` still tells readers to keep two forks in sync that this PR deleted. Plus an orphaned comment at `EntityRow.test.tsx:6`, and `.claude/rules/testing-patterns.md` now contradicts ~25 component-lib files on `afterEach(cleanup)`.

## Confirmed clean (independent verification)

The agents ran deeper equivalence checks than I did and cleared the bulk of the diff: the `utilities.ts`/`objects.ts` splits are byte-identical body-for-body (66 and 46 symbols mechanically diffed); `resolveChoiceView`'s legacy→unified rewrite produces **0 behavioural diffs across all 88 choices**; the parity validator emits byte-identical findings; `derivedStats` arithmetic survives the `statBonus`→`contributions` migration 1:1; every Convex `normalizeId` change is a strict tightening with the identity check unchanged and still before the write; the three deleted mutations have zero callers; `cn.ts`'s `theme.css`-parsing test closes the recurring black-on-black class rather than the instance; all removed props and package scripts have zero references.

Nothing here is fixed — this is review output only. Want me to take any of it?

PR #680, opened 2026-08-04
fix(data): stop two actions rendering their content twice

alxjrvs:
## What

`Can't Stop, Won't Stop` and `Mech-Gyver` each rendered their rules text **twice** on the entity card. Both records carried an un-split "blob" paragraph *alongside* the split version of the same prose.

- **Can't Stop, Won't Stop** — p1 was the entire ability text; p2–p4 were that same text split into its natural paragraphs.
- **Mech-Gyver** — p2 ended with the complete text of p3.

## Why it happened

Both were introduced by de972e71 (`Break out Actions (#66)`), the migration that lifted action prose out of `abilities.json` into `actions.json`. For these two records it emitted *both* its un-split source blob *and* its paragraph-split output instead of choosing one.

The giveaway is the trait markup: the `Can't Stop, Won't Stop` blob kept its `[[Vulnerable]]` links while the split copies had degraded to plain `Vulnerable` — two different passes over the same text.

## The fix

Each blob is trimmed to the text that is genuinely its own (the leading flavour sentence), keeping the split paragraphs as the mechanics. This lands both records on the book's flavour-then-mechanics shape rather than deleting a paragraph outright. The `[[Vulnerable]]` links are restored on the two `Can't Stop, Won't Stop` paragraphs that had lost them.

The data change is 8 lines.

## Why nothing caught it

Nine validators existed and **none looked inside a single record's own `content` array**, so this sat undetected since November. This PR adds a `content-dupes` check to close that gap:

- Flags a paragraph that repeats another **of the same record** verbatim, or fully contains it.
- Compares with `[[Trait]]` markers stripped — load-bearing, since the blob kept its links while the copies had not, so a marker-sensitive comparison would have missed this entirely.
- Ignores paragraphs under 8 words, so short stock phrases don't produce noise.
- **Scoped to within-record duplication only.** Repetition *across* records is normal and heavily used in this corpus — the Tech 2–6 Pilot Bay tiers, the Crawler/Pilot/Mech movement rates on the map guides, Minor/Major injury healing. A cross-record check would be all false positives.

Wired into the unified runner (`validate:all`) and available standalone as `bun --filter salvageunion-reference validate:content-dupes`.

## Verification

- Run against the **pre-fix** data, the new check reports exactly these two records (4 paragraph diagnostics) and exits non-zero; clean and exit 0 after the fix.
- 12 new unit tests cover both dupe shapes, trait-marker insensitivity, the word floor, nested records, and the no-cross-record-false-positives guarantee — plus a regression test asserting the real corpus stays clean.
- A repo-wide fuzzy sweep (Jaccard ≥ 0.5 over paragraph shingles) found no other affected records; all 25 near-matches are legitimate parallel prose.
- `bun run check:all` green.

https://claude.ai/code/session_019MXFAPGYPk765YeLGhNTYT

### DIFF
--- a/packages/salvageunion-reference/data/actions.json
+++ b/packages/salvageunion-reference/data/actions.json
@@ -1564,7 +1564,7 @@
     "content": [
       {
         "type": "paragraph",
-        "value": "You have learned to use the momentum and bulk of the vast weight of cargo you can carry to crush your enemies and see them driven before you. When you activate this Ability, choose a target in Range. You make a special charge attack against the target. Roll to hit as normal. On a hit, this deals HP damage equal to the amount of Inventory Slots your Pilot has filled, for example, if you have 6 Inventory Slots filled, it deals 6 HP damage. Non-Bio-Titan or Meld creatures hit by this attack are knocked Prone and gain the [[Vulnerable]] Trait. If you use this Ability whilst piloting a Mech, it deals SP damage equal to the amount of Cargo Slots you currently have filled on your Mech, for example, if you have 10 Cargo Slots filled, the attack deals 10 SP damage. Any target hit by this attack is knocked Prone and gains the [[Vulnerable]] Trait."
+        "value": "You have learned to use the momentum and bulk of the vast weight of cargo you can carry to crush your enemies and see them driven before you."
       },
       {
         "type": "paragraph",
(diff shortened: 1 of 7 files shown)


alxjrvs (commented):
## Review summary

Scope was one commit (`bc6ae441`) against `origin/main` — local `main` was stale, so `main...HEAD` would have pulled in an already-merged PR; I used `origin/main...HEAD`.

**The core of the change is sound.** I verified:

- **No prose was lost by the data edits.** The trimmed blob in `Can't Stop, Won't Stop` is fully recovered by paragraphs 2–4 (and the `[[Vulnerable]]` markers were correctly restored on the split copies); the sentence removed from `Mech-Gyver` p2 is still present verbatim as p3.
- **The check actually catches the bug it was written for.** Replaying `runContentDupeCheck` against `origin/main`'s `actions.json` reports exactly the 4 duplications across the 2 fixed records, and 0 on the fixed corpus.
- **The check is genuinely wired into CI** (root `validate:all` → `tools/validate.ts` → the `validate` job in `ci.yml`).
- **Everything is green**: `validate:all` (11/11), `bun test` in the package (1004 pass), `typecheck`, `knip`, `lint`, `format:check`, and `build:package` produces no generated-artifact drift.
- I also fuzz-scanned the corpus for near-duplicates the strict containment rule would miss. The only hits were `Parasitic Membrane`'s list items, which repeat a clause by design — correctly not flagged. And there is no aliased `[[X|y]]` link syntax anywhere in the data, so `normalizeParagraph`'s marker stripping has no gap.

Findings are all in the diagnostic/test layer, not the detection logic:

- `packages/salvageunion-reference/tools/validate.ts:187` — the diagnostic `path` puts a 1-based paragraph number inside 0-based array-subscript notation (`content[1]` for what is `content[0]`), unlike every sibling check in the same file. Pre-fix, it would have pointed at the wrong, innocent paragraph.
- `packages/salvageunion-reference/tools/validateContentDupesLogic.ts:82` — `excerpt()` quotes the *normalized* (lowercased, marker-stripped) text, so the string in the error message can't be found in the data file. With the off-by-one above, the report gives no working way to locate the offending block.
- `packages/salvageunion-reference/tools/validateContentDupes.test.ts:74` — 'ignores content blocks with no string value' passes vacuously; its fixture leaves only one eligible paragraph, so deleting the `typeof value === 'string'` guard still passes.
- `packages/salvageunion-reference/tools/validateContentDupes.test.ts:120` — comment claims the known #66 regressions are "Listed explicitly", but the assertion is a bare `toEqual([])` with nothing listed; looks like a leftover from a draft.

PR #682, opened 2026-08-04
audit: land the consistency audit — 8 live defects, the enforcement gaps, and the Biome rules that were switched off

alxjrvs:
Lands the [consistency audit](https://claude.ai/code/artifact/04a84d2e-e461-44c5-8632-fd8fcba1eb77) as four reviewable commits. `check:all` exits 0 at the tip with a clean tree.

The audit asked one question — *where does this codebase do the same job in more than one way?* — and the answer turned out to include eight things that are wrong on real data, not eight things that are untidy.

## The four commits

**1 · `fix(ci)` — unblock verification, and let the compiler see the backend**

`bun run lint` failed on clean `main`. Biome discovers nested root configs independently of `files.includes`, and every agent worktree carries its own `biome.jsonc` with `"root": true` — so `check:all`, the command CLAUDE.md and the `/validate` skill both tell you to run, exited 1 for anyone who had used the repo's own worktree workflow. No gate could see it: CI checks out fresh, pre-commit passes explicit paths, pre-push doesn't run lint.

`apps/itun/convex/` was in **no tsconfig program** — 36 files including auth, ownership, invites, proposals and the bot HTTP seam, the server of record for accounts. knip claimed the directory and Biome linted it, so three tools disagreed about whether it belonged to the workspace. Adding it surfaced a genuine latent defect: codegen had registered `convex/__tests__/harness.ts` as a *deployable module*, and the harness imports `_generated/api`, so `api`'s mapped type referenced itself and cascaded 26 implicit-anys into `src/components/games/`. Cut by giving the two `_generated` loaders a non-literal specifier. `BOT_OPS`'s `readonly string[]` annotation turns out to be load-bearing for the same reason and is now documented as such.

**2 · `fix` — the six remaining live defects**

| | Symptom | Measured |
|---|---|---|
| D-3 | Pilot sheet and Mech sheet disagree on the same mech's Max SP | Mule (TL1) + Beefcake: rail **12**, sheet **16** |
| D-4 | One dropdown spells every schema two ways | "Crawler Bays" vs "crawler-bays", **27/27** schemas |
| D-5 | Two mutations skip the Zod parse `schema.ts` calls mandatory | bodies are `v.any()`, so it was the only validation |
| D-6 | Neither browser app can report an exception to Sentry | 0 capture verbs, 0 `componentDidCatch` repo-wide |
| D-7 | Dashboard prints raw slugs where it means names | `"mule"` instead of `"Mule · TL 1"` |
| D-8 | ITUN links to SRD pages that were never generated | `hasSRDPage('actions')` → true, link 404s |

D-6 is the one with the widest blast radius: catching an error is exactly what stops it reaching Sentry's global handlers, so every deliberately-caught error in both browser apps was structurally unreportable — including a failed mirror to the server of record, which means IndexedDB silently diverging from Convex.

**3 · `ci` — close the gaps that let it accumulate**

The defects are symptoms; these are the reasons nothing noticed. The sharpest: `check-doc-drift` **already implements** the check for "a doc naming a deleted component" — its doc set just omitted `packages/component-lib/CLAUDE.md` and `README.md`, which is exactly where eight dead symbols and a non-compiling import example had accumulated. Two lines. Also: `su-assets` (a live site serving artwork for both production sites) was outside the Bun-version drift check; `srd` was the only workspace not extending the repo base tsconfig, and so the only one missing four strictness flags.

Both new guards were verified by **deliberate failure** — inject a dead symbol / a wrong pin, confirm the check fails, revert, confirm it passes.

**4 · `style` — turn on the Biome enforcement that was available and switched off**

`organizeImports` was `"off"`. `useImportType` and `noDefaultExport` were prose-only. Total fallout from enabling all of them plus five more rules: **two real violations**. `noDefaultExport` reported zero — the convention was already honoured, which is why locking it in is free. The rest of the commit is the mechanical `biome check --write` sweep (672 files, import order and type-import separation only, no logic).

`biome.jsonc` already said this about itself, on the one rule that was enforced: *"That convention was asserted in prose and enforced by nothing, so it held only where a reviewer happened to look."*

## Deliberately not done — these are yours, not mine

- **Which muted token wins** (`ink-2` vs `wk-muted`). The current split tracks which app authored the file, not what the text does. Needs a design call.
- **Meta schemas**: `/schema/actions.json` exists while `/schema/actions/` does not. Whether to add the HTML pages or drop the JSON endpoints is a product decision — the guard is fixed, the asymmetry is untouched.
- **Whether the unwrapped srd islands** are a considered exemption or an oversight.
- **`ruleset.md` §5**: six of eleven "irreducible atoms" resolve to no symbol. If they are an ontology rather than components, the doc should say so.
- **`structurePoints` positive-vs-nonnegative** across sibling schemas — a semantic question about whether a stat block may legitimately be 0.

## One judgement call worth a second opinion

D-5 parses NPC bodies as `EncounterNpcSchema.partial()` with `name` required, not the full strict schema. The only caller sends `{ name }`, and a strict parse would reject every write the app makes today. That trades some validation for not breaking the Mediator's Add button — flagging it rather than burying it.

## Commit 5 · `fix(convex)` — stop deploying the test harness, at the source

Commit 1 cut the type cycle by routing two loaders through a non-literal specifier — a symptom fix. The cause was that codegen registers every file under `convex/` as a **deployable module**, and the test harness was one of them, so test code shipped to the backend.

`convex/__tests__/` → `apps/itun/test/convex/`. `convex/` now holds only the 19 deployable modules and contains no test file at all. With the cause gone the workaround is deleted and the loaders are plain imports again.

`_generated/api.d.ts` loses exactly the two lines naming the harness. That is a hand edit of a generated file, so: `convex codegen` requires a live deployment (it calls `api.convex.dev` even with `--dry-run`), but its output is a pure function of the contents of `convex/` — removing a file removes exactly its import line and its `fullApi` entry, so the next real codegen run reproduces this file. Unlike the reference package's schemas, `convex/_generated` has no CI drift check.

`BOT_OPS` keeps its widened `readonly string[]` — **checked, not assumed**: narrowing it still produces 203 errors even with the harness moved, because `botHttp` is itself a registered module and `BotOp` closes an equivalent cycle through `api` on its own.

200 Convex tests pass from the new location.

## Verification

`lint`, `format:check`, `typecheck`, `test`, `knip`, `validate:all`, `check:audit`, `check:tokens`, `check:styling`, `check:schemas` — all exit 0. D-3, D-4, D-7 and D-8 were each proved by running the real code and reading the numbers, not by inspection.

https://claude.ai/code/session_01Cvbp9qqHE2fgxwE6dnzBbN


alxjrvs (commented):
## Review complete — 4 findings

**Scope:** `main...HEAD` (4 commits, 776 files) plus current working-tree changes. The bulk (`a7871b86`) is a mechanical Biome sweep; I verified it is import reordering and type-import separation only — no CSS or side-effect-only imports moved, no JSX/`let`/parameter-assignment rewrites, and the component-lib barrel lost zero exports (only `schemaPluralLabel` added). The real logic lives in `ac259f4d`, `e1f26c5c`, `b5593701`.

**Verification I ran:** `typecheck` (all 5 workspaces green; confirmed 23 Convex files are now actually in the itun program via `tsc --listFiles`), `lint` (exit 0, warnings only), `format:check`, `knip`, full `test` (all suites 0 fail), `check-bun-version`, and a real `srd` build with a dummy DSN to inspect the Sentry chunk graph.

### Findings

- `apps/srd/src/lib/observability.ts:66` — the new browser `captureException` is a silent no-op until the async `@sentry/browser` chunk resolves, so an island that throws on first render (the exact case D-6 claims to fix) still reports nothing. Confirmed from built output that the layout script and `IslandErrorBoundary` share one module instance, so this is a load-timing race, not a two-instance bug. No test covers the path.
- `apps/itun/src/components/sheet/railStats.ts:68` — `mechRailItems` now threads the piloting context into SP, EP **and** Heat, while `SheetMech.tsx:56-58` threads it into SP only. Both `mechMaxEPParts`/`mechMaxHeatParts` consume it, so this re-creates the D-3 divergence for EP/Heat. Latent today (no piloted-mech EP/heat contributions and no `perTechLevel` EP/heat amounts in the data), but by construction the two surfaces disagree.
- `tools/check-doc-drift.ts:614` — `siblingExports` widened from the sibling barrel to every export under `salvageunion-reference/lib` (internals and test helpers included), which excuses any doc symbol colliding with a sibling internal. I confirmed the narrow form fails on exactly one real name (`SURefEntity`), so a targeted exemption would preserve the guard.
- `biome.jsonc:33` — `!.claude` is broader than the `.claude/worktrees` that actually triggers Biome's nested-root error; future lintable files under `.claude/` would be silently uncovered. No coverage lost today.

### Notes, not findings

- Two things I chased and cleared: `claimLocal`'s new strict `MechPatternSchema` parse cannot silently drop real patterns (the IndexedDB read path returns `salvageSchema` output with unknown keys already stripped, so stored records satisfy the strict schema); and `mediator.updateNpc`'s now-required `name` has no production caller.
- `apps/srd/tsconfig.json` swapping `astro/tsconfigs/strict` for the repo base loses nothing — I diffed against Astro's actual `base.json`; every setting it contributes is either in the repo base or restated, and `jsx: react-jsx` was already srd's own override.
- A **concurrent session is editing this worktree**: mid-review, the Convex tests moved from `convex/__tests__/` to `test/convex/`, `_generated/api.d.ts` lost its harness entry, and `knip.json` widened its project globs. That in-flight change is sound (it resolves the "not fixed here" item from `e1f26c5c` and removes the need for the dynamic-specifier hack). A transient `knip` failure I saw was an artifact of catching that edit half-applied — it passes now.
- `b5593701` is titled "close the enforcement gaps" but its `ci.yml` change is 27 lines of comments; CodeQL remains outside the aggregate gate, which the comment states explicitly as ACTION REQUIRED. Documented known state, not a regression.

alxjrvs (commented):
Reviewed all five commits on `worktree-consistency-audit-fixes` (779 files vs `main`). The bulk — commit `a7871b86` — is a mechanical Biome sweep; I verified it programmatically by stripping import statements from every changed `.ts`/`.tsx` before and after and diffing the remainder: only 4 files differ outside imports (two comment moves, one `useTemplate` fix, one added test, plus the regenerated `etc/*.api.d.ts`, which is correctly excluded from Biome and mirrors its sources). Side-effect/CSS imports were not hoisted, no barrel export was dropped (`component-lib/src/index.ts` gains exactly `schemaPluralLabel`), and the two re-ordered `export *` barrels have no name collisions.

The six defect fixes check out: `mechRailItems` now takes the piloting context at both remaining call sites and matches `mechSheetModel`'s derivation exactly; `hasSRDPage`/`staticPaths` both go through `getEntitySchemas()`, which is literally `schemas.filter(s => !s.meta)`; `schemaPluralLabel` matches the category-row label; `chassisMeta`'s new `resolveChassisRef` returns the full entity (name + techLevel) and the throw path is still caught; the new Convex `parseBody`/`loadOwnable` consolidation preserves the `normalizeId` guard, and the `mechPatterns` edge parse is safe because that store already strict-validates on write.

Findings (4, all low/medium — no correctness bug in app logic):

- `apps/itun/convex/_generated/api.d.ts:23` — the hand-edited generated file drops the deleted harness entry but never adds `model/entities`, the module this PR created; `model/bot` and `model/permissions` are both listed despite exporting no Convex functions, so the next `convex codegen`/`deploy` will rewrite this file. That also falsifies the "the next real codegen run produces this same file" justification that made the hand edit safe.
- `tools/check-doc-drift.ts:614` — check 5's sibling skip-list widened from 40 barrel exports to 422 whole-tree exports (and the component-lib side now counts test/story-only exports); `ChassisStats` already collides, so a deleted public component named in a doc could pass. Narrow the sibling set back to the barrel.
- `apps/srd/src/lib/observability.ts:67` — `captureException` no-ops until the awaited `@sentry/browser` chunk resolves, so a hydration-time island crash (exactly what the new `componentDidCatch` targets) still produces no signal. Buffer events or await the init promise.
- `apps/discord-bot/src/itun/types.ts:15` — four stale references to the moved `convex/__tests__/` path (also `gameRoster.test.ts:8`, `discord-bot-game-client.md:21`, `game-invites-and-membership-plan.md:39,484`).

One process note: the worktree was clean when I started, but a separate 157-file in-flight change (test-preload consolidation, new `src/lib/runWrite.ts` / `safeLocalStorage.ts`) appeared in the working tree partway through — another session appears to be editing this same worktree. I reviewed the committed PR only; that unrelated, still-moving work is not covered above (its `bunfig.toml` preload paths do at least resolve to the new `test/reference-preload.ts`).

alxjrvs (commented):
Both background sweeps came back. Summary of what they added to the four findings already reported.

**Biome-style commit (`a7871b86`, 672 files) — semantics-preserving.** The audit did a per-file import-binding/export-name signature diff and a runtime module-load-graph diff (type-only imports erased) across all 671 modified sources: identical everywhere except one test file. No side-effect import moved, no default export added or removed, no `export *` name collisions in the four reordered barrels (checked pairwise), and the one cycle it touched (`actionResolution.js` hoisted to first in `lib/index.ts`) is declarations-only, so the TDZ window is unchanged. The dangerous autofix rules (`useOptionalChain`, `useExhaustiveDependencies`, `noUselessFragments`, `noArrayIndexKey`, `noUnusedImports`) were never enabled, so those classes can't be present.

Three non-mechanical things in it, none behavioural:

- Two net-new tests for `schemaPluralLabel` landed in the style commit (`packages/component-lib/src/utils/__tests__/schemaLabels.test.ts`, plus a case in `useSearchCombobox.test.tsx`) but cover a fix from the *previous* commit `ac259f4d`. Confirmed via `git log --diff-filter=A`. Reverting the style sweep silently drops that regression coverage.
- `packages/salvageunion-reference/lib/rules/index.ts:164-206` — alphabetizing the type-export block left eight hand-written section comments glued to whatever specifier they originally led. Confirmed by reading it: `HeatCheckEffect`/`HeatCheckResult` now sit under `// Take Damage`, `SoftWarning*` under `// Scrap`, `Roll` under `// Heat Check`. Cosmetic, but the taxonomy will rot on the next addition.

**CI/tooling sweep — four items, two worth acting on:**

- `apps/itun/convex/_generated/api.d.ts` is stale, and the commit's premise for hand-editing it ("the next real codegen produces this same file") is false. `convex/model/entities.ts` was added in `ac259f4d`; the file has 0 references to it, while it *does* list `model/bot` and `model/permissions` — both of which export only plain helpers, proving codegen lists every module regardless. Confirmed. Zero runtime impact today; there's no CI drift check on `convex/_generated`, which is why nothing caught it.
- `tools/check-design-tokens.ts:141` — the new radius guard matches `px` only, while the sibling rule 6 lines down matches `(?:px|rem)`. Confirmed by reading both. `rounded-[0.375rem]` — the natural spelling for anyone copying Tailwind's `rounded-md` default — passes the check.
- `tools/check-doc-drift.ts:611` widened `siblingExports` from a barrel read to `exportedSymbolsUnder(.../lib)` — 40 names → 422 — without the commit message mentioning it. The stated fix still works (the eight target symbols still fail), but the guard now excuses any component-lib name that collides with any internal or test-only export in the sibling package.
- `packages/component-lib/src/test-hygiene.test.ts:98` — the regex that's meant to stop bare `afterEach(cleanup)` coming back matches only two spellings; `afterEach(() => { cleanup() })` (the more common form) and the `async` variants slip through.

The sweep also verified the things it cleared by running them: `!.claude` genuinely fixes the nested-root-config error (biome A/B against the 19-worktree checkout), the moved `apps/itun/test/convex/` is outside Convex's default function root so `convex deploy` won't bundle it, `apps/srd/tsconfig.json`'s two removed options are both inherited from the repo base, and the CI aggregate still enumerates all 14 jobs with no `continue-on-error`.

One series-level note: `bun run test` is red at two intermediate SHAs (`3fde8839` points `bunfig.toml` at `test/reference-preload.ts` before it exists; `a40f806a` adds a guard whose assertions all fail at that SHA). Only `8bfcf31e` repairs both. Harmless for a squash merge, but a `git bisect` or a revert of `8bfcf31e` fails for unrelated reasons.

Stale paths worth a one-line sweep: `apps/discord-bot/src/itun/types.ts:15` and three docs still cite `apps/itun/convex/__tests__/`; `docs/architecture/package-contracts.md:363,375` still instructs contributors to create the per-workspace `test/preload-reference.ts` that `8bfcf31e` deleted and the new guard would flag.

alxjrvs (commented):
Two real updates from this re-report, both worth acting on:

**1. The lead's `check-doc-drift` finding doesn't hold as a regression — my exclusion of it was correct.** Its scenario ("remove `EntitySearcher` from `src/index.ts`, leave `EntitySearcher.tsx` in place, the citation still passes") already passed on `main`: `known` there is `barrelExports(index.ts) ∪ componentLibSourceNames(root)`, and `componentLibSourceNames` is a recursive walk of every *file and directory name* under `src/`. So the file's mere existence already excused the symbol before this PR. What the PR adds (`exportedSymbolsUnder`) only newly excuses exports whose name differs from their filename — a marginal further loosening of an already-loose set, not the defeat of the guard. That's consistent with the CI reviewer independently clearing it. It stays out of the findings; the underlying looseness is pre-existing and worth a separate look if you want that guard to actually pin the barrel.

**2. The branch has moved since the review.** There are now 11 commits, not 10 — `c6b26121 fix(itun): stop a refused write claiming the action succeeded` landed after I reported. It adds an optional `onApplied` callback to `runWrite` so seven Dashboard handlers stop printing a success readout after a refused write, plus a discriminating test. **That commit is not covered by the review.**

It does touch one finding's anchor: `ActiveItemBand.tsx` shifted, so the EP/Heat divergence finding now reads `ActiveItemBand.tsx:374-375` (was 370-371). `SheetMech.tsx:57-58` is unchanged. The finding itself still stands — neither call site passes piloting context.

The other six findings are unaffected. If you want the new commit reviewed, say so and I'll cover it.

alxjrvs (commented):
The tooling/CI review is the last one back, and unlike the other three it found real issues — three of them. I spot-checked all three mechanically and they hold up:

**1. `biome.jsonc:80` — the newly-enabled `organizeImports` assist is enforced by nothing in CI.** Assists only run under `biome check`. CI runs `bun run lint` (`biome lint .`) and `bun run format:check` (`biome format .`), and `check:all` (package.json:82) contains no `biome check` either — I confirmed both. So the exact drift this PR set out to fix (73 of 977 files re-opening a bare-package import after their first relative one) can be reintroduced by any PR and merge green. The only enforcement is lefthook's `biome check --write` on *staged* files, bypassed by `--no-verify`, absent for a GitHub web edit, absent for anyone who never installed hooks — and pre-push doesn't run it. Relatedly, `lefthook.yml:9`'s new comment asserts "pre-push/CI still run the full check," which is not true of any current step. Fix: add `biome check .` to the lint job, or swap `lint`+`format:check` for `biome ci .`.

**2. `test/reference-preload.ts` landed in the one directory neither Biome nor any tsconfig covers.** `files.includes` is `apps/**`, `packages/**`, `tools/**` — root `test/` isn't in it (confirmed at biome.jsonc:19-22), and no tsc runs at the root. All three files it replaced were covered on both axes. So a type error or lint violation in this now-load-bearing file — imported by all six workspaces' test runs — is reported as clean by `typecheck`, `lint`, `format:check`, and CI.

**3. `tools/check-doc-drift.ts:614` — widening `siblingExports` from the reference barrel to every export under `lib/` grows a *suppression* set from 40 to 422 names.** The diff confirms the change (`barrelExports(…/lib/index.ts)` → `exportedSymbolsUnder(…/lib)`), and it's used at line 643 as `if (known.has(symbol) || siblingExports.has(symbol)) continue`. A live collision exists today: `ChassisStats` is exported from both packages, so deleting it from component-lib while docs still name it would report clean. The `known`-set widening on line 608 is defensible; this one didn't need to change to fix the eight dead symbols the PR was after.

It also confirmed several things worth knowing were sound: the `!.claude` line in biome.jsonc is load-bearing (removing it makes `biome lint .` hard-error against a nested worktree root config), `arbitrary-radius` is genuinely enforced despite being absent from the tokens baseline (`baseline[rule.id] ?? 0` defaults to the strictest setting), every `noDefaultExport` override glob resolves as intended, and all 78 `.rejects` assertions in the new Convex tests are awaited.

**Where that leaves the review:** all four area passes are in. Total is 8 findings — the 5 refused-write bugs from the full-diff pass, plus these 3 enforcement gaps. Nothing is fixed; I've only reviewed.

alxjrvs (commented):
Reviewed the full `main...HEAD` range (12 commits, ~895 files — most of it import-order churn, regenerated JSON schemas, and data rewrites). I ran `bun run typecheck` (clean) and `bun run test` (all workspaces green) to separate real defects from noise, then hand-read every non-mechanical hunk in `apps/itun/convex`, `apps/itun/src`, `apps/srd/src`, `apps/discord-bot/src`, `packages/component-lib`, `packages/salvageunion-reference`, `tools/`, CI and tsconfig/biome config.

Findings (6):

- `apps/itun/src/components/dashboard/ActiveItemBand.tsx:214` — `doSalvage`: the readout moved inside `if (patch)`, and `areaSalvageOutcome` returns `patch: null` for both the "nothing found" band and the `requiresPlayerChoice` jackpot. Those two outcomes now render no overlay at all, so the "pick a Damaged Chassis, System or Module at Tech N" instruction is silently dropped. Highest-severity item; not offline-only, it affects every player.
- `packages/component-lib/src/components/shared/ControlButtons.tsx:179` — disabled secondary segment is `bg-wk-muted text-wk-muted`: identical token for fg and bg (1:1 contrast), so `segmentText` is invisible. Collateral from the `ink-2` → `wk-muted` sweep; the only such pair in the repo.
- `apps/itun/src/components/dashboard/ActiveItemBand.tsx:249` — `doScrap` still sets the "N scrap recovered" prompt unconditionally after `runWrite`. Newly reachable because this PR adds `requireWritableBackend()` to `entityStore.transfer`.
- `apps/itun/src/components/dashboard/ActiveItemBand.tsx:153` — `craft` prints "it is in the Hold" unconditionally after `runWrite`.
- `apps/itun/src/components/dashboard/ActionsDeck.tsx:148` (also 166, 203-205) — `setActivated(true)` and the push readout fire regardless of whether the write landed.
- `docs/adrs/ADR-026-entity-card-design-rules.md:51` — still documents `choiceType: "freeform"`, deleted from the (now `.strict()`) `ChoiceSchema` and added to `LEGACY_CHOICE_FIELDS` by this same PR.

The last four are all the same class the PR's own `runWrite(..., onApplied)` extension was written to fix — three sites in these two files were converted, five were not. Everything else I checked held up: the `parseBody`/`loadOwnable` extraction in `convex/model/entities.ts` is faithful, no export was dropped from either package barrel (verified by set-diffing the API report and `component-lib/src/index.ts`), the `getByName`/`resolveRef`/`techLevelRank` index swaps are behaviour-preserving, the print-CSS consolidation loses no rules, `hasSRDPage` now correctly mirrors `getItemStaticPaths`, and the srd tsconfig rebase does not drop `noImplicitReturns`/`noUncheckedIndexedAccess` (the repo base supplies both).

alxjrvs (commented):
Fourth background reviewer (CI, tooling, config) has finished. Its three findings map to the reported #7 (biome `organizeImports` assist), #5 (stale Convex codegen), and #11 (`package-contracts.md`), each with the empirical verification behind them — e.g. `bunx biome lint <file>` emits no organizeImports diagnostic while `bunx biome check <file>` does, so the rule is enforced only by the local pre-commit hook.

Two additions worth noting:

- The `package-contracts.md` drift has siblings. Four more files still cite the old `convex/__tests__/…` harness path: `apps/discord-bot/src/itun/types.ts:15`, `docs/design/game-invites-and-membership-plan.md:39,484`, and `docs/architecture/discord-bot-game-client.md:21`. `check-doc-drift` doesn't cover paths quoted in code fences, so none of it is caught by the guard this PR extends.
- **`.github/workflows/codeql.yml` is currently modified in this worktree and isn't from this reviewer.** Another agent appears to have edited it in the shared worktree mid-review. Worth checking `git status` before you act on the diff — the review's scope snapshot may no longer match the tree.

It also independently confirmed several things as sound: `!.claude` in `biome.jsonc` is load-bearing (tested by planting a fake nested config), the new `arbitrary-radius` design-token rule is fully enforced rather than silently baselined at 0, `check-doc-drift` check 5 still catches dead symbols, the preload works across the symlink boundary, no Convex test was dropped in the harness move (13 files / 200 tests, identical per-file counts vs `main`), and itun coverage sits at 89.96% against an 89 baseline.

All four background reviewers are now done; the 11 findings stand as reported.

alxjrvs (commented):
All three area agents have reported. Consolidated against my own pass: **two new confirmed findings**, four duplicates of what I already listed, and **one false positive I rejected**.

## New findings

**1. `apps/itun/src/components/dashboard/ActiveItemBand.tsx:214` — Area Salvage now gives no feedback on 10% of rolls, including the jackpot.**
`doSalvage` moved `setPrompt` into `runWrite`'s `onApplied`, but the write is only issued `if (patch)`. `areaSalvageOutcome` returns `patch: null` whenever `scrapQty <= 0`, which is **both** d20=1 (nothing) and d20=20 (jackpot → player picks a Damaged Chassis/System/Module) — `dashboardEconomy.ts:99` and the two tests at `__tests__/dashboardEconomy.test.ts:82-92` pin exactly that. Before this PR the readout fired unconditionally. Now the jackpot overlay never opens, the player is never told they may claim an item, and the roll is lost. Worse than the `onApplied` family below because it hits **Solo** too, not just Disconnected. Fix: keep `setPrompt` outside the `if (patch)`, or pass it as `onApplied` *and* call it in the null-patch branch.

**2. `packages/component-lib/src/components/shared/ControlButtons.tsx:179` — disabled segment text is invisible (1.00:1).**
The de-emphasis sweep turned `bg-wk-muted text-ink-2` into `bg-wk-muted text-wk-muted` — foreground and background are now the same token. Any disabled `ReferenceEntityControl` with a `segmentText` (cost/count chip) renders a blank grey block where the value should be. This is the only `bg-wk-muted` in the repo, which is why the mechanical swap didn't look wrong at the other ~20 sites.

**2b. Same file, line 163** — the paired swap put the disabled *label* at `bg-wk-faint text-wk-muted`. `theme.css:222-224` declares `--color-wk-faint` **"NON-TEXT ONLY … must never carry text"**; this is ~2.85:1, down from `text-ink-2`'s ~5.0:1. The PR's own rationale for retoning `wk-muted` was AA compliance, and this is the one site where the ground isn't paper.

## Rejected

`apps/srd/src/pages/schema/[schemaId].json.ts` — the srd agent reported the `preload('all')` import was dropped, with a repro. It isn't: the route imports `getEntitySchemas` from `'../../lib/gameData'` (only `getModel` comes from the package), and there's a comment above the import saying exactly why. Standalone evaluation returns 24 paths, no throw. False positive.

## Unchanged from my pass

`doScrap` (ActiveItemBand.tsx:249), `restore()` (DowntimeWizard.tsx:138), `craft()` (ActiveItemBand.tsx:153), `activate()`/`doPush()` (ActionsDeck.tsx:166, 203) — all independently reconfirmed by the itun agent; plus the `parseTraitReferences.tsx` lost `.trim()` (latent) and the stale `PageShell` docstring. The srd/bot agent additionally proved `linkifyTraitRefs`'s rewrite behaviour-identical across all 11,621 dataset strings, which is consistent with the trim loss being latent rather than live.

Net: **nine findings**. The five `onApplied` ones are one family and one commit; findings 1 and 2 are separate mechanical-sweep misses.

alxjrvs (commented):
Reviewed the full range diff (900 files, ~16k lines of non-test source diff) plus the test/CI changes. The vast majority is mechanical and safe: Biome-driven import ordering/`import type` separation, `text-ink-2` → `text-wk-muted`, `rounded-[Npx]` → radius tokens, `.find(e => e.name === x)` → indexed `getByName`/`getByNameIn`/`resolveRef`, and hand-rolled markup replaced by `PageShell`/`PageHeading`/`EmptyState`/`Button`.

I ran the repo's own gates as verification: `bun run check:all` exits 0 (typecheck, 3500+ tests, lint, format, validate, knip, audit, tokens, styling). So everything below is something CI cannot see.

Findings (4):

- `apps/itun/src/components/dashboard/ActiveItemBand.tsx:214` — Area Salvage now reports nothing when the roll yields no scrap. The readout moved into `runWrite`'s `onApplied`, which sits inside `if (patch)`; `areaSalvageOutcome` returns `patch: null` for *both* the "nothing" band and the jackpot that sets `requiresPlayerChoice`. Two of the three branches of the log expression are now unreachable — that's the tell.
- `apps/itun/src/components/dashboard/ActiveItemBand.tsx:153` — four handlers still state success unconditionally after a refusable write, which is the exact defect `onApplied` was introduced to fix: `craft` (:153), `doScrap` (:249), `ActionsDeck.doPush` (:203), `DowntimeWizard.restore` (:139). A Disconnected player gets the refusal toast *and* "Crafted X — it is in the Hold."
- `.github/workflows/codeql.yml:18` — the two halves shipped in the wrong order. I checked the live ruleset (`gh api …/rulesets/9182849`): `Analyze (javascript-typescript)` is already required and a merge queue is active, but `main`'s codeql.yml has no `merge_group:` trigger yet. Any *other* PR queued from current `main` hangs on a check that can never report. This PR unblocks itself; nothing else.
- `packages/component-lib/src/utils/parseTraitReferences.tsx:103` — the delegation to the package parser dropped the `.trim()` on trait name/param, while the bot's `linkifyTraitRefs` still trims. Latent (no padded refs in shipped data) but re-opens the divergence the change set out to close.

Two things I checked and deliberately did *not* file: the `claimLocal` mech-pattern edge parse is safe (the `mechPatterns` store sets no `updatedAt`, so records satisfy the strict schema), and dropping the three meta schemas from `/schema/*.json` is a real public-API removal but a deliberate, argued one.

alxjrvs (commented):
**Final review: 17 findings** (the 6 I reported, minus the now-moot CodeQL one, plus 12 from the four scoped subagents).

Two pieces of context that change how you act on this:

**The branch already merged.** `origin/main` is `71494ef7 audit: land the consistency audit … (#682)` — the squash of these 17 commits. Everything above is live on `main`, so this is a follow-up-PR list, not a merge gate.

**The CodeQL finding resolved itself.** `main` now carries the `merge_group:` trigger (it landed with #682), so `Analyze (javascript-typescript)` can report on `gh-readonly-queue/**` branches. The queue is not wedged. The `ci.yml:517` comment claiming the requirement "had to be walked back" is still wrong about current state, but that's prose, not a blocker.

**The one that actually breaks play:** Area Salvage. A natural 20 — the best result on the table — writes nothing and now says nothing, and its message branch is unreachable code.

**The pattern worth a single fix commit:** five sibling `runWrite` call sites (`ActiveItemBand` craft/scrap, `ActionsDeck` push/activate, `DowntimeWizard` restore) still assert success after a refused write. That's the exact defect class commit `c6b26121` was written to eliminate — the fix was applied to the mech/pilot handlers and missed their neighbours.

**Cheap and mechanical:** re-run Convex codegen (`model/entities.ts` is missing from `api.d.ts`), add `biome check .` to the lint job so the new `organizeImports` assist is actually enforced, narrow `check-doc-drift`'s sibling set back to the barrel, drop the no-op `apps/su-assets/bunfig.toml` `[test.env]` block, and restore `min-w-0` to `BandTitle`'s non-fill path.

**One decision for you, not a bug:** `/schema/actions.json` (686 records) and two siblings now 404. Deliberate, documented, test-pinned — but it's a public JSON endpoint disappearing with no redirect in `netlify.toml`.

Gates were green throughout (`typecheck`, all workspace tests, `lint`, `format`, `knip`, `validate:all`, `build`, `check:all`), and each subagent independently re-ran them. None of these are things CI would have caught.

alxjrvs (commented):
This one is the resumed lead review agent, and it's the only redelivery that adds something: an **8th finding**, low severity and latent.

**`packages/component-lib/src/utils/parseTraitReferences.tsx:103`** — swapping the local regex for the package parser dropped the `.trim()` the old code applied to `match[1]`/`match[2]`. `TraitRef` resolves via `t.name.toLowerCase() === name.toLowerCase()` with no trim, so `[[ Shield ]]` would render `" Shield "` and lose its tooltip. The Discord bot's `linkifyTraitRefs` still trims (`lookupEmbed.ts`), so the two renderers now disagree on the same markup. No padded refs exist in the data today (all 34 distinct refs checked), so this only bites when someone first authors one.

Worth noting the two agents differ slightly here: the component-lib pass called the dropped `.trim()` "safe," the lead pass calls it latent. They agree on the facts — zero divergence over current data — so it's a question of whether a divergence that only appears on future authoring is worth recording. I'd record it, since the bot and the web renderer now disagree.

Everything else in the report restates findings already covered. Updated tally: **8 findings** (7 previously reported plus this one), **2 non-blocking observations** remaining separate — the `resolveActionRef`/`resolveClassRef` dead exports have now been reported by two independent passes and are folded into the 8 as a ranked-low finding, leaving the stale `resolveClass` comment and the stale `choiceType`/`blackMarket` docs as advisory.

Branch still untouched — nothing has been edited.

PR #683, opened 2026-08-05
test: make the reference-model seam mean what it looks like it means

alxjrvs:
Follow-up to #682, which surfaced this class rather than introducing it.

## The seam did not mean what it looked like it meant

A `SalvageUnionReference` model is reachable four ways — `all()`, `getById`, `getByName`, `getBySlug` — and the last three read `BaseModel`'s lazily-built index rather than calling `all()`.

So `Model.all = mock(() => FIXTURE)` does not replace a model's rows. It replaces **one route to them**. Production code that reaches rows the other way keeps getting the real dataset, and the test asserting on fixtures has no idea.

**This already happened.** Converting the crawler lookups from `find(b => b.id === ref || b.name === ref)` to the indexed resolver broke **17 tests across five suites at once**. Nothing about the production change was wrong — the component rendered real bays while the tests asserted on fake ones, and every failure read as a component bug. The seam was what broke.

## What this changes

- **`AddNpcControl`** was the last suite still doing it, across six models. It passes *today* only because the encounter path reaches rows through `listCandidates` — but that resolves a pick with `find(c => c.slug === slug)`, which is exactly the linear scan that becomes `getBySlug` sooner or later. Same silent failure, waiting.
- **`patchModelRows` moves to `test/`** beside the other cross-workspace helpers. It was under `apps/itun/`, which made it unreachable from the other workspaces the guard polices — a guard whose remedy you cannot import is not much of a guard.
- **`test-hygiene.test.ts` gains a rule** that fails on any single-accessor assignment to a reference model, naming the offending file and accessor.

Proven by deliberate failure: reintroducing `NPCs.all = mock(...)` fails the guard with that file and accessor named; removing it passes.

## Deliberately not touched

`console.error =`, `req.onsuccess =`, `globalThis.fetch =` are a different thing — the first two capture output and wire event handlers, and the third replaces the only route to what it stubs. None can hand real data to one path and a fixture to another.

## Verification

`check:all` exits 0, and `bun --filter srd build` runs the Astro static build that `check:all` does not (the gap that let a `getStaticPaths` regression through in #682).

https://claude.ai/code/session_01Cvbp9qqHE2fgxwE6dnzBbN


alxjrvs (commented):
Reviewed commit `c87ea35b` (scope: `origin/main...HEAD` — 5 test files re-pointed at a moved helper, the helper's move to `test/patchModel.ts`, and a new hygiene rule). Ran `bun run typecheck`, `bun run lint`, `bun run knip`, `check-doc-drift`, the 5 affected itun suites and `test-hygiene.test.ts` — all green, so nothing here breaks CI. The findings are about the seam the change installs being weaker than it claims.

Findings:

- **test/patchModel.ts:53** — the local `slugify` is a third slug implementation and diverges from `nameToSlug`, which is what the real `BaseModel.getBySlug` and `getEntitySlug` use. Real dataset names differ: "Hunter's Sight" → `hunters-sight` vs `hunter-s-sight`, "H&V" → `hv` vs `h-v`, "M.A.D." → `mad` vs `m-a-d`. Import `nameToSlug`.
- **test/patchModel.ts:36** — `find`/`findAll` are left unpatched despite the "across EVERY accessor" contract; they read the real backing data. ~16 production call sites still use them (`dashboardEffects.ts:61-67`, `downtime.ts:343`). And the new guard bans patching them by hand, so there is no sanctioned workaround.
- **test/patchModel.ts:20** — the `model` parameter types every accessor as `() => unknown`, so `rows` is no longer checked against the model's entity type. Verified: `patchModelRows(SalvageUnionReference.NPCs, [{id,name,totallyNotAnNpc}])` compiles, while the now-banned `NPCs.all = mock(() => BOGUS)` is a TS2322. The migration trades away a real compile-time check on fixtures.
- **test/patchModel.ts:42** — the fake `getBySlug` matches an explicit `slug` field that `BaseModel.getBySlug` never reads (it keys purely on `nameToSlug(name)`), so a fixture can green-light a lookup production cannot perform.
- **packages/component-lib/src/test-hygiene.test.ts:92** — the guard regex requires the literal `SalvageUnionReference.X.acc =` path; one alias (`const bays = SalvageUnionReference.CrawlerBays; bays.all = mock(...)`) bypasses it — the same aliased shape AddNpcControl.test.tsx used in its restore loop before this commit.
- **packages/component-lib/src/test-hygiene.test.ts:14** — the header docblock still enumerates 3 rules for a file that now has 4, and `.claude/rules/testing-patterns.md` gained no prose for the new rule despite the header calling this file its executable half.

alxjrvs (commented):
## Review: reference-model test seam (2 commits, 7 files)

**Scope:** `git diff origin/main...HEAD` — moves `patchModel.ts` to the repo root `test/`, migrates `AddNpcControl.test.tsx` off hand-rolled `.all` mocks, and adds a `test-hygiene` guard banning single-accessor patches.

The direction is right and the change is small, but I verified the new seam empirically against the real ORM and it does not close the gap it claims to. All 43 affected tests and the 7 hygiene tests pass; `typecheck:itun` is clean. Every finding below is confirmed by execution or by reading `BaseModel.ts`.

### Findings

- `test/patchModel.ts:21` — `patchModelRows` does not patch `find`/`findAll`; `BaseModel` implements them against `this.data`, not `all()`. Proven: with the patch active, `all().length === 1` while `find(c => c.name === 'Mule')` returns the real Mule and `findAll(() => true).length === 51`. Production already uses that route in `dashboardEffects.ts:61-67`, `DisplayPanel.tsx:68`, `pilotSheetModel.ts:148`, `blankCreate.ts:98`.
- `packages/component-lib/src/test-hygiene.test.ts:93` — the guard bans assigning `find`/`findAll` and points authors at `patchModelRows`, which is a silent no-op for both. The rule forbids the only mechanism that works.
- `test/patchModel.ts:53` — the local `slugify` diverges from `nameToSlug`. Proven: `nameToSlug("Hunter's Sight") === 'hunters-sight'`, but the mock answers only `'hunter-s-sight'`. Real data has 13+ apostrophe names and 4 dotted acronyms (`God's Wrath`, `A.I. Personality`, `M.A.D.`).
- `test/patchModel.ts:42` — the `r.slug === slug` branch has no counterpart in `BaseModel.getBySlug` (name-derived only), so a fixture carrying an explicit `slug` passes a test that would return `undefined` in production.
- `packages/component-lib/src/test-hygiene.test.ts:73` — the glob covers only `**/*.test.ts{,x}`, so a half-patch in a shared `__tests__/` helper is invisible. `patchModel.ts` itself previously lived at exactly such a path.
- `packages/component-lib/src/test-hygiene.test.ts:14` — the header enumerates rules 1-3 and was not extended; `patchModelRows` appears in no `.claude/rules/`, `CLAUDE.md`, or `docs/` file, unlike the three existing rules.
- `apps/itun/src/components/encounter/__tests__/AddNpcControl.test.tsx:4` — header still says "Patches the reference model `.all`s (same seam as SalvageControl.test)"; it now patches every accessor, and `SalvageControl.test` no longer exists.

The single highest-value fix is adding `find`/`findAll` to `patchModelRows` and swapping the local `slugify` for the package's `nameToSlug` — that resolves findings 1-3 together.

PR #684, opened 2026-08-05
audit: land the dependency, monorepo and agentic-DX remediation

alxjrvs:
Executes the ordered plan from the dependencies / monorepo / framework audit. Every claim below was measured, and each fix is verified rather than assumed.

`bun run check:all` exits 0.

## The framework question: no change

The audit's headline question — Svelte/Preact/other for `component-lib` — was settled **keep React**, adversarially: the agent briefed to argue *for* migrating concluded against it after measuring. Two hard blockers (TanStack Router is React/Solid-only; Convex Auth ships React-only clients) and the arithmetic: React is 59 KB gz of srd's 215 KB gz eager page payload, while the largest chunk in both apps is app code. A migration attacks the smallest slice of payload while rewriting the largest surface. **No framework work in this PR.**

## Security — 10 advisories → 4

Raised the `overrides` security floors past four live advisories (undici 6.27.0→6.28.0, postcss 8.5.22→8.5.25, new floors for `yaml` and `@opentelemetry/core`).

#674 converted these overrides into caret security floors correctly, but **the lockfile was never regenerated**, so the fixes it enabled were never taken. CI gates at `--audit-level=high` and all ten findings were moderate/low, so nothing ever surfaced it.

The 4 remaining are deliberately not chased: 3 are the dead `astro@6.4.6` subtree `@vite-pwa/astro`'s peer range drags in (never executes), 1 is a low-severity esbuild issue affecting only the Windows dev server.

## Type-check holes — ~46 files across six locations

`bun run typecheck` iterated the six workspaces; anything outside them was checked by nothing. Closed via a new `tsconfig.tools.json` plus per-workspace `include` fixes: root `tools/**` (every `check-*.ts` CI guard), root `test/**` (the preloads every workspace bunfig loads), both `e2e/**` dirs + playwright configs, component-lib's `vite.config.ts` + `.ladle/components.tsx`, discord-bot's `test/env.ts`.

**This was not cosmetic.** Turning it on surfaced **19 real type errors** in four CI guards — including a genuine defect: `check-design-tokens` and `check-styling-ownership` both read `counts[rule.id]` unguarded, and `undefined > was` is always `false`, so a rule that vanished from `counts` could never be reported as a regression.

## Enforcement gaps

| Gap | Fix |
|---|---|
| `convex/_generated` committed with no drift gate | New `check-convex-codegen.ts`. It had **already drifted** — `model/entities.ts` added in #682 while `api.d.ts` was hand-edited in the same commit, so the module was never registered. `convex codegen` needs a `CONVEX_DEPLOYMENT` and can't run in CI, so the gate checks what *is* verifiable offline: registered module set == files on disk. |
| Render built the bot's shipped bundle on an unpinned Bun | Pinned in `render.yaml`; `check-bun-version.ts` now covers it (verified to fail on drift). |
| Biome `linter.enabled: false` didn't hold | Three such blocks sat *above* the rule blocks that reset `enabled` back to true — `salvageunion-reference/tools` was emitting 2 warnings while declared unlinted. They now sort last; the per-rule workarounds that papered over this are deleted. |
| doc-drift didn't scan `.claude/skills` | Now does, plus a new check that every documented `bun run <script>` exists. Both live drifts this audit found sat in that blind spot — and the new check immediately caught a third. |

## The `bun test` trap — 639 failures → 0

Raw `bun test` at the root reported **639 failures / 35 errors against a green tree**, purely from missing preloads. Worse, the checked-in `/verify` skill instructed exactly that, plus a mutating `bun run format` where a check belongs.

The root bunfig now preloads the union of what the workspaces need: **4712 pass, 0 fail**. `/verify` is now a true alias of `/validate`, which makes CLAUDE.md's "both do the same thing" claim actually true.

## Lint ratchet

~14 a11y/correctness/security rules were downgraded to `warn` during the ESLint migration as "a deliberate follow-up list" — but nothing ratcheted them, and `bun run lint` exits 0 while printing warnings. Cleared all **7 warnings + 2 infos** and promoted the set to `error`. The two genuinely-correct idioms that trip a rule carry a reasoned `biome-ignore`.

Also finished the 9-interface burn-down and deleted its temporary override; API report regenerated (declaration form only).

## Dependency majors

jest-dom 6→7 (with `@testing-library/dom` declared explicitly — now a required peer) and convex 1.42.3→1.43.0 with convex-test 0.0.55. Full suite green.

## itun boot path

`preload('all')` was created on first render, so the dataset fetch didn't start until React had booted and rendered down to the gate. Hoisted to module scope — the pattern srd already uses, and explicitly sanctioned by `check-architecture`, which exempts `.preload()` from the no-module-scope-ORM rule.

**Deliberately unchanged:** it still preloads `'all'` behind ONE gate. Narrowing to per-route lists is the footgun that file documents; making the preload non-blocking needs every ORM-touching component to tolerate a not-ready state. Both remain out of scope.

## Worktree hygiene — 13 GB reclaimed

`tools/prune-stale-worktrees.sh` existed but was referenced by nothing, and its "unpushed" guard skipped every branch worktree — so it had **never reclaimed anything** (measured: 15 worktrees / 13 GB, reporting "0 candidates, 15 skipped").

The guard now checks what actually matters — reachability *after* removal. A branch ref survives `git worktree remove`, so only a detached HEAD needs the check. Wired up as `bun run reap` and documented. Repo-wide `find` for tsconfigs went 97 hits → 9.

Also adds **`check:fast`** (~12s: lint + validate + knip in parallel, then typecheck) as the named inner loop, and puts `lint` + `check:schemas` on pre-push.

## Reviewer notes

- `apps/itun/convex/_generated/api.d.ts` is hand-edited by two lines. That file is generated, but `convex codegen` requires a deployment credential this environment doesn't have; the edit is exactly what codegen would emit, and the new gate verifies it.
- **Profiling finding, not fixed here:** itun's `src` chunk (191 KB gz, its largest) is 49% node_modules / 39% component-lib / 12% salvageunion-reference. Within node_modules: `@base-ui/react` 348 KB, `zod` 265 KB, `tailwind-merge` 103 KB, **`qrcode` 71 KB**, `sonner` 64 KB. `qrcode` sitting in the eager main chunk is a clear lazy-load candidate — worth its own PR.
- **Astro still cannot use TypeScript 7** (checked empirically, not cited): pointing srd at TS 7.0.2 makes `astro check` crash with `Cannot read properties of undefined (reading 'fileExists')` in `@astrojs/language-server`. The TS 6 foothold stays.

https://claude.ai/code/session_01BDWaNhy8qge17QvxvP7B7r


alxjrvs (commented):
Reviewed the audit-remediation diff (commit `9252ebfb`, ~40 files). I verified the tree is green (`bun run test`, `bun run check:fast`, `typecheck:tools`, all new validators pass), so these findings are about latent defects, not build breakage.

**Findings (5):**

- `tools/check-doc-drift.ts:199` — Adding `.claude/skills` to `LIVE_INSTRUCTION_DOC_DIRS` gates **zero files**: `markdownIn()` (line 75) is non-recursive and `.claude/skills/` holds only sub-directories with `SKILL.md` inside. Verified: appending `` `bun run this-script-does-not-exist` `` to `.claude/skills/verify/SKILL.md` leaves the check green at the same "67 references checked". The change's own stated purpose — catching the two skill drifts the audit found — is unmet.
- `lefthook.yml:49` — The new `schemas` pre-push command sits in a `parallel: true` group, but `check:schemas` runs `build:package`, which rewrites `lib/index.ts`, `lib/generated/*.generated.ts`, `schemas/*.schema.json` and `etc/*.api.d.ts` while the sibling `typecheck`/`test`/`lint`/`knip` are reading those exact files. `check:all` deliberately runs it first and sequentially for this reason; expect intermittent, retry-clean pre-push failures.
- `packages/component-lib/src/components/stat/StatProvenance.tsx:121` — Dropping the index from the `<li>` key makes duplicate React keys reachable. `pilotContributions` (`contributions.ts:120-147`) emits one entry per contribution with `copies: 1` and no dedup over `abilityRefs`, so one ability with two same-stat contributions of equal amount (or a repeated ability ref) yields two lines identical in `kind`/`label`/`detail`/`amount`. The comment's "cannot collide" claim is wrong.
- `tools/check-convex-codegen.ts:42` — `NOT_MODULES` is far narrower than Convex's real entry-point rules (`convex/dist/esm/bundler/index.js:279-360` also skips any basename with >1 dot, leading `.`/`#`, spaces, no-import/export files, and nested component dirs). Adding `apps/itun/convex/games.test.ts` — the layout `convex-test`, already a devDependency, documents — would fail `validate:all` in CI and pre-push with drift that doesn't exist, and the error's suggested fix would make `api.d.ts` diverge from real codegen.
- `apps/itun/src/components/shared/GameDataReady.tsx:63` (PLAUSIBLE) — The hoisted `preloadPromise` is a floating promise with no `.catch`; previously `use()` attached a handler in the same tick as creation. A preload rejection before `PreloadGate` mounts (chunk 404 after deploy rollover, cold PWA start) now fires `unhandledrejection`, which `@sentry/browser` reports on top of the intended Suspense/error-boundary path. `apps/srd/src/lib/gameData.ts` top-level-`await`s instead.

Things I checked and cleared: the `renovate.json` deletion (`.github/dependabot.yml` already covers it), the `yaml` / `@opentelemetry/core` overrides (no v1 consumers; otel unifies on 2.9.0 which every peer range accepts), the `@testing-library/jest-dom` 7.0.0 major (peer satisfied by the new `@testing-library/dom` dep; the `aria-query` 5.3.2→5.3.0 and `dom-accessibility-api` shuffles are correct nesting), the deleted `apps/discord-bot/tsconfig.build.json` (unreferenced), the biome override reordering (the "must sort last" fix is right, and the removed `playwright.visual.config.ts` entry points at a file that no longer exists), `tsconfig.tools.json` (tsbuildinfo is gitignored), the `?? 0` fixes in the two ratchet scripts, and `bun run reap` (dry-run behaves as documented).

alxjrvs (commented):
Reviewed the two commits on `worktree-audit-remediation` vs `main` (41 files: dependency/override changes, tsconfig coverage closures, new CI guards, bunfig/lefthook/biome rework). I verified the tree is green locally — `bun run lint` (0 findings), `bun run check:fast` (all six workspaces typecheck, including the new `typecheck:tools`), `bun run test:tools` (37 pass), `bun --filter component-lib test` (751 pass), and both new/changed guards (`check-convex-codegen`, `check-doc-drift`, `check-bun-version`) pass. I also confirmed the new tsconfig includes really do add coverage (17 itun e2e files, 7 `.ladle` files now in the type-check graph).

Findings:

- `tools/check-convex-codegen.ts:42` — The `NOT_MODULES` set is a hardcoded two-item approximation of Convex's real entry-point rule. Convex 1.43.0's `entryPoints()` skips any basename with >1 dot, plus `.`/`#`-prefixed files and paths with spaces. Adding `apps/itun/convex/games.test.ts` (natural — `convex-test` is already a devDep) makes this gate fail `validate:all`/CI/pre-push, and its own remediation advice can't fix it because real codegen produces exactly the `api.d.ts` it rejects.
- `lefthook.yml:49` — `check:schemas` runs `build:package`, which unconditionally `writeFileSync`s `lib/index.ts` and `lib/generated/*.generated.ts`, and it was added to a `parallel: true` pre-push group alongside `typecheck`/`test`/`knip` that read those exact files. Torn reads → intermittently-red pre-push. `check:all` deliberately sequences it first with `&&`; the hook doesn't.
- `tools/prune-stale-worktrees.sh:113` — Swapping the "on a remote" guard for "HEAD is on a branch" removes the last abandonment criterion; every ordinary worktree passes, so `reap --force` deletes all clean worktrees (including unlocked, hand-created ones with unpushed work and their gitignored local state). Header and CLAUDE.md still say "abandoned".
- `apps/itun/src/components/shared/GameDataReady.tsx:63` — Module-scope `preloadPromise` has no rejection handler until React renders `PreloadGate`; a failed dataset load now fires `unhandledrejection` (console error + spurious Sentry report) on top of the error-boundary path. One `.catch(() => {})` closes it.
- `packages/component-lib/src/components/stat/StatProvenance.tsx:121` — Content-derived keys can collide: `abilityContributions()` doesn't de-dup refs and emits one line per declared contribution with `copies: 1`, so two identical `{kind,label,detail,amount}` lines are reachable → duplicate React keys and undefined reconciliation on update. The comment's "harmless" claim holds only for first paint.
- `tools/check-design-tokens.ts:421` (same text at `tools/check-styling-ownership.ts:446`) — The `?? 0` justification is false twice over: `counts` is populated for every rule immediately above, and `0 > was` is false whenever `undefined > was` was, so it can never convert a missing rule into a regression. Harmless code, misleading invariant.

alxjrvs (commented):

---

I read the full diff and checked the surrounding code. No prompt-injection or instructions-to-reviewer text is present in it.

## Findings

**1. `packages/component-lib/src/components/stat/StatProvenance.tsx:121` — the new content-derived key can collide, reintroducing duplicate React keys where the old index-suffixed key was unique by construction.** `ProvenanceLine` is `{ kind, label, detail?, amount }` (lines 23–29) and nothing dedupes the array, so two copies of the same source contributing the same amount — e.g. two identical modules each granting +1, exactly the "two +1 bonuses of the same source type" case the new comment cites — produce byte-identical keys. React then logs `Encountered two children with the same key` and may drop one `<li>` on re-render. The comment's claim that such lines "render interchangeably, so a collision there is harmless" is not what React does with duplicate keys. If the goal was to stop keying on position, key on a stable per-line id from the caller, or fall back to appending the index only when the content key repeats.

**2. `lefthook.yml:86` — the new `schemas` step runs a code generator concurrently with four readers of the files it rewrites.** `pre-push` is `parallel: true` (line 31), and `check:schemas` is `bun run build:package && git diff --exit-code …`; `build:package` runs `generate:registry`, `generate:json-schemas`, `docs`, `generate:api-report`, which write `packages/salvageunion-reference/lib/generated/*`, `lib/index.ts`, `schemas/*.schema.json` and `etc/*.api.d.ts`. The sibling `typecheck`, `test`, `knip` and `validate:all` steps all read those same paths at the same time, so a push can fail intermittently on a partially-written `lib/index.ts` or schema file with an error that has nothing to do with the change. It needs to run before the parallel group (a `pre-push` `piped`/ordered stage, or fold it into the `typecheck` command) rather than beside it.

**3. `apps/itun/src/components/shared/GameDataReady.tsx:63` — hoisting the promise to module scope opens an unhandled-rejection window and makes the dataset load unconditional on import.** Previously the promise was created inside `PreloadGate` and handed to `use()` in the same tick, so a rejection always had a consumer. Now `SalvageUnionReference.preload('all')` starts at module evaluation and nothing observes it until React renders down to the gate; a failure in that window (offline, 404 on a data chunk) fires a global `unhandledrejection` that Sentry will report as an uncaught error before the error boundary ever sees it. It also means any module that transitively imports this file — including test files and any non-browser consumer — now triggers a full `preload('all')` on import. Attaching a `.catch()` that re-throws inside the `use()` path, or keeping the lazy getter and calling it once from the entry module, gets the earlier start without either side effect.

**4. `package.json:22` — the `astro: "^7.1.4"` override forces `@vite-pwa/astro@1.2.0` outside its declared peer range.** The `ci.yml` comment states that plugin peer-requires `astro ^1 || … || ^5` and has never been updated for 7; the override silences the peer check rather than satisfying it, so the PWA integration is running on an untested major. The manual verification recorded in the comment (sw.js + workbox chunk emitted) covers today's build only — a future `@vite-pwa/astro` patch or an Astro 7 minor can break service-worker generation with nothing in CI asserting the output. Worth an explicit assertion on `sw.js` in the srd build check, since the dedupe motive (removing the shadow astro 6 tree and its vulnerable esbuild) is otherwise sound.

None of these block merge: 1–3 are correctness/robustness regressions rather than breakage on the happy path, and 4 is a documented, verified trade-off.

alxjrvs (commented):

---

I read the full diff and checked the surrounding code. This is a large audit-remediation change (tooling, tsconfig coverage, Biome ordering, dependency overrides, a new Convex codegen gate) against an older base than the current worktree. No prompt-injection or instructions-to-the-reviewer text appears anywhere in it.

## Findings

- `lefthook.yml:86` — The new `schemas` pre-push step runs `bun run check:schemas`, which invokes `bun run build:package` and **rewrites** `packages/salvageunion-reference/lib/generated/*`, `lib/index.ts`, `schemas/*.json` and `etc/*.api.d.ts`. The `pre-push` block is `parallel: true`, so this writes those files while `typecheck`, `test`, `lint` and `knip` are concurrently reading them. Expect intermittent, unreproducible pre-push failures — and the worse case, a step that reads a half-written generated file and passes. It needs to be sequenced before the parallel group (as `check:all` already does), not added into it.

- `tools/check-doc-drift.ts:872` — New Check 7 resolves every `bun run <script>` found in a live-instruction doc against the **root** `package.json` only. The corpus includes `docs/architecture/**` and `.claude/**`, which legitimately document workspace-local commands (`cd apps/srd && bun run gate`, `bun run og:generate`). The first such reference hard-fails `validate:all` in CI and pre-push, with the fix being to delete accurate documentation. It needs the owning-manifest fallback the `bun --filter` branch already has in spirit.

- `package.json:24` — The `astro: ^7.1.4` override forces `@vite-pwa/astro@1.2.0` onto a major outside its declared peer range (`^1 || … || ^5`). The CI comment justifying it states the duplicate `astro@6.4.6` tree "never executes" — so the advisories being cleared are in dead weight, while the override makes a real behavioural change to the integration that generates `sw.js` for a production site. A PWA integration that degrades under an unsupported host major fails silently (no service worker, stale precache), which is exactly the failure class that is invisible to `bun audit`, typecheck and the build. Same shape, lower stakes: `@vitejs/plugin-react-swc: ^4.3.3` forces `@ladle/react@5.1.1` off its declared `^3.7.2`.

- `packages/component-lib/src/components/stat/StatProvenance.tsx:120` — Keying by `kind`/`label`/`detail`/`amount` drops the index that previously guaranteed uniqueness. Two provenance lines identical in all four fields (e.g. two `+1` bonuses from the same source type with no detail) now produce duplicate React keys — a console error and unreliable reconciliation, where the old key was unique by construction. The comment argues collisions are harmless, but React does not treat them as such.

- `apps/itun/src/components/shared/GameDataReady.tsx:63` — Hoisting `preload('all')` to module scope means the promise exists with no `.catch` attached until `PreloadGate` first renders. A rejection before that point is an unhandled rejection (browser console noise; in Bun's runner it can surface as an unrelated failure), and the dataset fetch now fires on any import of this module, including unit tests that only import `GameDataReady` and never render it.

- `tools/check-convex-codegen.ts:42` — The on-disk module set treats *every* `.ts` under `convex/` as a registered function module, with only `schema.ts` and `auth.config.ts` exempt. Convex itself skips colocated test files and other `*.config.ts`; `convex-test` is already a dependency of this app (its tests currently live in `apps/itun/test/convex/`), so a single colocated `convex/foo.test.ts` would fail the gate with no exemption short of editing the tool. Not a problem today — worth an extension of `NOT_MODULES` to a pattern.

alxjrvs (commented):

---

Reviewed the whole diff (2,012 lines). Note the diff's base predates this worktree by a lot — it still has Astro in `apps/srd` and a `/verify` skill — so I verified each hunk against the files where they survive in the tree.

No prompt-injection or reviewer-addressed text in the diff.

**Findings**

- `lefthook.yml:86` — the new `schemas` pre-push job runs `check:schemas`, whose first step is `bun run build:package`; under `pre-push: parallel: true` that **rewrites** `packages/salvageunion-reference/lib/generated/*`, `lib/index.ts`, `schemas/*.json` and `etc/*.api.d.ts` while the concurrent `typecheck`, `test`, `knip` and `validate` jobs are reading those exact files. Failure mode is a nondeterministic pre-push result (and, on drift, a hook that leaves the working tree modified). Every other pre-push job is read-only; this one is not. Non-blocking — CI is still the real gate — but it wants either `piped`/sequencing or a read-only variant of the check.

- `apps/itun/src/components/shared/GameDataReady.tsx:63` — hoisting the preload to module scope also hoists its *rejection*. Previously `getPreloadPromise()` created the promise and handed it straight to `use()` in the same synchronous call, so a failure always had a consumer; now the `.then()` chain exists from module evaluation with no handler until React first renders `PreloadGate`. A fast failure (ORM throwing synchronously, an immediately-rejected fetch) fires `unhandledrejection` before that render — which in itun means Sentry logs it in addition to the error boundary, and in a Bun/Node test context an unhandled rejection can fail the run. Secondary: importing this module now performs the full ~1–2 MB dataset load as an import side effect, whether or not the component renders. A `preloadPromise.catch(() => {})` sink alongside the real handle costs nothing.

- `tools/check-design-tokens.ts:458` and `tools/check-styling-ownership.ts:554` — the new comments claim `?? 0` is "load-bearing" because "a rule that disappeared from `counts` could never be reported as a regression." Both halves are wrong: `counts` is populated for **every** entry of `RULES` a few lines above (`check-design-tokens.ts:431`, `check-styling-ownership.ts:516`), so the key is never absent; and if it were, `0 > was` is still false — a missing rule can only ever land in `improvements`, never `regressions`. The code change is a harmless `noUncheckedIndexedAccess` appeasement, but the comment asserts a behaviour change that did not happen, in the exact class of doc-vs-code drift the rest of this PR exists to close.

- `packages/component-lib/src/components/stat/StatProvenance.tsx:121` — replacing the index-suffixed key with a content key makes duplicate keys reachable: two lines identical in `kind`, `label`, `detail` and `amount` (e.g. two unnamed `+1` bonuses of the same source type) now render under the same key, which React reports as "Encountered two children with the same key". The comment argues the render is interchangeable — true for a stateless `<li>` — but the console error is real and will surface in tests that assert a clean console.

- `tools/check-convex-codegen.ts:42` — `NOT_MODULES` only excludes `schema.ts` and `auth.config.ts`, so any `.ts` file under `convex/` that Convex does not register as a function module (a colocated `*.test.ts`, a `_lib/` helper dir) would be reported as "exists on disk but is NOT registered" and fail `validate:all` on a file that can never appear in `api.d.ts`. Latent today — this repo's Convex tests live in `apps/itun/test/convex/` — but the gate is now in `validate:all`, so the first colocated test file breaks the build.

Nothing here blocks the merge. The lockfile changes check out: the `astro: ^7.1.4` override collapses the `@vite-pwa/astro` shadow tree, and the top-level `@astrojs/compiler` dropping 4.0.0 → 2.13.1 with no nested `astro/` entry is consistent (astro 7 uses `@astrojs/compiler-binding`; only `@astrojs/language-server` still wants `^2`), not a broken resolution.

alxjrvs (commented):
**claude-review** — ⚠️ **2 blocking** of 8 finding(s)

---

Read the full diff and checked the surrounding code. Two findings are blocking; both are cases of the change not doing what its comment says.

## Blocking

- **`tools/check-doc-drift.ts:189` (adding `'.claude/skills'` to `LIVE_INSTRUCTION_DOC_DIRS`) — the new corpus entry matches zero files, so the headline fix is a no-op.** `markdownIn` (`tools/check-doc-drift.ts:76`) is non-recursive — `readdirSync(dir).filter(n => n.endsWith('.md'))` — and every skill lives one level down as `.claude/skills/<name>/SKILL.md`; `.claude/skills` itself contains no `.md` at all (verified: `component-refresh/`, `convex-deploy-verify/`, `knip-triage/`, `triage/`). `.claude/rules` and `.claude/agents` work only because they are flat. So checks 4–6 *and* the new check 7 still never see a single skill, including `verify/SKILL.md` and `generate/SKILL.md` — the two files this PR just hand-fixed and claims are now guarded. Fix: recurse (or glob `**/*.md`) in `markdownIn`, and add a scan-floor assertion so a corpus that collapses to zero is loud.

- **`tools/check-doc-drift.ts:845` (`checkReferencedScripts`) — validates every `bun run <script>` against the *root* `package.json` only, while the corpus includes each workspace's `CLAUDE.md`/`README.md`, so `validate:all` fails on a workspace doc naming its own package script.** `liveInstructionDocs` (`:230`) walks `apps/*/CLAUDE.md` and `packages/*/CLAUDE.md`, and `packages/salvageunion-reference/CLAUDE.md:138-139` documents `bun run scaffold:entity` twice — a package-level script, which is exactly why this PR rewrote the identical reference in `.claude/rules/package-development.md:9` to `bun --filter salvageunion-reference scaffold:entity`. The workspace copy was left alone, so the newly-added check goes red on the first CI run. Fix: resolve the doc's owning workspace manifest and accept a script defined there (`rootScripts.has(s) || localScripts.has(s)`), rather than rewriting docs to route around the checker.

## Non-blocking

- `lefthook.yml:86` — `schemas: bun run check:schemas` joins a `parallel: true` group alongside `typecheck` and `test`, and `check:schemas` runs `build:package`, which *rewrites* `lib/generated/*.generated.ts`, `lib/index.ts`, `schemas/*.json` and `etc/*.api.d.ts` while the other two jobs are compiling and importing those exact files — an intermittent, very confusing pre-push failure whose cause is invisible in the output. Either run it serially before the parallel group or have it generate into a temp dir and diff.
- `apps/itun/src/components/shared/GameDataReady.tsx:63` — hoisting `preload('all')` to module scope means the promise exists before any consumer attaches a handler, so a preload failure fires `unhandledrejection` (Sentry noise / a Bun test abort) in the window before `PreloadGate` renders, and any test that transitively imports this module now kicks off a full dataset load at import time rather than at render. A `.catch(() => {})` attached at creation (the `use()` still sees the original rejected promise) removes the window.
- `tools/check-design-tokens.ts:417` and `tools/check-styling-ownership.ts:443` — the `?? 0` is fine, but both comments assert it fixes a missed *regression*; a rule absent from `counts` means zero violations, so `undefined > was` and `0 > was` are both false. The change actually converts "silently ignored" into "reported as an improvement". Given this repo treats wrong prose as the defect, the comment should say that.
- `biome.jsonc` (final not-linted block) — `packages/component-lib/playwright.visual.config.ts` is silently dropped from the linter-disabled list while the block is being moved and re-commented; the comment explains the e2e dirs and both playwright configs but never mentions this path. If the file still exists on this branch it gains lint coverage with no tsconfig coverage (component-lib's `include` gained `vite.config.ts` and `.ladle/**/*`, not this).
- `package.json:20-22` (`astro: ^7.1.4`, `@vitejs/plugin-react-swc: ^4.3.3`) — both force a consumer past its declared range (`@vite-pwa/astro@1.2.0` peers astro `…|| ^5`; `@ladle/react@5.1.1` depends on `^3.7.2`). The CI comment records manual verification, but nothing re-runs it: a future plugin/API mismatch surfaces as a silently broken service worker or Ladle build, not as an install error. Worth a note in the block that these two are unverifiable-by-install pins.
- `tools/check-convex-codegen.ts:1631` — `NOT_MODULES` hardcodes only `schema.ts` and `auth.config.ts`, so any future `convex.config.ts` (component config) or `*.test.ts` under `convex/` will be reported as unregistered drift that regenerating cannot fix. Excluding `*.config.ts` and `*.test.ts` by pattern would keep the gate honest.

CLAUDE-REVIEW-SUMMARY findings=8 blocking=2

alxjrvs (commented):

---

Read the full diff and verified the claims against the tree.

## Findings

**BLOCKING**

- `tools/check-doc-drift.ts:871` — Check 7 resolves every `bun run <script>` against the **root** `package.json` only, but `liveInstructionDocs()` (line 230) also yields per-workspace `CLAUDE.md`/`README.md`, which legitimately document that workspace's own scripts. `packages/salvageunion-reference/CLAUDE.md:138-139` documents `bun run scaffold:entity`, which is a script in `packages/salvageunion-reference/package.json` and deliberately *not* in the root manifest — so the new check fails `validate:doc-drift` → `validate:all` → `check:all`, CI and pre-push, on docs the PR doesn't touch. (The same PR fixes exactly this reference in `.claude/rules/package-development.md` but not in the package's own CLAUDE.md.) `main` today carries an `owningManifest()` helper added for this, whose comment records that resolving these "against the root manifest alone reported four false failures" — the fix is to fall back to `<workspace>/package.json` for docs under `apps/*` / `packages/*`.

**Non-blocking**

- `tools/check-doc-drift.ts:192` — adding `.claude/skills` to `LIVE_INSTRUCTION_DOC_DIRS` matches **zero files**: `markdownIn()` (line 76) does a single non-recursive `readdirSync` filtered on `.md`, and every skill lives at `.claude/skills/<name>/SKILL.md` one level down. The stated motivation for the whole check — the `/verify` skill instructing raw `bun test` — is still unguarded, so the doc it is meant to police can regress silently again.
- `tools/check-doc-drift.ts:829` — `WORKSPACE_MANIFESTS` is a hardcoded six-entry map; a new workspace escapes the `bun --filter` half of the check with no signal. Discovering it from `apps/*`/`packages/*` (as `liveInstructionDocs` already does) costs the same lines.
- `lefthook.yml:48` — pre-push is `parallel: true`, and the new `schemas` command runs `check:schemas` → `build:package`, which rewrites `packages/salvageunion-reference/lib/generated/*`, `schemas/*.json` and `lib/index.ts` in place while `typecheck`, `test`, `knip` and `validate` are concurrently reading those same files. That is an intermittent, badly-misattributed pre-push failure (a truncate/write window on `lib/index.ts` surfaces as a random type or import error elsewhere). Either run it sequentially or make the generator write-if-changed.
- `apps/itun/src/components/shared/GameDataReady.tsx:63` — hoisting the promise to module scope means nothing is attached to it until `PreloadGate` first renders. If `preload('all')` rejects quickly (offline boot, a 404 on a rotated-away data chunk — the exact case `chunkRecovery.ts` exists for), the rejection settles with no handler and fires `unhandledrejection` (and a Sentry event) before React can route it to the Suspense/error path. A `preloadPromise.catch(() => {})` sink alongside the real promise keeps the earlier start without the stray rejection.
- `packages/component-lib/src/components/stat/StatProvenance.tsx:118` — the new content-derived key can collide where the old index-mixed key never could; two lines identical in `kind`/`label`/`detail`/`amount` produce duplicate React keys, which is a dev warning and can reuse the wrong DOM node across re-renders. The comment asserts a collision is "harmless" — it is harmless to *output*, not to React's reconciliation or the console.
- `package.json:23` — the `astro: "^7.1.4"` override forces `@vite-pwa/astro@1.2.0` outside its declared peer range (`^1 || … || ^5`), i.e. an unsupported combination that npm/bun will no longer warn about. The comment records a manual verification (sw.js + workbox chunk still emitted), which is the right evidence, but the failure mode after any `@vite-pwa/astro` or Astro bump is a broken service worker at deploy time with no install-time signal — worth a note in the ADR/CI comment that this pin must be re-verified, not just re-audited.

alxjrvs (commented):
**claude-review** — ⚠️ **1 blocking** of 5 finding(s)

---

Reviewed the diff at `/var/folders/.../claude-review.84306.diff` (43 files: tooling/config hardening, biome override reordering, tsconfig coverage, a dependency dedupe, plus doc/skill corrections). Checked the surrounding code for each behavioural hunk.

**Findings**

- **BLOCKING** — `tools/prune-stale-worktrees.sh:116` — the new detached-HEAD safety check can never fire: in a detached worktree `git branch -a --contains HEAD` always prints the `* (HEAD detached at …)` pseudo-entry alongside any real branches, so the `-z` test is never true. The guard the comment describes ("keep unless another ref contains it") is dead, and `bun run reap --force` will delete a detached worktree whose commits exist on no ref — taking that worktree's per-worktree reflog (`.git/worktrees/<n>/logs/HEAD`) with it, so recovery needs `git fsck`. Use a ref-only query, e.g. `git -C "$p" for-each-ref --count=1 --contains HEAD refs/heads refs/remotes`.

- `lefthook.yml:91` — the new `schemas` step runs under `pre-push: parallel: true`, and `check:schemas` is `bun run build:package && git diff --exit-code …`, which **rewrites** `packages/salvageunion-reference/lib/generated/*`, `lib/index.ts`, `schemas/*.schema.json`, `etc/*.api.d.ts` and the docs while `typecheck`, `test`, `knip` and `lint` are concurrently reading those same files. Expect intermittent, unreproducible pre-push failures (and a `git diff` that can observe a half-written tree); it needs to run outside the parallel group.

- `packages/component-lib/src/components/stat/StatProvenance.tsx:121` — switching from an index-mixed key to a pure content key makes duplicate React keys reachable (two lines identical in `kind`/`label`/`detail`/`amount`, which the comment concedes), where the old key was guaranteed unique. That is a dev-mode "two children with the same key" error and unstable reconciliation for those `<li>`s, in exchange for nothing the index key was doing wrong here.

- `apps/itun/src/components/shared/GameDataReady.tsx:63` — hoisting `preload('all')` to module scope means the full dataset fetch now fires on *import*, in every environment that pulls this module in (unit tests, any module transitively importing it), not only when `PreloadGate` renders; and a rejection that lands before first render escapes the `Suspense` boundary as an `unhandledrejection` instead of surfacing through `use()`.

- `tools/check-convex-codegen.ts:1653 (diff) / registered-set derivation` — the check only parses the `import type * as X from "../….js"` lines and never the `fullApi` object literal, so the hand-edit class it exists to catch is only half covered: a module that is imported but omitted from `fullApi` (or vice versa) still passes green.

No prompt-injection or instructions addressed to a reviewing agent appear in the diff. The `astro: ^7.1.4` / `@vitejs/plugin-react-swc: ^4.3.3` dedupe pins check out against the lockfile — the top-level `@astrojs/compiler` 4.0.0 → 2.13.1 move is a consequence of removing the shadow astro 6 tree (astro 7 uses `@astrojs/compiler-binding`), not a downgrade of the real one.

CLAUDE-REVIEW-SUMMARY findings=5 blocking=1

PR #685, opened 2026-08-05
feat(srd): server-render the entity card on the 82% of pages that are documents

alxjrvs:
## What

ADR-012 promises "entity pages are server-pre-rendered". They were not. The entity card was a `client:visible` island gated on a browser-side `SalvageUnionReference.preload()`, and `useGameData`'s server snapshot is hardcoded `false` — deliberately, to prevent a React #418 hydration mismatch — so Astro could only ever emit a **loading skeleton** where the card should be.

The readable content came from `StaticEntityContent`, a separate simplified renderer that the island hid on hydration and deleted once data arrived. Every entity page shipped three representations of one entity: skeleton, serialized props, shadow copy. The real card was unreachable to crawlers and no-JS readers.

The data is static JSON in this repo, so it is available during the build. `EntityView.astro` now awaits `preload()` in its frontmatter and renders the real `ReferenceEntityCard` with **no `client:*` directive**, so Astro emits it as markup and ships no JS for it — the same shape `SiteHeader` already uses.

## The gate

This could not be done for every page, and that is the important part of the change.

Most SRD entities are documents — prose, stats, links, nothing needing a click handler. A minority embed a real control: a roll table's Roll button, a disclosure, a nested-entity click target. Removing the island for those would leave a **rendered-but-dead rust action button**, which `docs/design-system/ruleset.md` treats as a defect ("rust means action, and only action"). Measured before gating, an ungated change would have shipped **257 dead buttons across 174 pages**, plus 140 nested-card click targets and 30 disclosures.

Hydrating the static ones is equally unsound — that is exactly the #418 mismatch the hardcoded `false` exists to prevent. So server-rendering is applied only where nothing hydrates afterwards.

`cardNeedsHydration` decides by **rendering the card and inspecting the markup** for anything needing a listener, rather than reading the entity's data shape. The card decides what controls to emit deep inside its own recursion, so a data-shape heuristic would drift the moment that logic changed — and drift here is silent, shipping a dead button instead of failing a build. It costs one extra static render per page against a build that runs in seconds, and it **fails safe**: if the probe throws, the answer is "hydrate", because the island is the status quo and the static path is only ever an optimization.

## Measured

Over the full 1,039-page build, verified against the built HTML:

|                                                | |
| ---------------------------------------------- | --- |
| entity pages server-rendering the real card    | **819 / 1,004 (81.6%)** |
| dead controls in those pages                   | **0** |
| pages keeping the island (unchanged)           | 185 |
| readable text, `/schema/chassis/item/forge/`   | ~490 → **1,939 chars** |

The chassis ability and all four patterns now appear in the HTML. Previously **neither did** — `grep -c "Pattern"` on that built page returned `0`. Those pages also no longer fetch the game-data corpus at all: the card's island chunk and its schema preload are gone, and `SearchIsland` already defers.

The 185 interactive pages are byte-for-byte unchanged, so this is a strict improvement with no user-visible regression.

## Notes for review

- `apps/srd/src/lib/cardNeedsHydration.tsx` is the load-bearing file. The comment there explains why it probes rather than guesses, and why `<a href>` is deliberately absent from the "needs a listener" pattern — anchors work fine in a document that never hydrates.
- The `hydration-directives` guard test now pins both halves: the static path carries no directive, the interactive path keeps `client:visible` **and** its no-JS fallback. Its helper reads the `.astro` template separately from the frontmatter, so a prose comment can no longer fail the very test asserting a symbol is gone.
- Two comments in `component-lib` claimed "the card is never server-rendered". That is now false for most srd pages, and the `narrow` stat-cluster wrap is the layout srd actually ships rather than an edge case. Both corrected.

## Base

Stacked on #684 rather than `main`, so the diff is this commit alone. **CI will not run while the base is a feature branch** — `.github/workflows/ci.yml` triggers on `pull_request: branches: ['main']` only. Re-target to `main` once #684 lands and checks will fire.

Full `bun run check:all` is green locally on this base: lint, format, typecheck across all 5 workspaces, tests, validate, knip, audit, tokens, styling.

### DIFF
--- a/apps/srd/src/components/EntityCardStatic.tsx
+++ b/apps/srd/src/components/EntityCardStatic.tsx
@@ -1,0 +1,61 @@
+import {
+  ClassAbilityTree,
+  EntityDetailLinkProvider,
+  EntityHrefProvider,
+  getClassSelections,
+  PatternHrefProvider,
+  ReferenceEntityCard,
+} from 'component-lib'
+import type { SURefEntity, SURefObjectPattern } from 'salvageunion-reference'
+import { srdEntityHref } from '../lib/entityHref'
+import { srdPatternHref } from '../lib/patternHref'
+
+type EntityCardStaticProps = {
+  item: SURefEntity
+  pattern?: SURefObjectPattern
+  titleAs?: 'span' | 'h1'
+}
+
+/**
+ * The reference entity card, rendered to HTML at build time.
+ *
+ * This is the same `ReferenceEntityCard` the builder app uses — NOT a
+ * simplified stand-in. It carries no `client:*` directive, so Astro renders it
+ * to static markup and ships no JS for it, exactly as `SiteHeader` does.
+ *
+ * That is only possible because the SRD is a Reference surface (ADR-021): it
+ * passes no `controls`, and `EntityDetailLinkProvider` puts nested entities in
+ * link mode, so every affordance on the card is a navigation rather than client
+ * state. The card must therefore never be given a `client:*` directive here —
+ * hydrating it would re-introduce the React #418 mismatch that the old
+ * island + `GameDataGate` arrangement existed to avoid.
+ *
+ * Callers must `await SalvageUnionReference.preload(...)` in their Astro
+ * frontmatter before rendering this, so the ORM lookups the card makes for
+ * nested entities resolve synchronously during the build.
+ */
+export function EntityCardStatic({ item, pattern, titleAs }: EntityCardStaticProps) {
+  const classSelections = getClassSelections(item)
+  const classEntity = classSelections.selectedClass || classSelections.selectedAdvancedClass
+
+  return (
+    <div className="mx-auto w-full max-w-6xl p-4">
+      <EntityHrefProvider value={srdEntityHref}>
+        <EntityDetailLinkProvider value={true}>
+          <PatternHrefProvider value={srdPatternHref}>
+            <ReferenceEntityCard
+              data={item}
+              pattern={pattern}
+              size="large"
+              titleAs={titleAs}
+              afterExtraContent={
+                classEntity ? <ClassAbilityTree classEntity={classEntity} /> : undefined
+              }
+              asideLead={!!classEntity}
+            />
+          </PatternHrefProvider>
+        </EntityDetailLinkProvider>
+      </EntityHrefProvider>
+    </div>
+  )
+}
--- a/apps/srd/src/components/EntityView.astro
+++ b/apps/srd/src/components/EntityView.astro
@@ -1,20 +1,32 @@
 ---
 import type { SURefEntity, SURefObjectPattern } from 'salvageunion-reference'
+import { SalvageUnionReference } from 'salvageunion-reference'
 import { ReferenceEntityIsland } from './islands/ReferenceEntityIsland'
+import { EntityCardStatic } from './EntityCardStatic'
 import { StaticEntityContent } from 'component-lib'
+import { cardNeedsHydration } from '../lib/cardNeedsHydration'
 import { extractStaticEntitySummary } from '../lib/gameData'
 import { patternStaticSummary } from '../lib/patternSummary'
 import { getSchemaPreloadList } from '../lib/schemaPreloadDeps'
 import { resolveTraitHref } from '../lib/staticLinks'
 
 /**
- * EntityView — the reference entity as ONE render unit: the interactive island
- * card plus its own static SEO / no-JS sub-content, rendered alongside it.
+ * EntityView — the reference entity, rendered the lightest way it can be.
  *
- * The static block is the Entity's sub-content, not a separate page concern:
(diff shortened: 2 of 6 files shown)


alxjrvs (commented):
## Review: `srd-ssr-entity-card` (3 commits vs `main`)

The bulk of the diff — CI/tooling hardening, tsconfig hole-closing, dependency dedupe floors, the worktree-reaper fix, the `?? 0` regression-detection fixes, the `interface`→`type` sweep — all reads correctly. `renovate.json`'s deletion is fine (`.github/dependabot.yml` already covers both ecosystems). `tools/check-convex-codegen.ts` runs green and its module walk matches Convex's registration rules for the current tree.

The SSR entity-card change is where the problems are. `cardNeedsHydration` decides "does this need JS?" by regexing the rendered markup for `<button|<input|<select|<textarea|role="button"|aria-expanded=|aria-haspopup=`. That catches roll-table buttons and card-click wrappers, but it misses three whole classes of JS-dependent markup. I confirmed each by preloading the ORM and running the real probe over 561 entities (500 classified static, 61 hydrating) — all srd tests stay green, which is exactly why these slipped through.

### Findings

- `apps/srd/src/lib/cardNeedsHydration.tsx:14` — Artwork renders at `opacity:0` forever on 29 static pages (28 chassis + the Wastelander NPC). `CardImage` only flips `loaded` via `onLoad`/effect, which never runs without hydration. `onError`'s hide-on-404 is dead too.
- `apps/srd/src/lib/cardNeedsHydration.tsx:15` — Keyword/trait tooltips are dead on 218 of the 500 static pages. `EntityTooltip`'s keyword branch uses `SimpleTooltip`, whose trigger is a bare `<span style="cursor:help">` with no `role="button"` — so the regex misses it and the page ships a help cursor that opens nothing.
- `packages/component-lib/src/components/referenceEntity/card/CatalogChoiceListing.tsx:106` — The read-only catalog `<details>` gates its children behind React state, so on static pages it renders literally `<details><summary>Armament Bay Weapons System</summary></details>`; the native disclosure opens to an empty box. Affects Armament Bay and Industrial Body Kit.
- `apps/srd/src/components/EntityView.astro:50` — Dropping `StaticEntityContent` on the static branch removes the page's only internal links. `.50 Cal Machine Gun`'s card emits zero `<a href>`; the removed fallback emitted three trait links. `resolveTraitHref` is now unreachable for those 500 pages.
- `apps/itun/src/components/shared/GameDataReady.tsx:63` — Module-scope `preload('all').then(...)` is rejectable before any `use()` attaches a handler, so a failed preload during React bootstrap surfaces as an unhandled rejection (double-reported to Sentry, without component context).

The cleanest fix for the first three is to widen `NEEDS_A_LISTENER` (or better, make the probe assert on a positive allowlist rather than a denylist of tags) — but note that adding `<img` and `<details` would push the chassis and bay pages back onto the island path, shrinking the "82%" claim. The alternative is to make `CardImage` and `CatalogChoiceListing` degrade correctly without JS, which is the more durable answer since the card is now a no-JS surface by design.

alxjrvs (commented):
Reviewed commit `83bc6992` (the branch's only commit vs `origin/main`; local `main` was stale and the upstream diff was empty). I verified the two top findings empirically by rendering all 850 entities through `cardNeedsHydration` + `EntityCardStatic` with the real ORM, not by reading alone. `bun --filter srd typecheck` and the changed test file both pass.

Findings:

- `apps/srd/src/lib/cardNeedsHydration.tsx:15` — `NEEDS_A_LISTENER` misses Base UI tooltip triggers, so 418 of the 672 statically-rendered cards ship a hovercard that can never open (280 of them the `cursor: help` keyword variant, e.g. `Uses (3)` on First Aid Kit). The keyword description is portal-rendered, so it isn't in the HTML either — content that worked on hover before this commit is now unreachable. The non-keyword `EntityTooltip` branch emits `role="button"` and *is* caught; the `isKeyword` branch emits a bare `<span>` and isn't.
- `apps/srd/src/lib/cardNeedsHydration.tsx:15` — the regex matches nothing React emits for `<details>`/`<summary>`, so `CatalogChoiceListing`'s read-only disclosure (body gated on `useState`) is classified static and opens to an empty box. Confirmed on `systems/Industrial Body Kit`; 4 pages. This is precisely the dead-control class the probe was built to exclude.
- `apps/srd/src/components/islands/__tests__/hydration-directives.test.ts:42` — `slice(indexOf('EntityCardStatic'))` becomes `slice(-1)` if the marker disappears, so "the static path carries no client directive" passes vacuously once there is no static path.
- `apps/srd/src/components/EntityView.astro:46` — the new `await preload(...)` can never load anything (`lib/gameData` already `preload('all')`s at module scope), but `preload()` unconditionally invalidates the action map and search index, so both are dropped ~1,004 times during the build.

Two things I checked and cleared: every static page emits exactly one `<h1>` (0 missing, 0 duplicated), and no static page emits an `<a>` without an `href` — the `cursor-pointer` cards on chassis pages are all wrapped in real anchors.

PR #689, opened 2026-08-05
feat(srd)!: migrate off Astro to an in-house Vite SSG

alxjrvs:
Implements **Plan B** of the "One Stack, Two Sites" migration plan: `apps/srd` leaves Astro for an in-house static-site generator built on Vite.

## Why — and one correction to the plan

The plan flagged a five-minute question with a 20-day consequence: *had a newer `@astrojs/check` added TypeScript 7 support?* If so, Option A closes the only gap Astro causes and this work is unnecessary.

**Checked: it has not.** `@astrojs/check@0.9.10` — the latest release, published 2026-07-27 — still declares `peerDependencies: { typescript: "^5.0.0 || ^6.0.0" }`. The TS 6 pin was unfixable locally, so Plan B is the honest route.

A second plan item was **half wrong**. Phase 4 says "delete the `typescript-classic` alias and the 6.0.3 pin". Only the pin can go: `tools/check-architecture.ts` and `packages/salvageunion-reference/tools/generateApiReport.ts` both import `typescript-classic` for the TS 6 compiler API. Astro was never its only consumer. The root alias stays, and now says why.

## The decision that made this tractable

**Islands mount with `createRoot`, never `hydrateRoot`.**

`useGameData`'s server snapshot is hardcoded `false` on purpose, so a hydrating card renders its fallback while the server rendered the real thing — the React #418 mismatch that `cardNeedsHydration` (#685) already exists to dodge. Client-only mounting has no mismatch class *at all*: placeholder markup is discarded and replaced.

That turns server-rendering an island into an **opt-in SEO choice per island** rather than a hydration contract, and removes the single largest correctness risk in the plan.

## Verification

`ssg/parity.ts` is the acceptance gate. It compares the new `dist` against an Astro baseline build semantically — emitted file set both directions, per-page `<title>`/description/canonical/`og:*`/`twitter:*`/robots, every JSON-LD block parsed and deep-compared, `<main>` visible text, all 899 JSON endpoints deep-compared, and byte-identical `llms.txt`.

```
PARITY OK — 1039 pages, 899 JSON endpoints, zero differences.
```

It passes the baseline against **itself**, and fails against **eight deliberately-injected defects** plus a synthetic Astro-dialect-vs-SSG-dialect fixture — so it demonstrably bites rather than merely passing. `bun run check:all` exits 0.

| | Astro | Vite SSG |
|---|---|---|
| HTML payload | 48.1 MB | **18.2 MB** (−62%) |
| A typical entity page | 54.5 KB | **26.4 KB** (−52%) |
| Build | 6.8s | **2.4s** |
| TypeScript | pinned 6.0.3 | **7.0.2** |
| Biome coverage of srd | 17 files excluded | **full** |

The payload win is `MobileNavIsland`: Astro inlined the same 16.6 KB catalog blob into all 1,039 pages (17.3 MB). It now computes its own props in its own chunk — one copy.

## Two bugs this found before shipping

**Sentry would have gone silently dark in production.** Vite only inlines `VITE_`-prefixed env, so `import.meta.env.PUBLIC_SENTRY_DSN` compiled to `undefined` — SDK initialised, reporting nothing, while build, bundle and deploy all looked healthy. Exactly the failure mode `CLAUDE.md` warns is silent. Fixed with `envPrefix: 'PUBLIC_'`, proven by grepping the DSN out of the emitted bundle.

**The bundle-budget guard was passing vacuously.** `bundle-budget.e2e.ts` filters on `/_astro/`, which Vite never emits — so the PR-blocking payload guard summed 0 bytes and every assertion passed while guarding nothing. Repointed, and given a non-empty-capture assertion first, since every other assertion in that file is "absent" or "under N" and a dead selector satisfies both.

## Review

An adversarial review ran four lenses (correctness, SEO/a11y, ownership, ops); each finding was then independently refutation-tested. **32 raised, 3 refuted, 29 confirmed** — all confirmed findings are fixed in `docs+fix`, including a `@view-transition` rule that two comments described as `ClientRouter`'s replacement but which had never actually been written, and `e2e-nightly.yml` still invoking the now-uninstalled `astro` binary.

## What this costs

~2,400 lines under `ssg/` that this repo now owns forever, against an Astro layer that cost it **three commits in its entire history**. ADR-031 states that plainly rather than selling it as free. The compiler dependency is the only thing that justifies it.

## Reviewing this

Start at `apps/srd/ssg/DESIGN.md` — the contract everything implements. Then:

```bash
bun install && bun run build:package
cd apps/srd && bun ssg/build.ts && bun ssg/parity.ts
```

ADR-031 supersedes ADR-012, and ADR-012's own `## Status` header now says so — several ADRs here record supersession only there.

https://claude.ai/code/session_019kVVMrzRkXMGpvcB44HyJM


alxjrvs (commented):
Reviewed `git diff origin/main...HEAD` (4 commits, 107 files) — the Astro → in-house Vite SSG migration for `apps/srd`. I read every hunk and also verified behaviour empirically: `bun ssg/build.ts` (1039 pages, 900 endpoints, 2.5s), `bun ssg/parity.ts` (**PARITY OK, zero differences**), `bun --filter srd typecheck`, `bun --filter srd test` (1084 pass), `bun run knip`, `bun run validate:all` — all green. I additionally drove the built `dist` through headless chromium: every island mounts (`SearchIsland`, `MobileSearchIsland`, `MobileNavIsland`, `ReferenceEntityIsland`, `SchemaViewerIsland`, `SearchResultsIsland`, `ColophonIsland`), no console/page errors, `.js` class applied, `[data-static-fallback]` removed after hydration, mobile drawer renders all six catalog sections, and the dev server serves all page + endpoint families. I also diffed head tags and non-`<main>` body chrome against the Astro baseline directly (only Astro-specific metas and `'`→`&#x27;` entity differences).

The migration itself is sound. Findings:

- `apps/srd/e2e/bundle-budget.e2e.ts:91` — `forbidden` chunk prefixes still use Astro's `name.hash.js` dot separator; Vite emits `name-hash.js`, so every "must not load" assertion passes vacuously. Measured: `/schema/traits/item/missile/` now requests `guides-xyPMdOvY.js` + `catalog-categories-CPgqFnVB.js` (MobileNavIsland's new `useGameData`), so with the separator fixed both this test and line 111 would fail — the dead guard is masking a real violation of its own invariant.
- `apps/srd/ssg/parity.ts:26` — `DEFAULT_BASELINE` hardcodes `/Users/jarvis/.claude/jobs/dbab36ff/tmp/baseline-dist`, an ephemeral agent-job temp dir; the documented "acceptance gate" exits 2 for anyone else, and Astro is deleted so the baseline can't be trivially rebuilt.
- `apps/srd/ssg/parity.ts:27` — `DEFAULT_CANDIDATE` uses `new URL(...).pathname` instead of `fileURLToPath`; percent-encoded under any path with a space, unlike every other `ssg/` module.
- `apps/srd/src/components/islands/MobileNavIsland.tsx:30` — `categories`/`currentPath` kept optional "so the (still-live) Astro `TopNavigation.astro` path … keep passing build-time values", but that file is deleted in this PR and no test passes them.
- `apps/srd/e2e/bundle-budget.e2e.ts:24` — comment says the old prefix was `/assets/`; it was `/_astro/`, in the very comment meant to prevent the bug above.

alxjrvs (commented):
Reviewed `origin/main...HEAD` (115 files, ~9.6k insertions) — the Astro → in-house Vite SSG migration for `apps/srd`. I read every hunk of the new SSG (`ssg/build.ts`, `render.tsx`, `document.tsx`, `dev.ts`, `preview.ts`, `endpoints.ts`, `sitemap.ts`, `pwa.ts`, `outputPath.ts`, the island runtime, all 14 ported page modules and 5 endpoints), then verified empirically: ran the build (1039 pages, 900 endpoints, 0 errors), `bun --filter srd test` (1152 pass), `bun run check:fast` (all green), started the dev server and exercised 13 route shapes, drove the built site in headless Chromium (island mounting, og-card, about lightbox, search, 404), and ran both PR-gating e2e specs.

Findings:

- `apps/srd/e2e/bundle-budget.e2e.ts:106` — Leaf-schema entity pages now load 703 KB of eager JS against the repo's 550 KB budget; the PR-blocking `e2e-web` CI job fails on this branch's own build. Cause: the single `islandRegistry.ts` entry collapses per-island tree-shaking into one 491 KB shared component-lib chunk that all three `client="idle"` chrome islands pull on every page (Astro's baseline was ~277 KB). This PR fixed the `/_astro/` → `/assets/` selector that had made the guard dead, but left the guard red.
- `apps/srd/ssg/dev.ts:238` — `decodeURIComponent` is outside `serve()`'s try/catch, so a request like `/%` throws a URIError that escapes `void serve(...)` and kills the dev-server process. Reproduced: server exits, all later requests are connection-refused. `ssg/preview.ts:47` guards the identical call correctly.
- `apps/srd/ssg/parity.ts:39` — The declared acceptance gate is wired into no package script, hook, or workflow, and its baseline dir is gitignored, absent, and unregenerable now that Astro is out of the lockfile. The "zero differences across 1,039 pages" claim is unverifiable by anyone else, and future `ssg/**` changes have no regression gate.
- `apps/srd/ssg/DESIGN.md:90` — The "contract" doc's inlined `types.ts` omits `PageResult.shell`, `RouteContext.builtAssets` and `DocumentShell`, and line 234 still names the per-machine absolute baseline path `parity.ts` deliberately removed.
- `.github/workflows/ci.yml:406` — `e2e-web`'s comment still describes `astro build` / `astro preview`; the sibling comments in `e2e-nightly.yml` were updated, these two were missed.

Things I checked that were clean: URL→file mapping and the `/404` special case, dotted-endpoint vs page pattern precedence in dev, path-traversal containment in `preview.ts`, sitemap count/filter/ordering (1036), island placeholder/props/registry coverage across the whole dist, `envPrefix: 'PUBLIC_'` (no `VITE_*` readers exist), the NODE_ENV save/restore around `viteBuild`, og-screenshot meta rewriting, and page-content fidelity (word-level diff of every ported page against its `.astro` original — differences were only dynamic content my extraction stripped).

alxjrvs (commented):
## Findings

- **`apps/srd/ssg/vite.config.ts:48`** — `manualChunks` only splits react-vendor, so all five island chunks statically import one 491 KB `src-*.js`; the three `client="idle"` chrome islands pull it onto *every* page. `/schema/traits/item/missile/` now loads **703 KB** of JS vs the 550 KB budget, so `e2e/bundle-budget.e2e.ts:106` **fails** — and `e2e-web` feeds `quality-checks` (ci.yml:517), the single required check. Verified by running the spec against the real build: 6 passed, 1 failed.
- **`apps/srd/src/runtime/islands.client.ts:84`** — `el.replaceChildren()` runs a task before React's first commit, so `ssr: true` islands paint an empty frame. Measured on `/about/`: colophon goes 487px/1269 chars → **0px/0 chars** → 487px within ~14 ms. Astro's `hydrateRoot` had no such flash.
- **`apps/srd/src/pages/schema/[schemaId]/index.page.tsx:81`** — `ssr` on `SchemaViewerIsland` emits 9 loading skeletons, not the entity grid: `<main>` of `dist/schema/traits/index.html` contains 6 characters of text and no entity names. The stated SEO rationale (here and in `ssg/DESIGN.md`'s per-island table) is not delivered, so the flag is pure cost plus exposure to the wipe-flash above.
- **`apps/srd/src/runtime/assets.entry.ts:29`** — the 1.8 MB `eldridge-coast-map.png` is still committed but imported by nothing after the switch to the pre-encoded `.webp`; knip doesn't cover assets.

Everything else checked out: `bun ssg/build.ts` succeeds (1039 pages, 900 endpoints, 1036 sitemap URLs, 27 MB dist), srd tests 1152/1152 pass, `check:fast` is fully green, the dev server renders routes/endpoints/404 correctly through the prod path, and `ssg/preview.ts` maps URLs correctly and rejects both `..` and `%2e%2e` traversal. The page ports are faithful to their `.astro` originals, and the `_astro/` → `assets/` prefix fix in the bundle-budget spec is correct (chunk names confirmed as `guides-HASH.js` etc.).

alxjrvs (commented):
Two review agents finished after the report was rendered. They surfaced findings the original three didn't cover; I verified the two concrete ones myself.

**New — real defect, `apps/srd/src/pages/discord.page.tsx:136`**

```
Looks up any entity in the reference — the same search as this site&rsquo;s
<kbd …>Cmd&nbsp;K</kbd>{' '}
```

JSX strips the newline-adjacent whitespace before an element, so the built page reads **`this site’sCmd K box.`** Confirmed in `dist/discord/index.html`:

```
this site’s<kbd class="rounded-card border …">Cmd K</kbd> box.
```

Note the asymmetry in the source: there's a `{' '}` *after* the closing tag but none before the opening one. Fix is appending `{' '}` to line 136.

The more important part: **`ssg/parity.ts` structurally cannot catch this class.** `normalizeText` (`ssg/parity.ts:403`) substitutes a space for every tag, so a missing space adjacent to a tag always normalizes into a match. A site-wide regex sweep of built HTML for the same shape found this as the only occurrence — but the gate would never have told you.

**New — pre-existing, now living in added files: `apps/srd/src/pages/api.page.tsx:116,204,210` and `src/endpoints/llmsTxt.ts:60`.** The documented example slug `iron-mongrel` doesn't exist — the dataset has `iron-wyrm`. Verified: `dist/schema/chassis/item/iron-mongrel` is absent, so the API page's `target="_blank"` example link and the URL advertised to LLM consumers in `llms.txt` both 404. It's verbatim from the deleted `api.astro`/`llms.txt.ts`, so it's a faithful port of an existing bug rather than a regression.

**Restated with sharper consequence — `apps/srd/src/components/TopNavigation.tsx:43`.** All three chrome islands are `client="idle"` with `Island`'s default `ssr={false}`, where Astro's `client:idle` server-rendered them. `dist` now contains `<div data-island="SearchIsland" …></div>` — empty. Crawlers and no-JS clients lose the header search box and the entire mobile nav catalog link list that previously existed in the HTML of all 1,039 pages, and on a slow connection the header paints without search until the idle callback fires. Parity only diffs `<main>`, and `NON_SSR_ISLANDS` strips these on both sides, so nothing in the gate sees it. Likely intentional (it's part of what kills the 16.6 KB per-page catalog blob) — worth an explicit confirm rather than an inference.

One correction to the second agent's report: the claim that `src/assets/eldridge-coast-map.png` is left unreferenced is wrong — `src/assets/` contains only the `.webp`. The `srcset` finding itself stands.

Both agents independently re-verified the green results already in the report (1,039 pages, 900 endpoints, 1,036 sitemap URLs, `registerSW.js` revision hash matching the baseline, 1,152 tests, typecheck/knip/`check:fast` clean), and both confirmed `src/lib/staticPaths.ts` is byte-identical to `origin/main`.

Nothing is fixed — say the word and I'll land the `{' '}` and, if you want, the `iron-mongrel` → `iron-wyrm` correction.

alxjrvs (commented):
## Findings

- **`apps/srd/src/runtime/islandRegistry.ts:31` — one island entry hoists all island code into a single 491 KB shared chunk; every page now ships 703,506 bytes of JS and the PR-blocking `bundle-budget.e2e.ts` spec fails (`Expected: < 550000 / Received: 703506`).** Confirmed by building `dist`, serving it with `ssg/preview.ts`, and running the spec. On `/schema/traits/item/missile/` — a page that deliberately does *not* hydrate its entity card — the loaded chunks are `src-CajHcoSg.js` (490,984 B, containing `ReferenceEntityCard`/`sonner`/`@base-ui` used only by ReferenceEntityIsland, SchemaViewerIsland and OgCardIsland) plus `react-vendor` (189,942 B). Because all eight islands are dynamic imports from one entry, Rollup cannot tell which co-occur on a page, so everything shared by ≥2 of them lands in one chunk that every page downloads. `/`, the leaf trait page and the chassis item page all transfer the identical 703,506 B. `.github/workflows/ci.yml`'s `e2e-web` job runs this spec against exactly this build+preview pair, so CI fails as-is. Fix direction: per-island entries, or a `manualChunks` rule in `ssg/vite.config.ts`.
- **`apps/srd/ssg/preview.ts:52` — `startPreview({ dir })` with a trailing slash 404s every request.** `normalize` keeps the trailing separator, so the `target.startsWith(dir + sep)` containment check can never pass. Verified live: default `dir` → 200 for `/about/`; `dir` with a trailing slash → 404. Latent (no in-repo caller passes `dir`), but `dir` is a documented option on an exported function and `parity.ts` uses that same trailing-slash spelling for its own dirs.

## What I verified as sound

Everything else held up under direct exercise, not just reading: `bun --filter srd test` (1152 pass), `bun ssg/build.ts` (1039 pages / 900 endpoints / 1036 sitemap URLs), `bun run knip`, `bun run validate:all` (all guards green including the new doc-drift script check and convex-codegen), `smoke.e2e.ts` and `json-api.e2e.ts` (pass), and manual probes of `ssg/dev.ts` (endpoint-before-route ordering, dotted patterns, 404, malformed `%` URL) and `ssg/preview.ts` URL→file mapping. The `</head>`/`</body>` string surgery in `document.tsx` is safe because `escapeJsonForScript` escapes `<` in every interpolated payload; the island props script is emitted before `</body>` and read by a deferred module script, so the ordering is correct; `su-assets`' new Sentry gate and 503-on-Blobs-failure path are correct and well tested.

One note not filed as a finding: `ssg/parity.ts` — the documented acceptance gate — cannot run without an Astro-era baseline that is gitignored and no longer producible from this branch, and it is not wired into CI. It is recoverable from git history, but the docs' "trust it over any agent's opinion" is currently unactionable for anyone who doesn't already hold the artifact.

alxjrvs (commented):
Four findings across the branch. The two new ones (su-assets + tooling) are the more consequential:

- **`asset.ts:63`** is the only one that changes shipped behavior: the new Sentry wrapper converts a `URIError` from `decodeURIComponent` into a 500 + an event, on exactly the scanner-probe paths the file's own comment says must stay silent as 404s.
- **`lefthook.yml:54`** is a newly-introduced write/read race — `check:schemas` regenerates `lib/index.ts` and friends while `typecheck`/`test`/`lint`/`validate`/`knip` read them in the same `parallel: true` group. Intermittent, non-reproducing pre-push failures on a clean tree.

The two from the SSG review are guard-coverage gaps rather than defects in the migration itself. Empirically the migration holds up: build is 1039 pages / 900 endpoints in 2.6s, `bun --filter srd test` is 1152 pass, parity reports zero differences, and the `ssr={false}` zero-area island placeholder still mounts under `client="visible"` in headless chromium.

PR #690, opened 2026-08-05
feat(observability): report su-assets Blobs failures to Sentry, document the Convex route

alxjrvs:
Closes the two SU-SRD surfaces that had no error tracking at all: the artwork host and the Convex backend. One of them needed code; the other needed a decision and a runbook.

## `apps/su-assets` — wired

`assets.salvageunion.io` serves every entity-card image in **both** srd and itun, and until now its failures were completely invisible — a Blobs outage surfaced as an unhandled 500 that nobody was watching, and the artwork simply stopped resolving in two apps at once.

- **`netlify/functions/_observability.ts`** — a near-verbatim mirror of `apps/itun/netlify/functions/_observability.ts`. `@sentry/node` pinned to `10.68.0`, identical to `itun`, `srd` and `discord-bot`. Entirely env-gated on `SENTRY_DSN`: with no DSN, `initObservability()` returns before `Sentry.init` and `captureException` is a no-op, so local dev, CI and forks ship and run exactly as before. No DSN is committed — it is already provisioned on the `su-assets` Netlify site.
- **`asset.ts`** — opening the store and reading from it are now one try/catch. That single catch covers both real failure modes: `getStore` throwing (missing Blobs runtime context or store binding — a deploy-shaped failure) and `get` rejecting (a Blobs outage). Both report with `{ fn, op: 'blobs.get', key }` and return a controlled **503** instead of an unhandled 500, mirroring `snapshot-retrieve.ts`. The default export calls `initObservability()` per cold start and catches anything unanticipated as a reported 500.

### What is deliberately NOT reported

**Every 404.** This function answers `/*` on a public, crawler-visible host, so all of these are ordinary outcomes, not errors:

| Outcome | Why it stays silent |
| --- | --- |
| Key absent from the store | A missing asset is a normal result, not a fault. |
| Extension outside the allowlist (`.txt`, `.html`, `.js`, extensionless) | Client asked for something this host does not serve. |
| Traversal probe (`..%2F…`) | Scanner traffic. Constant, and the guard already handled it. |
| Dotfile probe (`/.env.png`) | Same. |
| Site root / empty key | Same. |

Alerting on those is how the channel becomes noise — at which point the Blobs outage above is invisible again, which is the exact problem this PR exists to fix. There are tests asserting the silence, not just the reporting.

### Tests

`__tests__/_observability.test.ts` in the style of `apps/discord-bot/src/__tests__/observability.test.ts` — mocks `@sentry/node` and **restores it in `afterAll`** (`mock.module` is process-global in Bun), and deletes `SENTRY_DSN` before restoring so no later file inherits a live DSN. Covers the gate in both directions, including that it latches on the first init.

`__tests__/asset.test.ts` gains an error-reporting block. The reporter is injected through `makeAssetHandler`'s new second parameter — the same dependency-injection seam the store already used, chosen over `mock.module()` for the same reason.

## Convex — documentation, not code

**Decision: no application code. This is a Convex dashboard toggle plus a runbook.**

Convex has a first-party [Exception Reporting integration](https://docs.convex.dev/production/integrations/exception-reporting) that needs no application code. That is not merely the tidier option — it is the only one that covers the surface:

- **Queries and mutations cannot report from inside a function.** They run in Convex's deterministic runtime, which has no `fetch` and no network egress by design. That is most of `convex/` (`games.ts`, `invites.ts`, `proposals.ts`, `crew.ts`, `ownership.ts`, …), so a code-level SDK could never see the bulk of the errors we would be adding it for.
- **`@sentry/node` does not run there either.** The default Convex runtime is not Node, and this deployment has **no `'use node'` actions** (verified) — so adding the Node SDK would mean converting modules to the Node runtime purely to instrument them.
- **What is left is HTTP actions** (`http.ts`, `botHttp.ts`), where a hand-rolled `fetch` to Sentry's ingest endpoint would duplicate a built-in that already tags events with function name, function type, runtime, request id, deployment name, environment tier, and the caller's `tokenIdentifier` — none of which application code can reconstruct.

Documented in `docs/architecture/accounts-and-games.md` §6 (a new "Convex error reporting — a dashboard toggle, not code" subsection carrying the rationale and the enable-it steps), with a short "do not add a Sentry SDK to `convex/`" note in `apps/itun/CLAUDE.md` pointing at it.

## Human steps still required

1. **Convex (blocking, cannot be scripted):** create a Sentry project in `susrd` — EU region, platform **Node.js** (suggested slug `itun-convex`; confirmed absent today, the org has `srd`, `itun`, `itun-functions`, `su-discord`, `su-assets`), then Convex dashboard → Settings → Integrations → Sentry → paste the DSN. **Per deployment** — `dev/alex-jarvis` and `exuberant-porpoise-183` are configured separately.
2. **Convex caveat:** Exception Reporting is a **Convex Pro** feature. On the free plan the Sentry card is unavailable and Convex errors stay visible only in the Convex dashboard's function logs. The docs say so plainly rather than papering over it — and per the reasoning above, code would not be a workaround.
3. **su-assets:** nothing. The DSN is already provisioned on the Netlify site; it starts reporting on the next deploy.

## Follow-up (not done here, deliberately)

`tools/check-observability.ts` is untouched — it is being changed in in-flight #689. It currently guards the two browser apps' module→entry→CSP wiring. It could reasonably be extended to assert that `apps/su-assets`'s function imports its observability module (the same "the module exists but nothing calls it" failure it already catches elsewhere), and that `@sentry/node` stays version-locked across the now-four workspaces that pin it. Worth doing once #689 lands.

## Verification

`bun run check:all` exits **0** (lint, format:check, typecheck across all 6 workspaces, tests, validate:all, knip, audit, tokens, styling). `bun run knip` exits **0** with no report — the new `@sentry/node` dependency is not flagged. su-assets: **32 pass, 0 fail** (up from 17).

https://claude.ai/code/session_019kVVMrzRkXMGpvcB44HyJM

### DIFF
--- a/apps/itun/CLAUDE.md
+++ b/apps/itun/CLAUDE.md
@@ -106,6 +106,15 @@ and remains the account-free way to share a build.
   ([ADR-010](../../docs/adrs/ADR-010-srd-choices-ephemeral-vs-persisted.md)).
 - Backup nudge (`src/lib/backupNudge.ts`) tracks un-exported writes — the
   local-first analogue of durability.
+- **Do not add a Sentry SDK to `convex/`.** The browser bundle
+  (`src/lib/observability.ts`) and the Netlify Functions
+  (`netlify/functions/_observability.ts`) each own one; Convex instead uses its
+  first-party Exception Reporting integration, enabled in the Convex dashboard
+  with no application code. Queries and mutations run in a deterministic runtime
+  with no `fetch`, so an in-function SDK could not report from them at all —
+  rationale and the enable-it runbook are in
+  [docs/architecture/accounts-and-games.md](../../docs/architecture/accounts-and-games.md)
+  ("Convex error reporting — a dashboard toggle, not code").
 
 ## Commands
 
--- a/apps/su-assets/netlify/functions/__tests__/_observability.test.ts
+++ b/apps/su-assets/netlify/functions/__tests__/_observability.test.ts
@@ -1,0 +1,126 @@
+/**
+ * _observability — the artwork host's only channel for finding out that
+ * production broke. It is entirely env-gated, and the gate is the interesting
+ * part: with no DSN every export must be a silent no-op (local dev, CI and
+ * forks all run that way), and with one it must actually report. A regression
+ * in either direction is invisible until the day you need it — a host that has
+ * stopped reporting looks exactly like a host with no errors.
+ *
+ * `@sentry/node` is mocked and **restored in `afterAll`**: `mock.module` is
+ * process-global in Bun, not file-scoped, so leaving it in place would hand
+ * this fake to `asset.test.ts` too. The DSN is driven through `process.env`
+ * because that is what the module actually reads, and it is deleted again
+ * before the mock is restored so no later file inherits a live DSN.
+ *
+ * Assertion style: toBe() / toEqual() — not toBeInTheDocument().
+ */
+
+import { afterAll, beforeAll, describe, expect, mock, test } from 'bun:test'
+
+const sentryCalls: Array<{ fn: string; args: unknown[] }> = []
+
+const realSentry = { ...(await import('@sentry/node')) }
+
+mock.module('@sentry/node', () => ({
+  init: (...args: unknown[]) => sentryCalls.push({ fn: 'init', args }),
+  captureException: (...args: unknown[]) => sentryCalls.push({ fn: 'captureException', args }),
+}))
+
+let obs: typeof import('../_observability')
+
+beforeAll(async () => {
+  obs = await import('../_observability')
+})
+
+afterAll(() => {
+  delete process.env.SENTRY_DSN
+  delete process.env.COMMIT_REF
+  mock.module('@sentry/node', () => realSentry)
+})
+
+function calls(fn: string): Array<{ fn: string; args: unknown[] }> {
+  return sentryCalls.filter((c) => c.fn === fn)
+}
+
+/**
+ * These run in order against one module instance, because `initialized` /
+ * `enabled` are module-level state and the transition from off to on is
+ * exactly what is being tested.
+ */
+describe('with no DSN configured', () => {
+  test('init reports nothing and does not enable Sentry', () => {
+    delete process.env.SENTRY_DSN
+    obs.initObservability()
+    expect(calls('init')).toHaveLength(0)
+  })
+
+  test('captureException is a no-op rather than an unconfigured send', () => {
+    obs.captureException(new Error('boom'))
+    expect(calls('captureException')).toHaveLength(0)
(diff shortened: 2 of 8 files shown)


alxjrvs (commented):
Reviewed the actual PR scope (`origin/main...HEAD` — local `main` was stale, so the range is the single commit `d1890e33`, 8 files). Ran the workspace suite (32 pass), root typecheck, `knip`, and `validate:all` — all green, so the findings below are things the gates cannot see.

**Findings**

- `apps/su-assets/netlify/functions/asset.ts:63` — `decodeURIComponent` throws `URIError` on malformed percent-encoding and sits outside every try block, so `GET /foo%.png` (or `/%`, `/%E0%A4%A.png`) escapes `makeAssetHandler`, hits the new outer catch, and returns a **500 plus a Sentry event** instead of the intended silent 404. Verified empirically against the real handler — all three probes reject before the traversal/extension guards run. On a host with `config.path = '/*'`, routine scanner traffic will page you, which is the precise outcome the PR's own reporting taxonomy (the six-case `silent` table in `asset.test.ts`) is written to prevent.

- `apps/su-assets/netlify/functions/_observability.ts:43` — no `Sentry.flush()` before the serverless response returns. `captureException` only enqueues on the async transport; `asset.ts` returns the 503 immediately and Lambda freezes the container with the event still buffered, so the Blobs-outage report this PR exists to deliver can be dropped (most likely on a cold start — the outage case). The repo already documents this exact hazard and handles it in `apps/discord-bot/src/observability.ts` via `flushObservability()`. The same gap exists in `apps/itun/netlify/functions/_observability.ts`, the module this was mirrored from, so a fix should cover both.

**Checked and clean:** the DI seam and 404-vs-503 split in `asset.ts`; the `mock.module` restore discipline in `_observability.test.ts` (correctly spreads the namespace at capture time and deletes `SENTRY_DSN`/`COMMIT_REF` in `afterAll`, so nothing leaks into `asset.test.ts` — confirmed by running discord-bot, su-assets and itun function tests in one process); the `?fresh` cache-bust (fails loudly rather than silently if Bun ever normalizes it); knip's `includeEntryExports: false` for su-assets; `@sentry/node` pinned to `10.68.0` matching itun and discord-bot with no new lockfile entries. The doc claims in `docs/architecture/accounts-and-games.md` verify out — `convex/` has no `'use node'` modules, and `http.ts`/`botHttp.ts` are the only `httpAction` surfaces.

One non-blocking note: `tools/check-observability.ts` covers only `BROWSER_APPS`, so nothing detects an unprovisioned `SENTRY_DSN` on the su-assets site. That is a pre-existing gap for all server-side surfaces (itun's Functions and discord-bot are equally uncovered), not a regression, so I did not file it.

PR #693, opened 2026-08-05
ci: arm auto-merge on release-please PRs so releases stop stalling

alxjrvs:
## The actual bug

release-please is **healthy**. It ran successfully on every push to `main` and kept its open PRs current — #688 was refreshed at 17:54, two minutes after #689 merged at 17:52.

**Nothing ever merged its PRs.** That's the entire defect:

| PR | Release | Open since | Checks | Mergeable |
|---|---|---|---|---|
| #688 | `srd 2.0.0` | Aug 5 | all green | yes |
| #677 | `itun 1.0.0` | Aug 4 | all green | yes |

So `main`'s `CHANGELOG.md` stops at Jul 29 (`srd`) / Aug 3 (`itun`), and the Releases list matches. There is no gap in the *generation* — the lag is purely unmerged release PRs, and it grows until someone clicks merge by hand.

[ADR-024](docs/adrs/ADR-024-derived-release-changelogs.md) already specified that release PRs "land via the existing squash-only + auto-merge flow." That half was never built — nothing enables auto-merge on a PR unless something asks it to.

## The fix

One step in `release-please.yml` that re-arms auto-merge on every open `release-please--*` PR.

Three choices worth reviewing:

- **Not gated on `prs_created`.** That output is false whenever release-please merely *refreshes* an already-open PR — the common case. Gating on it would arm auto-merge once and never notice if it lapsed. Re-asserting every run is idempotent and self-healing.
- **PAT, not `GITHUB_TOKEN`.** `main` requires a merge queue, and `GITHUB_TOKEN` cannot add a PR to one. `GH_REPO` is set because this job never checks out the repo, so `gh` has no remote to infer.
- **Added `schedule:` + `workflow_dispatch:`.** Both release PRs edit `.release-please-manifest.json`, so whichever merges second gets rebased — and a PR ejected from the merge queue loses auto-merge silently. On `push:` alone it would sit dead until someone merged something unrelated. The daily run re-arms it unattended; `workflow_dispatch` is the manual kick.

A failed arm marks the run red rather than passing quietly, and one bad PR does not skip the others.

## Verification

- Workflow YAML parses; the new step's `run` body is **shellcheck-clean**.
- The `jq` selector was run against **live** repo data: it returns exactly `688` and `677`, and correctly excludes the two open Dependabot PRs.
- Prettier clean on both files. No source files touched, so no typecheck/test surface.
- `merge_group:` is already wired in `ci.yml`, so queued PRs do get their required `quality-checks` — the queue is not the blocker.

## ⚠️ Read before merging

When this lands, release-please runs and arms **#688 (`srd` 2.0.0)** and **#677 (`itun` 1.0.0)** — both **major** bumps, `srd`'s carrying the breaking Astro→Vite SSG migration. They will merge and publish on their own once green. That is the requested behavior, but it is the first thing this change does, so it should be a conscious call rather than a surprise.

If you'd rather majors stay manual, say so and I'll add a guard that skips arming when the PR title implies a major bump.

## Out of scope (noted, not fixed)

- Dependabot PRs #691/#692 sit open for the same class of reason — no auto-merge workflow, and the merge queue means `GITHUB_TOKEN` couldn't enqueue them anyway.
- `main` carries **both** a ruleset and classic branch protection. GitHub takes the union, so editing one leaves the other silently in force.

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -2,6 +2,14 @@ name: release-please
 on:
   push:
     branches: [main]
+  # Self-heal without needing a push. Both open release PRs edit
+  # `.release-please-manifest.json`, so whichever merges second gets rebased —
+  # and a PR ejected from the merge queue loses its auto-merge. On push alone,
+  # that PR would sit dead until someone happened to merge something else. The
+  # daily run re-arms it on its own; workflow_dispatch is the manual kick.
+  schedule:
+    - cron: '17 6 * * *'
+  workflow_dispatch:
 permissions:
   contents: write
   pull-requests: write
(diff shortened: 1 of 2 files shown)


alxjrvs (commented):
Reviewed the 2-file diff (`.github/workflows/release-please.yml`, `docs/adrs/ADR-024-derived-release-changelogs.md`) and verified the claims against live repo state (`gh repo view`, ruleset 9182849, open PRs, Actions fork-approval policy, `ci.yml`).

Findings:

- `.github/workflows/release-please.yml:54` — **Fork PR named `release-please--*` gets auto-merge armed.** The jq filter keys only on attacker-controllable `headRefName`. Confirmed context: repo is PUBLIC, `main`'s ruleset has no required-review rule and `bypass_actors: []`, and Actions approval is only `first_time_contributors`. So an outside contributor can name a fork branch `release-please--branches--main--components--srd`, and the next push or the new daily run arms auto-merge with `RELEASE_PLEASE_TOKEN`; once `quality-checks` + CodeQL go green the merge queue lands arbitrary code on `main` with no human in the loop. Fix: add `select(.isCrossRepository == false)` (plus an author and/or `--base main` constraint).
- `.github/workflows/release-please.yml:44` — **Arming step is skipped when the release-please step fails.** No `if: always()`, so a transient failure of `release-please-action@v5` skips the arming step entirely — including on the scheduled run that exists precisely to re-arm a stalled release PR (#688 is currently `clean` with `autoMergeRequest: null`).

Things I checked and deliberately did *not* flag: the branch prefix matches the real branch names; `allow_auto_merge` is on; the queue's `merge_method` is SQUASH so `--squash` is at most a gh warning, not an error; `set -euo pipefail` interacts correctly with the `|| { …; failed=1; }` block and the trailing `exit "$failed"`; and `gh` exits 0 on an already-queued PR.

PR #701, opened 2026-08-06
docs: document that main is behind a merge queue

alxjrvs:
## Why

`main`'s ruleset carries an active **`merge_queue`** rule (alongside `deletion`, `non_fast_forward`, `required_linear_history`, `required_status_checks`). `CLAUDE.md` said nothing about it, and the failure mode is genuinely misleading.

Running `gh pr merge <pr> --squash` prints only:

```
! The merge strategy for main is set by the merge queue
```

…**exits 0**, and enqueues the PR. Afterwards every obvious success signal is absent — the PR is still `OPEN`, `mergedAt` is `null`, `mergeCommit` is `none`, `autoMergeRequest` is `off`. That reads as a silent no-op and invites a pointless retry. It cost real time on #695 during this session before the queue was inspected directly.

## What this adds

A `### Merging — main is behind a merge queue` section to `CLAUDE.md` recording:

- the correct incantation, `gh pr merge <pr> --auto --squash`
- the exit-0-but-not-merged trap, and that PR state alone proves nothing while queued
- a GraphQL query for reading the queue directly (**verified against this repo** — it returns the live entries as written)
- that the queue re-tests each entry against the projected merge and *ejects* conflicts, so a green PR can still sit unmerged
- the Dependabot consequence: a merge queue can't be driven by `GITHUB_TOKEN`, so Dependabot PRs sit open until something PAT-backed enqueues them

## Scope

Docs only — one file, 33 insertions, no code or config touched.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -291,6 +291,39 @@ Pre-commit runs: lint --fix, format (parallel). Typecheck does NOT run pre-commi
 Pre-push runs (parallel): typecheck, test, validate:all, knip, check:tokens,
 check:styling, lint, check:schemas.
 
+### Merging — `main` is behind a merge queue
+
+`main`'s ruleset (`deletion`, `non_fast_forward`, `required_linear_history`,
+`required_status_checks`, **`merge_queue`**) means a PR is **enqueued, never merged
+directly.** Enqueue with:
+
+```bash
+gh pr merge <pr> --auto --squash
+```
+
+**A plain `gh pr merge <pr> --squash` does not fail — and does not merge.** It prints
+only `! The merge strategy for main is set by the merge queue`, **exits 0**, and adds
+the PR to the queue. Every obvious success signal is absent afterwards: the PR is still
+`OPEN`, `mergedAt` is `null`, `mergeCommit` is `none`, and `autoMergeRequest` is `off`.
+That combination looks exactly like a silent no-op, so **do not conclude the merge
+failed and retry** — read the queue itself:
+
+```bash
+gh api graphql -f query='{repository(owner:"SalvageUnion-io",name:"SU-SRD"){
+  mergeQueue(branch:"main"){entries(first:20){nodes{position state
+  pullRequest{number}}}}}}'
+```
+
+A queued PR stays `OPEN` until the queue merges it, so **PR state alone is never proof
+of anything** here. The queue is also the reason a PR can go green and still sit
+unmerged: it re-tests each entry against the projected merge, and ejects any that
+conflicts rather than merging it.
+
+Note that `gh pr merge --auto` is what Dependabot PRs need too. A merge queue is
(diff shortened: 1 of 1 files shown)


alxjrvs (commented):
## Review summary

The PR under review is `origin/main...HEAD` — a single docs commit (79afc10b) adding a "Merging — `main` is behind a merge queue" section to `CLAUDE.md`. (Note: local `main` is stale by one commit, so `git diff main...HEAD` misleadingly pulls in the Dependabot bump #695; the real scope is docs-only.)

Since `CLAUDE.md` is loaded into every session, wrong instructions here are the defect class. I verified the claims empirically rather than reading them:

**Confirmed accurate:** the ruleset rule list (`deletion`, `non_fast_forward`, `required_linear_history`, `required_status_checks`, `merge_queue`, `bypass_actors: []`); the GraphQL query runs and returns the expected shape; both required contexts (`quality-checks`, `Analyze (javascript-typescript)`) exist and — importantly — both `ci.yml` and `codeql.yml` already carry `merge_group:` triggers, so the queue-wedge hazard is genuinely avoided; no Dependabot auto-merge workflow exists, so that paragraph describes a real gap rather than a broken workflow.

### Findings

- `CLAUDE.md:304` — The `! The merge strategy for main is set by the merge queue` warning is attributed to the plain non-`--auto` form, but gh 2.94 gates it on `!m.opts.MergeStrategyEmpty` (any strategy flag) and forces `payload.auto = true` for both forms — so the *recommended* `--auto --squash` prints it too, and an agent will mis-diagnose its own successful enqueue. Dropping `--squash` (queue `merge_method` is already `SQUASH`) removes the warning and makes the cases distinguishable.
- `CLAUDE.md:305` — "prints only …" / "every obvious success signal is absent" holds for non-TTY runs only; in a terminal gh also prints `✓ Pull request …#N will be added to the merge queue for main when ready` (`merge.go:354`, suppressed by `infof`'s `isTerminal` check). The TTY dependence needs stating.
- `CLAUDE.md:307` — `autoMergeRequest` and `mergeCommit` read `null`, not `off` and `none` (verified against PR #701), so the documented diagnostic values can never be matched.
- `CLAUDE.md:296` — Rule *names* are listed but the `merge_queue` *parameters* that explain the symptom are not: `min_entries_to_merge_wait_minutes: 5` (a solo green PR always waits ~5 min), `grouping_strategy: ALLGREEN`, `check_response_timeout_minutes: 60`. Without them an agent polling a healthy `QUEUED` entry has no bounded expectation and will either declare it stuck or poll forever.

One thing I deliberately did not file as a finding, but worth raising: this section covers single PRs only, and says nothing about stacked PRs — which your global instructions make the default shape for multi-part work, and which land via `gh stack merge`, not `gh pr merge`. Whether `gh stack merge` works against a merge-queue-protected branch is something I could not verify without mutating repo state. Given `merge_queue` is now active on `main`, that interaction is worth confirming and documenting here before the next stack.

PR #702, opened 2026-08-06
docs(agents): service registry + Convex MCP

alxjrvs:
## What

An agent working in this repo had four MCP servers and no record of what they reach. Getting a Netlify site id meant listing every project on an account that also holds RANDSUM, Binfinite and JRVS Softworks work, then guessing which row was ours — and `suindex` being `apps/srd` is a trap that has cost real time.

- **New `docs/architecture/agent-tooling.md`** — the service registry. Every Netlify site id, the Render service + workspace id, the Sentry org/projects and their DSN env vars, the Convex deployments, the MCP auth model for each server, and a table of how to re-derive every claim. Scoped explicitly to public identifiers, with a rule for what must never be written into it.
- **`.mcp.json` gains `convex`.** Convex is the server of record for accounts and Games, but it was the one deployment target holding user data that an agent could not inspect. It authenticates through the Convex CLI's own device credentials, so the file stays secret-free. Production access and PII reads stay behind flags that are deliberately **not** set.
- **`github` is recorded as broken-by-default.** The endpoint does not support dynamic client registration, so the committed entry cannot OAuth and fails for everyone until a machine-local PAT header is supplied. That is a property of the endpoint, not a repo defect — but it was undocumented, so agents fell back to `gh` without knowing why.
- **One correction** in `accounts-and-games.md`: the `itun-convex` Sentry project was described as a suggested slug. It exists.

## Verification

- `claude mcp list` — `netlify`, `sentry`, `render`, `github` and `convex` all report **Connected**. `github` was failing before this branch; it now connects via a machine-local header override (applied outside the repo, as documented).
- Every identifier in the registry was read from the live platform via its MCP server, not copied from existing docs.
- `bun run check:fast` green (lint + validate:all + knip, then typecheck across all six workspaces).

## Known caveat, documented rather than hidden

The Convex MCP resolves its deployment from `CONVEX_DEPLOYMENT` in `apps/itun/.env.local` (gitignored). Before `bunx convex dev` has been run once, the server starts fine and every tool call fails. A green `claude mcp list` is therefore not proof the Convex tools work — the doc says so.

https://claude.ai/code/session_01L2121d7DtfikAGGUyDYKHB


alxjrvs (commented):
## Findings

- `docs/architecture/agent-tooling.md:83` — **"`convex` is read-only against dev by default" is false.** The two flags the body cites (`--dangerously-enable-production-deployments`, `--cautiously-allow-production-pii`) gate *production* only; against `dev/alex-jarvis` the full mutating tool set is live. Confirmed from `convex mcp start --help` on the pinned 1.43.0 and from the loaded tool schemas — `run` is "query, mutation, **or action**" and `envSet` writes deployment env vars. An agent trusting the heading can mutate dev data or overwrite `JWT_PRIVATE_KEY`/`SITE_URL` believing it made a read. The body text is right; only the heading overclaims, and the heading is what gets skimmed.
- `CLAUDE.md:81` — **"two of these authenticate outside the file" contradicts the doc added in the same commit.** agent-tooling.md:35 says "Every server below authenticates through something outside the file," and its table proves it for all five (Netlify CLI/OAuth, Sentry OAuth, Render OAuth, Convex device creds, GitHub PAT). CLAUDE.md is loaded every session, so it wins over a doc you have to open — an agent debugging a dead server will look for a repo defect instead of running the local login.
- `docs/architecture/agent-tooling.md:215` — **the Convex re-derivation recipe hangs.** `bunx convex mcp start` is a blocking stdio server with no `status` subcommand, and the invocation drops the `--project-dir apps/itun` the rest of the doc requires. Every other row in that table is a literal shell command, so this one gets run the same way and just hangs.

## Verified clean (no findings)

Worth recording, since a registry doc's value is entirely in its accuracy:

- **Sentry** — queried the live org: exactly six projects, slugs match the table 1:1 (`itun`, `itun-convex`, `itun-functions`, `srd`, `su-assets`, `su-discord`). `itun-convex` does exist, so the `accounts-and-games.md:266` "this step is done" edit is accurate. DSN env-var column checks out against code, including the non-obvious `PUBLIC_SENTRY_DSN` for `srd` (Vite `envPrefix` override at `apps/srd/ssg/vite.config.ts:25`).
- **Render** — queried the live workspace: `suref-discord-bot` / `srv-d600r21r0fns73el5kk0` / `tea-cspvcb0gph6c739fv6s0`, background worker, dashboard URL all exact. The named unrelated services (`hermuz`, `randsum-discord-bot`, `gear`) are really there, and `list_services` does take a `workspaceId`.
- **Convex** — the quoted `{"error":"No CONVEX_DEPLOYMENT set…"}` reproduces verbatim; `--project-dir` is a real flag; deployment names match `accounts-and-games.md`; `apps/itun/convex/` really has 17 `.ts` modules.
- **`.mcp.json`** — I specifically suspected `bunx convex` would miss the pinned version, since `convex` is only in `apps/itun/node_modules` and absent from root `node_modules/.bin`. Tested it: `bunx convex --version` from the repo root returns `1.43.0`. Not a bug.
- `--live` is a real flag on `tools/check-observability.ts`, and `SENTRY_INGEST_HOST` is `https://*.ingest.de.sentry.io` as claimed.

Netlify site ids are the one table I could not verify — the Netlify MCP never finished connecting in this session and the ids appear nowhere else in the repo. No reason to think they're wrong; just untested.

One note outside the diff: the `claude.ai Google Calendar` and `plugin:expo:expo` MCP servers need authorization via an interactive session, and `claude.ai Netlify` / `netlify` never came up here.

alxjrvs (commented):
## Review complete

**Scope.** `origin/main...HEAD` — two commits (`99f3f428`, `efc20a44`): a new service registry at `docs/architecture/agent-tooling.md`, a new `convex` entry in `.mcp.json`, plus CLAUDE.md / docs/README.md / accounts-and-games.md / a11y-scan SKILL.md edits. (`git diff main...HEAD` is misleading here — local `main` is 3 commits stale, so it pulls in unrelated Dependabot bumps. `origin/main` is the real base.)

### Findings

- `docs/architecture/agent-tooling.md:83` — heading "`convex` is read-only against dev by default" is false; `.mcp.json` passes no `--disable-tools`, so `run`/`envSet`/`envRemove` are live on the dev deployment. **CONFIRMED**
- `.mcp.json:19` — `envGet`/`envList` are enabled, returning dev deployment secret *values* into the transcript, contradicting this PR's own "never write down a Convex env var value" rule. **CONFIRMED**
- `.mcp.json:21` — `--project-dir apps/itun` is relative to the server process cwd; from a subdirectory it fails with the exact error the doc attributes to "you haven't run `bunx convex dev`", sending you down the wrong debugging path. **PLAUSIBLE**
- `CLAUDE.md:81` — "two of these authenticate outside the file" contradicts the registry it points at ("Every server below"); all five authenticate externally. **CONFIRMED**

### What I verified as correct (no findings)

Worth noting since the PR's whole value is identifier accuracy — I re-derived what I could reach:

- **Sentry**: org `susrd` has exactly the six projects listed (`srd`, `itun`, `itun-functions`, `itun-convex`, `su-assets`, `su-discord`); `itun-convex` does exist, so the accounts-and-games.md "this step is done" edit is true. DSN env vars match source (`PUBLIC_SENTRY_DSN` for srd, `VITE_SENTRY_DSN` for itun, `SENTRY_DSN` elsewhere). `SENTRY_INGEST_HOST` is `https://*.ingest.de.sentry.io` in `tools/check-observability.ts` and matches both `netlify.toml` CSPs; `--live` is real.
- **Render**: service id `srv-d600r21r0fns73el5kk0`, workspace `tea-cspvcb0gph6c739fv6s0`, `background_worker`, dashboard URL, and the unrelated-services claim (`hermuz`, `randsum-discord-bot`, `gear`) all check out against the live API; `buildFilter` paths match `render.yaml`.
- **Convex**: the `--dangerously-enable-production-deployments` / `--cautiously-allow-production-pii` flags exist as described, and I reproduced the documented `{"error":"No CONVEX_DEPLOYMENT set…"}` verbatim. `apps/itun/.env.local` is gitignored as claimed. `apps/itun/convex/` really is 17 top-level modules.
- **Repo**: all seven relative links in the new doc resolve; `tools/a11y-scan.ts` really is Playwright (the SKILL.md fix is correct); `/triage` and `/component-refresh` skills exist. `bun run validate:doc-drift` passes all 7 checks and `format:check` is clean (markdown is deliberately unformatted by tooling, so the table misalignment in the new doc is cosmetic only).
- **Netlify** site ids are the one block I could not verify — the Netlify MCP never finished connecting and the CLI call was denied by the sandbox. Someone should spot-check those three UUIDs.

alxjrvs (commented):
# Review: agent tooling docs + service registry (`worktree-agent-tooling-docs`)

Scope: 3 commits vs `main` — a new service registry doc, two new skills, a deleted `/commit` skill, two `.mcp.json` servers (Convex + Context7), CLAUDE.md updates, and a Dependabot-style dependency bump. No application code changed, so I verified the doc/skill claims against the platforms and the repo rather than reading logic.

## Findings (6)

- `.claude/skills/convex-deploy-verify/SKILL.md:33` — Every `bunx convex env …` command is given with no working directory, and `convex env` has no `--project-dir`/`--cwd` flag. Verified by running it: from the repo root it fails with `✖ No CONVEX_DEPLOYMENT set…` — the *same* error the companion doc blames on a missing `.env.local`, so the agent misdiagnoses and runs `bunx convex dev` at the root, which offers to create a **new** Convex project. Step 4's presence-by-length check also reports `0` ("absent") for variables that are set. The `.mcp.json` entry added in this same PR gets it right (`--project-dir apps/itun`); the skill needs `cd apps/itun`.
- `.claude/skills/knip-triage/SKILL.md:43` — Says "**Four** workspaces opt out with `includeEntryExports: false`", then lists three; `knip.json` has exactly three. The natural guess for the phantom fourth is `apps/itun`, which deliberately does *not* opt out — leading an agent to `@knipignore` a genuinely dead itun entry export, the exact laundering the skill exists to prevent.
- `CLAUDE.md:212` (also `:189`, `.claude/rules/monorepo-patterns.md:28`) — Still claim `build:package` regenerates only `schemas/*.schema.json`, directly contradicting the `/build-package` skill added in the same PR. The real script also writes `lib/generated/*.generated.ts`, `etc/*.api.d.ts`, and a marker block in `lib/index.ts`. Following CLAUDE.md (loaded every session, overriding) means committing 1 of 4 paths → `check:schemas`, the first step of `check:all`, fails in CI.
- `docs/architecture/agent-tooling.md:107` — Ships "knip 6.29" while the same PR bumps knip to `6.31.0`. Stale on merge; `check-doc-drift.ts` only guards Vite/React/Tailwind majors, so nothing catches it.
- `docs/architecture/dashboard.md:651` — The "a11y-scan uses puppeteer" claim was fixed in the skill and CLAUDE.md but survives here, in a directory `check-doc-drift.ts` itself treats as live instruction. puppeteer is not a dependency anywhere.
- `.claude/skills/validate/SKILL.md:3` — Frontmatter `description` (the only text in the skill listing) still omits knip/audit/tokens/styling, which the rewritten body specifically calls out as easy to forget; `/verify`'s description lists them, making two "literal aliases" look like different commands.

## Verified correct (worth noting — this PR is unusually well-checked)

- All three Netlify site ids, names, production URLs and the team id `6a3b41d74a67a34e3aae3ede` match the Netlify API exactly, including the `suindex` → `apps/srd` mismatch the doc flags.
- Sentry org `susrd` has exactly the six projects listed, no more.
- `main`'s ruleset has exactly the five rule types the new merge-queue section names, including `merge_queue`.
- The `/validate` `check:all` execution order matches `package.json` step for step, and `/build-package`'s four paths match `check:schemas`' `git diff` paths exactly.
- `ITUN_BOT_SECRET` / `ITUN_CONVEX_SITE_URL` match `apps/discord-bot/src/config.ts`, `apps/itun/convex/botHttp.ts` and `render.yaml`; `--project-dir`, `--prod`, and the two production-gating flags are real `convex mcp start` options.
- `bun tools/check-doc-drift.ts` and `bun tools/check-architecture.ts` both pass on the branch, and no stale `/commit` references remain after the skill deletion.

alxjrvs (commented):
Review complete. Scope was `git diff origin/main...HEAD` — 13 files, effectively a documentation/tooling-registry PR (new `docs/architecture/agent-tooling.md`, two new skills, `.mcp.json` gaining Convex + Context7, and a rewritten header comment in `apps/srd/ssg/parity.ts`). No executable logic changed, so "correctness" here means whether the claims match reality; I verified them against the repo and against the live platforms.

**Verified correct** (worth stating, since these are the load-bearing bits): all three Netlify site ids, names, production URLs and the team id match `get-projects`; the Render service id `srv-d600r21r0fns73el5kk0`, its worker type, workspace id and the unrelated-service list match `list_services`; all six Sentry project slugs match `find_projects` in the `susrd` org; `PUBLIC_SENTRY_DSN`/`VITE_SENTRY_DSN`/`SENTRY_DSN` map to the right surfaces; the documented `check:all` ordering matches the root script exactly; the four `check:schemas` paths in `/build-package` are exactly right; `ITUN_BOT_SECRET`/`ITUN_CONVEX_SITE_URL` exist as named; the Convex MCP connects and fails with the *exact* error string the doc quotes; `make-parity-baseline.ts` really does derive the Astro commit rather than hardcode it; and `bun tools/check-doc-drift.ts` passes.

Findings:

- `CLAUDE.md:253` — Root CLAUDE.md still says parity "reports zero differences across 1,039 pages today", directly contradicting this PR's own new claim that `/changelog` mismatches permanently and the gate exits 1. This is the always-loaded file, so it's the highest-impact one.
- `docs/architecture/agent-tooling.md:79` — "which is why `render` connects while `github` does not" is false: both have machine-local `headersHelper` overrides and `claude mcp list` reports github Connected.
- `.claude/skills/knip-triage/SKILL.md:43` — Says "Four workspaces" opt out of `includeEntryExports`, then lists three; knip.json has three.
- `docs/architecture/dashboard.md:651` — Still describes `/a11y-scan` as puppeteer-based; the PR fixed that claim in two other places but missed this live-instruction doc.
- `docs/architecture/agent-tooling.md:108` — Cites knip 6.29; root package.json pins 6.31.0 already on origin/main.

alxjrvs (commented):
Reviewed the 5-commit branch (docs/agent-tooling registry + MCP wiring, skill edits, the `/changelog` parity exemption, and a Dependabot dep bump). I verified behavior rather than reading only: ran the parity test file (72 pass), `validate:all`, `knip`, `typecheck`, `lint`, `format:check`, and `bun install --frozen-lockfile --dry-run` — all green. I also audited `isInsertionOf` for false negatives/positives (the greedy-prefix approach is sound as a single-insertion test) and cross-checked the new docs' factual claims against `knip.json`, `package.json` scripts, `check-observability.ts`, `check-doc-drift.ts`, `outputPath.ts`, and `parseChangelog.ts` — most claims check out.

Findings:

- `.claude/skills/convex-deploy-verify/SKILL.md:75` — the verification curl pipes headers through `grep -i location`, discarding the status line; the 500 and 404 cases print nothing at all and are indistinguishable, and even the 302 case never shows the status code the Report section demands.
- `apps/srd/ssg/parity.ts:472` — `isInsertionOf` accepts unbounded appends at the *end* of the page, but `/changelog` is newest-first, so any end-append is a regression that now passes and the gate still prints "zero differences".
- `apps/srd/ssg/parity.ts:869` — the append-only failure message asserts "baseline content was removed, reordered or reworded. A real finding.", but the check also rejects non-destructive non-contiguous growth (a backfilled historical entry sorts into the middle), sending the reader after a nonexistent regression.
- `package.json:21` — `@biomejs/biome` bumped to 2.5.6 without updating `biome.jsonc`'s `$schema` pin (2.5.5); `bun run lint` now emits a config-mismatch diagnostic on every run (exit 0, so CI stays green).
- `.claude/skills/knip-triage/SKILL.md:43` — says "Four workspaces opt out" with `includeEntryExports: false`, then lists three; `knip.json` has exactly three.

Non-findings worth noting: the `.mcp.json` Convex entry's missing `"type"` matches the existing `netlify` entry's style (stdio default); the `changelog/index.html` key format is correct on all platforms because `walk()` normalizes separators; and the changelog page's head meta and JSON-LD are static, so the exemption doesn't need to cover them.

alxjrvs (commented):
Reviewed `main...HEAD` (36 files; Convex claim/dedupe work, the client `serverError` helper, srd parity + PWA + Sentry changes, docs/deps). I read every code hunk, opened the surrounding Convex modules and schema, and verified the three findings below by running the real functions under `convex-test` (throwaway probe files, deleted afterward). The existing suites are green: `apps/itun` convex tests 30/30, `apps/srd` parity + observability 76/76.

Findings:

- `apps/itun/convex/maintenance.ts:79` — `dedupeAppIds` deletes another account's entity. Groups are keyed by appId across the whole deployment and `rankForKeeping` never checks for two *distinct* non-null owners, so a cross-account duplicate (the export/import case `claimLocal`'s own header documents) loses one player's character permanently. Confirmed: pilot 'p1' owned by A and by B → A's row deleted, B's kept. `GroupReport` has no owner field, so the dry run doesn't reveal it either.
- `apps/itun/convex/entities.ts:472` — the crawler pre-pass re-opens the duplicate-appId hole it was added to close. Because it resolves all crawlers against the DB before inserting any, `claimLocal({crawlers: [c, c]})` returns `claimed: 2, alreadyPresent: 0` and writes two rows with the same appId — after which `patchCrawlerByAppId`'s `.unique()` throws on every mirrored crawler write, forever. Pilots/mechs are immune (insert-as-you-go); crawlers are not.
- `apps/itun/convex/entities.ts:538` — soft-link dedupe is global rather than scoped to the claimer. A (from, to, kind) triple present in anyone else's game marks the claim `alreadyPresent`; confirmed that user B's identical imported link is silently dropped. The comment above it notes nothing calls `.unique()` on this table, and the pattern check below correctly scopes by `by_owner`.

Nothing else rose to a real defect. Checked and cleared: `isInsertionOf` is sound for "one contiguous insertion" (greedy prefix cannot produce a false negative, and deletion/reorder/reword cases all fail as the tests claim); the append-only early `return` is inside the per-page `mapLimit` callback, not a loop exit; `NotAuthorized extends ConvexError` keeps `.message` intact so `botClient.asFailure`'s in-process `instanceof` still works; the PWA `.catch()` is not asserted anywhere and parity never byte-compares `.js`; Sentry `ignoreErrors` substring matching is used correctly.

alxjrvs (commented):
## Review complete

Scope: `origin/main...HEAD` — 6 commits, 14 files (711 insertions). Mostly agent-facing documentation, plus one real code change: the `/changelog` append-only exemption in `apps/srd/ssg/parity.ts`.

### Finding

- `.claude/skills/knip-triage/SKILL.md:43` — Says "Four workspaces opt out per-workspace with `includeEntryExports: false`", then lists three. `knip.json` has exactly three (`packages/component-lib`, `apps/srd`, `apps/su-assets`). An agent following the skill either hunts for a nonexistent fourth or "fixes" `knip.json` by adding an opt-out to a workspace that is deliberately gated — the exact regression the gate exists to prevent.

### What I verified and found clean

**`parity.ts` / `isInsertionOf`** — I checked the algorithm for both soundness and completeness. Greedy-longest-common-prefix followed by `candidate.endsWith(baseline.slice(i))` is provably correct here: if a valid decomposition `baseline[0..k) + INS + baseline[k..)` exists, the greedy `i >= k`, so `baseline.slice(i)` is a suffix of `baseline.slice(k)` and the check passes (no false negatives); and any passing pair admits a valid decomposition since `cand.len >= base.len` (no false positives). The `if/else` (rather than early `return`) is the right call and is correctly justified in the comment. `bump()` is correctly skipped only on the tolerated path. The `+=` counters are safe under `mapLimit(…, 24)` since they run synchronously between awaits.

I also traced the contiguity assumption through `mergeChangelogs` (`packages/component-lib/src/changelog/parseChangelog.ts:98`), since `/changelog` merges **two** prepend-only sources — two prepend points could in principle produce two insertions and a false FAIL. It holds: the merge is date-descending and every post-baseline entry is strictly newer than every baseline entry, so all growth lands as one contiguous head block. The only breaking case needs a new entry dated *exactly* equal to the frozen baseline's max date, which is not reachable going forward. Confirmed `/changelog` content feeds no other compared artifact (`llms.txt`, the 899 JSON endpoints, head metadata are all unaffected). 72 tests pass; Biome clean; `ssg/parity.ts` is a knip entry with `includeEntryExports: false`, so the two new exports won't trip `check:all`.

**Registry doc (`docs/architecture/agent-tooling.md`)** — this is the PR's main payload, and its value is entirely accuracy, so I checked it against the live platforms rather than reading it:

- Netlify: all three site ids, names, primary URLs, and team id `6a3b41d74a67a34e3aae3ede` match exactly, including the `suindex` → `apps/srd` name mismatch the doc flags.
- Render: `srv-d600r21r0fns73el5kk0`, background worker, workspace `tea-cspvcb0gph6c739fv6s0`, plus the three unrelated services (`hermuz`, `randsum-discord-bot`, `gear`) it warns about.
- Sentry: `susrd` org returns exactly the six projects listed, in the EU region.

**Other doc claims** — `check:all` step order matches `package.json:85` exactly; `check:schemas` guards exactly the four paths the build-package skill names; `tools/a11y-scan.ts` does use Playwright; Prettier is genuinely absent from every `package.json` and `lefthook.yml`; `.mcp.json` declares six servers, is valid, and stays secret-free; the skill roster in `CLAUDE.md` matches `.claude/skills/` after the `/commit` deletion, and nothing else references `/commit`.

One judgment call I made and did not report: `CLAUDE.md` carries forward "17 modules" for `apps/itun/convex/` where there are 18 top-level `.ts` files. The sentence was rewritten in this PR, but "module" is ambiguous enough (excluding `auth.config.ts` gives 17) that it isn't provably wrong — flagging it would have been noise.

PR #703, opened 2026-08-06
fix(srd): stop two unactionable browser errors filling Sentry

alxjrvs:
Every srd issue open in Sentry today was one of these two, and neither is
actionable — both are features working as specified.

## `Error: Rejected` (SRD-2, 9 events in a week)

`registerSW.js` registered the service worker with no `.catch()`. Any
rejection outside this site's control — service workers disabled, a
locked-down enterprise profile, an extension intercepting the request —
became an *unhandled* rejection, which Sentry's `globalHandlers`
integration reported as `Error: Rejected`: no stack worth reading, no user
impact, nothing anyone could do.

Swallowing is the correct handling here rather than the lazy one. A failed
registration means no offline caching, which is a progressive enhancement
the site works fine without — there is no fallback to attempt and nothing
to tell the user.

## `AbortError: Transition was skipped` (SRD-A)

srd opts into *cross-document* view transitions declaratively, via
`@view-transition { navigation: auto }` in `global.css`. When a second
navigation supersedes an in-flight transition (a fast click, or a click
during page load) the browser rejects the transition's `finished` promise
with an `AbortError`. **That promise is the user agent's** — there is no JS
handle of ours to `.catch()` — so the SDK's `ignoreErrors` is the only
place it can be dropped. The navigation itself completes normally.

The new `IGNORED_ERRORS` list is deliberately hard to add to: a filtered
error is one nobody will ever see again, so each entry names the mechanism
rather than just the string.

## On parity

The `.catch()` moves `registerSW.js` off the Astro baseline's 134 bytes.
**The gate is unaffected**: `ssg/parity.ts` compares the emitted file set,
head metadata, JSON-LD, `<main>` text, the 899 JSON endpoints and
`llms.txt` — never `registerSW.js` or `sw.js` bytes. Only the precache
revision hash in `sw.js` moves, and nothing asserts on it. The comment's
now-false "byte-identical to the baseline's" claim is corrected in place
rather than left to mislead the next reader.

## Verification

- `bun run check:all` green (lint, format, typecheck ×6, test, validate, knip, audit, tokens, styling)
- `bun --filter srd test` — 1152 pass
- `bun ssg/build.ts` + `bun ssg/parity.ts` against a freshly regenerated baseline: the only mismatch is `changelog/index.html`, which is **pre-existing drift unrelated to this PR** — `676078fa release srd 2.0.1` added a changelog entry that the frozen Astro-era baseline cannot contain. Flagged separately; it will fail on every PR until the baseline story for that page is decided.

### DIFF
--- a/apps/srd/src/lib/__tests__/observability.test.ts
+++ b/apps/srd/src/lib/__tests__/observability.test.ts
@@ -82,6 +82,16 @@ describe('observability', () => {
     // `tools/check-observability.ts` asserts against both netlify.tomls.
     expect(options.tracesSampleRate).toBe(0)
 
+    // Cross-document view transitions reject their `finished` promise with an
+    // AbortError whenever a navigation supersedes an in-flight one. It is not
+    // ours to catch (the promise is the browser's) and it means nothing went
+    // wrong, so it must never reach Sentry. Asserted on the substring Sentry
+    // actually matches against `${type}: ${value}`, so the real event shape
+    // "AbortError: Transition was skipped" is covered.
+    const ignored = options.ignoreErrors as string[]
+    expect(ignored).toContain('Transition was skipped')
+    expect('AbortError: Transition was skipped').toContain(ignored[0] as string)
+
     const boom = new Error('boom')
     captureException(boom, { where: 'island' })
 
--- a/apps/srd/src/lib/observability.ts
+++ b/apps/srd/src/lib/observability.ts
@@ -23,6 +23,26 @@ let initialized = false
 let sentryModule: typeof import('@sentry/browser') | null = null
 
 /**
+ * Browser noise that is not this site's to fix, dropped before it is sent.
+ *
+ * The bar for adding to this list is deliberately high: a filtered error is one
+ * nobody will ever see again, so anything here has to be something we could not
+ * act on even if we wanted to. Each entry names the mechanism, not just the
+ * string.
+ *
+ * - **`Transition was skipped`** — srd opts into *cross-document* view
+ *   transitions declaratively, with `@view-transition { navigation: auto }` in
+ *   `src/styles/global.css`. When a second navigation supersedes an in-flight
+ *   transition (a fast click, or a click during page load) the browser rejects
+ *   the transition's `finished` promise with an `AbortError`. That promise
+ *   belongs to the user agent — there is no JS handle of ours to `.catch()`, so
+ *   the SDK is the only place it can be dropped. The navigation itself
+ *   completes normally; the "error" is the feature working as specified
+ *   (issue SRD-A).
+ */
+const IGNORED_ERRORS = ['Transition was skipped']
+
+/**
  * Initializes browser Sentry when `PUBLIC_SENTRY_DSN` is configured.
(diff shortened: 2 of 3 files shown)


alxjrvs (commented):
## Review scope note

The environment shifted under me mid-review: the worktree started clean at `fix/srd-sentry-noise`, then acquired staged changes, then committed them as **`202286fe fix(itun): stop claimLocal duplicating rosters, and let refusals say why`** on `fix/itun-claimlocal-duplicates`, and by the end had been reset again to a *new* branch `fix/itun-starter-set-unique-ids`. I reviewed commit `202286fe` (13 files, +988/−32) — the diff that was in scope when the review began. That commit object still exists; the branch it was on does not currently match the checkout.

Verification I ran against it: `bun run knip` clean, `bun run typecheck:itun` clean, `bun test test/convex/ src/lib/connection/__tests__/serverError.test.ts` → 215 pass. I also confirmed against the `convex@1.43.0` source that the `NotAuthorized extends ConvexError` mechanism actually works end to end (`Symbol.for("ConvexError")` identity, and `browser/sync/client.js` overwriting `.data` with `result.errorData`) — that part of the change is sound.

## Findings

- `apps/itun/convex/entities.ts:318` — Global `appIdTaken` check makes the built-in Starter Set unclaimable for every account after the first; its ids are hard-coded literals seeded into every browser, and the UI then falsely reports them as "already in your account".
- `apps/itun/convex/entities.ts:538` — Soft-link dedupe queries `by_from` globally with no owner/game scoping, and `templates.ts` inserts `STARTER_SOFT_LINKS` with the same fixed triples into every starter Game, so players' own links are discarded with no `.unique()` hazard to justify it.
- `apps/itun/convex/entities.ts:549` — Gating shelf-Game creation on `claimableCrawlers.length > 0` leaves `shelfGameId === null` on a re-claim, so newly claimed soft links land with `gameId: null` and are unreachable (`listForGame` reads `softLinks` only via `by_game`).
- `apps/itun/convex/maintenance.ts:116` — `dedupeAppIds` deletes duplicate crawler rows but not the duplicate `games`/`memberships` the same bug created, leaving one empty "Claimed crawler" Game per re-claim forever.
- `apps/itun/convex/maintenance.ts:60` — The comment claims an orphaned proposal "answers 'That entity no longer exists'", but `model/entities.ts:93/96` and `proposals.ts:172` throw plain `Error`, so Convex redacts it to "Server Error" — the two player-facing messages the `ConvexError` conversion missed.
- `apps/itun/src/stores/entityBackend.ts:194` — Refusals are now identified via `refusal` but still sent through `captureException`, so a routine per-write refusal floods Sentry with unactionable events — the opposite of the branch's purpose.
- `apps/itun/convex/maintenance.ts:147` — Both dry run and apply `.collect()` three whole tables in one mutation; on a deployment large enough to have duplicates this trips Convex's transaction read limits and the documented dry-run first step fails outright instead of reporting.

The two highest-ranked findings share one root cause — deterministic, deployment-wide-shared ids in `src/lib/starterSet/starterSet.ts` colliding with the new *global* identity checks. Notably, the branch the worktree was reset to (`fix/itun-starter-set-unique-ids`) suggests that root cause is already being addressed independently.

PR #704, opened 2026-08-06
fix(itun): stop claimLocal duplicating rosters, and let refusals say why

alxjrvs:
Both findings from today's Sentry triage. They presented as one thing: an
opaque `[CONVEX M(entities:upsertByAppId)] Server Error`, 12 events over
~50 minutes, spanning a deploy that did not fix it. The real cause existed
only in the Convex logs:

```
Uncaught Error: unique() query returned more than one result from table pilots:
 [m172sf321pxbr3hv51y732f7j98br9t7, m1725mycjgy3k35dp7ezsb8xv98btb52, ...]
    at async byAppId (convex/entities.ts:524)
```

---

## 1. `claimLocal` duplicated rosters, and the duplicate broke writes forever

`claimLocal` inserted unconditionally. The only thing stopping a second run
was a `localStorage` marker — and the card that writes it is **deliberately
re-offered on a second device**, so a laptop after a phone, a fresh browser
profile, or cleared site data ran the whole claim again. The file's own doc
comment said as much ("Claiming twice would duplicate the roster") while
relying on a client-side flag to prevent it.

The consequence was not a cosmetic double-up. Pilots, mechs and crawlers
are addressed by `appId`, and `byAppId` / `patchCrawlerByAppId` use
`.unique()`, which **throws** on a second row. From the moment a roster was
claimed twice, every mirrored write for those entities failed — silently,
because the mirror is fire-and-forget. The local write still succeeded, the
app looked completely healthy, and IndexedDB drifted from the declared
server of record (ADR-030) permanently. It does not self-heal.

**Every insert in the claim is now guarded on identity**, per kind:

| kind | identity |
| --- | --- |
| pilots, mechs, crawlers | `appId` |
| soft links | the (from, to, kind) triple — a link has no id, it *is* its endpoints |
| patterns | the body's own id (no `appId` column exists) |

It reports a new `alreadyPresent` count, kept distinct from `skipped`
(which means "could not be read") — telling someone their roster was
unreadable when it is simply already there would be a worse lie than the
silence it replaces. The UI now says so, so "0 copied" reads as "already
done" rather than "nothing happened".

The crawler pass also resolves what it will write **before** raising the
placeholder Game, which was itself a repeated write: a re-claim used to
leave behind an empty "Claimed crawler" Game and a membership every time.

### Repairing what is already broken

`convex/maintenance.ts` — operator-only (`internalMutation`, reachable only
via `bunx convex run`), **dry-run by default**:

```bash
bunx convex run maintenance:dedupeAppIds --prod              # report only
bunx convex run maintenance:dedupeAppIds '{"apply": true}' --prod
```

It keeps one row per app id (an owned row over an unclaimed one, then most
recently written) and reports how many `changeLog` rows — audit history and
pending Mediator proposals alike — still point at a copy it would delete.
Those address entities by **Convex id**, not `appId`, so they do not follow
the survivor; it degrades gracefully rather than corrupting, but it is a
real consequence and is reported up front instead of discovered later.

**Not run against production in this PR** — that is a destructive operation
on player data and wants a human on the dry-run output first.

### Known limitation, deliberately not fixed here

`appId` is a client-minted UUID, and export/import copies a build between
people *keeping its id* — so two accounts can legitimately hold the same
app id, which the `by_app_id` + `.unique()` pairing assumes cannot happen.
This PR takes the safe side (report `alreadyPresent`, never insert a
second) rather than corrupting the mirror for two people. Closing it
properly means keying the index on `(ownerId, appId)`, which is a schema
change and its own piece of work.

## 2. Server errors could not say anything

Convex redacts every non-`ConvexError` throw to `"[CONVEX M(fn)] Server
Error"` before the client sees it. The repo used `ConvexError` **nowhere**,
so all ~35 carefully-worded refusals — "This game has no Union Crawler yet
— the Mediator raises one before the crew joins it" — were written, thrown,
and discarded at the boundary. A player who tried something the rules
refuse got the same opaque string as a genuine crash. (A real
`NotAuthorized: Only the Mediator can do that` fired during the same
session and reached neither the player nor Sentry.)

`NotAuthorized` now extends `ConvexError`, so those messages arrive.
Genuine defects — parse failures, broken invariants — stay plain `Error`:
opaque to clients on purpose, legible in the Convex logs and now in Sentry.

`src/lib/connection/serverError.ts` is the client half, and the only
sanctioned way to ask which kind you have. Never string-match
`'Server Error'`, and never render `String(err)` from a mutation — that is
the redacted one. Both mirror-write handlers now tag their Sentry reports
with the refusal text, so an operator sees *why* rather than "Server Error".

`ClaimLocalData` had no `.catch()` at all: a failed claim just stopped
saying "Copying…" and left the player to guess. It now says what happened
and stays on offer.

## Notes

- **No Sentry SDK was added to `convex/`** — `apps/itun/CLAUDE.md` forbids it, correctly (queries/mutations run in a deterministic runtime with no `fetch`). Convex's first-party Exception Reporting integration was enabled in the dashboard instead; docs updated to record it as live and to state the `ConvexError` contract.
- `convex/_generated/api.d.ts` is regenerated (`convex codegen`) because the new module must be registered — `tools/check-convex-codegen.ts` catches this, and did.

## Verification

- `bun run check:all` green — lint, format, typecheck ×6, test, validate, knip, audit, tokens, styling
- `bun --filter itun test` — **1711 pass, 0 fail**
- 11 new tests. The premise is pinned, not assumed: `maintenance.test.ts` first proves a duplicate row really does make `upsertByAppId` throw `/unique/i`, then that the repair restores writability — a repair whose premise is untested is one nobody can trust.

### DIFF
--- a/apps/itun/CLAUDE.md
+++ b/apps/itun/CLAUDE.md
@@ -110,11 +110,26 @@ and remains the account-free way to share a build.
   (`src/lib/observability.ts`) and the Netlify Functions
   (`netlify/functions/_observability.ts`) each own one; Convex instead uses its
   first-party Exception Reporting integration, enabled in the Convex dashboard
-  with no application code. Queries and mutations run in a deterministic runtime
-  with no `fetch`, so an in-function SDK could not report from them at all —
-  rationale and the enable-it runbook are in
+  with no application code (**on for production since 2026-08-05**). Queries and
+  mutations run in a deterministic runtime with no `fetch`, so an in-function SDK
+  could not report from them at all — rationale and the enable-it runbook are in
   [docs/architecture/accounts-and-games.md](../../docs/architecture/accounts-and-games.md)
   ("Convex error reporting — a dashboard toggle, not code").
+- **Throw `ConvexError` when the message is for a player; plain `Error` when it
+  is not.** Convex redacts every non-`ConvexError` throw to
+  `"[CONVEX M(fn)] […] Server Error"` before the client sees it, so a
+  user-facing message thrown as a plain `Error` is written and then discarded.
+  `NotAuthorized` (`convex/model/permissions.ts`) extends `ConvexError` for
+  exactly this reason; parse failures and broken invariants stay plain. On the
+  client, read it with `serverMessage()` / `isServerRefusal()` from
+  `src/lib/connection/serverError.ts` — never by string-matching `'Server Error'`,
+  and never by rendering `String(err)` from a mutation.
+- **Never insert into an `appId`-addressed table without checking first.**
+  `pilots`, `mechs` and `crawlers` are looked up by the client's `appId` with
+  `.unique()`, which **throws** on a second row — and since mirrored writes are
+  fire-and-forget, that throw is invisible: local writes keep succeeding while
+  the server of record silently stops receiving them. `convex/maintenance.ts`
+  repairs rows already in that state (`dedupeAppIds`, dry-run by default).
 
 ## Commands
 
--- a/apps/itun/convex/_generated/api.d.ts
+++ b/apps/itun/convex/_generated/api.d.ts
@@ -19,6 +19,7 @@ import type * as entities from "../entities.js";
 import type * as games from "../games.js";
 import type * as http from "../http.js";
 import type * as invites from "../invites.js";
+import type * as maintenance from "../maintenance.js";
 import type * as mediator from "../mediator.js";
 import type * as model_bot from "../model/bot.js";
 import type * as model_entities from "../model/entities.js";
@@ -45,6 +46,7 @@ declare const fullApi: ApiFromModules<{
   games: typeof games;
   http: typeof http;
   invites: typeof invites;
+  maintenance: typeof maintenance;
   mediator: typeof mediator;
   "model/bot": typeof model_bot;
   "model/entities": typeof model_entities;
--- a/apps/itun/convex/entities.ts
+++ b/apps/itun/convex/entities.ts
@@ -301,6 +301,33 @@ export const removeCrawler = mutation({
 })
 
 /**
+ * Whether this table already holds a row carrying this app id.
+ *
+ * Deliberately `.first()` and not `.unique()`, which is the opposite of
+ * `byAppId` two hundred lines down, and the difference is the point: this
+ * function's job is to run correctly **on a table that is already duplicated**.
+ * `.unique()` throws when it finds more than one row, so using it here would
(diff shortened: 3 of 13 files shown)


alxjrvs (commented):
I reviewed commit `202286fe` (`fix(itun): stop claimLocal duplicating rosters, and let refusals say why`) — the only commit on this branch. Note: local `main` is stale, so `main...HEAD` pulls in unrelated dependency bumps; I scoped to `git show HEAD`.

The core fix is sound — `appIdTaken` correctly uses `.first()` not `.unique()`, pilots/mechs are safe against within-batch duplicates via read-your-writes, and `NotAuthorized extends ConvexError` does what the doc claims (field-initializer order means `name` ends up `'NotAuthorized'`, and `botClient.asFailure`'s in-process `instanceof` still works). Findings, most severe first:

- `apps/itun/convex/maintenance.ts:90` — `dedupeAppIds` groups by `appId` alone with no owner/game scoping, so `--apply` permanently deletes a *different* account's legitimate entity. `claimLocal`'s own header documents that app ids are not globally unique (export/import preserves them), and pre-fix cross-account claims are exactly the state this tool is run against. For crawlers it can strip a Game's only crawler, after which `assertMayAddToContainer` refuses every new entity in it.
- `apps/itun/convex/entities.ts:549` — moving the placeholder-Game gate from `crawlers.length` to `claimableCrawlers.length` leaves `shelfGameId` null on every re-claim, so soft links written then get `gameId: null` and are never returned by `listForGame` (entities.ts:154) for the Game that holds the crawler. First-claim links are visible, later ones silently are not.
- `apps/itun/src/components/account/ClaimLocalData.tsx:164` — the new "N were already in your account and were left as they are" copy is false when `appIdTaken` matched *someone else's* row. The player's account holds nothing, and every later mirrored write dies in `assertMayWrite` and is swallowed by the fire-and-forget mirror.
- `apps/itun/convex/entities.ts:468` — the crawler pass runs all `appIdTaken` checks before any insert, so two crawlers with the same app id in one payload both insert, recreating the `.unique()` breakage. Pilots/mechs are immune because their checks interleave with their inserts.
- `apps/itun/convex/entities.ts:538` — the soft-link `duplicateLink` check scans `by_from` deployment-wide with no game/owner scoping, so another account's identical link suppresses this user's.
- `apps/itun/convex/maintenance.ts:147` — the repair deletes duplicate rows but leaves the empty "Claimed crawler" Games and memberships that the repeated claims created, so the visible symptom (and crawler-less Games that reject new entities) survives.

PR #705, opened 2026-08-06
fix(itun): stop game entities silently desyncing, and give the game roster its missing verbs

alxjrvs:
Play-session feedback:

> The game mechs didn't save / You couldn't assign them to a crawler / U can't delete pilots or mechs / I think the shelf ones were fine / But the game specific ones desynced from the main page

**Rebased onto #704**, which has since merged into `main`. Redundancies resolved — see the reconciliation section.

## Root cause, confirmed in production

`claimLocal` inserted blindly, guarded only by a **per-device localStorage marker**, so claiming the same shelf from a second browser duplicated the roster. `byAppId` then asked for `.unique()` on `by_app_id`, an ordinary Convex index and **not** a uniqueness constraint, so it threw. `mirrorWrite` swallowed the throw, because a mirror failure must not undo a local write.

Net effect: the local copy kept accepting edits, **every surface kept rendering them as saved**, and nothing reached the game.

Measured, not inferred:

| Evidence | Finding |
|---|---|
| Prod snapshot (read-only) | 31 pilots under **13** appIds (`Babe` ×4, `Stucky Bolter` ×4); 31 mechs under **14** (`Reaper` ×4, `Harlequin` ×3) |
| Convex prod logs | 12× `unique() query returned more than one result from table pilots at byAppId` |
| Sentry ITUN-2 / ITUN-3 | 12 events, `source: mirrorWrite`, `handled: yes` — the two event URLs are **exactly** the two duplicated pilot appIds |

## What this adds on top of #704

#704 closed the front door (prevention) and cleans up behind it (`maintenance.dedupeAppIds`). This adds the parts it did not cover:

1. **`byAppId` / `crawlerByAppId` survive a duplicate** — resolve to the oldest match and log, instead of throwing. ⚠️ *See the note below; this reverses a deliberate choice in #704.*
2. **Mirror failures reach the player.** #704 made the Sentry report better (refusal vs defect); the player was still told nothing. `reportMirrorFailure` now warns, reports **and** toasts — showing #704's `serverMessage(err)` when the server said why, generic copy when it did not, throttled to one per 30s.
3. **Soft links mirror.** They were excluded as "derived" while `listForGame` read them back, so a Game froze its wiring at claim time. `upsertSoftLink`/`removeSoftLink` address a link by endpoints (idempotent for free); `remove`/`removeByAppId` cascade link pruning as the client always has.
4. **Game roster gains Delete** — owner only, own danger confirm, kept distinct from "Offer to the crew" (releasing is reversible, deleting is not). `entities.remove` addresses by **server id**, because the roster sees rows this browser never held and template pre-gens with no `appId` at all (a dozen in prod).
5. **The crawler sheet gains "+ Add Crew."** A mech reaches a bay only through its pilot (`resolveLinkType` throws for mech→crawler), and the crawler had no way to add an *existing* pilot — beside an empty rail that said "dock one".

## ⚠️ One deliberate override of #704 — worth a look

#704 keeps `.unique()` on the `appId` lookups as an intentional tripwire, and documents that in `apps/itun/CLAUDE.md`. **This PR softens it**, and rewrites #704's characterization test (`maintenance.test.ts` → "duplicate app ids no longer break the mirror").

Reasoning: a throw there is *invisible* — mirrored writes are fire-and-forget, so it never fails an edit the player can see; it just stops their edits reaching the server while the UI reports success. That is the exact failure that lost an evening of play. Refusing to sync is a worse answer to "there are two rows" than syncing to one and logging it loudly. Prevention and repair still do the real work; this only makes the window survivable.

The two rules are pinned as agreeing: `byAppId` keeps the **oldest**, which is the row `dedupeAppIds` keeps, so a write landing before the repair is not discarded by it. There is a new test for exactly that.

Happy to drop this piece if you'd rather keep the tripwire hard.

## Reconciliation with #704

- **Dropped my `convex/repair.ts`** in favour of #704's `maintenance.ts`.
- **Dropped my `claimLocal` idempotency** in favour of #704's `appIdTaken` + `alreadyPresent` counter (better — it separates "already there" from "malformed").
- **Kept #704's `serverError.ts`** and fed it into the toast, so a refusal shows the backend's own words.
- Claim-time soft-link dedupe now shares my `findSoftLink` helper instead of an inline copy.
- One note, **not** actioned: `dedupeAppIds` keeps the oldest row, so for `Babe` (3 shelf copies + 1 in a game) it keeps a *shelf* row and the character leaves the crew roster until the client re-mirrors. Self-heals via `upsertByAppId`'s re-homing, but you may want to prefer the in-game copy.

## Testing

- `check:all` green (exit 0), 1734 itun tests pass.
- New tests verified to **bite**: reverting `byAppId` fails them.

## Not done

**The production data repair was NOT run** — only a read-only snapshot export. `maintenance:dedupeAppIds` is ready behind an explicit `apply: true`.

### DIFF
--- a/apps/itun/CLAUDE.md
+++ b/apps/itun/CLAUDE.md
@@ -125,11 +125,22 @@ and remains the account-free way to share a build.
   `src/lib/connection/serverError.ts` — never by string-matching `'Server Error'`,
   and never by rendering `String(err)` from a mutation.
 - **Never insert into an `appId`-addressed table without checking first.**
-  `pilots`, `mechs` and `crawlers` are looked up by the client's `appId` with
-  `.unique()`, which **throws** on a second row — and since mirrored writes are
-  fire-and-forget, that throw is invisible: local writes keep succeeding while
-  the server of record silently stops receiving them. `convex/maintenance.ts`
-  repairs rows already in that state (`dedupeAppIds`, dry-run by default).
+  `pilots`, `mechs` and `crawlers` are looked up by the client's `appId`, and
+  `by_app_id` is an ordinary index — **not** a uniqueness constraint — so
+  nothing in the database stops a second row. Prevention is the rule:
+  `appIdTaken` before any insert. `convex/maintenance.ts` repairs rows already
+  in that state (`dedupeAppIds`, dry-run by default).
+
+  The **lookups themselves no longer throw** on a duplicate. `byAppId` /
+  `crawlerByAppId` resolve to the oldest match and `console.warn`, because
+  mirrored writes are fire-and-forget: a throw here is invisible to the player,
+  so it does not fail the write, it just stops the write from ever reaching the
+  server while every surface keeps rendering it as saved. That is how an
+  evening of play went missing. A duplicate is a repair job — it is not a reason
+  to refuse the write that would have kept client and server in step. The
+  warning is the tripwire, and it fires without taking the player's data with
+  it. Resolving to the *oldest* is deliberate: it is the row `dedupeAppIds`
+  keeps, so a write that lands before the repair runs is not discarded by it.
 
 ## Commands
 
--- a/apps/itun/convex/entities.ts
+++ b/apps/itun/convex/entities.ts
@@ -18,6 +18,7 @@ import {
   requireTableRunner,
   requireUser,
 } from './model/permissions'
+import { entityRefType, softLinkType } from './schema'
 
 /**
  * Entity reads and writes against the server of record (ADR-030 §1).
@@ -242,6 +243,28 @@ export const update = mutation({
     const body = parseBody(args.table, args.body)
 
     await ctx.db.patch(doc._id, { body, updatedAt: Date.now() })
+  },
+})
+
+/**
+ * Delete an entity, addressed by its **server** id. Owner only.
+ *
+ * The twin of `removeByAppId`, which the mirror uses, and necessary because the
+ * Game roster cannot use that one: it is looking at rows this browser may never
+ * have held, and at pre-gens seeded from a template that have no `appId` at all
+ * (production holds a dozen). Addressing by `_id` is the only way to name those,
+ * and it is exactly how `update`, `ownership.release` and `removeCrawler`
+ * already address a row from that surface.
+ */
+export const remove = mutation({
+  args: { table: OWNABLE, entityId: v.string() },
+  handler: async (ctx, args): Promise<void> => {
+    const userId = await requireUser(ctx)
+    const doc = await loadOwnable(ctx, args.table, args.entityId)
+
+    assertMayWrite(doc, userId)
+    await ctx.db.delete(doc._id)
+    if (doc.appId !== undefined) await pruneSoftLinksFor(ctx, doc.appId)
   },
 })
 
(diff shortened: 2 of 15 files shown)


alxjrvs (commented):
Reviewed commit `865f27b7` (the branch's only real commit — `git diff main...HEAD` also picks up already-merged dependency bumps and the CLAUDE.md merge-queue doc because the local `main` ref is stale; those are not part of this PR).

Findings, most severe first:

- `apps/itun/convex/repair.ts:95` — `pickSurvivor` keeps the **newest** duplicate while `byAppId` resolves every mirrored write to the **oldest**. Both behaviors are pinned by tests in this same PR, so `repair:dedupe --apply` deletes precisely the row that has been absorbing all post-fix edits and keeps the frozen pre-duplication snapshot. Data loss in a script whose stated purpose is repair.
- `apps/itun/convex/repair.ts:113` — `collectDuplicates`/`pickSurvivor` group by `appId` and never consult `ownerId`, so cross-account duplicates (the shared-browser claim path the module header itself describes) collapse into one row and one player's character is permanently deleted.
- `apps/itun/convex/entities.ts:322` — `removeCrawler`/`removeCrawlerByAppId` never call `pruneSoftLinksFor`, so scrapping a crawler leaves mirrored `pilot-to-crawler` links pointing at a row that no longer exists; `listForGame` keeps serving them. Inconsistent with the cascade this PR added to `remove`/`removeByAppId`.
- `apps/itun/convex/entities.ts:622` — `upsertByAppId` re-homes an entity's `gameId` but not its links' `gameId`. Moving an entity out of a Game leaves wires stamped with the old game; moving one in leaves its wires invisible to the crew. The PR's own re-home test only passes because it re-issues `upsertSoftLink` manually after the move.
- `apps/itun/convex/entities.ts:393` — the new `claimLocal` idempotency guard skips a row when *any* row holds that `appId`, including another user's. A second account claiming on a device that holds a first account's cached roster claims nothing, then every mirrored write is refused with `NotAuthorized` — now surfacing the new user-facing toast on a 30s loop.

Things I checked and found sound: `byAppId`/`crawlerByAppId` duplicate tolerance and deterministic oldest-wins; `upsertSoftLink`/`removeSoftLink` permission anchoring and endpoint idempotency; `pruneSoftLinksFor` index coverage (`by_from`/`by_to` both exist); `entities.remove` owner-only gating plus the client `forget` (not `delete`) pairing in `GameRoster`; the mirror-failure toast throttle; `AssignPilotToCrawler`'s hook ordering, `from`-end direction, and already-crewed filtering (soft links are hydrated by the sheet route, so its direct `useEntityStore` read is safe).

alxjrvs (commented):
Reviewed commit `414763d3` (the only commit ahead of `origin/main`; no uncommitted changes) — soft-link mirroring, `entities.remove`, duplicate-`appId` tolerance, the Game-roster Delete verb, and the new `AssignPilotToCrawler` control.

Findings (most severe first):

- `apps/itun/src/components/wiring/AssignPilotToCrawler.tsx:90` — the crew picker filters only on "already crewing this crawler", not on container or ownership; picking a shelf pilot files the mirrored link under `gameId: null`, and picking a crewmate's pilot adopted via roster drill-in makes `upsertSoftLink` throw `NotAuthorized`, which now surfaces as an error toast while the local sheet keeps showing the pilot aboard.
- `apps/itun/src/stores/entityBackend.ts:303` — nothing reads `listForGame().softLinks`: `makeFrozenStore` hardcodes `softLinks: []` (`src/components/sheet/frozenSheet.ts:108`) and `GameRoster.tsx` ignores the field, so the headline fix ("nobody else ever saw the assignment") is plumbed but not observable. The comments, `apps/itun/CLAUDE.md` and `docs/architecture/accounts-and-games.md` all assert a read side that does not exist.
- `apps/itun/src/stores/entityStore.ts:312` — links are mirrored only on link create/update/delete, so `MoveToContainerControl` re-homes the entity server-side and leaves its links in the old container permanently; the re-home branch in `upsertSoftLink` is unreachable for the move case it was written for.
- `apps/itun/convex/entities.ts:316` — `pruneSoftLinksFor` was wired into `remove`/`removeByAppId` but not `removeCrawler`/`removeCrawlerByAppId`, so scrapping a crawler leaves every `pilot-to-crawler` row pointing at it, while the client still cascades locally.
- `apps/itun/convex/entities.ts:676` — `byAppId` picks the oldest duplicate by `_creationTime`, but `dedupeAppIds`' `rankForKeeping` ranks owned-before-unowned first; if the oldest duplicate has been released (`ownerId: null`), every mirrored write refuses with `NotAuthorized` forever — the exact outcome this change set out to eliminate. The new maintenance test's "lands on the row dedupeAppIds will keep" only holds when ownership and `updatedAt` tie.

Nothing wrong found in the toast throttling, `findSoftLink` identity semantics, `loadOwnable` id normalisation, `crawlerByAppId`, or the `entityRefType`/`softLinkType` schema exports (Convex `api` types pick up the new mutations without regenerating `_generated/api.d.ts`).

PR #707, opened 2026-08-06
fix(itun): give every seeded Starter Set row its own UUID

alxjrvs:
Follows #704. That PR stopped duplicate `appId`s from *corrupting* anything;
this one removes the reason they existed.

## A copy is a new thing, and gets a new id

Every path in the app already worked that way. The investigation checked all
of them:

| path | behaviour |
| --- | --- |
| `crud.create()` | mints `crypto.randomUUID()`; its input type is `Omit<T, 'id' \| …>`, so a caller **cannot** supply an id |
| `mergeImport` (export → import) | strips `id` off every row, creates fresh, remaps soft links through an old→new map |
| snapshots (`/s/$id`) | read-only; explicitly never touch entityStore |
| moving Shelf ↔ Game | patches `gameId` only — same row, same id |
| `seedStarterSet` | **wrote 12 hardcoded ids** |

So this corrects the record on something I'd written in #704: export/import
does *not* preserve ids. It never did. The one real offender was the Starter
Set.

## Why the Starter Set did it, and why that stopped working

`seedStarterSet` bypassed `create()` and `put` the template's own ids —
`starter-pilot-bonesaw`, `starter-mech-scrapper`, … — because id equality
doubled as the idempotence guard: a `put` of a known id overwrites, so a
double-click could not duplicate.

That reasoning was sound while the ids never left the device. It stopped
being sound when accounts landed. A seeded row's `id` becomes its `appId` on
the server of record, where rows are looked up with `.unique()` — so **every
player who seeded this roster carried the same twelve ids**, and any two of
them claiming collided on all twelve:

- before #704 → `unique()` threw, and both players' mirrors broke permanently
- after #704 → the second player's starter crew is silently declined

Neither is acceptable, and neither is fixable downstream. The ids have to be
distinct at the point of copying.

## The fix

Seeding mints a UUID per row and records the template slug as **`seedRef` —
provenance, never identity**. That keeps the idempotence guard's original
shape (the roster is its own evidence, so a partially-deleted set re-seeds to
whole and an intact one is skipped) without borrowing identity to express it.
Soft links are rebuilt against the remapped endpoints, skipping any whose
endpoint is missing — the same rule `mergeImport` follows, for the same
reason.

The crawler placeholder-Game bug's sibling is fixed here too: only genuinely
missing rows are written, so a re-seed adds what was deleted and touches
nothing else.

**Migration 14** stamps `seedRef` onto rows seeded the old way, so an existing
roster still reads as present instead of being laid down a second time beside
itself. Their **ids are deliberately not re-minted** — a row already claimed
exists on the server under its old `appId`, and changing it here would strand
that row and re-mirror the entity as a new one, turning a collision that is
now merely declined into duplicated data.

`mergeImport` also drops `seedRef` now, for the same reason the field exists:
it records that a row was spawned from a built-in template *on this device*,
and an imported row was not. Carrying it across would make someone else's
starter pilot read as your own seeded one and quietly suppress the Starter Set
offer.

## Two invariants, now pinned by tests

They pull in opposite directions, which is exactly why neither should be left
to the reader:

- **Moving between containers keeps the id.** A Game is a viewport onto the
  same sheet, not a duplicate of it.
- **Importing a bundle mints a new id.** That file may have been shared, and
  two people holding one id is precisely what the server cannot represent.

## Verification

- `bun run check:all` green — lint, format, typecheck ×6, test, validate, knip, audit, tokens, styling
- `bun --filter itun test` — **1722 pass, 0 fail** (rebased onto main with #704 merged)
- 12 new tests, including: seeded rows get UUIDs not template ids; two devices seeding the same roster produce **disjoint** id sets; soft links point at the minted ids; a partially deleted roster re-seeds only what is missing; the migration stamps legacy rows and leaves a player's own builds alone.
- One pre-existing flake exists in this suite — `ShareSnapshotScreen › hides Publish…` times out at 5s. Verified identical on clean `origin/main`; unrelated to this change.

### DIFF
--- a/apps/itun/CLAUDE.md
+++ b/apps/itun/CLAUDE.md
@@ -141,6 +141,20 @@ and remains the account-free way to share a build.
   warning is the tripwire, and it fires without taking the player's data with
   it. Resolving to the *oldest* is deliberate: it is the row `dedupeAppIds`
   keeps, so a write that lands before the repair runs is not discarded by it.
+- **A copy gets a new UUID; a move keeps its own.** These pull in opposite
+  directions, so both matter:
+  - **Copy → new id.** Importing a bundle (`mergeImport`) and seeding the
+    Starter Set (`seedStarterSet`) each mint a fresh UUID per row and remap the
+    soft links onto them. An entity id becomes its `appId` on the server, so a
+    copy that kept its id would put one id in two accounts — and since a
+    duplicate now resolves to the oldest row, the second account's mirrored
+    writes aim at the first account's entity and are refused by `assertMayWrite`
+    as somebody else's. Never write a fixed or template id into
+    `pilots`/`mechs`/`crawlers`; record provenance in `seedRef`, which is what
+    it is for.
+  - **Move → same id.** Shelf ↔ Game is a container change, patching `gameId`
+    and nothing else. A Game is a viewport onto the same sheet, not a duplicate
+    of it — never re-mint on a move, and never treat one as a fork.
 
 ## Commands
 
--- a/apps/itun/src/lib/db/__tests__/starter-set-seed-ref.test.ts
+++ b/apps/itun/src/lib/db/__tests__/starter-set-seed-ref.test.ts
@@ -4,7 +4,8 @@
  * Seeding now mints a fresh UUID per row and records the template slug in
  * `seedRef`, because the old fixed ids (`starter-pilot-bonesaw`, …) were
  * identical in every player's browser — and a seeded row's id becomes its
- * `appId` on the server of record, where rows are looked up with `.unique()`.
+ * `appId` on the server of record, where two accounts sharing one resolve to a
+ * single row and the later player's writes are refused as somebody else's.
  *
  * This migration exists so that change does not cost anyone a duplicate
  * roster: without it, a browser that seeded the old way has no `seedRef`
--- a/apps/itun/src/lib/db/__tests__/starter-set-seed-ref.test.ts
+++ b/apps/itun/src/lib/db/__tests__/starter-set-seed-ref.test.ts
@@ -1,0 +1,128 @@
+/**
+ * Unit tests for the v14 record rewrite (Starter Set → `seedRef`).
+ *
+ * Seeding now mints a fresh UUID per row and records the template slug in
+ * `seedRef`, because the old fixed ids (`starter-pilot-bonesaw`, …) were
+ * identical in every player's browser — and a seeded row's id becomes its
+ * `appId` on the server of record, where rows are looked up with `.unique()`.
+ *
+ * This migration exists so that change does not cost anyone a duplicate
+ * roster: without it, a browser that seeded the old way has no `seedRef`
+ * anywhere, the "already seeded?" check answers no, and the next visit to the
+ * Roster lays a whole second crew down beside the first.
+ *
+ * The classifier is deliberately narrow, and both halves are pinned here: a row
+ * sitting under a known template id is stamped, and nothing else is touched.
+ */
+import { describe, expect, test } from 'bun:test'
+import { migrate } from '../migrations/14-starter-set-seed-ref'
+
+/**
+ * A stand-in for the versionchange transaction.
+ *
+ * The migration only ever opens a cursor and updates through it, so a cursor
+ * over a plain array is a faithful double — and it keeps these cases about the
+ * rewrite rule rather than about IndexedDB. The full ladder against a real
+ * database is `migration-ladder.test.ts`'s job.
+ */
+/** A raw stored row, exactly as it sits on disk — no Zod involved. */
+type Row = Record<string, unknown>
+
+/** The three stores this migration walks. Named, not indexed, so `[0]` narrows. */
+type Fixture = { pilots: Row[]; mechs: Row[]; crawlers: Row[] }
+
+function fakeTx(stores: Record<string, Row[]>) {
+  return {
+    objectStore(name: string) {
+      const rows = stores[name] ?? []
+      let i = 0
+      const cursorAt = (index: number): unknown => {
+        if (index >= rows.length) return null
+        return {
+          value: rows[index],
+          async update(next: Record<string, unknown>) {
+            rows[index] = next
+          },
+          async continue() {
+            i = index + 1
+            return cursorAt(i)
+          },
+        }
+      }
+      return {
+        async openCursor() {
+          i = 0
+          return cursorAt(0)
+        },
+      }
+    },
+  }
+}
+
+describe('v14 starter-set seedRef backfill', () => {
+  test('stamps a row still sitting under a template id', async () => {
+    const stores: Fixture = {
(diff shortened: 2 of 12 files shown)


alxjrvs (commented):
## Review complete — 2 findings

Scope: `git diff origin/main...HEAD` on `fix/itun-starter-set-unique-ids` (12 files; working tree clean). The change mints per-device UUIDs for seeded Starter Set rows, records provenance in a new `seedRef` field, adds DB migration v14 to backfill it, and strips `seedRef` on import.

**Findings**

- `apps/itun/src/lib/starterSet/seedStarterSet.ts:160` — **Legacy re-seed duplicates every surviving starter soft link.** Migration 14 stamps `seedRef` on pilots/mechs/crawlers but never touches `softLinks`, so a pre-v14 device keeps link ids in the old `starter-link-mech-<pilotId>` format. The re-seed loop derives ids in a new format (`starter-link-<type>-<from>-<to>`) and dedupes against `existingLinkIds`, which can never match. Delete one starter pilot from the Roster and click "Load Starter Set" again → 22 links instead of 12, with 10 duplicate `type:from->to` pairs. Reproduced against fake-indexeddb. The file's own comment at lines 150–153 asserts precisely the invariant that breaks. Fix: dedupe on the endpoint triple, or rewrite legacy link ids in migration 14.
- `apps/itun/src/lib/starterSet/seedStarterSet.ts:87` — **`isStarterSetSeeded()` widened from one store to three but is still a synchronous read.** `entityStore.list` returns `[]` for an unhydrated type, and `Roster.tsx:310` drives the "Load Starter Set" button from three independently-hydrating hooks, so a fully-seeded user gets a button flash. Cosmetic only — clicking is safe because `ensureStarterSetSeeded` awaits all hydration first.

**Verified clean:** `bun run typecheck:itun` passes; `bun test` in `apps/itun` is 1722/1722 on the committed tree. Migration 14 is safe against store existence (pilots/mechs/crawlers are created at `oldVersion < 1`), its inlined `STARTER_IDS` match the real template ids exactly, and its await-only-IDB constraint is respected. `mergeImport`'s `seedRef` strip is correct on all three entity paths. No server→local pull path exists that could import someone else's `seedRef`.

**Noted, not filed:** the migration deliberately does not re-mint legacy ids, so every user who already loaded the Starter Set keeps the colliding `starter-*` appIds and their claim is still declined server-side. That is a documented, reasoned scope decision (re-minting would strand already-mirrored rows), not a defect — but it does mean the PR fixes the collision only for new seeds.

**Housekeeping:** two untracked scratch test files (`src/lib/starterSet/__tests__/scratch-legacy.test.ts`, `scratch-fresh.test.ts`) appeared and then vanished mid-run during this review — presumably from a concurrent session. The tree is clean now; worth confirming they don't get committed, since the legacy one fails by design.

alxjrvs (commented):
Reviewed commit `a06b0de3` ("give every seeded Starter Set row its own UUID") — the only commit on this branch not in `origin/main`. Scope: seedStarterSet rewrite, migration 14, `seedRef` on the three entity schemas, `mergeImport` stripping, plus tests.

Verification run: `bun test src/lib/starterSet src/lib/db/__tests__ src/lib/export` (154 pass), `bun run typecheck:itun` (clean), `bun run knip` (clean). I also wrote two throwaway scratch tests to exercise the upgrade path and deleted them afterwards — the worktree is clean.

Findings:

- `apps/itun/src/lib/starterSet/seedStarterSet.ts:160` — CONFIRMED. Partial re-seed on a pre-v14 (upgraded) device duplicates every surviving crew's soft links. The derived link id `starter-link-<type>-<from>-<to>` never matches the legacy `starter-link-mech-<pilotSlug>` / `starter-link-crew-<pilotSlug>` ids, and migration 14 doesn't touch the `softLinks` store, so `existingLinkIds` never hits. Reproduced: legacy roster + delete one pilot + re-seed → 22 links instead of 12, 10 duplicated endpoint pairs; `resolveSheetComposition` then lists 5 of 6 pilots twice on the crawler sheet. A fresh v14 device is fine (verified 12 links), so this lands squarely on the upgrade path migration 14 exists to protect. Fix direction: have migration 14 also rewrite the legacy link ids into the new derived form, or dedupe links by endpoint pair rather than by id.
- `apps/itun/src/lib/starterSet/seedStarterSet.ts:117` — PLAUSIBLE, lower severity. Dropping deterministic ids turns the seed into a check-then-write, losing the old "a re-entrant seed overwrites rather than duplicates" property. Two Roster tabs each clicking "Load Starter Set" (the disable is per-tab local state) both see an empty `present` set and both write a full roster — 12 pilots / 12 mechs / 2 crawlers / 24 links, with no dedupe path. The module header still claims idempotence.

Everything else checked out: migration 14's inlined id list matches `starterSet.ts` exactly and follows migration 13's cursor pattern; `seedRef` is accepted by the export bundle, frozen-sheet and Convex `parseBody` paths (all reuse the same Zod schemas); `mergeImport` strips it on all three entity types; entity deletes do cascade-prune soft links, so the re-mint doesn't strand orphans on a fresh device.

alxjrvs (commented):
Reviewed the 3-commit diff (`origin/main...HEAD`): the Starter Set unique-ID change (`seedStarterSet.ts`), migration 14, the `seedRef` schema fields, and the `mergeImport` strip. I read the surrounding store/DB/Roster code and empirically exercised both the legacy and fresh re-seed paths against the real IndexedDB layer (scratch test written, run, and removed — working tree is clean).

One real bug found:

- `apps/itun/src/lib/starterSet/seedStarterSet.ts:167` — Legacy re-seed duplicates every starter soft link. The new derived link id (`starter-link-<type>-<from>-<to>`) never matches the ids a pre-change device already stored (`starter-link-mech-<pilotId>` / `starter-link-crew-<pilotId>`), so the `existingLinkIds` skip at line 168 misses and a partial re-seed writes a second copy of every surviving crew link. Verified: on a device seeded the old way + stamped by migration 14, deleting one starter pilot and clicking "Load Starter Set" takes soft links from 10 → 22 (expected 12); each surviving pilot ends with two `mech-to-pilot` and two `pilot-to-crawler` links, so `Roster.tsx:512` renders every pilot twice in the crawler's crew line and `dashboardLinks.ts`'s one-outgoing-link assumption is broken. The same sequence on a device seeded with the new code correctly stays at 12 — this hits only the migrating cohort, i.e. exactly the users migration 14 exists to protect.

Suggested fix direction: treat a legacy link as equivalent by endpoints rather than by id (skip when a link of the same `type`/`from`/`to` already exists), or have migration 14 also rewrite starter link ids into the new derived form.

Everything else checked out: Convex entity bodies are `v.any()` so the new `seedRef` field mirrors fine; `entityStore.delete` does cascade link pruning, so the fresh-device partial re-seed leaves no dangling or duplicate links; the captured `store` reference is safe because `list()` reads through `get()`; migration 14 is correctly registered and `DB_VERSION` bumped; `mergeImport` strips `seedRef` on all three entity kinds. The existing 154 tests in the touched areas pass.

PR #710, opened 2026-08-06
docs(itun): an entity is never copied into a game — amend ADR-030's fork clause

alxjrvs:
> I do not want games to "Copy" these fields. Entities in your shelf can or can not belong to a game, but they are always the same entity, not copies that need to be mirrored.

The code already worked this way. **ADR-030 did not**, and that mismatch was pointed the wrong direction: `MoveToContainerControl` re-stamps `gameId` in place and `entities.upsertByAppId` re-homes the existing row, while the ADR called a cross-container move *"an explicit **fork** that copies and records its origin"* — and the control's own header apologised for not having built one yet. That is the setup for somebody later "fixing" correct code to match a doc.

## The fork clause did not follow from its own premise

It justified forking with: *a pilot's crawler level, scrap, TP and injuries are Game-specific, so a shared pilot would be incoherent.*

But `gameId` is a **single nullable column**, so an entity is in at most one Game *by construction*. There is no shared pilot to be incoherent about. Forking guarded a state the schema cannot represent, and would have paid for it with two records of one character to keep in step.

## Move and copy, as two different verbs

| Verb | Result |
|---|---|
| **Move** | The **same** entity. Same id, body and history; only `gameId` changes. Nothing to reconcile. |
| **Copy** | A **new** entity. New id, `COPY OF <name>`, always `gameId: null`, no tie to the source or its Game. |

Copying is explicitly fine and wanted — it is a deliberate, user-initiated "I want my own one of these." What makes it safe is that it keeps **no relationship**: it does not track its origin, sync with it, or follow it into or out of a Game. That is also exactly what separates it from the fork being removed here, which "records its origin" — and a retained relationship between two records of one character is the part that rots.

Copy-to-shelf is **not built**; it is recorded in the phase list in `accounts-and-games.md`.

## Changes

- `ADR-030` §2 — amended in place with a dated note quoting the original clause, so the reasoning is auditable rather than just gone.
- `accounts-and-games.md` — decision table and phase list corrected ("Fork into Game" → move/copy).
- `MoveToContainerControl.tsx` — header rewritten so in-place re-stamping reads as the design, not a stopgap, and notes that it and `upsertByAppId` must keep agreeing.

**No behaviour change.** `check:fast` green.

## One open question, deliberately not decided here

The request said a copy should land "unclaimed". Taken literally as `ownerId: null` that collides with a documented schema invariant — `gameId: null && ownerId: null` is listed in ADR-030 as the one **invalid** combination ("a shelf is a person's, so an unclaimed shelf entity would be owned by nobody and reachable by nobody"), and `entities.create` throws on it. This PR therefore reads "unclaimed" as *unattached to any Game* and leaves `ownerId` as the copier. Flagging it rather than quietly picking.

### DIFF
--- a/apps/itun/src/components/container/MoveToContainerControl.tsx
+++ b/apps/itun/src/components/container/MoveToContainerControl.tsx
@@ -6,14 +6,26 @@
  * not a dialog, so a move is one tap — with Workspaces swapped for the two real
  * containers.
  *
- * ## What this does NOT do
+ * ## A move is one field, and that is the whole design
  *
- * ADR-030 says a cross-container move is an explicit **fork**, and no fork
- * mutation exists yet (`convex/entities.ts` has create/update/claimLocal, not
- * fork). So this keeps exactly the semantics of the control it replaces: it
- * re-stamps the local record's container in place. Turning that into a fork is
- * a separate change with its own server surface, not something to smuggle in
- * behind a rename.
+ * Re-stamping `gameId` in place is not a shortcut pending something better —
+ * it is the model. There is **one entity**. The pilot on your shelf and that
+ * same pilot in a Game are one record with one field set differently, so a move
+ * changes that field and nothing else: same id, same body, same history.
+ *
+ * This header used to say the opposite, because ADR-030 §2 originally called a
+ * cross-container move an explicit **fork** and this control was documented as
+ * holding the line until a fork mutation existed. That clause was amended
+ * (2026-08-06): forking guarded against a "shared pilot" that the schema makes
+ * unrepresentable — `gameId` is a single nullable column, so an entity is in at
+ * most one Game by construction — and it would have bought that non-protection
+ * with a duplicate character to reconcile.
+ *
+ * So there is no missing fork mutation to add. `entities.upsertByAppId` re-homes
+ * the existing server row for the same reason (`existing.gameId !== args.gameId`
+ * patches the column rather than inserting), and the two halves must keep
+ * agreeing: if either ever starts copying, a player ends up with two of
+ * themselves and no way to tell which one the table can see.
  *
  * ## Solo renders nothing
  *
--- a/docs/adrs/ADR-030-accounts-games-server-of-record.md
+++ b/docs/adrs/ADR-030-accounts-games-server-of-record.md
@@ -97,9 +97,50 @@ An entity is in exactly one container, encoded as two nullable columns:
 | null     | set       | on the owner's shelf                                   |
 | null     | null      | **invalid** — must be unreachable through any mutation |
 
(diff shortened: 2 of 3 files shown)


alxjrvs (commented):
**Scope note:** `git diff @{upstream}...HEAD` was empty (branch already pushed) and local `main` is ~6 commits stale, so `main...HEAD` pulled in 50 already-merged files. The real PR (#710, base `main`) is one commit against `origin/main`: 3 files, +67/−12, and it is **documentation only** — an ADR amendment, an architecture-doc update, and a rewritten JSDoc header. No executable code changed, so there are no runtime-correctness findings; the three below are defects in what the diff asserts.

**Claims I verified as accurate** (worth stating, since they carry the argument): `entities.upsertByAppId` really does re-home via `existing.gameId !== args.gameId` → `patch` (convex/entities.ts:729-731), never insert; `gameId` really is a single nullable column on `pilots`/`mechs`/`crawlers` (convex/schema.ts:256, 279, 319) and on the local Zod schemas (lib/schemas/pilot.ts:152); copy-to-shelf really is unbuilt; and no stale "moving is a fork" instruction survives anywhere else in `docs/`, `apps/itun/src`, or `apps/itun/convex`. Markdown is deliberately untooled here (lefthook has an explicit "no Markdown/YAML formatting step" comment, and `format:check` is `biome format .`, which does not parse `.md`), so the new tables' column misalignment is not a CI failure and I did not file it.

Findings:

- `docs/adrs/ADR-030-accounts-games-server-of-record.md:126` — The amendment's rebuttal disproves *simultaneous* dual-Game membership but not the sequential Game A → Game B move, which is where the deleted clause's concern actually bites. `moveTo()` returns `{ gameId }` alone (lib/container.ts:85) and `upsertByAppId` patches `body` unchanged, so a pilot arrives in Game B still carrying Game A's `crawlerLevel`, `trainingPoints`, and `injuries`. The ADR now tells implementers the in-place move "IS the design" with that case unaddressed and no longer documented anywhere.
- `docs/adrs/ADR-030-accounts-games-server-of-record.md:3` — The `## Status` block was not updated to record the 2026-08-06 amendment, though ADR-022 sets the convention (`**Amended 2026-07** — …` in Status, linking a headed `## Amendment` section) and CLAUDE.md instructs readers to consult Status first. A reader following the documented convention can still come away believing the fork clause stands — the exact outcome this PR exists to prevent. The amendment is also an unheaded blockquote, so nothing can link to it.
- `docs/architecture/accounts-and-games.md:101` — A bullet reading "Not built yet" was added inside `### Phase 1 — Accounts, Games & sync ✅`, whose exit criterion is declared met. It replaced a line that read as delivered, so the phase now both claims completion and enumerates an unshipped deliverable; either defer the item to a later phase or drop the ✅.

PR #711, opened 2026-08-06
feat(itun): copy an entity to your shelf, and stop calling a shelved build unclaimed

alxjrvs:
> A game can have either a null ownerId or a null gameId. if ownerId null, it is unclaimed. If gameId null, it is on the owners shelf. if both are set, it is a claimed entity in a game. if neither are set, that is invalid.

Both changes here fall straight out of that table, which is ADR-030 §2 verbatim. I checked the invariant is already enforced at all four paths that can write `ownerId: null` (`entities.create` unassigned, `templates` seeding, `ownership.release`, `ownership.leaveGame`) — it is. These are the two places the *model* was not being told correctly.

## 1. Copy to shelf

`ownership.leaveGame` has documented this act since before it existed:

> Entities stay behind, unclaimed, so the crew survives a departure... **a player who wants to keep a character copies it to their shelf first, which is a separate, explicit act.**

That act did not exist. So leaving a game meant your character stayed behind unclaimed and you had **no way to keep one** — a flow shipped without its own documented escape hatch.

A copy is a **new entity with no relationship to its source**: new id, `COPY OF <name>`, always `gameId: null`, owned by the copier. It does not track its origin, sync with it, or follow it into or out of a Game. That absence of a relationship is precisely what makes copying safe — and what separates it from the **fork** ADR-030 dropped in #710. A fork "records its origin", and a retained link between two records of one character is the part that rots.

**Where it appears:** every row in the Game roster, including a crewmate's and an unclaimed pre-gen. It is derived from what you may already read — membership grants the frozen crew view of every row, so copying what is on your screen escalates nothing. No confirm dialog: copying destroys nothing, so a modal would be friction guarding an undo-by-delete.

**Crawlers are excluded**, and not arbitrarily: `crawlers.gameId` is **non-nullable** in the Convex schema, so a shelved crawler has no server row it could be. That is the same constraint that makes `claimLocal` park a claimed crawler on a placeholder Game rather than shelving it.

**On "unclaimed".** The request said a copy should land unclaimed. Read as `ownerId: null` that is row 4 — the invalid one — and `entities.create` already refuses it ("A shelf has no crew to pick anything up — shelved builds are yours"). Unclaimed is a thing a character can be *inside a Game*, waiting for somebody to take it; a shelf is already somebody's. So a copy lands row 3: `gameId: null`, owner = copier.

## 2. `release` was telling players the opposite of that table

Its refusal on a shelved entity read:

> An entity on a shelf is already unclaimed

A shelved build is `gameId: null, ownerId: set` — **on the owner's shelf**, not unclaimed. The guard itself was correct (releasing it would set `ownerId: null` alongside `gameId: null`, which is the invalid row, so refusing is right) — only the sentence was wrong. It matters more now that #704 made `NotAuthorized` extend `ConvexError`, so this string is *shown* to the player rather than redacted.

Rewritten, and the existing test strengthened to assert the invariant and that the message no longer says "unclaimed", rather than asserting the old wording.

## Testing

- `check:all` green (exit 0).
- 9 new unit tests on `copyForShelf` — the ones that matter are what a copy does **not** keep: no `id` (a copy carrying its id is a duplicate `appId` by hand — the exact condition that took the play session down), no `gameId`, and no legacy `workspaceId`, which `containerOf` still reads as a fallback and could resolve a copy back into the Game it was copied out of.
- 3 new `GameRoster` tests; 2 rewritten in `permissions.test.ts`.

### DIFF
--- a/apps/itun/convex/ownership.ts
+++ b/apps/itun/convex/ownership.ts
@@ -184,7 +184,23 @@ export const release = mutation({
   handler: async (ctx, args): Promise<void> => {
     const doc = await loadOwnable(ctx, args.table, args.entityId)
     if (doc.gameId === null) {
-      throw new NotAuthorized('An entity on a shelf is already unclaimed')
+      /*
+       * A shelved entity is NOT unclaimed, and this used to say it was.
+       *
+       * ADR-030 §2 allows three states, and they are not interchangeable:
+       * `gameId` set + `ownerId` null is **unclaimed** — in a Game, waiting for
+       * somebody to take it; `gameId` null + `ownerId` set is **on the owner's
+       * shelf** — already somebody's. Both null is the one invalid combination,
+       * which is exactly what releasing a shelved entity would produce, and is
+       * why this refuses rather than no-ops.
+       *
+       * The message matters now that it reaches a player: `NotAuthorized`
+       * extends `ConvexError`, so this string is shown rather than redacted, and
+       * it was teaching the wrong model to the person least able to check it.
+       */
+      throw new NotAuthorized(
+        'A build on your shelf is already yours — there is no crew here to release it to. Move it into a game first.'
+      )
     }
     if (doc.ownerId === null) return
 
--- a/apps/itun/src/components/games/GameRoster.tsx
+++ b/apps/itun/src/components/games/GameRoster.tsx
@@ -47,6 +47,7 @@ import {
   ModalShell,
   PageHeading,
   Text,
+  toast,
 } from 'component-lib'
 import { useMutation, useQuery } from 'convex/react'
 import { Bot, UserRound, Warehouse } from 'lucide-react'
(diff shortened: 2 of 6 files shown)


alxjrvs (commented):
Reviewed `origin/main...HEAD` (6 files: new `copyEntity.ts` + "Copy to shelf" on `GameRoster`, and the `ownership.release` message change). Typecheck, knip, and both new test files are green locally.

Findings:

- `apps/itun/src/lib/copyEntity.ts:59` — `DROPPED_FIELDS` omits `seedRef`. Copying a Starter-Set-derived pilot out of a Game makes `seededRefs()` treat the copy as that template row, so `ensureStarterSetSeeded()` never writes the real one and wires the starter soft links to the copy. `mergeImport.ts:157` drops `seedRef` for exactly this reason and documents it.
- `apps/itun/convex/ownership.ts:202` — the new refusal says "A build on your shelf is already yours", but `loadOwnable` does no authorization and this branch runs before any owner/member check, so it fires with that wording for *another* user's shelved entity. `claim`'s equivalent branch is ownership-neutral ("on somebody's shelf").
- `apps/itun/src/components/games/GameRoster.tsx:356` — `copyToShelf` toasts success for a build the Roster filters out: in Connected mode the Roster scopes to the persisted `activeContainer`, which is the Game after any use of `enterGameContainer`, and copying never changes it.
- `apps/itun/test/convex/permissions.test.ts:519` — `.catch((e) => e)` yields `undefined` on success, and `String(undefined)` doesn't match `/unclaimed/i`, so this test passes even if the whole shelf guard is deleted. Assert the rejection first.

No issues found with the id/createdAt/updatedAt/workspaceId drops, the explicit `gameId: null` (correctly survives `withActiveContainer`), the crawler exclusion, or the server-side shelf insert path (`assertMayAddToContainer` short-circuits on `null`).

PR #713, opened 2026-08-06
refactor(itun): route every reference-entity picker through EntitySearcher

alxjrvs:
## The bug

The **Change Chassis** modal laid its options out in a fixed `220px` master pane
beside a preview pane. A chassis is one of the largest entities in the data —
artwork, stat block, chassis ability, patterns — and that track is narrower than
the card's own artwork, so every option rendered clipped and unreadably tall. The
preview pane had no height cap or scroll container of its own either, so it and
the option list scrolled at different rates inside the dialog.

## What changed

Three modals ran that bespoke master/detail shape instead of the shared picker:

| Modal | Was | Now |
|---|---|---|
| Change Chassis | `grid-cols-[220px_1fr]` + `ChassisOptionList`/`ChassisDetail` | `EntitySearcher` |
| Change Crawler Type | `grid md:grid-cols-2` + `CrawlerTypeOptionList`/`Detail` | `EntitySearcher` |
| Change Class (pilot sheet) | `grid-cols-[320px_1fr]` + `ClassOptionList`/`ClassDetail` | `EntitySearcher` |

All three now use `EntitySearcher` — the same picker the equipment, ability,
system, module, crawler-weapon and catalog-choice modals already used. There is
one picker now, not two layouts drifting apart. The searcher's two-column masonry
gives each card its natural width, and the body scrolls internally.

## Three additions to `EntitySearcher`

Each is driven by what these pickers actually needed:

- **`mode="single"`** — exactly-one selection with real radio semantics
  (`radiogroup` + `aria-checked`). Emission is unchanged from `toggle`, so this
  is an a11y difference, not a wiring one.
- **`hide`** — merged over the searcher's own defaults, letting a caller drop
  sections a picker cannot act on (`{ patterns: true }` for chassis; patterns are
  a separate step, and swapping chassis wipes them anyway).
- **`railActions` + `subtitle`** — a destructive picker needs a confirm affordance
  and its warning copy, and `bare` ModalShell paints no header of its own, so the
  "Swapping chassis clears the current loadout" caveat had nowhere else to live.

## Dead code removed

`ChassisStep.tsx` and `ClassStep.tsx` lost their last production consumers.
`ClassStep` was left importable only by its own story, which the package's
"nothing lives only in the catalog" rule forbids, so both are deleted along with
the barrel exports and the stale `package-contracts.md` entry.

## Ladle

`Compositions/Catalog/Entity Searcher` gains a **Big Entity Single Select** story
driving the real chassis pool — the exact case that looked broken.

## Verification

- `check:all` exits 0 (lint, format, typecheck, test, knip, audit, tokens,
  styling, schema-drift, doc-drift).
- New `MechChassisPickerModal` tests pin the part most likely to break silently:
  the picker must emit `nameToSlug(name)`, since the mech record stores a chassis
  slug while the searcher's default identity is the entity name. They also cover
  the pre-selected rail, the disabled-until-changed Apply, and that `onConfirm`
  fires only after the danger dialog.
- Rendered the story in Ladle at 1440 and 390 — no console errors, no horizontal
  page overflow.
- **Pre-existing e2e failures, not from this PR:** 8 specs (print-preview,
  segment-switch, wired-build-and-snapshot, pilot-build, workspace-crud) fail
  identically on unmodified `origin/main` — verified by checking out `main` in the
  same worktree and re-running them. `workspace-crud` tests the retired
  workspaces feature. Untouched by this change; flagging rather than fixing.

## Known, out of scope

- `AddNpcControl` (encounter tray) is also a hand-rolled reference-entity picker,
  but it is an inline control rather than a modal and spans **six** schemas, which
  `EntitySearcher` cannot express (it takes a single `schema`). Left alone.
- A chassis card's nested roll-table "Show" control overflows its box below
  ~400px. Reproduces on the untouched `chassis-card` story, so it predates this
  work.

https://claude.ai/code/session_01NgN11qoWHcJjY22b1MhfnQ


alxjrvs (commented):
Reviewed commit `400a780f` — "refactor(itun): route every reference-entity picker through EntitySearcher" (11 files: `EntitySearcher` gains `mode="single"` / `hide` / `railActions` / `subtitle`; the chassis, crawler-type and pilot-class pickers migrate onto it; `ChassisStep.tsx` and `ClassStep.tsx` deleted).

Verification run: `bun run typecheck` (all 6 workspaces green), `bun run knip` (green), `bun --filter component-lib test` (751 pass), `bun --filter itun test` (1762 pass). The high-severity finding is untested and I confirmed it by rendering the component in a throwaway test (since deleted).

Findings:

- `apps/itun/src/components/mech/MechChassisPickerModal.tsx:50` — **CONFIRMED, high.** `changed = selected !== currentChassisRef` doesn't guard the empty case. Clicking the already-selected chassis card (or the rail's ✕ Remove) sets `selected` to `''`; contrary to the inline comment, Apply then *enables*, and confirming calls `onConfirm('')` → `patch({ chassisRef: '', systems: [], modules: [] })`. `chassisRef` is a bare `z.string()`, so an unresolvable chassis and a wiped loadout persist. Both sibling pickers migrated in the same commit guard this correctly (`selected !== null && …`, `!pendingClass || …`) — chassis is the one that doesn't.
- `packages/component-lib/src/components/shared/SheetSection.tsx:178` — low. `SheetPickerModal`'s `footer` prop lost its last call site here; the `floating` branch returns before rendering it, so a future `floating` + `footer` combination (which both docstrings still recommend) silently drops the confirm buttons. Not caught by typecheck or knip.
- `packages/component-lib/src/components/wizard/CrawlerTypeStep.tsx:22` — low. `CrawlerTypeOptionList` / `CrawlerTypeDetail` are now imported only by their own story file, yet still exported from the barrel — the "nothing lives only in the catalog" rule says delete them, and knip can't see it (`includeEntryExports: false` for component-lib).

Everything else in the diff checked out: the identity contracts (`nameToSlug` for chassis, `item.id` for crawler type and class) match what each record stores; `SheetPickerModal floating` correctly injects `title`/`onClose` into the searcher via `cloneElement`; `cardHide` merging, radiogroup semantics, and the deleted-module references are all clean.

alxjrvs (commented):
Reviewed the two-commit branch (`main...HEAD`): the three ITUN reference-entity pickers (chassis, crawler type, pilot class) migrated onto the shared `EntitySearcher`, plus a new `single` mode / `hide` / `subtitle` / `railActions` API in component-lib, the deletion of `ChassisStep` + `ClassStep` + the crawler master/detail pair, and `SheetPickerModal`'s `footer` prop.

Gates I ran: `typecheck` (all 5 workspaces green), `knip` (clean, hints only), `component-lib test` (751 pass — story-coverage/taxonomy guards included), and the new `MechChassisPickerModal.test.tsx` (5 pass). No dangling importers of the deleted components; the empty-selection guard is present in all three pickers (`selected !== ''`, `selected !== null`, `!pendingClass`), so the "Apply an empty ref" class of bug is genuinely closed.

Findings (3):

- `apps/itun/src/components/mech/MechChassisPickerModal.tsx:91` — CONFIRMED: chassis pool cards contain nested roll-table `Show`/`Roll` buttons inside the card's own `role="radio"` click handler. Rendering the picker and clicking Jackhammer's nested `Show` emits `onToggle('jackhammer')`; done on the already-chosen card it clears the selection and disables Apply. `main`'s `extent="catalog"` option pane had 0 nested buttons on all 51 chassis, so this is new. Fix: add `rollTable: true` to the `hide` config (it isn't force-overridden), or stop propagation in the searcher.
- `apps/itun/src/components/sheet/PilotIdentity.tsx:265` — CONFIRMED: the Change Class picker now shows name + flavour + source only. `ClassDetail`'s `afterExtraContent={<ClassAbilityTree/>}` is gone and the searcher force-hides `choices`, so no class's ability trees are visible anywhere in the swap flow; the "Advanced / Hybrid · Requires 6 core abilities" caveat was lost too (the `Kind` facet filters, it doesn't state the rule).
- `packages/component-lib/src/components/shared/EntitySearcher.tsx:487` — PLAUSIBLE: `subtitle` renders through `BandTitle`, which is hard-wired to `truncate`, and it is now the sole home of each picker's destructive-change warning ("Swapping chassis clears the current loadout.", "Resets the special NPC · keeps bays and their crew") — clipped to an ellipsis at phone width.

PR #714, opened 2026-08-06
feat(itun): represent hybrid pilot classes

alxjrvs:
Every Hybrid Class was offered **two of its four ability trees**. Advancing a pilot made them *less* representable: a base Hacker reaches five trees, a Cyborg reached two.

```
Cyborg      2 of 4   missing Augmentation, Gladiatorial Combat
Fabricator  2 of 4   missing Forging, Electronics
Ranger      2 of 4   missing Survivalist, Sniper
Smuggler    2 of 4   missing Sleuth, Salvaging
Union Rep   2 of 4   missing Leadership, Mechanical Knowledge
```

`treesFor()` began from `cls.coreTrees`, which hybrid records do not have, so it started from nothing and recovered only trees the pilot already held picks in. A freshly-created Cyborg could never be given a Gladiatorial Combat ability — the picker never offered one.

## The mapping was already in the dataset

`ability-tree-requirements.json` holds each hybrid's adjacency trees and was already parsed, for one display-only call site. Nothing routed it into logic. **No new data, no new schema.**

The five advanceable Core classes form a ring; each Hybrid sits on an edge between two of them, reachable from either end through whichever of its two trees that class owns. From any origin: **keep one tree, gain one from your neighbour, seal two.** Sealed never means lost — abilities already trained stay usable (p. 321).

## Origin is derived, not stored

A hybrid's sealed pair depends on where they came from, and `classRef` holds only the destination. But the origin is *recoverable*: advancing legally takes 6 Core abilities — 3 in the gate tree, which the hybrid grants and so proves nothing, plus 3 more that only the origin's own two trees can supply. Those are disjoint across all five hybrids (asserted in tests), so a rules-legal pilot always carries decisive evidence.

So `originClassRef` is an **override**, not the record of truth: absent on the ordinary pilot, written only when a player answers a disambiguation. Additive-optional — no migration, no `schemaVersion` bump.

When the origin cannot be resolved — no abilities, only granted trees, or a free-edited pilot whose trees point two ways — **nothing is sealed**. Closing the wrong two is worse than closing none, and that state is reachable by a row rather than a click: the class picker has been ungated for as long as it has existed.

## The picker

Grouped by where this pilot can actually go — the two hybrids on their edges (each labelled with its gate tree), a re-home to any Core class, then the rest, demoted and labelled `asks origin` rather than withheld. Free Edit gates nothing; those rows are not inert, they just cost one extra question.

Beside it a Gain / Lose / Keep ledger. LOSE names what forecloses and what seals; KEEP always leads with the retained abilities, because the likeliest misreading is "do I lose what I already have?"

The origin usually needs no asking: the pilot's **current class is the origin**. Class and origin are written as one patch, since `patchPilot` routes `classRef` changes through the soft-warning confirm.

## Not in this PR

The sheet's path band and the sealed-above / open-below split, plus the sealed-trees facet in the ability picker. The trees are correct now; showing *which are sealed* on the sheet is the follow-up.

## Verification

Full suite green. New tests pin all ten ring edges against the book, both directions, plus the disjointness the inference rests on — real dataset throughout, since a fixture would only verify the derivation against itself.

Design report: https://claude.ai/code/artifact/98df4163-e82b-4d4e-8d12-eecc96e5604c
Mockups: https://claude.ai/code/artifact/75fea047-0640-4971-bc2b-ee16d256dc19

https://claude.ai/code/session_01HpMLz1zhRq9TRHRjLitAJe


alxjrvs (commented):
Reviewed the 5-commit hybrid-classes branch (advancement rules in `salvageunion-reference`, the ITUN `ClassPathPicker` / `pilotAdvancement` layer, the `originClassRef` schema field, and the hybrid ability-tree fix). All 41 tests in the touched files pass; every finding below was verified by executing the code against real reference data, not by reading alone.

Findings:

- `apps/itun/src/components/pilot/classPathOptions.ts:189` — When the origin is implied by the current class, `needsOrigin` is false and `classChangePatch` writes `originClassRef: undefined`, while the picker tells the player "Recorded as Hacker — derived from this pilot's trees". Verified: a pilot holding only gate-tree abilities resolves to `originSource: 'unresolved'`, `sealed: []`. The implied origin is lost.
- `apps/itun/src/components/pilot/classPathOptions.ts:82` — Current class is matched by strict `c.id === currentClassRef`, but seeded/imported pilots store a slug (`classRef: 'hacker'`). Verified: `classPathGroups(all(), 'hacker')` produces no "From Hacker" group, demotes Cyborg/Fabricator to "asks origin", and the consequence reports `needsOrigin: true` with no seals.
- `apps/itun/src/lib/rules/pilotAdvancement.ts:80` — `heldTreesOf` matches `ability.id` only; `pilotAbilities.ts` uses `matchesRef` specifically because seeded pilots store kebab ability slugs. Verified: non-id ability refs yield zero held trees, so inference is always `ambiguous` and nothing seals.
- `apps/itun/src/components/sheet/PilotIdentity.tsx:148` — `confirmClass` skips the patch when `pendingClass === pilot.classRef`, but the picker asks the origin question for a pilot already in that hybrid. Answering and confirming silently writes nothing.
- `apps/itun/src/lib/rules/pilotAdvancement.ts:92` — `resolvePilotAdvancement` has no production consumer and `originClassRef` is read by no surface, so the sealing the picker promises ("each answer seals a different pair of trees") does not happen anywhere. Knip stays green only because the test counts as a consumer.

The other half of the PR — `offeredAbilityTrees` giving hybrids all four trees, and the de-duplication of the two identical `treesFor`/`editTreesFor` copies — checks out against the data (`Cyborg` correctly yields Augmentation, Gladiatorial Combat, Cyborg, Legendary Cyborg), as does the ring derivation in `advancement.ts` (candidate exclusivity is genuinely disjoint across all five hybrids).

alxjrvs (commented):
## Review complete — 3 findings

Scope: `git diff main...HEAD` (6 commits, 25 files) — the pilot advancement ring in `salvageunion-reference/lib/rules/advancement*.ts`, the ITUN Change Class picker, the class-flag data/schema fill, and the entity guards.

**Findings**

- `apps/itun/src/components/pilot/classPathOptions.ts:216` — `classChangePatch` never persists the *implied* origin. A Hacker with only Augmentation abilities (or none) picking Cyborg gets `needsOrigin: false`, so `answeredOrigin` stays undefined and the patch carries no `originClassRef`. I ran the real functions: the resulting pilot resolves to `originState: 'ambiguous'`, `origin: undefined`, `sealed: []` — the ledger's "Hacking and Electronics seal" promise is unreproducible, and the origin is unrecoverable (the change log is local-only, as this PR's own export test notes). `classChangePatch` isn't even given `currentClassRef`.
- `apps/itun/src/components/sheet/PilotIdentity.tsx:148` — `confirmClass` bails when `pendingClass === pilot.classRef`, so an already-hybrid pilot (e.g. a Cyborg) that the picker *does* ask "where did they come from?" can answer, hit Confirm, and have the answer silently discarded. That's the one case `originClassRef` exists for, and it has no reachable write path.
- `apps/itun/src/lib/rules/pilotAdvancement.ts:95` — `resolvePilotAdvancement` / `heldTreesOf` have zero non-test importers, and `originClassRef` is read nowhere outside its schema declaration. The Add Abilities picker still runs through `offeredAbilityTrees`, which appends `selectedTrees` verbatim and never consults `sealed` — so sealed trees remain fully pickable while the ledger says "no new picks".

**Verified clean:** typecheck, full `bun run test` (0 fail), `knip`, `validate:all`, and `check:schemas` (no drift) all pass. The `isBaseAdvancedClass` semantic change (from "has `advancedTree`" to `hybrid === true`) is behaviour-preserving at its one call site — `ClassAbilityTree` tests `isCoreClass` first, so core classes never reach it. `resolveClassKind`'s ordering (hybrid before `advanceable === false`) correctly survives the new `advanceable: false` on hybrids, and the added test is a genuine tripwire. The `ability-tree-requirements` preload that `offeredAbilityTrees` newly depends on is covered by ITUN's `preload('all')` gate and Ladle's.

PR #715, opened 2026-08-06
feat(itun): a mech's drones are granted, live, and tied to their chassis

alxjrvs:
## What

A chassis pattern's drone had nowhere to live on the mech sheet. The controller ability rendered in the identity band carrying its stat block **as free text** — Big Brother's prose literally reads `"3 SP, 4 EP, 4 Heat Cap…"` beside the mech's own lozenges — while the drone itself was never instantiated at all.

That second half was the real gap: **nothing in the app ever created a `PartnerInstance`.** The only writers were `removePartner` and the v11/v12 migrations, so every partner in existence was a converted legacy record. Building a Little Sestra dropped its Sestra Drone and a Big Brother on DronTek dropped all four, which made `MechSheet`'s Partners region live code that could never render. Equipping a pilot's Survey Drone produced an inert equipment card with no structure, energy or loadout.

## How

`apps/itun/src/lib/rules/partnerGrants.ts` answers "what does this host's configuration grant?" as a list of seeds, and reconciles that list against the instances the host already carries. It settles the lifecycle question ADR-027/028 left open.

**The grant is the lifecycle.** A partner is created when its grant appears and destroyed when it goes — unequip the Survey Drone equipment and the drone goes with it; change a mech's chassis and its drones change with it. There is therefore **no standalone remove control**: one would be unrecoverable, since nothing would ever grant it back. `removePartner` is gone from both sheet action sets.

**The two hosts reconcile differently, and the asymmetry is the rules.**

| | roster | on reconcile |
|---|---|---|
| **Mech** | exact — the pattern says how many and what each carries | re-cuts the roster and each drone's loadout, as the wizard already re-cuts the mech's own |
| **Pilot** | additive — `equipment` says *which* stat blocks, never *how many* (Mecha Packmaster) | a surplus instance is left alone rather than reaped |

Reconciliation never costs live state: a surviving drone keeps its id, structure, energy, heat, conditions, name and cargo.

## Also

- **A drone's systems are the union of two sources** — integrated hardware on the `drones` record (both player drones carry a Hover Locomotion System) plus the pattern's fitted picks. A deliberate divergence from the SRD pattern card, which describes a pattern rather than accounting for slots in play.
- **`partnerCap` gained a second raiser.** It knew Mecha Packmaster but assumed every stat block capped at one, so Big Brother's four would each have read "4 of 1" the moment seeding worked. The Big Brother controls four (False Flag p. 62).
- **Stripped the duplicated stat sentence** from the Big Brother controller's prose in `actions.json`. Every number in it is already structured on the drone record, including the integrated system — it was the same block twice, once as data and once as text in the mech's identity band.
- **ADR-028 gains an amendment** recording the lifecycle rule, and loses its now-false claim that `controls` carries removal.

## Not fixed, and it was not fixed before either

Mecha Packmaster's second Mecha Companion still has no creation path, because a pilot's equipment list is a set and the cap lives on an ability. The additive rule exists so reconciliation cannot destroy one if it ever gains one.

## Verification

`bun run check:all` green — 14/14 steps: lint, format, typecheck ×6, test, validate:all, knip, audit, tokens, styling, schemas.

New coverage (28 tests): `partnerGrants.test.ts` (seeds + reconciliation), `MechSheet-partners.test.tsx` (the region renders a live drone with its own stats and loadout, four named drones at "4 of 4", no remove control), `pilotSheetActions-partnerGrants.test.tsx` (equip/unequip grants and revokes in one write), plus create-payload cases on both wizard form-state suites.

Note: `apps/srd/ssg/parity.ts` will report a diff on the Big Brother action text. That is the intended data change, and parity runs against an archived Astro baseline — it is not wired into CI.

https://claude.ai/code/session_01JZ5hMJuWv9acvpkJebhZKg

---

# Follow-up in this PR: Mecha Packmaster's second companion

The original version of this PR left one gap open, and it turned out the fix was
already sitting in the data.

**`Mecha Packmaster.grants` lists `Mecha Companion` twice** — and it is the only
ability in the whole dataset that grants the same thing more than once. Nothing
read it: `grants` fed the SRD display layer and nothing else. So the second
companion had no way to exist, and `partnerCap` restated the same fact by
string-matching `'mecha-packmaster'`, letting the cap and the number actually
fielded drift apart.

**Equipment gates, abilities count.** `pilot.equipment` is a set, so it could
never express "two Mecha Companions" — but it still decides whether the companion
exists at all. `partnerGrantCount` reads the grants, `pilotPartnerSeeds` emits
that many, and `partnerCap` derives from the same call, so a future ability that
grants two needs no code change.

**MAX, not sum.** The Ranger tree's L1 ability is `Mecha Companion` (one grant)
and Packmaster is Legendary Ranger, so a pilot holding Packmaster normally holds
both. Summing fields three, which Packmaster's own text denies — *"Gain a second
Mecha Companion in addition to your first."* The two entries encode the **total**.

Abilities become a second lifecycle edge: `toggleAbility` reconciles alongside
`toggleEquipment`, so taking Packmaster fields a companion and dropping it
retires one (the newest — reaping takes from the tail).

## The data-loss bug this had to fix first

Making the pilot roster exact means a partner with no seed is reaped — which
turned an unnoticed invariant breach into real data loss. A pilot partner's
`hostRef` must appear in `pilot.equipment`. **v11 was careful about this** and
says so in its own header; **v12 was not** — it converted companion-mech rows
without ever adding the granting slug, which is precisely the case where the
player had *not* equipped the item. Left alone, Eldridge Coast's Custos,
Incitatus, PR-1 and Rek Jet would have disappeared on their pilot's next edit.

**Migration v15** backfills the missing slug rather than deleting the partner:
the partner is the evidence that something once granted it, so the honest repair
is to restore the grant. Append-only, so it is a no-op on a consistent database,
and it never touches mech-hosted partners or invents a partner. One real cost —
granting equipment occupies an inventory slot, so a healed pilot's usage rises by
one and may read as over capacity. Advisory only (ADR-007/021), and it is the
true reading: under-reporting it was the bug.

## Also in the follow-up

- **`syncPartners`'s `exact` flag split in two.** It conflated "the seeds are the
  whole roster" with "re-cut the loadout". Pilots need the first and must *not*
  have the second, since a pilot's partner is kitted out in play and its seed is
  bare. Only `reseedLoadout` survives as an option.
- **Seed claiming is now two passes.** A single pass let an unnamed seed claim the
  instance a named seed needed, renaming a drone the player had customised.
- **Indexed lookups.** `partnerCap` runs per card per render and was reaching the
  ORM via `Model.find((e) => matchesRef(e, ref))` — the shape the package's
  CLAUDE.md calls "a SEARCH wearing a predicate's clothes". Now `resolveRef`.

## Why this is not a stacked PR

I said I'd stack this. I can't: `ci.yml` fires `pull_request` only on
`branches: ['main']`, so a PR based on this branch runs **no CI** and could never
satisfy the required `quality-checks` gate.

Verification: `bun run check:all` green. Partner coverage is now 45 tests across
four files, including the max-vs-sum rule and the v15 backfill classifier.

### DIFF
--- a/apps/itun/src/components/mech/MechWizard.tsx
+++ b/apps/itun/src/components/mech/MechWizard.tsx
@@ -34,6 +34,7 @@ import type { MechWizardFormState } from '../../lib/wizard/mechFormState'
 import {
   EMPTY_MECH_FORM_STATE,
   mechFormToCreateInput,
+  mechFormToPartners,
   mechFormToUpdatePatch,
 } from '../../lib/wizard/mechFormState'
 import {
@@ -42,6 +43,7 @@ import {
   useWizardDraftSync,
   wizardDraftKey,
 } from '../../lib/wizard/wizardDraft'
+import { WIZARD_TXN } from '../../stores/surfaceProvenance'
 import { useWizardFlow } from '../wizard/useWizardFlow'
 import { CraftItemsStep } from './CraftItemsStep'
 import { InstallStep } from './InstallStep'
@@ -218,6 +220,17 @@ export function MechWizard({
     schema: MechSchema,
     toCreateInput: mechFormToCreateInput,
     toUpdatePatch: mechFormToUpdatePatch,
+    // Drones follow the chassis that grants them: change the chassis or the
+    // pattern and the roster changes with it — a drone the new chassis does not
+    // grant is dropped, one it does is added, and a drone that survives keeps
+    // its structure, energy, heat, name and cargo while its loadout is re-cut
+    // from the new pattern (exactly as the mech's own loadout already is).
+    // Runs here rather than in the patch because that live state has to be read
+    // before it can be preserved, and `before` is where it lives.
+    afterUpdate: async (store, id, before) => {
+      const partners = mechFormToPartners(form, before?.partners)
+      await store.update('mech', id, { partners }, WIZARD_TXN)
+    },
     failureMessage: 'Failed to save mech. Please retry.',
     onComplete,
   })
--- a/apps/itun/src/components/pilot/PilotWizard.tsx
+++ b/apps/itun/src/components/pilot/PilotWizard.tsx
@@ -34,6 +34,7 @@ import type { PilotWizardFormState } from '../../lib/wizard/pilotFormState'
 import {
   EMPTY_PILOT_FORM_STATE,
   pilotFormToCreateInput,
+  pilotFormToPartners,
   pilotFormToUpdatePatch,
 } from '../../lib/wizard/pilotFormState'
 import {
@@ -42,6 +43,7 @@ import {
   useWizardDraftSync,
   wizardDraftKey,
 } from '../../lib/wizard/wizardDraft'
+import { WIZARD_TXN } from '../../stores/surfaceProvenance'
 import { pilotInventoryCapacity, pilotInventoryUsed } from '../sheet/pilotInventory'
 import { useWizardFlow } from '../wizard/useWizardFlow'
 import { ReviewStep } from './ReviewStep'
@@ -231,6 +233,15 @@ export function PilotWizard({
     schema: PilotSchema,
     toCreateInput: pilotFormToCreateInput,
     toUpdatePatch: pilotFormToUpdatePatch,
+    // A partner cannot outlive its grant: dropping the Survey Drone equipment
+    // drops the drone, and adding it back grants a fresh one. Runs here rather
+    // than in the patch because a partner holds live-play state (structure,
+    // energy, heat, cargo) that reconciliation has to preserve, and that means
+    // reading the stored partners — which `before` carries.
+    afterUpdate: async (store, id, before) => {
+      const partners = pilotFormToPartners(form, before?.partners)
+      await store.update('pilot', id, { partners }, WIZARD_TXN)
+    },
     failureMessage: 'Failed to save pilot. Please retry.',
     onComplete,
   })
--- a/apps/itun/src/components/sheet/MechSheet.tsx
+++ b/apps/itun/src/components/sheet/MechSheet.tsx
@@ -456,13 +456,6 @@ export function MechSheet({
                 fielded={(mech.partners ?? []).filter((p) => p.hostRef === partner.hostRef).length}
                 readOnly={readOnly}
                 store={store}
-                onRemove={
-                  readOnly
-                    ? undefined
-                    : () => {
-                        actions.removePartner(partner.id)
-                      }
-                }
               />
             ))}
           </div>
--- a/apps/itun/src/components/sheet/PartnerCard.tsx
+++ b/apps/itun/src/components/sheet/PartnerCard.tsx
@@ -76,7 +76,6 @@ type PartnerCardProps = {
   /** How many of this same stat block the host fields, for the "2 of 2" note. */
   fielded?: number
   readOnly?: boolean
-  onRemove?: () => void
   store?: typeof useEntityStore
 }
 
@@ -87,7 +86,6 @@ export function PartnerCard({
   hostAbilityRefs = [],
   fielded = 1,
   readOnly = false,
-  onRemove,
   store = useEntityStore,
 }: PartnerCardProps) {
   const { partner, hostKind, host } = found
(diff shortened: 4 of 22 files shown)


alxjrvs (commented):
Reviewed the `mech-drones-partners` branch (18 files: the new `partnerGrants.ts` reconciler, wizard/sheet wiring, the removal of the standalone partner-remove control, the `partnerCap` extension, and an `actions.json` text trim). Ran the new tests (70 pass) and itun typecheck (clean), and verified two of the three findings with throwaway probes.

Findings:

- **apps/itun/src/lib/rules/partnerGrants.ts:182** — `pilotPartnerSeeds` sets `hostRef` from `pilot.equipment`, which stores reference-record **UUIDs** (`idOf={(item) => item.id}`), not slugs. `partnerCap`'s slug comparisons (`hostRef === 'mecha-companion'`, `a === 'mecha-packmaster'`) therefore never match, so the Mecha Packmaster cap-of-2 is dead and two Mecha Companions render as over-cap "2 of 1". Verified by probe. The mech side escapes this only because `seedFromDrone` uses `nameToSlug`, which is also why the new `BIG_BROTHER_DRONE_REF` check works. Also violates the schema's own "Slug of the … record" doc and the repo's slugs-never-UUIDs rule.
- **apps/itun/src/lib/rules/partnerGrants.ts:149** — clearing a mech's pattern ("Custom build", reachable in EDIT mode) falls back to one bare seed, and `exact` mode then drops the surplus. A Big Brother/DronTek with 4 kitted, damaged drones collapses to 1 on save — verified by probe — irrecoverably, since there is no add-partner control and `partnerCap` itself says the chassis fields 4. The survivor also keeps the stale name "Shield Drone".
- **apps/itun/src/components/mech/Pattern/InstantiateFromPattern.tsx:42** (and `dashboard/dashboardLaunch.ts:58`) — both saved-pattern instantiation paths hand-build the create input and skip partner seeding, so a mech made from a saved Little Sestra/Big Brother pattern has no drones. `createBlank('mech')` on the same chassis does seed one, so the shortcut paths now disagree.

Things I checked and cleared: `syncPartners` additive/exact claim logic (order, surplus retention, `[]` vs `undefined`), the integrated-vs-fitted system dedupe (including DronTek's doubled Chaff Launcher), `store.update` merge semantics with `partners: undefined` (no spurious change-log entries — `jsonEqual` guards it), the removed `removePartner` (no dangling references), and the `actions.json` trim (the removed stat line is redundant — `ChassisAbilitiesContent` renders the linked `drones` record from `ability.drone`). Note the data edit will register as a diff in `apps/srd`'s `ssg/parity.ts`, which is not wired into CI.

alxjrvs (commented):
Reviewed the full `main...HEAD` diff (36 files: the new `partnerGrants` reconciliation module, DB migration v15, wizard/live-sheet wiring, and the `EntitySearcher` single-select picker migration). Typecheck, lint, knip and the whole test suite (4952 tests) are green, so everything below is behavioural, found by reading.

**Findings (4)**

- `apps/itun/src/lib/rules/partnerGrants.ts:265` — Renaming a named mech drone destroys it on the next mech-wizard save. `matchesSeed` demands exact `name` equality for named seeds, yet `PartnerCard` ships an editable Name field; rename one of Big Brother/DronTek's four drones and the next save drops it (id, SP/EP/heat, conditions, cargo, A.I. personality) and mints an empty replacement. This PR also removed the only manual partner control, so there is no recovery.
- `apps/itun/src/lib/wizard/mechFormState.ts:119` — `reseedLoadout: true` is unconditional and `MechWizard.afterUpdate` runs on *every* update, so editing a mech's name or appearance re-cuts its drones' systems/modules from the pattern. A system the player removed on the live sheet comes back — and, since `systemConditions` is keyed by slug and not reset, comes back still flagged `destroyed`. The mech's own loadout survives such an edit (it round-trips through the form state); only its drones' does not.
- `apps/itun/src/lib/rules/partnerGrants.ts:236` — `pilotPartnerSeeds` emits `partnerGrantCount` seeds *per equipment entry*, but `pilot.equipment` is not a set: the create-mode count stepper pushes duplicates. Two Mecha Companion entries plus Mecha Packmaster yields four companions against a cap of two ("4 of 2"). The module header's stated premise that equipment is a set is what the code relies on, and it's false.
- `apps/itun/src/lib/db/migrations/15-partner-equipment-backfill.ts:80` (plausible) — the backfill's membership test is exact-string against a hardcoded slug, while every picker writes reference-entity UUIDs into `pilot.equipment` (`idOf={(item) => item.id}`). A pilot holding Mecha Companion under its UUID plus a v12 partner with slug `hostRef` gets the slug appended anyway: two inventory slots for one item, and a second empty companion minted on the next toggle. Everywhere else refs are compared with `matchesRef`.

Nothing wrong found in the `EntitySearcher` `single` mode work, the `SheetPickerModal` `footer` removal (no callers remain), the deleted master/detail step components (no dangling imports), or the `actions.json` prose trim (the removed Big Brother Drone stat line is fully present in `drones.json`).

PR #716, opened 2026-08-07
feat(about): add a Special Thanks section to both about pages

alxjrvs:
Adds a **Special Thanks** section to the About page, between "About the Dev" and the Ko-fi support block.

It goes into the shared `Colophon`, so it appears on **both** about pages — `salvageunion.io/about` and ITUN's `/about` — the same way the author bio and LLM statement already do.

## Content

New repo-root `SPECIAL_THANKS.md`, following the existing `ABOUT_JRVS.md` / `LLM_STATEMENT.md` contract: one document, no second copy, read by srd at build time via `node:fs` and by ITUN via a Vite `?raw` import. Editing the wording changes both sites together, with no component change.

## Rendered order on `/about`

```
About the Dev
Special Thanks      ← new
Support the Project (Ko-fi)
```

Verified against the real built output (`apps/srd/dist/about/index.html`), not just the source.

## Verification

- `bun run check:fast` — lint, `validate:all`, knip, typecheck across all six workspaces: green
- `bun run test` — 0 failures
- `bun run format` — clean
- `bun ssg/build.ts` — 1,039 pages, 900 endpoints

## Note on `ssg/parity.ts`

Not run: the baseline is gitignored and absent in a fresh worktree. It would report a `<main>` text difference on `about/index.html` either way — this PR deliberately adds visible prose to that page, and the parity baseline is frozen at the last Astro commit. That is expected for any intentional post-migration content edit, not a regression introduced here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_0123FhUCSo53Qx3aiJ3k71t8

PR #721, opened 2026-08-07
docs: update the LLM statement copy

alxjrvs:
Rewrites `LLM_STATEMENT.md` — the single canonical file rendered verbatim in the colophon on both about pages (srd reads it at build time with `node:fs`; ITUN inlines it via a Vite `?raw` import), so both sites change together.

## What changed

- **Names the specific tools** used: AugmentCode, Kilo, Claude, and open weight models.
- **Adds a link to this repo** as the receipt for the review claim — `[GitHub](…)` is the one inline markdown form the colophon renderer interprets.
- **Draws an explicit boundary around game content and runtime**: no LLM authored the rules, none is called while using the site, and none belongs in play.
- **States the judgment boundary on data**: the schema and entity dataset are hand-authored; an LLM only ever applied a specified change there, never originated one.
- Keeps the open-weights paragraph verbatim, and closes on "Software, like games, will eternally be a human endeavor."

## Why this shape

Checked against [Anthropic's AI diligence statement guidance](https://claude.com/resources/tutorials/writing-an-ai-diligence-statement). The revision covers all five of its essential elements — what AI assisted with, which tools, review and involvement, changes the author made, and accountability — where the previous copy left elements 2 and 4 unaddressed.

## Verification

- `bun --filter component-lib test` → **761 pass / 0 fail**, including `MarkdownSection.test.tsx`, which reads the real `LLM_STATEMENT.md` off disk and asserts its heading parses.
- Format contract held: one `#` heading, blank-line-separated paragraphs, no bold/italic/list markup (which would ship as literal punctuation).

https://claude.ai/code/session_018wq3syswSkrVtcY1YBJBVe

### DIFF
--- a/LLM_STATEMENT.md
+++ b/LLM_STATEMENT.md
@@ -15,14 +15,10 @@
 
 # LLM Statement
 
-This was started as a hand-made project, and it predates the LLMs now used to work on it. The site and its schema are still shaped by hand.
+This project was conceived and created before the popularization of LLM coding tools. In its continued development, I have made use of various LLM coding tools to aid in the production of code, features, design, API documentation, deployments, and bug fixes. These have included AugmentCode, Kilo, Claude, and various open weight models. All LLM changes were directed, reviewed, and deployed by me, in addition to any actions taken by the LLMs. The [GitHub](https://github.com/SalvageUnion-io/SU-SRD) shows the full extent of LLM involvement. Bugs and errors are my responsibility, and the blame for them falls on me.
 
-Much of the code since then has been written with LLM assistance. Every change is directed, read, run, and merged by me, and the commit history shows which parts.
-
-The content described in the SRD is all sourced from Leyline Press, and none of it has been authored by me or any of my LLM coding assistants.
-
-My job requires proficiency in these tools, and I best gain proficiency through passion projects like this. It has been my intention to use LLMs as a force multiplier, not as a replacement for my own judgment: to more consistently and efficiently implement my changes, not to decide on the changes themselves. Bugs or issues in this project are solely my responsibility and my fault.
+The data schema and the entity dataset are hand-authored and hand-corrected; where an LLM has touched them, it was to apply a change I specified, not to originate one. All game content is sourced from Leyline Press, and no LLMs have been used in the production of game rules or content on this site. No LLMs are in use or being called while using this site. No LLMs are used in the playing of this game, nor should they be.
 
 I support open-weight models. I am against consolidated ownership of LLM infrastructure, and against the obtrusive, environmentally unsound data centers built to serve it.
 
-Software is, and will always be, a human endeavor.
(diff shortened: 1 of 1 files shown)


alxjrvs (commented):
**No findings.** The reviewable scope is a single commit (`dd0f2788`, "docs: update the LLM statement copy") vs `origin/main` — note that the worktree's local `main` ref is stale, so `git diff main...HEAD` misleadingly shows 34 files; the real diff is `LLM_STATEMENT.md` only, and `git diff HEAD` is empty. No code changed.

What I verified rather than assumed:

- **Format contract holds.** `LLM_STATEMENT.md`'s own header comment restricts the file to one `#` heading, blank-line-separated paragraphs, and `[label](href)` as the only interpreted inline markdown. The new copy adds exactly one link (`[GitHub](https://github.com/SalvageUnion-io/SU-SRD)`), which `parseInline` handles (`https://` passes `isLinkableHref`), and introduces no bold, italics, or `- ` lists that would have shipped as literal punctuation.
- **The repo-root-document guards pass.** `packages/component-lib/src/markdownSection/MarkdownSection.test.tsx` reads the real file and asserts the heading is still `LLM Statement`, that no block text matches `/\*\*|^[-*]\s|^#/`, that every link resolves to an `http(s)`/root-relative target with a non-empty label, and that every parsed block renders. `bun --filter component-lib test src/markdownSection` → 26 pass, 0 fail.
- **Both consumers pick the change up with no code edit.** srd reads it at build time with `readFileSync` (`apps/srd/src/pages/about.page.tsx:41`) and itun via `?raw` (`apps/itun/src/routes/about.tsx:4`); no second copy of the prose exists anywhere (grepped the removed sentences — zero hits outside the file).
- **The new factual claim is true.** "No LLMs are in use or being called while using this site" — grepped `apps/` and `packages/` for `openai|anthropic|@ai-sdk|gemini`: no LLM SDK or API call anywhere. `apps/srd/src/endpoints/llmsTxt.ts` serves docs *to* LLM consumers, which is not the site calling one, and `apps/itun/convex/mediator.ts` (the one suspicious module name) is the Mediator/NPC-tray surface, not an AI mediator.

One piece of context worth passing on, which I deliberately did **not** file as a finding: `/about` is SSR'd inside `ColophonIsland` (`ssr` at `apps/srd/src/pages/about.page.tsx:287`), so this copy edit changes `<main>` visible text on a page that `apps/srd/ssg/parity.ts` compares strictly (`/about` is not in `APPEND_ONLY_PAGES` — only `changelog/index.html` is). That gate is already red from the earlier about-page work in #716/#719, since the baseline is frozen at the last Astro commit and can never contain a Special Thanks section; this diff doesn't newly break it, and `parity` is not wired into CI (only `apps/srd/package.json` scripts). Don't read a parity failure on this branch as caused by this change.

PR #722, opened 2026-08-07
docs: scope the no-LLMs-in-play line to my own playing

alxjrvs:
One-word copy change in `LLM_STATEMENT.md`:

> No LLMs are used in ~~the~~ **my** playing of this game, nor should they be.

The original read as a claim about every table running this game; the revision states what the author can actually vouch for. The prescriptive tail ("nor should they be") is unchanged.

## Verification

- `bun --filter component-lib test` → **761 pass / 0 fail**, including `MarkdownSection.test.tsx`, which reads the real `LLM_STATEMENT.md` off disk.

## Deploy note

⚠️ This commit touches **only** a repo-root prose file, and neither `netlify.toml` `ignore` filter includes the repo root — so merging this will produce **no deploy on either site** (same as #721). Both sites need a cache-cleared build afterward:

```
netlify api createSiteBuild --data '{"site_id":"<id>","clear_cache":true}'
```

A plain re-trigger does not work: the ignore command runs on every build regardless of trigger, and the path diff is empty. The durable fix is adding the root prose files to both ignore filters — not done here, as it is out of scope for a copy change.

https://claude.ai/code/session_018wq3syswSkrVtcY1YBJBVe

### DIFF
--- a/LLM_STATEMENT.md
+++ b/LLM_STATEMENT.md
@@ -17,7 +17,7 @@
 
 This project was conceived and created before the popularization of LLM coding tools. In its continued development, I have made use of various LLM coding tools to aid in the production of code, features, design, API documentation, deployments, and bug fixes. These have included AugmentCode, Kilo, Claude, and various open weight models. All LLM changes were directed, reviewed, and deployed by me, in addition to any actions taken by the LLMs. The [GitHub](https://github.com/SalvageUnion-io/SU-SRD) shows the full extent of LLM involvement. Bugs and errors are my responsibility, and the blame for them falls on me.
 
-The data schema and the entity dataset are hand-authored and hand-corrected; where an LLM has touched them, it was to apply a change I specified, not to originate one. All game content is sourced from Leyline Press, and no LLMs have been used in the production of game rules or content on this site. No LLMs are in use or being called while using this site. No LLMs are used in the playing of this game, nor should they be.
+The data schema and the entity dataset are hand-authored and hand-corrected; where an LLM has touched them, it was to apply a change I specified, not to originate one. All game content is sourced from Leyline Press, and no LLMs have been used in the production of game rules or content on this site. No LLMs are in use or being called while using this site. No LLMs are used in my playing of this game, nor should they be.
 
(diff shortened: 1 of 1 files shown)


alxjrvs (commented):
No findings.

**Scope:** the branch is one commit (`99322828`) with a single-line prose edit to `/Users/jarvis/Code/SU-SRD/.claude/worktrees/llm-statement-copy/LLM_STATEMENT.md:20` — "No LLMs are used in **the** playing of this game" → "in **my** playing of this game". Working tree is clean; no code changes.

**What I verified:**
- **Single source of truth holds.** The sentence appears nowhere else in the repo. The file is consumed by exactly two renderers — `apps/srd/src/pages/about.page.tsx:41` (build-time `readFileSync`) and `apps/itun/src/routes/about.tsx:4` (Vite `?raw`) — both of which pass the raw string through `MarkdownSection`, so both sites pick the change up with no second copy to sync.
- **Format contract satisfied.** The file's own header comment restricts the doc to one `#` heading, blank-line-separated paragraphs, and `[label](href)` links only. The edit adds no bold/italics/list markers and no new link, and stays on one physical line, so `parseMarkdownSection` produces the same block structure as before.
- **No test asserts this prose.** `packages/component-lib/src/markdownSection/MarkdownSection.test.tsx` reads the real file but only asserts structure (heading text, no stray bullets/bold, links resolve to real targets) — nothing keyed to this sentence.
- **The srd parity gate is not newly at risk.** `apps/srd/ssg/parity.ts` compares `<main>` text against a frozen pre-Astro-removal baseline, and `/about` is not in `APPEND_ONLY_PAGES` — but the same file was already reworded post-baseline in `030402db` (#721) and in the colophon PRs, so any `/about` divergence predates this commit. Parity is also local-only (gitignored 56 MB baseline, not wired into CI), so this PR introduces no new gate failure.

**One editorial note, not filed as a defect:** the trailing clause now reads "No LLMs are used in **my** playing of this game, nor should they be." Narrowing the first half to the author's own play while leaving "nor should they be" as an unscoped prescription is a slight mismatch with the commit's stated intent (scoping the claim to the author). It reads fine and is plausibly deliberate, so it's a wording call for the author rather than a bug.

PR #720, opened 2026-08-07
chore(srd): retire the Astro-migration parity gate

alxjrvs:
Deletes `ssg/parity.ts`, `ssg/make-parity-baseline.ts` and their tests (**2,109 lines**), plus every config entry and doc claim that pointed at them.

This is the shelf life [ADR-031](https://github.com/SalvageUnion-io/SU-SRD/blob/main/docs/adrs/ADR-031-srd-vite-ssg.md) called in its own Consequences section arriving — it is executing the ADR's plan, not overturning it:

> "once that baseline is gone … `ssg/parity.ts` is a historical record of a clean migration, not a live gate"

## That point had already passed, silently

- The baseline was a ~56 MB Astro-era `dist`, **gitignored and present in no checkout on this machine** — so the gate exited `2` with `baseline directory does not exist` everywhere it was invoked. I hit exactly this trying to use it on an unrelated PR.
- It was **never wired into CI** (no `.github/**` reference).
- Regenerating it meant checking out the last Astro commit and installing ~2,200 Astro-era packages.

So what actually existed was a documented instruction, repeated across nine files, that nobody could follow — including one telling agents to *"trust it over any agent's opinion about whether output changed."* A gate that cannot run is worse than no gate, because the docs still sell it as coverage.

## What is honestly lost

Nothing that was running. But parity was the only thing that **ever** compared whole-page output, so rather than quietly dropping the claims I made each touched comment state what checks its constraint — and say "nothing does" where that is the truth:

| Constraint | Status now |
| --- | --- |
| page count held at 1,039 | **nothing** — `build.ts` only refuses at zero |
| `llms.txt` byte-stable | **nothing** — reflowing the literal ships silently |
| no og/canonical on the two `bare` pages | 4 assertions in `render.test.tsx` |
| ~880 entity pages | **no output comparison at all** |
| URL → dist file mapping | `outputPath.test.ts` (now the sole check) |

I have **not** added a replacement gate — that is a separate decision, not a silent rider on a deletion. A page-count floor in `build.ts` would be the cheapest partial recovery if you want one.

## ADR-031 is amended, not rewritten

Decision 6 is marked expired, the "shelf life" consequence records that it arrived, and the line-count claim now separates the retired 897-line harness from the ~1,491-line generator this repo still owns.

## Not touched

The **rules-parity audit** in `salvageunion-reference` (`validate:parity`, ADR-029). Same word, unrelated subsystem — left completely alone.

## Verified

[...]

PR #719, opened 2026-08-07
feat(about): rework the Special Thanks copy, and render bullet lists

alxjrvs:
Updates the Special Thanks copy on both about pages, and teaches the shared markdown renderer to render bullet lists.

## The copy

```
Thank you to Panny and all the folks at [Leyline Press] for their continued support.

A special thanks to the #salvage-union-io crew:

- Tommy_Dude
- lunarsignals
- the puppetter
- Grandowl
- kindalas
- decivre
- Saving Throw

[Join the Discord] to join the conversation!
```

The Discord invite (`discord.gg/kMwD2bWgtC`) is the one **Leyline Press publishes on its own Salvage Union Discord page** — there was no invite anywhere in this repo, and it is not a guessed code.

## Why a renderer change was needed

`MarkdownSection` had no list support. Its `blocksOf` joins consecutive lines with a space, so the crew block would have rendered as a single paragraph reading `- Tommy_Dude - lunarsignals - the puppetter …`, dashes and all. The copy as requested was not expressible.

So `parseMarkdownSection` now returns **ordered blocks** (`{kind:'paragraph'} | {kind:'list'}`) instead of a flat `paragraphs: string[]`, and `MarkdownSection` renders a list block as a real `<ul>`.

**All-or-nothing by design:** a block becomes a list only when *every* line in it is a `- ` item. A block mixing prose and dashes stays a paragraph, so every document written before lists existed parses to exactly the blocks it did before, and a wrapped line beginning with a hyphen can't silently swallow the paragraph around it. Prefix checks only — this file is deliberately regex-free after CodeQL's `js/polynomial-redos` alerts, and that property is preserved.

## Verification

Checked against the **real built output** (`apps/srd/dist/about/index.html`), not just source:

```html
<p>Thank you to Panny and all the folks at <a href="https://leyline.press" …>Leyline Press</a> for their continued support.</p>
<p>A special thanks to the #salvage-union-io crew:</p>
<ul class="flex list-disc flex-col gap-1 pl-5">
  <li>Tommy_Dude</li> … <li>Saving Throw</li>
</ul>
<p><a href="https://discord.gg/kMwD2bWgtC" …>Join the Discord</a> to join the conversation!</p>
```

Still sits between "About the Dev" and "Support the Project", on both srd `/about` and ITUN `/about`.

- `bun run check:fast` — lint, `validate:all`, knip, typecheck across all six workspaces: green
- `bun run test` — 0 failures; component-lib 751 → **761** (10 new tests)
[...]

PR #723, opened 2026-08-07
feat(srd): replace the retired parity gate with a self-hosted output snapshot

alxjrvs:
#720 deleted `ssg/parity.ts` and documented, honestly, that nothing then compared the built site against a reference. This restores that coverage with a gate designed so it cannot rot the way parity did.

`ssg/snapshot.ts` diffs the built `dist` against `ssg/output-snapshot.json`, a committed 684 KB digest of our own output. **It runs in CI** — in the existing `build-srd` job, which is already path-filtered on everything that can change srd's output (`apps/srd/**`, `packages/component-lib/**`, `packages/salvageunion-reference/**`).

## Why a snapshot rather than a new oracle

Parity's baseline was an archived **Astro** dist: foreign, ~56 MB, gitignored, regenerable only by installing ~2,200 Astro-era packages. It was absent from every checkout, exited `2` wherever invoked, and was never in CI — while nine files told people to run it.

| | parity (retired) | snapshot (this PR) |
| --- | --- | --- |
| baseline | archived Astro `dist`, ~56 MB | our own output, 684 KB digest |
| in a fresh checkout | absent | committed |
| regenerate | install ~2,200 Astro-era packages | `bun --filter srd snapshot:update`, ~3s |
| in CI | **never** | yes |
| catches wrong-from-the-start output | yes | no |

The trade is stated in the code, the ADR and the docs: parity compared against an **oracle** and could catch output that was wrong from the first build. This compares against what was last blessed, so a wrong output committed as the snapshot stays wrong. **A weaker check that runs beats a stronger one that cannot.**

## Recovered, not rewritten

Parity's HTML analysis survives as `ssg/htmlDigest.ts`. Those scanners carry comments documenting real bugs found while diffing 1,039 pages — malformed closing tags (`</script\t\n foo>`), comment-stripping not being a fixed point — and a fresh implementation would have reintroduced them one at a time. **Verified identical to the deleted code** on head metadata, JSON-LD, `<main>` inner HTML and `<main>` text across all 1,039 real pages.

`isInsertionOf` — parity's best idea — carries over for `/changelog`: its full text is stored and growth must be one contiguous insertion, so deleting, reordering or rewording an old entry still FAILS. Without it every release PR would fail the gate and deadlock the release (cf. #700).

## Covered / not covered

**Covers:** the emitted file set both directions (this is what holds the page count at 1,039), per-page title/description/canonical/robots as plaintext, an og/twitter digest, the JSON-LD `@type` list, a digest of `<main>` text, all 899 JSON endpoints, `llms.txt`, sitemaps, `_headers`.

**Does not cover**, deliberately and documented rather than implied:
- bundle bytes (content-addressed; would churn on every dependency bump)
- **attributes inside `<main>`** — verified concretely: the artwork regression from #717 (`style="opacity:0"`) does **not** fail this gate. Hashing full markup would catch it and fire on every Tailwind class change. Visual regressions are not this gate's job.

## It bites — including once for real

22 tests inject a defect and assert it is reported. 8 more were injected into the **real 1,039-page build** and all 8 caught, with readable output (`title: "About - …" -> "WRONG"`).

Then it caught one in the wild, unprompted: rebasing onto a main containing #719 produced

```
about/index.html
    [page] main text changed (2461 -> 2697 chars)
SNAPSHOT FAILED — 1 difference(s) across 1 path(s).
```

Re-blessing that produced **exactly one changed line** in the 4,008-line snapshot — the property the custom serializer exists for, demonstrated rather than asserted. (Biome is excluded from the file: its JSON formatter re-wrapped it to 13,734 lines and destroyed that property.)

## Verified

- Deterministic — a clean rebuild reproduces the snapshot exactly.
- `check:fast` green, including the repo's own doc-drift check, which caught that `gate`/`snapshot:update` are workspace scripts so root-scope docs must say `bun --filter srd gate`.
- Full suite green, 0 failures; srd 1,086 → 1,108 = exactly the 22 new tests.
- ADR-031 amended: decision 6 is marked replaced, with the oracle-vs-snapshot trade recorded.

https://claude.ai/code/session_01QunKeBj1xCvWTL8kVK3mvs


alxjrvs (commented):
Reviewed the diff (new `ssg/snapshot.ts` + `ssg/htmlDigest.ts` output gate, its test suite, the committed 683 KB `output-snapshot.json`, the CI wiring, and the doc/comment rewrites replacing the retired parity gate).

What I verified working: the 22 new tests pass, `bun ssg/snapshot.ts` reports `SNAPSHOT OK — 1039 pages` against the committed snapshot, typecheck passes, knip is clean, `validate:all` (incl. doc-drift) passes, the biome ignore keeps the one-line-per-page format intact, and the doc counts (1,039 pages / 899 JSON endpoints / ~680 KB) are accurate. I also checked the two release-deadlock risks specifically: the changelog page's title/description are static constants and no changelog or version text leaks into any digested JSON, so the append-only exemption really does cover the release path.

Findings:

- `apps/srd/ssg/snapshot.ts:430` — The two new files add ~496 measured lines to `apps/srd` with the CLI half untested, dropping the workspace from ~97.0% to 92.68% line coverage. `bun tools/coverage-report.ts` reports `| apps/srd | 92.68% | 95.00% | FAIL regressed |`, so the `coverage` job and `quality-checks` fail as-is.
- `.github/workflows/ci.yml:85` — The `web` path filter omits repo-root `ABOUT_JRVS.md` / `LLM_STATEMENT.md` / `SPECIAL_THANKS.md`, which `about.page.tsx` reads at build time. A PR touching only those (main has two such commits from yesterday, #721 and #722) skips `build-srd` entirely, merges with a stale snapshot, and makes the *next* srd PR fail the gate on `/about` diffs it didn't cause.
- `apps/srd/ssg/snapshot.ts:412` — `parseArgs` returns `'help'` on an unknown flag and the CLI exits 0; `bun ssg/snapshot.ts --updat` prints usage and exits 0 without comparing anything, so a future CI flag typo turns the gate into a green no-op.
- `apps/srd/ssg/htmlDigest.ts:76` — `decodeEntities` checks `Number.isFinite` but not the 0x10FFFF ceiling, so `&#1114112;` throws an uncaught `RangeError` that aborts the whole run (confirmed by direct call), breaking both `gate` and `snapshot:update`.

PR #729, opened 2026-08-10
fix(itun): make a stale Convex backend impossible to ship silently

alxjrvs:
## What happened

Production ran a **four-day-stale Convex backend** from 2026-08-06 to 2026-08-10 — surfaced by a player report, not by us:

> "I'm getting this unaccepted change error when modifying a pilot associated with a game"

The published client called `entities:upsertSoftLink`; prod had no such function, so every soft-link write (pilot ↔ mech ↔ crawler assignment) failed outright. `byAppId` was still the pre-#705 version that throws on a duplicate `appId`, so mirrored pilot and mech writes died too — with the redacted `Server Error` string, while every surface kept rendering them as saved. **39 failures in one evening.**

Nothing was red. Netlify was green, the site served the current commit, and the only trace was a Convex deployment log stream nobody tails. The backend has since been pushed by hand; this PR is about why nothing noticed.

Three seams were open. This closes all three.

## 1. The deploy could silently not happen

`apps/itun/netlify.toml` gated the backend push on `if [ -n "$CONVEX_DEPLOY_KEY" ]`, which made a **misconfiguration indistinguishable from an intended Solo build**: an absent key silently took the else branch, the build went green, and the site shipped a current client against a backend nobody pushed.

In the `production` context an absent key is now fatal, loudly, before anything is built. Deploy previews and branch deploys keep the Solo fallback they legitimately need.

## 2. Nothing compared what was deployed to what was committed

New `tools/check-convex-parity.ts`, following the two-half shape `tools/check-observability.ts` already established:

- **static** (`validate:all`, so every PR) — asserts the netlify.toml guard is still there. Hermetic, no credentials. A guard nothing checks is a guard that gets deleted in a cleanup six months from now.
- **`--live`** (nightly) — asks the deployment for its function list and fails on anything this repo defines that is not on it. That direction is the one that breaks players: a backend *behind* the client means the client calls things that aren't there.

Verified both ways: it fails when the guard is removed, and run against the spec I captured *before* the manual deploy it names exactly the three functions that were missing (`entities:upsertSoftLink`, `entities:removeSoftLink`, `maintenance:dedupeAppIds`).

The nightly job is gated on a `CONVEX_PARITY_PROBE` repo variable for the same reason `observability-probe` is — an alarm that rings for an unprovisioned credential is a broken window. **To turn it on:** add a Convex *production* deploy key as the `CONVEX_DEPLOY_KEY` repo secret, then set `CONVEX_PARITY_PROBE = on`.

## 3. The Sentry events were unusable

Sentry did capture these — 20 events — but a redacted defect arrives as `[CONVEX M(fn)] [Request ID: …] Server Error`, so the request id made every message unique. One condition became two issues, neither titled with anything searchable, and the mutation name — the only fact a redacted error still carries — appeared in no title.

Mirror failures now fingerprint on function + refusal/defect and tag the same, via a new `convexFunctionName()` that is **explicitly a diagnostic label, not a refusal classifier** (that stays `instanceof ConvexError`, per the rule in `apps/itun/CLAUDE.md`). `captureException` grew an options arg for fingerprint/tags; the two existing call sites are unchanged.

## Also: a false claim in two docs

`apps/itun/CLAUDE.md` and `docs/architecture/accounts-and-games.md` both recorded Convex Exception Reporting as "on for production since 2026-08-05". The `itun-convex` Sentry project has received **zero events in 30 days** — across a window containing those 39 failures. A reporting integration that reports nothing looks exactly like a healthy one with no errors, and this one was believed for five days. Both now say what is measurably true and point at `bunx convex logs`. The Pro-plan gate the doc already mentions is the first suspect.

## Verification

`bun run check:fast` green (lint, `validate:all` incl. the new check, knip, typecheck all packages); `bun --filter itun test` 1830 pass / 0 fail, including new coverage for `convexFunctionName` and the fingerprint/tags plumbing.

## Not in this PR

`maintenance:dedupeAppIds` still needs a run against prod to clear the duplicate rows — it's cleanup rather than an outage now that `byAppId` tolerates duplicates.


alxjrvs (commented):
Read the diff and checked it against the repo (Convex CLI internals in `node_modules`, `apps/itun/convex/*`, Sentry v10 types, knip config, the nightly workflow).

Nothing in the diff addresses instructions to a reviewing agent; it is all ordinary code and prose.

**Findings**

- `apps/itun/src/lib/observability.ts:104` — **BLOCKING.** `const hint: Parameters<typeof sentryModule.captureException>[1] = {}` resolves to Sentry v10's `ExclusiveEventHintOrCaptureContext | undefined`, which is the union `(CaptureContext & Partial<Record<keyof EventHint, never>>) | (EventHint & Partial<Record<keyof ScopeContext, never>>)`; `extra`, `tags` and `fingerprint` are all keys of `ScopeContext`, so in the second constituent they are typed `never`, and a property *write* through a union takes the intersection of the member types — `hint.extra = context` / `hint.tags` / `hint.fingerprint` are assignments to `undefined`/`never` and will not compile. Independently, `typeof sentryModule.captureException` is a type query on the *declared* type (`typeof import('@sentry/browser') | null`) — type queries do not honour the `if (!sentryModule) return` narrowing — which is a second error on the same line. The existing call sites passed the object literal inline (`context ? { extra: context } : undefined`) precisely because contextual typing picks the right union member; building it as a pre-typed variable loses that. Verify with `bun run typecheck:itun`; the fix is to build a plain local (`const hint: { extra?: …; tags?: …; fingerprint?: string[] } = {}`) and pass it at the call.

- `apps/itun/netlify.toml:50` — the new guard breaks the `&&` chain that used to gate the build on a successful install. The command is now `bun install … && if <guard>; fi; if <build>; fi`: `;` ends the AND-list, so when `bun install --frozen-lockfile` fails the guard is *skipped* and the build branch runs anyway. The stated property — "in the production context an absent key is fatal, loudly, before anything is built" — silently does not hold on an install failure. Chain the guard's `fi` to the build with `&&` (or put the guard first, before `bun install`).

- `tools/check-convex-parity.ts:157` — `functionsDefinedInRepo()` uses a non-recursive `readdirSync` and skips every dirent that is not a top-level `.ts` file, so a Convex function added in a nested module (`convex/games/list.ts` → `games/list.js:foo`, which Convex supports) is invisible to the parity set: the nightly check passes green while that function is undeployed, which is the exact failure mode the tool exists to catch. Currently latent — `convex/model/*` holds only helpers — but the check silently narrows as soon as the backend grows a subdirectory. Recurse, or fail loudly on an unexpected subdirectory.

- `.github/workflows/e2e-nightly.yml:403` — `convex-parity` is compared with `== 'failure'` in `notify-failure` and `!= 'failure'` in `notify-recovery`, so a job that exhausts its `timeout-minutes: 5` (reported as `cancelled`, not `failure`) opens no issue *and* satisfies the recovery condition, closing an existing tracking issue. This is the same hazard the comment 90 lines above documents for the e2e jobs and calls "the suite broke silently for days"; it is inherited from `observability-probe`, but this job does `checkout` + `bun install` + a network round-trip to Convex inside those 5 minutes, so a timeout is more likely here. Treating `cancelled` as a failure for both probe jobs (while still exempting `skipped`) closes it.

alxjrvs (commented):
Reviewed the diff against the surrounding code (`apps/itun/netlify.toml`, `apps/itun/convex/**`, `knip.json`, the Convex 1.43 CLI bundle to confirm `function-spec`'s output shape, `--deployment` selector semantics, and that its JSON goes to stdout uncontaminated). No prompt-injection text addressed to a reviewer is present in the diff.

The workflow gating logic is correct: a skipped `convex-parity` (`CONVEX_PARITY_PROBE` unset) yields `result == 'skipped'`, which neither trips `notify-failure`'s `== 'failure' || == 'cancelled'` clauses nor blocks `notify-recovery`'s `!=` clauses. The netlify.toml one-liner is POSIX-valid (`if …; fi && …` is an AND-list term, so the guard cannot be orphaned). `CaptureOptions` is used in-file so `ignoreExportsUsedInFile` keeps knip quiet, and `tools/**/*.ts` is already a root knip entry.

Findings:

- `tools/check-convex-parity.ts:212` — the live check prints `spec.url` but never asserts it is the production deployment, and when `CONVEX_DEPLOY_KEY` is set it deliberately passes no deployment selector. A dev or preview deploy key in the repo secret makes the nightly compare `main` against a deployment `convex dev` keeps current, so it goes green every night while production stays four days stale — the exact silent-misconfiguration shape this PR exists to eliminate. `DEFAULT_DEPLOYMENT` is right there to check the resolved URL/deployment against.
- `tools/check-convex-parity.ts:231` — parity is a set comparison of *names* only. The documented incident had two halves, and only one is caught: `byAppId` running the pre-#705 implementation that throws on a duplicate `appId` has an unchanged identifier, as does any schema drift, so a backend that is stale but name-compatible passes with "all N repo-defined functions are deployed". The header comment and the workflow comment both describe this as catching "the backend is behind the client", which overstates it.
- `apps/itun/src/stores/entityBackend.ts:211` — `source` and `op` are tags but not part of the fingerprint, and `convexFunctionName` returns `'unknown'` for anything without the `[CONVEX …]` prefix. So every non-Convex-prefixed failure across all three mirror paths — a WebSocket/network outage in `mirrorWrite`, a transient failure in `mirrorCrawlerWrite`, and a genuine client-side throw in `mirrorSoftLinkWrite` — collapses into the single issue `itun-mirror-write-failed / unknown / defect`. A new client bug arrives buried in the same issue as last week's connectivity blip, which is a grouping regression against Sentry's default stack-based fingerprint for exactly the cases the prefix does not cover.
- `apps/itun/netlify.toml:58` — the guard makes an absent `CONVEX_DEPLOY_KEY` fatal in the `production` context, so merging before that variable is provisioned on Netlify's production scope hard-fails *every* production ITUN deploy, including ones that touch nothing Convex-related. That is the intended design, but it is a merge-ordering precondition on out-of-repo state that nothing in CI can verify; confirm the variable is set (scope: builds, context: production) before landing.
- `tools/check-convex-parity.ts:201` — `execFileSync` throws on a non-zero exit and `JSON.parse` throws on any non-JSON stdout, both outside the `failures` collector, so an expired deploy key or a Convex outage surfaces as a raw Node stack trace rather than the diagnostic the rest of the tool is careful to produce. There is also no `timeout` on the child; only the job's `timeout-minutes: 5` bounds it, and a local run has no bound at all.

PR #735, opened 2026-08-11
test(itun): land the Convex backend suite that was written and never merged

alxjrvs:
## The gap

`apps/itun/convex/` — the server of record for every pilot, mech and crawler, 18 modules with `entities.ts` alone at 962 lines — had **zero tests in `main`**. Not thin coverage. None:

```
$ git ls-tree -r --name-only origin/main -- apps/itun/convex | grep -i test
$
```

It was not missing because nobody wrote it. It was written, on the `itun-accounts-*` stack, and those PRs were closed when the work landed via #647/#656 — taking ~1,700 lines of tests with them. This restores **109 of them**, reconciled against ten weeks of drift.

## Why this is worth doing, concretely

`appId.test.ts` contains a test named:

> `'a second write updates rather than duplicating'`

That is precisely the invariant that broke in production on 2026-08-09. Duplicate `appId` rows made `byAppId` throw; a player's pilot edits stopped reaching their game while every surface rendered them as saved; 39 mirrored writes failed in one evening and it surfaced as a Discord message rather than a red build. **The test that would have caught it existed the whole time, on a branch.**

Two more of the same shape. `crew.ts` carries a comment explaining that it once read `currentHp` while the Zod schema defines `currentHP`, so every pilot's HP came back null and the crew strip rendered a row of em-dashes — indistinguishable from "nobody has taken damage yet" — and it closes with "nothing but this comment and **the test below** will catch it next time." That test is in this PR. `proposals` had the identical `currentSp` slip.

## Reconciliation

Each of these is a real change since the branch, not a test that was wrong:

- **`patchCrawler` was deleted.** The communal-merge and member-gate tests now go through `patchCrawlerByAppId`.
- **`assertMayAddToContainer` now requires a crawler** before a player may add builds, so `seedGame` raises one. Without it every player-side write fails with "This game has no Union Crawler yet", which reads like a permissions bug and is not.
- **`proposals.apply` re-parses the merged body**, so its mech fixture had to become a complete `MechSchema` record rather than a two-field stub.
- **Two refusal messages were reworded.** Both assertions now match the load-bearing phrase rather than the whole sentence, with a comment saying why — this copy is player-facing and will drift again.
- **`presence` was deliberately removed** from `mediator.ts` (no client ever called `heartbeat`, so the table had no writer and `botClient.channel` rendered a permanently-false "0 at the table"). Those tests are dropped with the feature rather than left skipped, with a note pointing at the module header that asks any rebuild to land the writer in the same change.

## A tooling fix that may explain the gap

`tools/check-convex-codegen.ts` **failed on the first test file placed under `convex/`**. It skipped only `_generated`, while Convex ignores *any* path segment starting with `_` — so `convex/__tests__/` tripped it, and it demanded that codegen register modules Convex refuses to see. There is no way to satisfy that. Anyone who hit it would reasonably conclude tests do not belong in `convex/` at all.

`convex-test@0.0.55` was already a dependency in `main`, unused — the harness needed nothing new.

## Verification

- `bun --filter itun test`: **1830 → 1939 pass, 0 fail.** Picked up by the workspace runner automatically; no bunfig change needed.
- `bun run test` (canonical, all workspaces): green.
- `bun run check:fast`: green — lint, `validate:all` (including the now-passing `validate:convex-codegen`), knip, typecheck across all packages.

### DIFF
--- a/apps/itun/convex/__tests__/account.test.ts
+++ b/apps/itun/convex/__tests__/account.test.ts
@@ -1,0 +1,215 @@
+import { describe, expect, test } from 'bun:test'
+import { api } from '../_generated/api'
+import type { Id } from '../_generated/dataModel'
+import { testConvex } from './harness'
+
+/**
+ * Account management, and in particular the promise D27 makes:
+ * **a campaign never dies because one person quit.**
+ *
+ * Deletion is the operation with no undo, so the cases below are less about
+ * "does it delete" and more about "does it delete *only* the right things" —
+ * the crew's Game, the communal crawler, and other people's characters all
+ * have to survive somebody exercising their right to be forgotten.
+ */
+
+type Ctx = ReturnType<typeof testConvex>
+
+async function makeUser(t: Ctx, name: string) {
+  const userId = await t.run(
+    async (ctx) => await ctx.db.insert('users', { name, displayName: name })
+  )
+  return { userId, as: t.withIdentity({ subject: userId }) }
+}
+
+async function seedPilot(t: Ctx, gameId: Id<'games'> | null, ownerId: Id<'users'> | null) {
+  return await t.run(
+    async (ctx) =>
+      await ctx.db.insert('pilots', { gameId, ownerId, body: { callsign: 'X' }, updatedAt: 1 })
+  )
+}
+
+describe('profile', () => {
+  test('displayName falls back to the Discord name when unset', async () => {
+    const t = testConvex()
+    const u = await makeUser(t, 'Roach-Boy')
+    await u.as.mutation(api.account.updateProfile, { displayName: null })
+
+    const me = await u.as.query(api.account.me, {})
+    // Cleared means "fall back", not "blank" — a nameless owner chip is worse
+    // than a stale one.
+    expect(me?.displayName).toBe('Roach-Boy')
+  })
+
+  test('a set displayName overrides the Discord name', async () => {
+    const t = testConvex()
+    const u = await makeUser(t, 'Roach-Boy')
+    await u.as.mutation(api.account.updateProfile, { displayName: '  Beefcake  ' })
+
+    const me = await u.as.query(api.account.me, {})
+    expect(me?.displayName).toBe('Beefcake')
+  })
+
+  test('an anonymous caller has no profile to read', async () => {
+    const t = testConvex()
+    await expect(t.query(api.account.me, {})).rejects.toThrow(/not signed in/i)
+  })
+})
+
+describe('export is scoped to what you own', () => {
+  test('excludes a crewmate pilot you can merely see', async () => {
+    const t = testConvex()
+    const owner = await makeUser(t, 'Owner')
+    const other = await makeUser(t, 'Other')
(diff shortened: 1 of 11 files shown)


alxjrvs (commented):
I read the diff, then checked it against the repo, the Convex 1.43 CLI bundler, and the existing test suite.

## Findings

- **BLOCKING** — `apps/itun/convex/__tests__/harness.ts:1` — Putting the harness back under `convex/` makes it a **deployable Convex module**: convex 1.43's entry-point walker (`node_modules/.bun/convex@1.43.0*/node_modules/convex/dist/cli.bundle.cjs:65342-65405`) skips only `_generated/`, dotfiles, `#` tempfiles, `schema.ts`, basenames with more than one dot, and paths containing a space — it does **not** skip `_`-prefixed directories (only `_deps/`, which is a hard error). The sibling `*.test.ts` files are excluded by the two-dot rule, but `harness.ts` is not, so `convex dev`/`convex deploy` bundles it — dragging the `convex-test` devDependency into the production backend — and `convex codegen` emits `import type * as __tests___harness from "../__tests__/harness.js"` into `_generated/api.d.ts`. `apps/itun/test/convex/harness.ts:19-39` records this exact failure (harness shipped to the backend, plus a TS2615 self-referencing `api` type that cascaded implicit `any` through `src/`) and says in as many words: "This harness lives OUTSIDE `convex/` on purpose. Do not move it back."

- **BLOCKING** — `tools/check-convex-codegen.ts:59` — The `entry.startsWith('_')` skip encodes a rule Convex does not have, and it breaks the gate in both directions: after a legitimate regeneration (`bunx convex dev`, or the `convex deploy` in `apps/itun/netlify.toml`) `api.d.ts` will register `__tests__/harness`, which the walk no longer sees on disk, so the check fails with "api.d.ts registers convex/__tests__/harness.ts, which no longer exists" — a drift gate failing on correct output; and because the skip applies to files as well as directories, a future `convex/_helper.ts` — which Convex *would* register — is now invisible to the check, which is the add/remove class it exists to catch.

- `apps/itun/convex/__tests__/account.test.ts:1` — All nine files here already exist at `apps/itun/test/convex/` (which is a superset: 16 test files plus the harness, including `bot`, `botHttp`, `duplicateAppId`, `invites`, `maintenance`, `softLinkMirror`, `tableSetup`), with matching describe blocks down to the two `requireMediator` blocks in `permissions.test.ts`; merging adds no coverage, doubles the suite's runtime, and creates a second module map that must be kept in step with `convex/` — the very thing the harness comment warns fails open.

- `knip.json:52` — The checked-in comment "`test/**` covers the Convex test suite, which lives OUTSIDE `convex/` so that codegen does not register the harness as a deployable module" is now false, and nothing in this diff updates it, so the file that records *why* the suite sits outside `convex/` will read as stale advice to the next person who moves it.

The safe shape is to drop `apps/itun/convex/__tests__/` entirely (the suite is already landed at `apps/itun/test/convex/`) and revert the `check-convex-codegen.ts` walk to skipping `_generated` only. If a `convex/`-local test location is genuinely wanted, the harness still cannot live there under a single-dot filename.

PR #741, opened 2026-08-11
docs: record bun why, markdown profiling, and the --parallel/--isolate result

alxjrvs:
Three additions to CLAUDE.md, all of them things a session currently gets wrong
by guessing.

`bun why` in the audit-gate section. That section and the long comment in
ci.yml carry hand-written dependency archaeology ("it reached the graph via
workspace:srd -> astro"), which is exactly the kind of prose that goes stale
while reading as authoritative. `bun why image-size` prints the real chain from
the lockfile in one command, so a suppression's justification can be CHECKED
rather than believed.

Markdown profiling in the Debugging section. `--cpu-prof-md`, `--heap-prof-md`
and `--metafile-md` emit grep-friendly markdown instead of a binary
`.cpuprofile` that needs a flamegraph UI — which matters because most of the
work here is done by an agent reading a terminal. Verified: the CPU profile
renders a "Hot Functions (Self Time)" table, and the metafile report includes a
"Raw Data for Searching" section. Explicitly noted as diagnostics that must not
be wired into check:all or CI.

The measured cost of --parallel and --isolate, recorded next to the test
commands so nobody "optimizes" into them:

  bun test              34.4s   4991 tests / 365 files
  bun test --parallel   65.1s   + 2 spurious 5s-timeout failures
  bun test --isolate   234.1s   6.8x slower

Both flags look like the obvious way to speed a 4991-test suite up and are large
regressions here, because the preloads (happy-dom, testing-library, a full ORM
schema load, fake-indexeddb) get re-paid per isolated file.

Also corrects a claim that had gone stale: the bare-root `bun test` line
asserted "green (4712 pass)". Measured today it is 4990 pass / 1 fail —
component-lib's CardImage.ssr.test.tsx, which passes in its own workspace. It
fails identically under --isolate, so it is a preload-set difference (the root
bunfig adds fake-indexeddb and the bot env shim), NOT cross-file leakage. The
failure itself is left alone; this only stops the doc from asserting green.

---

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,6 +31,13 @@ coverage-output.txt
 # (the committed snapshot lives at packages/salvageunion-reference/etc/)
 packages/salvageunion-reference/.api-dts/
 
+# Profiling artifacts (bun --cpu-prof / --heap-prof, bun build --metafile-md).
+# These default to the cwd, so without this they land beside the source they
+# profiled and get swept into the next `git add -A`.
+/.profiles/
+*.cpuprofile
+*.heapsnapshot
+
 # Vite
 .vite
 apps/srd/.vite
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -387,18 +387,29 @@ profiles as grep-friendly markdown, which is readable in a terminal and in a
 transcript — a binary profile that needs a flamegraph UI is close to useless to
 an agent, and most of the work in this repo is done by one:
 
+Write the artifacts to `.profiles/`, which is gitignored — the profilers and
+`--metafile-md` otherwise drop files in the cwd, and the next `git add -A`
+sweeps them into a commit.
+
 ```bash
-# where the time went (any bun-run script; --cpu-prof-dir to place the output)
-bun --cpu-prof --cpu-prof-md tools/check-doc-drift.ts
+# where the time went (any bun-run script)
+bun --cpu-prof --cpu-prof-md --cpu-prof-dir=.profiles tools/check-doc-drift.ts
 # what was retained — from apps/srd, since ssg/build.ts resolves relative to it
-bun --heap-prof --heap-prof-md ssg/build.ts
+bun --heap-prof --heap-prof-md --heap-prof-dir=../../.profiles ssg/build.ts
 # module graph + a "Raw Data for Searching" section, for bundle-size questions
-bun build --metafile-md=graph.md --target node apps/discord-bot/src/index.ts
+bun build --metafile-md=.profiles/graph.md --outdir=.profiles/build \
+  --target node apps/discord-bot/src/index.ts
 ```
 
+**`--outdir` is not optional there.** `bun build` with no `--outdir` writes the
+bundle to *stdout*, so omitting it dumps the whole 2 MB Discord bot into your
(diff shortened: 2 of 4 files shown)


alxjrvs (commented):
Docs-only diff (CLAUDE.md). I verified the claims against the repo and the Bun 1.3.14 docs shipped in `node_modules/bun-types`.

Verified correct: the `bun why image-size` output matches `bun.lock` exactly (`image-size@2.0.2` → `@netlify/dev-utils@4.4.7` → `@netlify/blobs@10.7.12`, consumed by `itun` + `su-assets` at `^10.7.11`); `--cpu-prof`, `--cpu-prof-md`, `--cpu-prof-dir`, `--heap-prof`, `--heap-prof-md` and `--metafile-md` are all real flags (`docs/project/benchmarking.mdx`, `docs/bundler/index.mdx`); `tools/check-doc-drift.ts`, `apps/srd/ssg/build.ts`, `apps/discord-bot/src/index.ts` and the named `CardImage.ssr.test.tsx` all exist; the new prose adds no `bun run <script>` reference that would trip `check-doc-drift.ts` check 7. No instructions-to-the-reviewer text in the diff.

Findings:

- `CLAUDE.md:399` — `--cpu-prof-interval` does not appear anywhere in the Bun 1.3.14 docs bundled with `bun-types`, whose CPU-profile flag table lists exactly `--cpu-prof`, `--cpu-prof-md`, `--cpu-prof-name`, `--cpu-prof-dir`; the "1000µs default" is likewise unsourced. Every other flag in this new section was checkable against those docs and checked out, so this one line is the likely invention, and an agent following it hits an unknown-flag error (or a silently ignored flag plus a wrong belief about the sampling default). Non-blocking.
- `CLAUDE.md:396` — `bun build --metafile-md=graph.md --target node apps/discord-bot/src/index.ts` passes no `--outdir`, and `bun build` without `--outdir` prints the entire bundled output to stdout (`docs/bundler/macros.mdx:29`). The recommended command therefore dumps the whole bundled Discord bot into the terminal and the transcript — the exact failure mode the section opens by arguing against. Add `--outdir` (a gitignored dir) or redirect stdout. Non-blocking.
- `CLAUDE.md:390-397` — the outputs these commands write (`graph.md` in the repo root, and the markdown CPU/heap profiles, which default to the cwd) are covered by nothing in `.gitignore`; `dist`, `coverage`, `.vite` etc. are listed but no profile or metafile pattern is. An agent told to run these and then commit will sweep the artifacts in with `git add -A`. Either name a gitignored output dir in the examples (`--cpu-prof-dir`/`--metafile-md` both take a path) or add the patterns. Non-blocking.
- `README.md:64` and `CONTRIBUTING.md:60` — this PR rewrites the CLAUDE.md paragraph about a bare root `bun test`, but both sibling docs still assert the pre-fix claim it replaces ("it skips the per-workspace bunfig preloads and fails by the hundreds"), which the root `bunfig.toml`'s union preload made false. Neither file is scanned by `check-doc-drift.ts`, so nothing will catch it later; updating them to "prefer `bun run test`; a bare root run is viable but not identical" belongs in this change. Non-blocking.

PR #743, opened 2026-08-11
feat(deps): hoist shared versions into a workspace catalog

alxjrvs:
19 dependencies were pinned identically in two or more manifests — react,
react-dom, vite, @sentry/browser, @sentry/node, playwright, @playwright/test,
@base-ui/react and 11 others — so keeping them consistent was manual, and a
major bump opened a separate Dependabot PR per workspace (the flood
.github/dependabot.yml's own comment describes). They now live once in the root
`workspaces.catalog` and are referenced as `catalog:`. 44 references rewritten.

This changed ZERO resolved versions: `bun install` reports "Checked 953 installs
across 1186 packages (no changes)". Only where the version is written changed.
`--frozen-lockfile` works unchanged, so CI and every deploy target are
unaffected.

Two consequences worth knowing, both documented in CLAUDE.md:

  - `overrides` beats the catalog. Bumping a catalogued dep that also carries an
    exact override leaves the catalog stating a version that is not what
    resolves. `sharp` is exactly that case, so it is in the updater's `ignore`
    list and must be bumped in both places by hand.

  - Dependabot cannot read `catalog:` — it will not update these deps
    (dependabot-core #14320) and may strip the `catalog` field from a manifest
    it rewrites (#12522). Both are open. That is why this change ships with a
    replacement updater rather than on its own; without it, moving 19 deps into
    the catalog would silently stop them being updated.

`.github/workflows/catalog-update.yml` covers them instead, pinned to a COMMIT
SHA rather than a moving `@v1`: brandhaug/catalog-update-action is a young,
low-adoption composite action (created 2026-02-27, 0 stars/forks, MIT) and a
composite action runs in this repo's runner with the permissions granted to it,
so a floating tag would be unattended code execution on every scheduled run.
Grouped so deps that must move in lockstep move together (react/react-dom,
both @sentry packages, playwright/@playwright/test, both @fontsource fonts).
Its PRs use GITHUB_TOKEN and therefore do not trigger ci.yml — which fails
closed, since `quality-checks` is required on main and an unchecked PR cannot
merge.

Also fixes tools/check-doc-drift.ts, which learned framework majors by reading a
version literal out of a workspace manifest and reported "has the dependency
moved?" for Vite and React once those became `catalog:`. It now resolves one hop
through the catalog, handling both the default (`catalog:`) and named
(`catalog:<name>`) forms.

Verified: check:all exits 0; itun and discord-bot build; no routeTree drift.

---


alxjrvs (commented):
Read the diff and checked it against the repo (root `overrides`, CI path filters, Netlify/Render install commands, knip's own workspace declarations, release-please config, and the `check-doc-drift` consumers).

The mechanical part is sound: all 19 catalog entries match the literals they replaced, all 44 references resolve, `workspaces` object form breaks nothing here (knip declares its own workspaces, `check-doc-drift`'s workspace check reads directories, release-please has no `node-workspace` plugin), and CI's `shared` path filter already includes `package.json` + `bun.lock`, so a catalog-only bump still triggers every app build. No prompt-injection text in the diff.

Two findings:

- `.catalog-updaterc.json:22` — the `sharp` ignore rests on a premise that isn't true, and the dep it should be protecting is unignored. Root `overrides` (`package.json:34-43`) contains **no `sharp` entry**; it does contain `"@vitejs/plugin-react-swc": "^4.3.3"`, which *is* catalogued (`package.json:21`). So the updater is free to raise `@vitejs/plugin-react-swc` in the catalog to a version the override's caret pins away from — producing exactly the "catalog states a version that is not what resolves" failure the new docs warn about — while `sharp`, which has no override at all, is excluded from updates for no reason. Concretely: a `@vitejs/plugin-react-swc` 5.x release lands, the workflow writes `"^5.0.0"` into the catalog, the `^4.3.3` override keeps 4.x resolved tree-wide, and the catalog silently becomes fiction. (`CLAUDE.md:100` encodes the same wrong premise — "`sharp` is exactly this case (exact override, per the audit-gate section)" — and the audit-gate section it cites documents no such override either; `.github/workflows/ci.yml:326` carries the same stale claim, which predates this PR.)

- `.github/workflows/catalog-update.yml:112` — the job never sets up Bun, unlike every other job in this repo (`uses: ./.github/actions/setup-bun`), yet passes `"packageManager": "bun"`. `ubuntu-latest` ships no `bun`, so if the action shells out to it to regenerate `bun.lock` the step fails; if it instead only rewrites `package.json`, the PR carries a catalog/lockfile mismatch that `bun install --frozen-lockfile` rejects. Either way the scheduled run produces nothing mergeable. This one fails closed (the PR can't merge without `quality-checks`, which the file already notes never fires for `GITHUB_TOKEN` PRs), so it costs a silent weekly no-op rather than a bad merge — but it is worth a dry-run (`workflow_dispatch` with `dry-run: true`) before trusting the schedule.

PR #744, opened 2026-08-11
fix(ci): run CI and CodeQL on stacked PRs, not just PRs based on main

alxjrvs:
Both workflows filtered `pull_request` with `branches: ['main']`. That filter
matches the pull request's BASE branch, so any PR based on another PR — every
layer of a stacked PR above the bottom one — got no run at all.

Because `quality-checks` and `Analyze (javascript-typescript)` are both required
statuses on main, the effect was worse than "unscanned": those PRs were
permanently unmergeable, and nothing on the PR explained why. `gh pr checks`
reports "no checks reported" and the required status simply never arrives.

Observed directly on stack #742: #738, based on main, got a full run and passed;
#739 and #740, based on bun-bot-sourcemaps and bun-install-cooldown, had zero
checks of any kind.

Stacked PRs are the preferred shape for multi-part work in this repo, so CI has
to run regardless of base. Dropping the filter (rather than enumerating branch
patterns) is what makes that true for stacks that do not exist yet. The `push`
trigger keeps its `branches: ['main']` filter, so branch pushes still do not
double-run.

The cost is that a PR targeting any branch now runs the full suite. That is the
intended behaviour here, not a side effect — the path filters in `changes` still
skip the heavy jobs for changes that cannot affect them.

https://claude.ai/code/session_01S7FKAMEU6h1MxzrmRMnkEC

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,8 +3,21 @@ name: CI
 on:
   push:
     branches: ['main']
+  # NO `branches:` filter here, deliberately. That filter matches the PR's BASE
+  # branch, so `branches: ['main']` silently gave ZERO checks to any PR based on
+  # something other than main — i.e. every layer of a stacked PR above the
+  # bottom one. Since `quality-checks` is a required status on main, those PRs
+  # were not merely unchecked, they were unmergeable, with nothing on the PR
+  # saying why: `gh pr checks` reports "no checks reported" and the required
+  # check simply never arrives.
+  #
+  # Observed on stack #742: #738 (base main) got a full run, while #739 and #740
+  # (bases bun-bot-sourcemaps / bun-install-cooldown) got none at all.
+  #
+  # Stacked PRs are the preferred shape for multi-part work here, so CI has to
+  # run regardless of base. The cost is that a PR targeting any branch now runs
+  # CI, which is the intended behaviour rather than a side effect.
   pull_request:
-    branches: ['main']
   # Run on the merge queue's temporary branches so the required `quality-checks`
   # gate reports on queued PRs; without this the merge queue hangs forever
   # waiting for a check that never fires.
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -6,8 +6,11 @@ name: CodeQL
 on:
   push:
     branches: ['main']
+  # Unfiltered for the same reason as ci.yml: `branches:` matches the PR's BASE
+  # branch, so it gave no run at all to a PR stacked on another PR — and
+  # `Analyze (javascript-typescript)` is its OWN required status on main, so
+  # that left those PRs permanently unmergeable rather than merely unscanned.
   pull_request:
-    branches: ['main']
   # Required before `Analyze (javascript-typescript)` may be made a required
   # status context on `main`, and the order matters: this repo merges through a
   # merge queue, whose temporary `gh-readonly-queue/**` branches fire NEITHER
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -24,6 +24,16 @@ on:
     # top of the hour to avoid the scheduling stampede.
     - cron: '30 5 * * 1'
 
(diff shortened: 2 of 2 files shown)


alxjrvs (commented):
I read the diff and checked both workflow files plus everything that depends on their triggers (`quality-checks` aggregate gate, `dorny/paths-filter` base resolution, the other two workflows, and the ruleset assumptions recorded in the repo docs).

**Verdict: the change is correct.** `pull_request.branches` does filter on the PR's *base* ref, so `branches: ['main']` genuinely gave zero runs to any PR stacked on a non-main branch, and dropping it is the right fix. The resulting YAML is valid (`pull_request:` with a null value = default activity types, all bases), `push`/`merge_group` are untouched, and nothing else in `.github/` keys off the base branch. `dorny/paths-filter` on `pull_request` resolves the diff from the PR's own base, so a stacked layer still gets correctly-scoped path filtering rather than the union of the layers below it. No prompt-injection or agent-directed text in the diff.

One non-blocking issue:

- `.github/workflows/codeql.yml:13` — CodeQL has no `concurrency` block (unlike `ci.yml`, which cancels superseded runs), so now that it fires on every PR regardless of base, a single `gh stack sync` force-pushing an N-layer stack spawns N CodeQL analyses per sync with every earlier in-flight analysis left running to completion (`timeout-minutes: 360`); this was bounded to main-based PRs before and is now unbounded. Suggested fix: add `concurrency: {group: codeql-${{ github.ref }}, cancel-in-progress: true}`.

Two things I checked and am *not* reporting, for the record: the new comment's claim that the unchecked stacked PRs were "unmergeable" holds for native `gh stack` stacks (GitHub evaluates the stack base's ruleset on every layer) even though it would not hold for a hand-retargeted PR; and the pre-existing merge-queue prose in both files may be stale relative to `main`'s current ruleset, but that text is not touched by this diff.

alxjrvs (commented):
Read the diff and both workflow files at their post-change state.

**What the change does:** drops the `branches: ['main']` filter from `pull_request` in `ci.yml` and `codeql.yml` (that filter matches the PR's *base*, so stacked layers got zero checks while `quality-checks` / `Analyze (javascript-typescript)` are required on `main` — genuinely unmergeable), and adds a workflow-level `concurrency` block to `codeql.yml` to bound the extra runs. Both changes are correct for the stated problem. `pull_request:` with a null value is valid YAML and keeps the default activity types (`opened`, `synchronize`, `reopened`), so no trigger coverage is lost. `push: ['main']` is untouched, so no duplicate main runs. The `merge_group:` trigger and the ordering constraint it documents are preserved in both files.

I checked the concurrency key against every trigger in `codeql.yml`: for `pull_request` `github.ref` is `refs/pull/N/merge` and for `merge_group` it is the unique `gh-readonly-queue/**` ref, so cancellation stays scoped to re-pushes of the same PR / same queue entry — it cannot cancel a sibling stack layer or a queued PR. No other workflow depends on the removed filter (`e2e-nightly`, `release-please`, `catalog-update` have their own triggers), and no doc under `docs/` asserts the old trigger shape, so `check-doc-drift` won't go stale on this.

No security defect: both workflows use `pull_request` (not `pull_request_target`), so fork code keeps a read-only token; widening the base set doesn't change that trust boundary. The diff contains no text addressed to a reviewing agent.

Findings, both non-blocking:

- `.github/workflows/codeql.yml:3` — the file's header comment still says the workflow scans "every push to main, every PR into main, and weekly", which is exactly the claim this PR invalidates; a future reader reconciling the header against the new inline comment gets contradictory statements about which PRs are scanned.
- `.github/workflows/codeql.yml:34` — `group: codeql-${{ github.ref }}` resolves to `codeql-refs/heads/main` for both the `push: main` and the `schedule` triggers, so a Monday-05:30 merge and the weekly scan land in the same group and cancel each other; impact is low (both analyze main's HEAD and the survivor covers the same code), but the weekly run's purpose — re-scanning *unchanged* code against new query packs — is the one that gets dropped if the push wins.

PR #736, opened 2026-08-11
fix(srd): re-bless the output snapshot for the About page

alxjrvs:
## `main` is red for every PR that builds srd

#731 added Aevox to `SPECIAL_THANKS.md`, which the About page renders — so `about/index.html` grew by 6 characters. That PR touched only the one Markdown file, and CI's path filtering **skipped `build-srd` entirely**, so the output gate never ran on the change that moved the output.

The drift landed on `main` unnoticed and now fails on the next PRs unlucky enough to trigger that job:

| PR | what it changes | why it's red |
|---|---|---|
| #732 | a codeql-action version bump | `build-srd` — snapshot drift |
| #733 | Dependabot dev-dependencies | same |
| #734 | Dependabot production-dependencies | same |

None of them introduced it. This is the only reason all three are blocked.

## The diff

Two entries, both expected:

- **`about/index.html`** — main-text digest. The Aevox line (2697 → 2703 chars).
- **`changelog/index.html`** — digest + append-only text, picking up the Site v2.3.0 entries. The gate already classifies this page as append-only growth and was passing it; re-blessing just records the current tail.

Verified: `bun --filter srd gate` **FAILS** on `356645ef` and **passes** here.

## Worth noting the shape

Path-filtered CI makes "which file did I change" and "which output changed" two different questions, and only the second is what the gate exists to answer. A prose file at the repo root can move a rendered page without tripping any of the filters that decide whether to look — so the gate is skipped precisely when a docs-only change quietly alters the site.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_019Ms18wRLjK2W62PGUk8XWj

PR #748, opened 2026-08-11
ci: run build-srd when the root prose the About page renders changes

alxjrvs:
## The gap

`apps/srd/src/pages/about.page.tsx` reads three files off disk at **build time** and renders them into `about/index.html`:

```ts
const aboutJrvsMd     = readFileSync(repoRootFile('ABOUT_JRVS.md'), 'utf8')
const llmStatementMd  = readFileSync(repoRootFile('LLM_STATEMENT.md'), 'utf8')
const specialThanksMd = readFileSync(repoRootFile('SPECIAL_THANKS.md'), 'utf8')
```

All three live at the repo root, outside `apps/srd/**`, so a change to one matched **no filter** in the `changes` job. `build-srd` was skipped — and with it the output snapshot gate, the one check whose entire job is to notice that a page moved.

The gate was never broken. It was simply never asked, and only on the changes most likely to need it: a docs-only edit that quietly alters the site is exactly where "which file did I touch" and "which page changed" come apart.

## The worked example

#731 added one name to `SPECIAL_THANKS.md` and touched nothing else. CI skipped `build-srd`, `main` went green carrying a stale snapshot, and the next three PRs to trigger that job — #732 (a codeql-action bump), #733 and #734 (Dependabot groups) — were all red on a **six-character** difference in the About page that none of them introduced. Three PRs blocked for a day, with `main` reporting healthy throughout. #736 cleaned up the drift; this stops it recurring.

## Scope

Added to `web` rather than `shared`: these files affect the srd build and nothing else. The changelog page's two `CHANGELOG.md` sources need no entry — they sit under `apps/srd/` and `packages/salvageunion-reference/`, both already matched.

**Residual risk, stated in the comment:** the list is hand-maintained, so a new root file rendered by a page has to be added here too. The comment sits at the point someone would be editing.

## Verification

This PR touches `.github/workflows/**`, which is in `shared`, so every job including `build-srd` runs on it — the change exercises the surrounding config on its own merge. The filter itself is verified by the next `SPECIAL_THANKS.md`-only change running `build-srd` instead of skipping it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -91,6 +91,39 @@ jobs:
               - 'bunfig.toml'
               - 'patches/**'
               - 'coverage-baseline.json'
+              # Root prose with THREE consumers, none of them under a path
+              # matched above: `apps/srd/src/pages/about.page.tsx` reads it off
+              # disk at build time and renders it into `about/index.html`,
+              # `apps/itun/src/routes/about.tsx` imports it `?raw` into the ITUN
+              # bundle, and
+              # `packages/component-lib/src/markdownSection/MarkdownSection.test.tsx`
+              # reads the real files and asserts on their headings.
+              #
+              # `shared` rather than `web`, for two reasons. The obvious one is
+              # that ITUN and component-lib are consumers too, so `web` alone
+              # would leave `build-itun` and `test` skipped on a prose edit that
+              # renames a heading. The subtle one is that `web` alone does not
+              # even fix srd: `build-srd` has `needs: [build-package, lint,
+              # typecheck, test, validate, knip, audit]`, every one of them
+              # gated on `code`. A prose-only PR would resolve to `web=true,
+              # code=false`, those jobs would skip, and GitHub skips a dependent
+              # job when any `needs:` job is skipped — so `build-srd` would skip
+              # anyway and the snapshot gate would still never run.
+              #
+              # That is not hypothetical. #731 added one name to
+              # SPECIAL_THANKS.md and nothing else; CI skipped build-srd, main
+              # went green carrying a stale snapshot, and the next three PRs to
+              # trigger that job (#732, #733, #734 — a codeql bump and two
+              # Dependabot groups) were all red on a six-character difference
+              # none of them introduced. The gate was working; it was never
+              # asked.
+              #
+              # Keep this in step with what those three modules read. The
+              # changelog page's two CHANGELOG.md sources need no entry: they
+              # sit under `apps/srd/` and `packages/salvageunion-reference/`.
+              - 'ABOUT_JRVS.md'
+              - 'LLM_STATEMENT.md'
+              - 'SPECIAL_THANKS.md'
             itun:
               - *shared
               - 'apps/itun/**'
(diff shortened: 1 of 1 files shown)


alxjrvs (commented):
**Review — `.github/workflows/ci.yml` (paths-filter: root prose → `web`)**

The premise is right and verified: `apps/srd/src/pages/about.page.tsx:40-42` reads exactly those three repo-root files at build time, and the comment's claim about the two `CHANGELOG.md` sources is accurate (`apps/srd/CHANGELOG.md` matches `apps/srd/**`, the package one matches `shared`). No injected instructions in the diff.

Two problems with the fix itself:

- **BLOCKING** — `.github/workflows/ci.yml:119` — the three paths are added only to `web`, but they match **no** pattern in the `code` filter (`shared` + `apps/**` + `packages/**` + `tools/**` + `docs/**` — a repo-root `.md` matches none). Until this change, `web ⊆ code` held for every path, so `build-srd`'s `needs: [changes, build-package, lint, typecheck, test, validate, knip, audit]` was always satisfied. On a prose-only PR it now resolves to `web=true, code=false`: `build-package`, `typecheck`, `test`, `validate`, `knip` and `audit` are all skipped by their own `if:`, and GitHub skips a dependent job when any `needs:` job is skipped (only an `always()`/`!cancelled()` `if` escapes it — which is exactly why `quality-checks` carries `if: always()`). So `build-srd` — and `e2e-web`, which needs it — is still skipped, the snapshot gate still never runs, and `quality-checks` still reports green on a stale snapshot. The #731 scenario reproduces unchanged. Note this PR cannot demonstrate otherwise: it touches `.github/workflows/**`, which is in `shared`, so every job runs here. Fix: put the three files in `shared` (or add them to `code` as well as `web`).

- `.github/workflows/ci.yml:119` — root prose is an input to more than srd's build, and the other consumers are left uncovered by the same gap. `packages/component-lib/src/markdownSection/MarkdownSection.test.tsx:9-12` reads all three real files off disk and asserts on their content (e.g. headings `['About the Dev', 'LLM Statement', 'Special Thanks']`), and `apps/itun/src/routes/about.tsx:3-5` imports them with `?raw` into the ITUN bundle. With `code` and `itun` unextended, a prose edit that renames a heading or reshapes the crew list still merges without `test`/`typecheck` or `build-itun` ever running. Adding the entries to `shared` instead of `web` closes this and the blocking item together.

alxjrvs (commented):
Read the diff and verified each claim in it against the repo.

**What the change does, verified:** the three added paths are exactly the repo-root prose consumed by builds — `apps/srd/src/pages/about.page.tsx:40-42` reads all three with `readFileSync` at build time, `apps/itun/src/routes/about.tsx:3-5` imports all three `?raw`, and `packages/component-lib/src/markdownSection/MarkdownSection.test.tsx:9` reads the real files. A grep for every root-level `.md` reference from `apps/**` returns nothing else, so the list is complete for its stated scope, and the comment's note about the two `CHANGELOG.md` sources is right (`apps/srd/CHANGELOG.md` → `web`/`code`, `packages/salvageunion-reference/CHANGELOG.md` → `shared`).

**The `shared`-not-`web` reasoning holds.** `build-srd` at `.github/workflows/ci.yml:404` is `needs: [changes, build-package, lint, typecheck, test, validate, knip, audit]` with `if: needs.changes.outputs.web == 'true'`, and those upstream jobs gate on `code` — so `web` alone would indeed leave `build-srd` skipped-by-dependency and the snapshot gate unrun. Putting the paths in the `&shared` anchor turns on `code`/`web`/`itun`/`bot` together, which is what makes the gate actually fire. YAML anchor placement is valid; comments inside the sequence are fine for `dorny/paths-filter`.

No correctness or security defects in the diff. One adjacent gap:

- `.github/workflows/ci.yml:150` — `code` matches `tools/**` and `docs/**` so `validate` runs on the docs it guards, but `tools/check-doc-drift.ts` also reads root `CLAUDE.md`, root `README.md` (`frameworkVersionDocs`, lines 809-811) and `.claude/rules|agents|skills` (`LIVE_INSTRUCTION_DOC_DIRS`, line 196) — none of which match any filter. A PR editing only those (e.g. writing a stale framework major into root `CLAUDE.md`) skips `validate` entirely, merges green, and then reddens the next unrelated code PR — the identical #731 failure mode this diff fixes one directory over. Non-blocking, pre-existing, and a natural companion entry to the ones being added here.

Nothing in the diff addresses instructions to a reviewing agent; it is all ordinary explanatory comment prose.

PR #749, opened 2026-08-12
docs(itun): Convex exception reporting is delivering — verified, not assumed

alxjrvs:
## It works now

Both `apps/itun/CLAUDE.md` and `docs/architecture/accounts-and-games.md` asserted this integration was configured on 2026-08-05 and **not delivering**. The DSN is now in the Convex dashboard and events are arriving.

**Confirmed end to end, not by the save button:**

1. Forced an error against production — `maintenance:dedupeAppIds '{"apply":"not-a-boolean"}'`
2. Saw it in `convex logs` at `20:36:51Z`
3. Saw it in Sentry as **`ITUN-CONVEX-1`** — the first event that project has *ever* received, after **zero in 90 days**

The Pro-plan caveat the docs flagged as the prime suspect turned out not to apply. Step 1 of the runbook (create the Sentry project) had been done; step 2 (paste the DSN) never had.

It also settles a question the docs left open: Convex forwards **`ArgumentValidationError`**, not only handler throws.

## Why the wording changed shape

The old text failed in a specific way worth not repeating. "Enabled" was recorded the moment the *Sentry project* existed — one of two steps — and nothing about the result looked different afterwards, because a reporting integration that reports nothing is indistinguishable from a healthy one with no errors. It stayed wrong for a week, through 39 failed `entities:upsertByAppId` mutations in one evening, and that incident reached a player on Discord before it reached us.

So both files now carry **the probe** instead of a status line:

```bash
cd apps/itun
CONVEX_DEPLOYMENT=dev:perfect-donkey-72 bunx convex run --prod \
  maintenance:dedupeAppIds '{"apply":"not-a-boolean"}'
bunx convex logs --deployment alex-jarvis:suref-itun:prod --history 6 --jsonl
```

In the log but not in Sentry ⇒ it has stopped delivering again.

## No automated check, deliberately

`tools/check-observability.ts` can assert the browser SDK is present in the shipped bundle. There is no equivalent here: **"zero events" is also what a healthy quiet week looks like**, so it cannot be asserted without producing a false alarm on every quiet day. Saying so in the doc is the honest alternative to a check that would get muted.

## Flagged, not fixed

There is **no alert rule** on the `itun-convex` project, and Sentry collects silently by default. Events arriving is not the same as anyone reading them — which is the actual failure mode from 8/9. Worth a rule before the next incident.

Also: `ITUN-CONVEX-1` is my synthetic probe. Resolve it whenever — it's evidence, not a real defect.

## Verification

`bun run validate:doc-drift` green (ADR supersession markers, component-lib symbols, framework majors, and all 79 documented bun-script references).

### DIFF
--- a/apps/itun/CLAUDE.md
+++ b/apps/itun/CLAUDE.md
@@ -111,12 +111,16 @@ and remains the account-free way to share a build.
   (`src/lib/observability.ts`) and the Netlify Functions
   (`netlify/functions/_observability.ts`) each own one; Convex instead uses its
   first-party Exception Reporting integration, enabled in the Convex dashboard
-  with no application code. **It was configured on 2026-08-05 and is not
-  delivering** — the `itun-convex` Sentry project has had zero events in 30 days
-  (measured 2026-08-10, over a window containing 39 real mutation failures), so
-  a quiet project is not evidence of a healthy backend. Read Convex errors with
-  `bunx convex logs --deployment alex-jarvis:suref-itun:prod`; the Pro-plan gate
-  is the first suspect. Queries and
+  with no application code. **Delivering since 2026-08-12**, verified end to end
+  rather than assumed: a forced error appeared in `convex logs` and then in
+  Sentry as `ITUN-CONVEX-1`, the first event that project had ever received.
+  It also forwards `ArgumentValidationError`, not only handler throws.
+  Before that it had been recorded as "enabled" on 2026-08-05 while the DSN
+  had never actually been pasted into the dashboard, and it sat silent for a
+  week — through 39 real mutation failures. **A quiet Sentry project is not
+  evidence of a healthy backend**; re-verify by forcing an error and comparing
+  against `bunx convex logs --deployment alex-jarvis:suref-itun:prod`, which
+  remains the ground truth. Queries and
   mutations run in a deterministic runtime with no `fetch`, so an in-function SDK
   could not report from them at all — rationale and the enable-it runbook are in
   [docs/architecture/accounts-and-games.md](../../docs/architecture/accounts-and-games.md)
--- a/docs/architecture/accounts-and-games.md
+++ b/docs/architecture/accounts-and-games.md
@@ -322,25 +322,44 @@ repo):
   Convex errors will not carry a commit SHA the way the Netlify and Render
   surfaces do.
 
-**Status: configured 2026-08-05, and NOT delivering.** Measured 2026-08-10: the
-`itun-convex` Sentry project has received **zero events in 30 days** — over a
-window that contains 39 failed `entities:upsertByAppId` mutations in a single
-evening, every one of which should have landed there.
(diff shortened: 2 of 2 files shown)


alxjrvs (commented):
Docs-only diff (two files). I read the changed prose against `apps/itun/convex/maintenance.ts`, `docs/architecture/agent-tooling.md`, and the installed `convex@1.43.0` CLI source to check the new runbook commands. No instruction-shaped text addressed to a reviewing agent appears in the diff.

**Findings**

- `docs/architecture/accounts-and-games.md:352` — The new probe recipe dropped the `timeout 120 … > logs.jsonl` wrapper that the removed version carried, but `convex logs` still never exits: `node_modules/…/convex/dist/esm/cli/lib/logs.js:47` is a `for (;;)` poll loop, and `--history N` only changes what the *first* poll prints (line 58-69). Anyone (or any agent) following this runbook literally hangs their shell after the probe, which is exactly the failure the deleted comment existed to prevent.
- `docs/architecture/accounts-and-games.md:350` — The documented liveness probe invokes `maintenance:dedupeAppIds` against **production**, a mutation that permanently deletes `pilots`/`mechs`/`crawlers` rows when `apply` is true (`apps/itun/convex/maintenance.ts:169-174`). It is safe today only because `v.optional(v.boolean())` rejects `"not-a-boolean"` before the handler runs — so a single typo (`'{"apply":true}'`), a copy-paste that drops the args, or any future loosening of that validator turns a routine verification step into prod data loss. A function that cannot mutate anything (or a deliberately nonexistent function name) would force the same `ArgumentValidationError`-class event with no blast radius.
- `docs/architecture/accounts-and-games.md:350` — `CONVEX_DEPLOYMENT=dev:perfect-donkey-72` and `--prod` in the same command point at different deployments; the flag wins, so the env var is at best noise and at worst misleading. Since step 2 of the runbook (line 308-311) requires the DSN be pasted **per deployment** and the diff only claims prod was verified, a reader who drops `--prod` to "match the env var" probes dev, gets no Sentry event, and concludes delivery has broken again.
- `docs/architecture/accounts-and-games.md:328` — "the Pro-plan caveat turned out not to apply" contradicts the caveat block directly above it (lines 316-319), which is unedited and still states as fact that Exception Reporting is Pro-only and is framed as the first thing to check. A reader cannot tell whether the caveat is now known-false or merely satisfied by this account's plan; likewise line 306-307 still instructs "check the dashboard rather than assuming" for a step this diff declares done.
- `docs/architecture/accounts-and-games.md:355` — Sentry's issue-stream `project` query param takes a numeric project id, not a slug, so `?project=itun-convex` most likely lands on the unfiltered issue stream rather than the `itun-convex` project — which weakens the "in the log but not in Sentry" comparison the very next sentence depends on.

Nothing here breaks code, CI, or a build — both files are prose, and the surrounding claims (deployment names, `--deployment team:project:prod` selector form, `susrd.sentry.io` web host, `itun-convex` slug) check out against the repo. The two commands in the new probe block are the parts I'd fix before merging.

PR #750, opened 2026-08-13
fix(itun): stop printed sheets losing the pilot class, mech chassis and crawler type

alxjrvs:
Fixes #616 — the nightly `e2e-itun` failure.

## The bug

A picker `Field` renders its **value inside the `<button>`** rather than beside it, so both apps' print rule `button:not([data-print='keep'])` took the field's *content* along with its affordance.

On paper, each sheet lost its single most identifying field:

| Sheet | Field | Renders as |
|---|---|---|
| Pilot | **Class** | `PilotIdentity.tsx:200` picker |
| Mech | **Chassis** | `MechIdentity.tsx:105` picker |
| Crawler | **Type** | `CrawlerIdentity.tsx:134` picker |

Every *sibling* field printed fine, which is what made this read as a quirk of one field rather than a rule hitting a whole variant: `InlineEditField` renders a `<span role="button">`, which the element selector never matched. Only the editable branch was affected — a read-only `Field` already rendered a plain `<div>`.

## The fix

`data-print="keep"` on the picker button — the escape hatch `print.css` documents for exactly this case ("a control that carries printed STATE… opts back in"), shared verbatim with `apps/srd/src/styles/global.css`, and until now unused. The value box prints as a bordered box, which is what a printed form field should look like.

## Verification

- **Causal, not correlational**: both print-preview specs (A4 + US Letter) **fail with the fix reverted** and **pass with it applied**.
- Added a unit-level guard so this doesn't rely solely on the nightly. jsdom applies no `@media` rules, so it asserts the opt-in attribute *and* the containment that makes it necessary.
- `check:fast` (lint + validate:all + knip + typecheck, 6 workspaces) — clean.
- Full suite `bun run test` — green.
- `bun --filter srd gate` — **SNAPSHOT OK, zero unexpected differences** (srd renders these fields read-only, so its output is untouched).

## Notes

Pips were checked and are **not** affected — they render as `span[data-pip]`, so the print stylesheet's pip rules were always live. The "filled pip button" in that file's comment was hypothetical.

### DIFF
--- a/apps/itun/src/components/sheet/__tests__/print.test.tsx
+++ b/apps/itun/src/components/sheet/__tests__/print.test.tsx
@@ -145,6 +145,36 @@ describe('Print markup — PilotSheet', () => {
     const backLink = container.querySelector('header a[aria-label="Back to roster"]')
     expect(backLink).toBeTruthy()
   })
+
+  /**
+   * Regression (#616). A picker `Field` renders its VALUE *inside* the button
+   * rather than beside it, so the print rule `button:not([data-print='keep'])`
+   * hid the pilot's class on paper — the one field saying what the pilot IS.
+   * The sibling fields printed fine because `InlineEditField` renders a
+   * `<span role="button">`, which that element selector never matched, so the
+   * loss looked like a quirk of one field instead of a rule hitting a whole
+   * variant. Same shape on the mech (chassis) and crawler (type) sheets.
+   *
+   * jsdom applies no `@media print` rules, so assert the opt-in attribute and
+   * the containment that makes it necessary — not a computed style.
+   */
+  test('class picker opts into print, so its value survives the button-hide rule', () => {
+    const { container } = render(
+      <Sheet
+        kind="pilot"
+        id="print-pilot-1"
+        entityStore={makeEntityStore([fakePilot])}
+        softLinkStore={makeEmptySoftLinkStore()}
+      />
+    )
+    const picker = container.querySelector('button[aria-label="Change class"]')
+    expect(picker).toBeTruthy()
+    // The value is inside the button — this containment is why the hatch is
+    // required, so assert it too; if it ever moves out, this test should be
+    // revisited rather than silently passing on the attribute alone.
+    expect(picker?.textContent).toContain('Engineer')
+    expect(picker?.getAttribute('data-print')).toBe('keep')
+  })
 })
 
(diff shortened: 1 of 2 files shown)


alxjrvs (commented):
Reviewed the diff (2 files: a `data-print="keep"` opt-in on `Field`'s picker button, plus a regression test). The fix itself is correct: `apps/itun/src/styles/print.css:126` hides `button:not([data-print='keep'])`, the picker's value lives inside that button (`Field.tsx:172`), and the three affected call sites — `PilotIdentity.tsx:200`, `MechIdentity.tsx:105`, `CrawlerIdentity.tsx:134` — all route through this one branch, so a single shared-component attribute covers all three. The read-only path (`canEdit === false` → `onEditClick` undefined) falls to `InlineEditField`'s `<span role="button">` and was never hidden, so nothing regresses there. No blocking issues.

Findings, all non-blocking:

- `packages/component-lib/src/components/chrome/Field.tsx:162` — the comment says the hatch "is shared verbatim by both apps' print CSS", but srd's print block (`apps/srd/src/styles/global.css:220-247`) hides only `nav, footer, [data-print='hide']` and has no button-hide rule at all; only the *attribute vocabulary* is shared, so a reader chasing the srd side will look for a rule that does not exist.
- `packages/component-lib/src/components/chrome/Field.tsx:163-164` — "a read-only Field renders the `<div>` below" describes an unreachable branch: the block is already inside `if (onEditClick)` and the ternary at line 149 re-tests the same truthy value, so that `<div>` is dead code and a read-only `Field` actually takes the `InlineEditField` path. The conclusion ("was never hidden") is right, the stated reason is not.
- `apps/itun/src/styles/print.css:28` (and the mirrored claim at `apps/srd/src/styles/global.css:218`) — both assert "Neither is used in markup yet"; this PR makes that false for `data-print="keep"` and neither file is updated, leaving the two stylesheets stating the hatch is unused while a shared component now depends on it.
- `apps/itun/src/components/sheet/__tests__/print.test.tsx:161` — the guarded behaviour lives in `component-lib`, but the only assertion is in the itun suite, so `bun --filter component-lib test` stays green if the attribute is deleted; `packages/component-lib/src/components/chrome/__tests__/chrome.test.tsx:22` already has a `Field` describe block and is the natural home for it (CI's full-suite run and the pre-push shared-surface rule do still catch it, which is why this is minor).

PR #752, opened 2026-08-13
fix(ci): stop the nightly observability probe hanging, and bound every job

alxjrvs:
Follow-up to #750, which fixed the `e2e-itun` half of #616. This fixes the other nightly job and makes timeout policy consistent across every workflow.

Three findings, all the same shape: **a failure that produces no signal.**

## 1. The observability probe has hung silently since 2026-08-05

`observability-probe` is `cancelled` on every nightly, having printed **nothing at all**. That reads as "cancelled" in the run list, which looks like someone stopped it — so nobody looked.

The probe itself is healthy. It runs in **3–21s locally** and reports the DSN inlined and the CSP permitting on both sites. Two defects made a working check unrunnable:

| | |
|---|---|
| `fetch` has no default timeout | one stalled socket blocks forever |
| retry ladder is 1+2+4 = **7s per URL** | × ~134 chunks (srd 49, itun 85) = up to **938s** |

Against `timeout-minutes: 5`, the job could *only* die at its timeout — and did, at 315s. No output, because the tool printed only at the end: no way to tell which app, which URL, or even which phase.

**Fixed by bounding both axes without weakening the retry policy** (which exists so the probe doesn't cry wolf):

- 15s per-attempt `AbortSignal.timeout`
- 120s whole-probe budget
- progress lines naming each app before its first request, and the chunk count after its crawl
- job timeout → 10m, so it always **loses the race** to the probe's own budget; the tool reports the reason instead of the runner killing it mute

Budget exhaustion raises a distinct `ProbeBudgetExhausted`, reported as *"this is a PROBE failure, not evidence the deploy is dark"*. Without that it surfaces as "SDK absent" / "production unreachable" and opens a tracking issue against an innocent deploy — the exact cry-wolf outcome the retry policy was written to prevent. Verified by forcing the budget to 1ms.

New output:

```
  [srd] fetching https://salvageunion.io…
  [srd] scanned 49 chunk(s) from 3 entry point(s)
  [srd] DSN inlined in …/islands-C4xH-Zp7.js (ingest o…de.sentry.io)
  [itun] fetching https://intheunionnow.com…
  [itun] scanned 85 chunk(s) from 34 entry point(s)
✓ observability probe OK
```

## 2. `convex-parity` could open the tracking issue but never named itself

It sat in `needs:` and in the notify `if:` — so it can file the issue on its own — while having **no line in the issue body**. A parity-only failure therefore filed an issue reading "e2e-itun: **success**, e2e-srd: **success**" and nothing else.

That is verbatim the defect the comment directly above it warns about for `observability-probe`; the lesson just wasn't applied when this job was added. Verified by running the real extracted script against stubbed GitHub APIs:

```
- `e2e-itun`: **success**
- `e2e-srd`: **success**
- `convex-parity`: **failure**
```

## 3. 13 of 15 `ci.yml` jobs had no timeout at all

GitHub's default is **360 minutes**. A hung job doesn't fail — it holds the `quality-checks` gate, and therefore the PR, open for six hours. Only the two e2e jobs were bounded.

Every job in every workflow is now bounded — **25 jobs across 5 workflows**; CodeQL keeps its deliberate 360. 15m is ~7× the slowest measured job on main (2.1m), so it fires only on a genuine hang, never on a slow runner. `convex-parity` keeps its 5m: measured at 42s, 7× headroom.

## Verification

- Probe passes live (6.6s) and its budget-exhaustion path was exercised directly.
- Static `validate:observability` unchanged and still green.
- Both `github-script` bodies extracted and `node --check`ed, then behaviourally tested.
- All five workflows re-parsed; a script asserts every job is bounded.
- `check:fast` clean.

### DIFF
--- a/.github/workflows/catalog-update.yml
+++ b/.github/workflows/catalog-update.yml
@@ -42,6 +42,10 @@ permissions: {}
 jobs:
   update:
     runs-on: ubuntu-latest
+    # Bounded like every job in ci.yml and e2e-nightly.yml — GitHub's default is
+    # 360 minutes, and an unattended weekly job is the least likely place anyone
+    # would notice one hanging.
+    timeout-minutes: 15
     permissions:
       # The minimum the action needs: commit a branch, open a PR against it.
       contents: write
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,6 +34,18 @@ concurrency:
   group: ci-${{ github.ref }}
   cancel-in-progress: true
 
+# EVERY job below declares `timeout-minutes`. GitHub's default is 360 (six
+# hours), so an unbounded job that hangs does not fail — it sits, burning runner
+# minutes and holding the `quality-checks` gate, and therefore the PR, open for
+# the rest of the day. Until now only the two e2e jobs were bounded and the
+# other thirteen were not, which is the same hole that let the nightly
+# observability probe hang undiagnosed: nothing is red, so nothing is read.
+#
+# 15 minutes is deliberately generous — the slowest job on main measures 2.1m,
+# so this is ~7x headroom and will only ever fire on a genuine hang, never on a
+# slow runner. The e2e jobs keep their own 20: they boot a dev server and drive
+# a browser, so their spread is wider. Raise a specific job's number when its
+# real work grows; do not remove the key.
 jobs:
   # ---------------------------------------------------------------------------
   # Path filter — decide which areas a change touches so we can skip heavy jobs
@@ -45,6 +57,7 @@ jobs:
   # ---------------------------------------------------------------------------
   changes:
     runs-on: ubuntu-latest
+    timeout-minutes: 15
     outputs:
       itun: ${{ steps.filter.outputs.itun || steps.all.outputs.all }}
       web: ${{ steps.filter.outputs.web || steps.all.outputs.all }}
@@ -154,6 +167,7 @@ jobs:
     # Always runs — format:check covers docs (markdown/yaml) too, and the two
     # static design guards below are unconditional by design.
     runs-on: ubuntu-latest
+    timeout-minutes: 15
     steps:
       - name: Checkout
         uses: actions/checkout@v7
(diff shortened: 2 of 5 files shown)


alxjrvs (commented):
I read the diff and the full post-change state of `tools/check-observability.ts`, `.github/workflows/e2e-nightly.yml`, and `ci.yml`. The workflow-bounding half is sound — I verified all 15 jobs in `ci.yml` now carry `timeout-minutes` (13 new + the two e2e jobs at 20), so the header comment's claim is accurate. Nothing in the diff addresses instructions to a reviewing agent; the prose is maintainer-facing history.

Findings, none blocking:

- `tools/check-observability.ts:224` — `PROBE_BUDGET_MS` is 120s for the *entire* run, shared across both apps and a fully serial chunk crawl, and it was not sized against a measured CI run: this same PR states two different crawl sizes (`.github/workflows/e2e-nightly.yml:255` says "~134 chunks", `tools/check-observability.ts:205` says "~43 chunks the two apps actually serve"). At 134 serial CDN round-trips a healthy production on a slow runner can exhaust 120s, which exits 1 and files the `nightly-e2e-failure` issue — the cry-wolf outcome this file exists to prevent. Reconcile the two counts and set the budget from an observed CI timing, or make the crawl concurrent.
- `tools/check-observability.ts:311` — when the budget expires mid-crawl, `reachableChunks` rethrows and discards every chunk body it already downloaded, so a run that had already found the DSN and SDK in chunk 3 of 134 still reports a probe failure. Returning the partial map (and only failing if the accumulated bodies lack the DSN) would let an already-proven-healthy deploy pass.
- `tools/check-observability.ts:244` — `AbortSignal.timeout(REQUEST_TIMEOUT_MS)` is created inside `fetchWithRetry` but its 15s clock also covers the `.text()` body read that happens *outside* it (`:306`, `:352`). In `reachableChunks` an abort during that body read is caught by the generic `catch` and swallowed as `continue` ("chunk missing"), so a slow-but-successful chunk becomes invisible and can contribute to the "Error tracking is DARK in production" verdict at `:404`. Read the body inside the retry loop so a timed-out read is retried rather than silently dropped.
- `.github/workflows/e2e-nightly.yml:377` — the new `ProbeBudgetExhausted` distinction ("this is a PROBE failure, not evidence the deploy is dark") exists only in the job log; the tracking issue body still renders `observability-probe: failure`, identical to a real dark-production result. The alarm channel the diff is trying to keep honest cannot tell the two apart — e.g. a distinct exit code surfaced into the issue body would close that gap.

PR #753, opened 2026-08-13
feat(bot): fold the live sheet into /su sheet, and unbreak its links

alxjrvs:
`/su sheet` rendered three fields — HP/AP/Class, SP/Heat/Chassis — with `classRef` and `chassisRef` printed as raw slugs. It now renders the live sheet region-for-region, because the two turn out to be the same structure:

| Live sheet | Embed |
| --- | --- |
| identity band fields | description |
| vitals rail | inline fields |
| section slab + count | one full-width field, count in the field **name** |
| `ReferenceEntityCard` | one linked line |
| sheet accent | colour strip |
| reserved image seat | thumbnail |

Slugs resolve to the names the book prints, linked to salvageunion.io — the bot already had the whole dataset in memory, so this is a map lookup. An unknown slug falls back to the slug rather than vanishing: the embed must not disagree with the app about what a player owns.

Abilities group by ability tree, as the sheet groups them under its dashed sub-slabs. Faithfulness first, field-cap fix second — a Salvager may take 12 abilities, and 12 worst-case linked names exceed Discord's 1024-per-field limit.

## Three things landed alongside it

**The crawler is openable.** `botClient.sheet` typed `table` as `'pilots' | 'mechs'`, so the crawler sat on the crew board and could be opened nowhere. It is communal (no `ownerId` column at all), so it reports no owner rather than an absent one. `normalizeId` still keeps the widened union from reaching the Mediator's NPC tray — there is now a test that says so explicitly.

**The deep links were broken for everyone but the owner.** Both `/su crew` and `/su sheet` emitted `/sheet/<kind>/<appId>`, which resolves out of the **clicker's** IndexedDB. A crewmate does not have that entity locally, so every link the bot handed them opened a page with nothing on it — and nothing errored. They now use `/games/<gameId>/view/<kind>/<convexId>`, the read-only Game view, which is addressed by row id precisely because the viewer has no local copy. `sheet` gained `gameId` to make that addressable. The shelf keeps the local route, where the reader *is* the owner.

A side effect worth knowing: an **unclaimed** entity is now linkable, where it previously rendered bare. It exists server-side whether or not anyone has claimed it into a browser, so there is a real page to open.

**Embed limits were not enforced at all.** `gameEmbed.ts` declared its own limits omitting `footer` and `total`, so nothing checked the 6000-character ceiling — and an oversized embed is rejected by Discord with a 400, which fails the reply rather than degrading it. Harmless at three fields; not harmless now. `format.ts` now owns the shared caps, `stripDanglingLink` (lifted from `lookupEmbed.ts`, since collection fields are lists of links and truncating one mid-URL ships visible broken markdown), and a field-shedding `enforceEmbedLimits` applied once in `toEmbed` so no builder can forget.

## Measured, not estimated

Assembled against the real dataset and counted: a fully-linked pilot embed is ~2000 of 6000 characters, mech ~1000, crawler ~1200. So the 6000 total is **not** the binding constraint — the 1024-per-field cap is, which is what the tree grouping addresses.

## Note for review

Per-sheet accents are a deliberate amendment to the "colour carries one meaning only" rule in `docs/architecture/discord-bot-game-client.md` §7 rule 4. The values are the design system's own `--color-sheet-*` tokens rather than a bot palette, critical state still wins wherever both apply, and every non-sheet embed is still rust. The rule and the delivery-phases section are both updated in this PR.

Slot fractions (`6/10 slots`) are deliberately **not** rendered — deriving used-slots needs per-item slot costs, and a count is honest where a wrong fraction would not be.

## Verification

- `bun run test` — full canonical suite, 0 failures
- `bun run check:fast` — lint, validate:all, knip, typecheck across all six workspaces, all clean
- 39 new tests: embed-limit enforcement (incl. the mid-markdown-link truncation case), sheet rendering against real SRD data, and server-side coverage for the crawler widening

`check:all` shows an intermittent `SNAPSHOT FAILED` from `apps/srd`; it is a pre-existing race between `bun run test` and `check:srd-output` both building `dist` concurrently, reproduces without this diff, and this PR touches no `apps/srd` file.


alxjrvs (commented):
Reviewed against the surrounding code (ITUN schemas, the reference ORM's ref-resolution rules, and the Convex `botClient` path). The Convex-side widening to `crawlers` is sound — `normalizeId` keeps the NPC tray closed, and the `gameId` addition plus the `/games/$gameId/view/$kind/$entityId` route match. The defects are all in the new crawler rendering and in the limit-enforcement seam.

**Findings**

- `apps/discord-bot/src/gameEmbed.ts:845` — **BLOCKING.** `refName`/`refArtwork`/`collectionLines` resolve refs with `findEntityBySlug` (slug-only), but crawler refs were *never* slug-migrated: `crawler.type` is written as the SRD **id** (`CrawlerTypeStep` → `onSelect(type.id)`, and `data/crawlers.json` ids are UUIDs) and `crawler.systems` likewise (`SystemsList` → `toggle(system.id)`; migration `6-mech-refs-to-slugs.ts` covers mechs and mechPatterns only). So a real crawler's `/su sheet` renders its identity line as `*8bffb508-8c8f-418d-b6ce-f24f7266e41b · Tech Level tech-3*`, every Armament line as a bare UUID, and no thumbnail. `packages/salvageunion-reference/lib/rules/resolveRefs.ts:88` documents this exact trap. Fix by resolving through `resolveRef`/`resolveCrawlerRef` (id → name → slug) from `salvageunion-reference/rules`, which the bot already imports from. The new tests miss it because `sheetEmbed.test.ts:481` feeds bay slugs and no `type` at all.
- `apps/discord-bot/src/gameEmbed.ts:812` — `techLevel` is stored as the slug `tech-3` (`crawlerFormState.ts:168` writes `` `tech-${n}` ``), so the embed prints "Tech Level tech-3" in both the field and the subtitle; `sheetEmbed.test.ts:479` uses `techLevel: '3'`, a value the app never writes, so the test asserts a shape that cannot occur.
- `apps/discord-bot/src/gameEmbed.ts:805` — crawler max SP is read from `maxSpOverride` alone, but the real derivation is `crawlerMaxSP` (tech-level `structurePoints` + type bonus + `maxSpModifier`), exported from `salvageunion-reference/rules` and used everywhere else. With no override — the normal case — the field degrades to a bare number, and since absent `currentSP` means *full* (schema comment, `crawler.ts:155`) an undamaged crawler renders `SP —` while pilots and mechs get gauges. This is the same "derive maxima via the package" rule `pilotStats`/`mechStats` follow two functions above.
- `apps/discord-bot/src/gameEmbed.ts:505` — `buildCrewEmbed` (and `buildShelfEmbed:390`) pre-truncate their link lists with the plain `truncate` at exactly `LIMIT.fieldValue`, so the value arriving at `enforceElementLimits` is never `> fieldValue` and the new link-aware `stripDanglingLink` pass (`format.ts:137`) never fires. A crew field cut mid-`[name](url)` still ships broken markdown — and this PR makes it likelier by swapping the short `/sheet/pilot/<appId>` for the longer `/games/<gameId>/view/pilot/<convexId>`.
- `apps/discord-bot/src/format.ts:138` — `stripDanglingLink(truncate(...))` can return `''` (it drops back to before the last `[`), and Discord rejects a field with an empty `value` with a 400, which is the failure mode this whole helper exists to prevent. Reachable when a single collection line exceeds 1024 chars and opens with `[` — bodies are `v.any()` and equipment entries are unvalidated strings. Guard the post-strip result before writing it back.
- `apps/discord-bot/src/gameEmbed.ts:1128` — `motto`/`quirk` and unresolved slugs are interpolated into the description and field values verbatim, and Discord renders markdown there, so an entity's owner can plant a masked link (`[Free scrap](https://…)`) that every crewmate reading `/su sheet` sees as bot-authored text. Low severity and same-table scope, but these are the first user-authored free-text fields the bot renders.

One doc nit, not a defect: `apps/discord-bot/src/itun/types.ts:62`'s `OwnedEntity.appId` comment still says "Deep links use THIS, never `id`", which the crew board no longer does.

alxjrvs (commented):
Reviewed the full diff plus the surrounding code (`gameEmbed.ts`, `format.ts`, `botClient.ts`, the itun crawler schema/wizard, and the `/games/$gameId/view/$kind/$entityId` route). The link repointing, the `gameId` addition, the `crawlers` widening (with `normalizeId` still gating it), and the shared limit enforcement all check out. No instructions addressed to a reviewing agent appear in the diff.

**Findings**

- `apps/discord-bot/src/gameEmbed.ts:823` — **BLOCKING.** Crawler refs are not slugs: `crawlerFormState.ts:205` writes `bayRef: bay.id`, `type` is documented and stored as the SRD id ("Chosen crawler-type ref (SRD id)"), and `CrawlerBuilder` stores system **ids** — all UUIDs in `data/crawler-bays.json` / `data/crawlers.json`. `findEntityBySlug` only consults the slug index, so for a real crawler the Bays field renders `233d7930-1c4d-475d-9ea8-c88a1c70350c` per line, Armament (`:828`) does the same, the subtitle (`:845`) prints a UUID, and `refArtwork` (`:852`) always returns undefined. The app itself resolves these through `resolveCrawlerBay`/`resolveCrawlerType`/`resolveCrawlerSystem`, all of which wrap `resolveRef` (id → name → slug) from `salvageunion-reference/rules` — the subpath this file already imports. `sheetEmbed.test.ts:481` passes only because its fixture uses `'command-bay'`, a value no stored crawler holds, so the test asserts the opposite of production. Pilot and mech refs are genuinely slugs and are unaffected.
- `apps/discord-bot/src/gameEmbed.ts:812` — `techLevel` is a slug (`CrawlerSchema`: `techLevel: \`tech-${n}\``), so the Tech Level field renders `tech-3` and the subtitle reads `Tech Level tech-3`; the new test's `techLevel: '3'` fixture is not a shape the app writes.
- `apps/discord-bot/src/gameEmbed.ts:805` — crawler max SP is derivable via `crawlerMaxSP` (tech level + type bonus + `maxSpModifier`, already exported from `salvageunion-reference/rules`), but only `maxSpOverride` is read, so a normal crawler gets a bare number instead of a gauge; and unlike `pilotStats`/`mechStats`, an absent `currentSP` is not treated as full, so an undamaged crawler shows `—`.
- `apps/discord-bot/src/format.ts:718` — `stripDanglingLink` is applied only when a value exceeds `fieldValue`, but `buildCrewEmbed:504` and `buildShelfEmbed` already `truncate(..., LIMIT.fieldValue)` before `toEmbed` runs, so those values arrive at ≤1024 and skip the fix entirely; with the longer `/games/<id>/view/<kind>/<convexId>` URLs this change introduces, a crew bucket cut mid-link now renders a literal `[Name](https://…`.

PR #755, opened 2026-08-13
feat(itun): public read-only sheets, and a Discord link that isn't a 404

alxjrvs:
A shareable link to a character had two shapes and neither fit. The **Game view** needs an account and a membership, which a Discord channel routinely has neither of. An **ADR-004 snapshot** needs nobody — but it is *frozen* and has to be minted per share, which is wrong on both counts for a link that lives somewhere durable and wants to be current.

So: a public, read-only rendering of the **live** sheet at `/p/:kind/:appId`. No publishing step, no blob, no expiry to manage.

**[ADR-032](docs/adrs/ADR-032-public-read-only-sheets.md) records the decision**, because the boundary was already drawn twice and this crosses it: ADR-004 says anything needing mutable shared state belongs in a new ADR, and ADR-030's consequences list "any change to snapshot sharing" among what it deliberately did *not* adopt.

## Opt-in, which is the one real decision here

ADR-030 §5 says visibility begins at membership. Rather than make row ids the capability — they are not secrets, they travel in crew payloads and Game-view URLs — a nullable `publicRead` column sits beside `gameId` and `ownerId`. **Absent means private, so every existing row is unchanged.**

It costs the owner one action *ever* rather than one per share, and switching it off revokes everywhere at once, because the URL is derived rather than minted.

The **owner** decides for a pilot or mech, with no Mediator override: making somebody's character world-readable is the clearest possible case of a thing that is theirs to decide. The **crawler** has no `ownerId` at all, so it follows §5a and is the table runner's act.

## No new renderer

`frozenSheet.ts` already parses a bare entity and wraps it in a store whose every write throws. The snapshot page and the Game crew view are its two consumers; this is the third. Three consumers, one renderer.

## A private sheet is never advertised, and never posted

Two halves of one rule, both held by `sheetVisibility.test.ts`:

- The bot renders the `/p/…` link **only** when the server reports `publicRead: true`. A link to an unpublished sheet renders "this sheet isn't available", so offering one would hand the reader a 404.
- **`/su sheet` stays ephemeral**, including for a published sheet. The owner chose to publish a page; they did not choose to have it posted into this channel by whoever ran the command.

That test exists because "make /su sheet public so the table can see it" is a plausible one-line change that would break both silently.

## A defect the frozen path structurally cannot fix

A published mech snapshot has to carry `context.pilotAbilities` alongside the entity, because a bare frozen mech cannot see the pilot flying it and computes Max SP and Cargo without their contributions (ADR-029). A live query runs with the whole `softLinks` graph in reach.

## UI

The Share screen gains a panel beside the snapshot one, so a player choosing between *frozen* and *live* sees both. It mounts only when connected — and the gate is in the **parent**, because the panel calls Convex hooks, which cannot be called conditionally and throw with no provider. That is exactly the Solo case, and `.claude/rules/tanstack-query-hooks.md` names the pattern.

The copy is deliberately blunt: publishing makes bio and appearance text world-readable, and the person deciding should be told rather than left to infer it.

## Note for review

`convex/_generated/api.d.ts` is **edited by hand** for the new module. Codegen requires a deployment this environment cannot reach; the file is a mechanical module enumeration and this edit is byte-identical to what `convex codegen` emits. Worth a glance. `routeTree.gen.ts` is regenerated normally by the Vite plugin.

## Verification

- `bun run test` — full canonical suite, 0 failures
- `bun run check:fast` — lint, validate:all, knip, typecheck across all six workspaces, clean
- New Convex tests cover the refusals, which are most of the surface: unpublished reads as null, non-existent is indistinguishable from private, a stranger cannot publish, an anonymous caller cannot publish, an unclaimed entity cannot be published at all, unpublishing takes effect immediately, and no `kind` argument reaches `encounterNpcs`


alxjrvs (commented):
**Review — ADR-032 public read-only sheets**

The Convex surface is the careful part of this change and it holds up: `get` is null-for-both-cases, the kind union is the whole table surface, `setPublic` is owner-gated with a table-runner branch for the communal crawler, and `publicRead` survives mirrored writes because every entity mirror uses `db.patch`, never `replace`. The bot side is fine too — `publicSheetUrl` is null-safe on an absent flag, and the ephemerality tests match the real `fakeExecute`/`respondWithItun` contracts. One blocking defect on the client route, plus three smaller ones.

- **BLOCKING** `apps/itun/src/routes/p.$kind.$appId.tsx:56` — `useQuery` is called unconditionally with no `isConvexConfigured` gate, so in a build with no `VITE_CONVEX_URL` the route throws `Could not find Convex client!` and renders `RootErrorComponent` instead of the "This sheet isn't available" page. `'skip'` does not help: `useQuery` calls `useQueries` on every path (`convex/dist/esm/react/use_queries.js:7`) and `AppConvexProvider.tsx:22` deliberately mounts no provider in that build. This is not a hypothetical config — `apps/itun/netlify.toml:58` makes **every deploy preview and branch deploy** a Solo build, as is a checkout that has never run `bunx convex dev` and the unit-test env. `PublicSheetPanel` documents this exact rule and is gated by its parent (`ShareSnapshotScreen.tsx:320`); the route needs the same treatment — an inner component rendered only when `isConvexConfigured`, falling back to `<NotAvailable />`. A component test rendering the route in the (Solo) test env would have caught it; none was added.
- `apps/itun/src/components/sheet/PublicSheet.tsx:61` — `Sheet` is rendered without `pilotAbilities`, and the frozen store carries no pilot and no soft links, so a published **mech** computes Max SP and Cargo without its pilot's contributions (ADR-029) and reads lower than the same mech on its owner's sheet. `SnapshotSheet.tsx:93` passes `context.pilotAbilities` for precisely this reason, and ADR-032's "Serving live fixes a defect the frozen path cannot" claims the live path resolves the piloting pilot — but `publicSheet.get` returns the bare body and never touches `softLinks`, so the live path is currently *worse* than the snapshot it is offered as an upgrade to.
- `apps/itun/convex/publicSheet.ts:158` — publishing an unparseable pilot/mech goes through `parseBody`, which throws a plain `Error` (`convex/model/entities.ts:73`); Convex redacts that to `Server Error`, so `isServerRefusal` is false at `PublicSheetPanel.tsx:61` and the owner sees "That could not be saved. Try again." with no reason. The crawler branch three lines above throws `ConvexError` and does surface its reason, so the module's own promise — fail "where the owner is standing and can see it" — holds for one of the three kinds. Wrap the `parseBody` call, or throw `ConvexError` here for both branches.
- `apps/itun/convex/publicSheet.ts:108` — dead guard: `crawlers.gameId` is `v.id('games')` (`convex/schema.ts:334`), never null, so after the `'ownerId' in row` narrowing this branch cannot fire and `requireTableRunner` is always reached. Harmless at runtime, but comparing `Id<'games'>` to `null` under strict TS is the "no overlap" comparison (TS2367) — if `bun run typecheck` is red on this file, that line is why.

alxjrvs (commented):
I read the full diff and checked the surrounding code (Convex permissions/soft-link mirroring, the bot embed path, the Share screen layout, the frozen-sheet renderer, and the connection gating).

Nothing in the diff attempts to instruct a reviewing agent — the comments and ADR prose are addressed to human maintainers.

## Findings

- `apps/itun/src/components/sheet/ShareSnapshotScreen.tsx:320` — `PublicSheetPanel` is inserted as a **third direct child** of `<main className="grid … lg:grid-cols-[1fr_360px]">`, between the preview `Panel` and the `<aside>`. On `lg` the auto-placement becomes preview → (r1,c1), public panel → (r1,c2, the 360px rail), aside → (r2,c1), so the existing "Anonymous link / Shared links / QR / Print" rail is pushed out of the sidebar into the wide column and the rail below the new panel is empty. Only connected users see this, so a Solo-rendered screenshot won't show it. Non-blocking (visual), but it changes an existing screen's layout unintentionally — the panel needs to live inside the left column (or carry an explicit `lg:col-start-1`).

- `apps/itun/convex/publicSheet.ts:119` — `pilotAbilitiesForMech` follows *any* `mech-to-pilot` soft link out of the published mech and reads the target pilot's `body.abilities` with no check that the link's `to` end is related to the caller. `entities.upsertSoftLink` (`apps/itun/convex/entities.ts:882-901`) deliberately validates only the `from` anchor ("wiring your own mech to a crewmate's pilot is your business"), and `to.id` is a free-form string. So any signed-in user can point a soft link from their own published mech at an arbitrary pilot `appId` — no shared Game, no membership — and the unauthenticated `/p/mech/<their-mech>` query then returns that pilot's ability slugs to the world. The module comment reasons about this ("discloses no pilot… only a set of ability slugs already implied by the mech's own numbers"), but that argument holds only when the linked pilot is actually the mech's pilot, which nothing here enforces. Exploiting it needs the victim's `appId`, which is a UUID but is also the id that appears in the owner's own `/sheet/pilot/<appId>` URL. Non-blocking given the narrow payload and the UUID precondition; the fix is to require the linked pilot to share owner (or Game) with the mech before resolving abilities.

Everything else checked out: `publicRead` is `v.optional` so existing rows default private; every mirror write is a `patch`, so publishing survives later edits; `botClient.sheet` coerces a missing flag with `=== true`; `kindOf` does map `pilots → pilot` (the bot test's expectation is right); `parseBody`'s narrowing to `'pilots' | 'mechs'` after the crawler branch is sound; the `ConvexError` re-throw is genuinely load-bearing given Convex's redaction; and the `isConvexConfigured` gate in `p.$kind.$appId.tsx` plus the route test correctly cover the provider-free Solo build (`useConnection` also defaults to `SOLO_STATE` outside a provider, so existing Share-screen tests won't break).

alxjrvs (commented):
I read the full diff and checked the surrounding code (`convex/publicSheet.ts`, `convex/ownership.ts`, `convex/entities.ts`, `gameEmbed.ts`, `format.ts`, `ShareSnapshotScreen.tsx`).

Nothing in the diff attempts to address or instruct a reviewing agent — the prose is ordinary in-repo documentation.

## Findings

**BLOCKING**

- `apps/itun/convex/publicSheet.ts:187` — `assertMayPublish` refuses *any* `setPublic` call on an entity whose `ownerId` is `null`, including `isPublic: false`. So a published sheet that later becomes unclaimed can never be un-published by anyone, and keeps serving to anonymous readers. This is reachable through normal play: `ownership.release` (`ownership.ts:214`) and the leave-Game sweep (`ownership.ts:252`) both patch `ownerId: null` on entities that are still in the Game. Concretely — Alice publishes her pilot, then leaves the table; her pilot is now unclaimed and still world-readable at `/p/pilot/<appId>`, and when she presses "Stop sharing" on the Share screen she gets `That entity is unclaimed — it must be assigned before it can be shared` while the URL keeps working. The only escape is a Mediator reassigning the row to a real owner who then unpublishes. This directly falsifies the property ADR-032 §6 and the panel copy both promise ("Turning it off takes effect immediately, everywhere"). The unclaimed branch should be skipped when `args.isPublic === false` (mirroring `assertMayWrite` is right for edits, where the dead-end is harmless; it is not right for revocation).

**Non-blocking**

- `apps/itun/convex/publicSheet.ts:156` — the `sameGame` arm of the pilot-link check lets a crewmate expose another player's pilot ability slugs to *anonymous* readers. `upsertSoftLink` validates only the `from` anchor and `to.id` is free-form, so a member can point `mech-to-pilot` from their own mech at a crewmate's `appId`, publish the mech, and `get` returns that pilot's `abilities` array to anyone with the link. The cross-game case is caught (and tested); the same-Game case is not. In-Game members can already read each other's sheets, but this republishes that data outside the membership boundary without the pilot's owner consenting — the exact consent argument the module header makes. Tightening to "the pilot's owner is the mech's owner, or the pilot is unclaimed in the same Game" would keep the legitimate cases.

- `apps/itun/src/components/sheet/ShareSnapshotScreen.tsx:332` — `lg:col-start-1` does not "leave the rail alone" as the comment claims. Under grid sparse auto-placement the cursor is left at row 2 / column 1 after the pinned panel, so the following `<aside>` is placed at row 2 / column 2 rather than row 1 — the rail's top cell goes empty and the anonymous-link/QR/print column drops down beside the public-sheet panel. Pinning the aside as well (`lg:col-start-2 lg:row-start-1`) is the fix. I could not measure this in a browser, so treat it as a layout claim to verify at 1440px rather than as confirmed.

- `apps/discord-bot/src/gameEmbed.ts:916` — the Share field is `push`ed last, and `enforceEmbedLimits` sheds fields from the end (`format.ts:165`). On a large sheet (a Salvager with many ability trees, a fully-fitted crawler) the public link is the first thing dropped, and it is counted in the "N sections omitted" notice as if it were a sheet section. Inserting it at the front of `fields`, or exempting it from shedding, would make the one link that survives account-free access the one that isn't trimmed.

alxjrvs (commented):
I read the diff, then checked every hunk against the current branch. **Important context first:** the diff you handed me is stale relative to `public-live-sheet` HEAD — it predates commits `dec7c164` and `8eac98bb`. Three real defects that are visible *in the diff* are already fixed in the working tree, so I am not reporting them as findings:

- `assertMayPublish` refusing on `ownerId === null` in both directions — a pilot released via `ownership.release` or the `leaveGame` sweep would have stayed world-readable with "Stop sharing" permanently refused. Fixed at `apps/itun/convex/publicSheet.ts:197-210` (unpublish now always allowed).
- `pilotAbilitiesForMech`'s `sameGame` branch — a Game member could point a `mech-to-pilot` link at a crewmate's pilot, publish their own mech, and hand that pilot's abilities to anonymous readers. Narrowed to same-owner-or-unclaimed at `publicSheet.ts:161-166`.
- The Share screen grid — `lg:col-start-1` alone left the aside on row two; the aside is now pinned too (`ShareSnapshotScreen.tsx:333,345`).

What still holds, all non-blocking:

- `apps/discord-bot/src/itun/types.ts:129` — `publicRead` is typed as a required `boolean` on a wire payload, but an ITUN deployment running an older `botClient` sends no such key; the file's own convention two fields up (`gameId`'s comment, and `gameSheetUrl`'s tolerant parsing) is to type network fields as absent-able. Runtime is safe (`publicSheetUrl` treats falsy as private), but the new test has to write `undefined as unknown as boolean` to express the case the type says cannot happen.
- `apps/itun/src/components/sheet/ShareSnapshotScreen.tsx:320` — the panel mounts only in `connected`, so in `disconnected` (signed in, offline) and during `connecting` the Share screen shows the snapshot panel alone with no indication that the sheet is currently public. Revocation being unavailable offline is inherent; silently hiding the *fact of publication* is not, and this is the one screen that would tell the player.
- `apps/itun/convex/publicSheet.ts:164` — `unclaimedInSameGame` still widens past "opt-in per entity": a crewmate publishing their mech makes an unclaimed pilot's ability slugs world-readable with nobody having opted in. The reasoning in the comment (no owner to ask, already visible to the table) is defensible, but it is a second path to public that ADR-032 §1 does not describe.
- `docs/adrs/ADR-032-public-read-only-sheets.md:57` — the decision says publishing "is `assertMayWrite` for a pilot or mech". The implementation is a separate `assertMayPublish` that now deliberately diverges (unpublish is permitted on an unclaimed row, which `assertMayWrite` would refuse). Since ADRs here are cited as authority later, the §6 revocation promise and this clause should say what the code does.

Everything else I checked out clean: the schema columns default absent-means-private and every write path in `convex/` is a `patch`, so publishing survives entity edits; `game.destroy` deletes unclaimed rows and crawlers; the unauthenticated `get` cannot reach `encounterNpcs`; the Solo gating on `/p/$kind/$appId` is correct (`useQuery` genuinely throws without a provider, and `PublicSheetView` calls no hooks before its guards); the Netlify SPA fallback covers `/p/*`; and the bot's `kindOf` maps `pilots → pilot`, which the new test asserts.

alxjrvs (commented):
I read the full diff and checked it against the surrounding code (`gameEmbed.ts`/`format.ts` shedding order, `convex/schema.ts`, `model/permissions.ts`, `model/entities.ts`, `ownership.ts`, `frozenSheet.ts`/`Sheet.tsx`, `connectionContext.ts`, `knip.json`).

The core of the change holds up. The unauthenticated `get` refuses by default (`row.publicRead !== true`), not-found and not-public are the same answer, the kind union genuinely cannot reach `encounterNpcs`, `assertMayPublish` has no path that lets a non-owner turn publishing *on*, and `pilotAbilitiesForMech`'s ownership guard does close the `upsertSoftLink`-only-validates-`from` hole the comment describes. The Share-field insertion really does survive both `enforceElementLimits`' 25-field slice and `enforceEmbedLimits`' shed-from-the-end, `kindOf` handles `pilots` correctly, and no mirrored write uses `db.replace`, so `publicRead` isn't silently cleared by ordinary edits. No text in the diff is addressed to a reviewing agent.

Findings, none blocking:

- `apps/itun/convex/publicSheet.ts:265` — `publicRead` is a plain column that no ownership mutation touches, so an entity published by its owner stays world-readable across `ownership.release`, the leave-Game sweep, `ownership.claim` and a Mediator's `ownership.assign` (`convex/ownership.ts:119,162,214,252` all patch `ownerId` alone); a player who claims or is assigned a previously-published pilot silently inherits a public URL nobody asked them about, which is the one thing ADR-032's consent argument says must be the owner's decision. Clearing `publicRead` on any owner change, or surfacing it on claim, would close it.
- `apps/itun/convex/publicSheet.ts:206` — the un-publish exemption is gated only by `requireUser`, so *any* signed-in account that knows an appId can flip `publicRead` off on any unclaimed pilot or mech, including one in a Game they are not a member of; it can only reduce exposure, but it is an unauthenticated-adjacent write by a stranger and the ADR's justification (the owner who walked away) would be satisfied by `requireMember(row.gameId)` for game-held rows.
- `apps/itun/src/components/sheet/ShareSnapshotScreen.tsx:345` — pinning the aside to `lg:row-start-1` while the new panel takes `lg:row-start-2` makes row 1 as tall as the aside (four panels), so on `lg` the "Public sheet" panel renders below the *entire* rail with a large blank gap under the snapshot panel whose copy points at it as "below"; `lg:row-span-2` on the aside (or wrapping the two left-column panels in one grid child) gives the intended stacking without the gap.
- `apps/itun/src/components/sheet/PublicSheetPanel.tsx:55` — both `publicSheet.get` and `setPublic` resolve an appId to the *oldest* row, so where a duplicate appId exists across accounts (the legacy fixed-id Starter Set case `apps/itun/CLAUDE.md` documents and `maintenance.dedupeAppIds` repairs), the later account's panel reports *their* entity as published and shows a URL that actually serves the other account's sheet, while "Stop sharing" throws "You cannot publish another player's entity".

PR #758, opened 2026-08-14
fix(component-lib): drop the card fold chevron's width-keyed tap floor

alxjrvs:
## The bug

On narrow screens the fold ("enlarge") chevron on entity cards in the live sheets rendered as an oversized white square that hung down into the card header and covered the TL/SV stat cells.

## Cause

The icon-only rail control carried `min-h-11 min-w-11 sm:min-h-0 sm:min-w-0` — a **width**-keyed duplicate of the **pointer**-keyed touch floor `theme.css` already applies to every `button` under `@media (pointer: coarse)`.

That rule is unlayered, so it beats any Tailwind utility (those live in `@layer utilities`). Two consequences, and the second is the bug:

- Where there **is** a touch pointer, the utility was redundant — the 44px box survives with or without it, at every width.
- Where there is **not** — a narrow desktop window, a responsive-mode preview, a small laptop — it applied a touch floor to a mouse. That is the one case the pointer rule deliberately skips, and the only case the class could actually change.

So all it ever did was distort geometry for pointers that never needed it. This button renders inside `CardControlRail`, an **absolute** row centred on the card's top frame line, so a 44px box grew the rail with it.

## Measured (Playwright, real Ladle render, 390px)

| | chevron | rail | status badge |
|---|---|---|---|
| `hasTouch=false` **before** | 44×44 | 44 | 22 |
| `hasTouch=false` **after** | **24×20** | **22** | 22 |
| `hasTouch=true` before / after | 44×44 | 44 | 44 |

The rail's premise is that every cell is the same stamp height, so a 44px square beside a 22px status badge read as broken. After the fix the rail is a uniform 22px and the TL/SV cells are clear.

Coarse-pointer output is **byte-identical to `main`** — `theme.css` governs there either way, so there is no tap-target regression.

## Note on the earlier revisions of this PR

The first two commits added a `::before` hit area and a `data-tap-floor` opt-out in `theme.css`. The automated review's blocking finding was correct and invalidated the measurement they rested on: I had measured by viewport width only, which never matches `pointer: coarse`.

Both are reverted. The opt-out was also verified to achieve nothing on touch: `ControlButtons` renders a `flex gap-1` row whose default `align-items: stretch` stretches children to the rail's line height, and that height is set by the status badge — itself still floored at 44px. Exempting one button while its neighbour sets the line is a no-op, and would have shipped a global stylesheet change for it.

## Known, not addressed here

On a **real touch device** the whole rail is 44px tall and still overlaps the TL/SV stats. That is pre-existing and unchanged by this PR. Fixing it means exempting every rail cell from the global floor and giving each its own hit area — trading tap-target size for rail geometry across `StatusBadge`, `CountStepper` and the worded controls. That is a design call, not a bug in this diff.

Same root cause, also unchanged: the section chevron (`SheetSectionSlab`, declared `size-7`) measures 28×28 on a fine pointer but **44×44** on a coarse one, contradicting its own documented size contract.

## Checks

`bun run check:all` exits 0 — 5,064 tests pass / 0 fail, plus lint, knip, format, styling, audit and the srd output snapshot gate.

https://claude.ai/code/session_016YGBUwi4LZ4bs3FzzgRr2c

### DIFF
--- a/packages/component-lib/src/components/shared/ControlButtons.tsx
+++ b/packages/component-lib/src/components/shared/ControlButtons.tsx
@@ -103,39 +103,32 @@ function ControlButton({
   // replaces, so an icon button and a worded one are interchangeable in the
   // row.
   //
-  // The coarse-pointer tap target is an invisible ::before, NOT a bigger box.
-  // This used to be `min-h-11 min-w-11 sm:min-h-0 sm:min-w-0` — the repo-wide
-  // 44px touch floor, which is correct everywhere it sits in normal flow
-  // (SmallButtons, the Stat steppers, Field). Here it is not: this button
-  // renders inside `CardControlRail`, an ABSOLUTE row centred on the card's top
-  // frame line, so growing the box grows the rail with it. Measured at 390px
-  // the rail went 22px → 44px tall and the button 24×20 → 44×44, which (a) hung
-  // 22px down into the header band, covering the TL/SV stat cells the rail is
-  // z-30 over, and (b) put a 44px square next to a 22px status badge in a row
-  // whose whole premise is that every cell is the same stamp height.
-  // Expanding the HIT AREA instead keeps the rail one uniform stamp row at
-  // every width while still giving a finger 44px of vertical reach.
+  // It carries NO tap-target floor of its own. It used to:
+  // `min-h-11 min-w-11 sm:min-h-0 sm:min-w-0`, which was a WIDTH-keyed
+  // duplicate of the POINTER-keyed floor `theme.css` already applies to every
+  // `button` under `@media (pointer: coarse)`. Two consequences, and the second
+  // is the bug:
   //
-  // Its size is bounded by the rail's own gaps, in BOTH axes — an invisible
-  // target that reaches past a neighbour steals that neighbour's taps, and one
-  // of the neighbours here is the `danger` remove:
+  //  - Where there IS a touch pointer, the utility was redundant. That rule is
+  //    unlayered, so it beats any Tailwind utility (those are in
+  //    `@layer utilities`) — the 44px box survives with or without this class,
+  //    at every viewport width.
+  //  - Where there is NOT — a narrow desktop window, a responsive-mode preview,
+  //    a small laptop — the utility applied a touch floor to a mouse. That is
+  //    the one case the pointer rule deliberately skips, and the only case this
+  //    class could actually change.
   //
-  // - width: siblings in this group sit 4px away (`flex gap-1`) and the rail's
-  //   other groups 6px (`gap-1.5`), so `w-7` — 28px on a 24px button — spends
-  //   2px per side and stops 2px short of the nearest edge.
-  // - height: the rail is `flex-wrap`, and its documented crowded case (status
-  //   seal + selection seal + count seal + three controls) wraps at exactly the
-  //   widths this pseudo is live at. Wrapped lines are 6px apart, and the line
-  //   is ~22px tall around a 20px button, so anything over 34px overhangs into
-  //   the line above and below. `h-8` (32px) stops ~1px short of the gap.
-  //   The former `min-h-11` had no such ceiling only because it grew the line
-  //   itself — the very thing that broke the geometry.
+  // So all it ever did was distort the geometry for pointers that never needed
+  // it. This button renders inside `CardControlRail`, an ABSOLUTE row centred
+  // on a card's top frame line, so a 44px box grew the rail with it: measured
+  // at 390px, rail 22px → 44px and button 24×20 → 44×44, which hung 22px down
+  // into the header band over the TL/SV stat cells the rail is z-30 above, and
+  // put a 44px square beside a 22px status badge in a row whose whole premise
+  // is that every cell is the same stamp height.
   //
-  // So the reach is 28×32, not the 44px of the repo-wide floor. That is the
-  // most this rail's geometry actually affords, and it still clears the 24×24
-  // minimum in WCAG 2.2 §2.5.8, which measures the target, not the ink. On a
-  // FOLDED card the whole surface is an expand target as well (`cardClick`), so
-  // the chevron is never the only way in.
+  // Measure this with touch emulation, not just a narrow viewport — the two
+  // disagree here, and a fine-pointer run alone will tell you a coarse-pointer
+  // change worked when it did nothing.
   //
   // A `danger` icon (the ✕ remove) is a RED stamp — filled, not outlined —
   // because it is the one control in the rail that destroys something, and as
@@ -148,8 +141,7 @@ function ControlButton({
       <button
         type="button"
         className={cn(
-          'relative inline-flex shrink-0 items-center justify-center border-2 transition-colors',
-          "before:absolute before:left-1/2 before:top-1/2 before:h-8 before:w-7 before:-translate-x-1/2 before:-translate-y-1/2 before:content-[''] sm:before:hidden",
+          'inline-flex shrink-0 items-center justify-center border-2 transition-colors',
           RUNG_INLINE_PADDING.mini,
           RUNG_TYPE.mini.label,
           isDisabled
--- a/packages/component-lib/src/components/shared/ControlButtons.tsx
+++ b/packages/component-lib/src/components/shared/ControlButtons.tsx
@@ -101,8 +101,27 @@ function ControlButton({
   // remove/swap/fold cluster. It is a STAMP like every other control in the
   // rail: same padding, same height, the glyph sized to the label line it
   // replaces, so an icon button and a worded one are interchangeable in the
-  // row. `min-h-11`/`min-w-11` still holds the 44px coarse-pointer tap target
-  // without changing the desktop size.
+  // row.
+  //
+  // The coarse-pointer tap target is an invisible ::before, NOT a bigger box.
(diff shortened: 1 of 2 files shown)


alxjrvs (commented):
Reviewed the diff (2 files, one behavioral change: the icon-only control's 44px coarse-pointer floor moves from `min-h-11 min-w-11` on the box to an invisible `::before`). The mechanism itself is sound — a transparent, sized pseudo-element is hit-testable, `<button>` supports generated content, the pseudo is out of flow so the rail keeps its stamp height, and nothing between the button and `CardControlRail`'s absolute row clips it (`ReferenceEntityCard`/`Card` both render the rail outside the `overflow-hidden` inner). No instructions-to-the-reviewer text in the diff.

Two findings, both about the geometry the new comment reasons through only horizontally:

- `packages/component-lib/src/components/shared/ControlButtons.tsx:138` — the hit box is 44px tall on a ~20px line, but `CardControlRail` is `flex-wrap` with `gap-1.5`, so when the rail wraps (its documented crowded case: status seal + selection seal + count seal + three controls, at exactly the <640px widths where the pseudo is live) each icon control's invisible box overhangs 12px into a 6px gap and covers ~6px of the cell on the line above and below. Taps at the edge of the status badge or selection seal are then routed to the icon control — one of which is the `danger` remove. The former `min-h-11` did not have this failure mode: it made the whole flex line 44px, so wrapped lines could not overlap. The comment's clearance argument (`w-7` leaves 2px against a 4px gap) covers only same-line siblings.
- `packages/component-lib/src/components/shared/ControlButtons.tsx:138` — the rail is `top-0 -translate-y-1/2`, so the upper half of the 44px box sits *outside* the card, in the gap above it. The height is unchanged from `main`, but it used to be a visible 44px button; it is now an unmarked band in which a tap fires the destructive remove with nothing on screen indicating a target. Worth either dropping to `h-9`/`h-10` (still ≥24px, keeps the overhang inside the gap) or accepting explicitly in the comment alongside the horizontal reasoning.

Not findings, but checked: `cn()`'s extended twMerge config does not touch `before:*`/`content-*` groups so nothing is stripped; Tailwind scans these as literal strings; `tools/a11y-scan.ts` runs at 1280px with no `wcag22` tags, so no target-size gate regresses; the `min-h-11` assertions in `apps/itun/.../mobile-responsive.test.tsx` target `InlineEditField` and the Share link, not this button.

alxjrvs (commented):
## Review

**`packages/component-lib/src/components/shared/ControlButtons.tsx:152` — BLOCKING.** The fix does not remove the 44px floor on the devices it targets. `packages/component-lib/src/styles/theme.css:465` carries a global, **unlayered** rule:

```css
@media (pointer: coarse) { button, [role='button'], … { min-height: 44px; min-width: 44px } }
```

`apps/itun/src/index.css:1-2` imports `tailwindcss` (which puts every utility in `@layer utilities`) and then imports `theme.css` with no layer, so that unlayered `button` rule wins over any Tailwind utility regardless of specificity — it beat the old `min-h-11`/`sm:min-h-0` too. Consequence: on a real touch device the rail button is still forced to 44×44, the rail is still ~44px tall, and it still hangs down over the TL/SV stat cells — exactly the defect the PR is fixing. The 22px→44px measurement quoted in the comment is reproducible only in a **fine-pointer** narrow viewport (DevTools responsive without touch emulation), which is the one case where a bigger tap target was never needed. So after this change: coarse pointers keep the broken geometry (the new `before:` box is inert there, being smaller than the 44px button), and fine pointers get an invisible 28×32 target no one needs. The real fix has to exclude these rail buttons from `theme.css:465` (e.g. scope that floor, or give the rail an opt-out class), and the block comment at lines 106-138 plus the new tests assert geometry that is false wherever `pointer: coarse` matches.

**`packages/component-lib/src/components/shared/ControlButtons.tsx:152` — non-blocking.** The width budget in the comment ("`w-7` … stops 2px short of the nearest edge") measures against the neighbour's *box*, not the neighbour's own expanded hit area. Two adjacent icon controls sit 4px apart (`flex gap-1` in `ControlButtons:268`) and each `::before` extends 2px per side, so the two invisible targets abut exactly at the gap midpoint with zero clearance. No overlap, so no tap is stolen outright, but the 4px dead gap between the fold chevron and the `danger` remove now resolves to one of the two buttons instead of falling through — a 2px mis-tap toward the remove destroys something where it previously did nothing. Worth narrowing to `w-6` (24px = the box) if that gap is meant to stay neutral.

Note on scope: `ControlButtons` also renders outside the rail (`packages/component-lib/src/components/dashboard/DisplayPanel.tsx:139`), where growing the box was harmless; the comment's justification only covers `CardControlRail`. No injected instructions to a reviewing agent were present in the diff.

alxjrvs (commented):
Reviewed the diff against the surrounding code. I verified the central claim: `packages/component-lib/src/styles/theme.css:465` applies `min-height/min-width: 44px` to every `button` under `@media (pointer: coarse)`, the file contains no `@layer` and is `@import`ed without `layer()` in both apps (`apps/srd/src/styles/global.css:2`, `apps/itun/src/index.css:2`), so it does beat `@layer utilities`. The removal is therefore a no-op on touch and only changes fine-pointer rendering, as the comment says. Nothing in `apps/` asserts on this floor, and the new token check passes (the button's classes are `inline-flex shrink-0 … border-2 px-1 py-0.5 text-badge …` — no `h-*`/`w-*` token). No injected instructions in the diff.

Findings:

- `packages/component-lib/src/components/shared/ControlButtons.tsx:144` — Non-blocking, but the headline: because the surviving 44px floor is pointer-keyed and unlayered, on a real phone this `<button>` is still forced to 44×44, so `CardControlRail` (`absolute … top-0 -translate-y-1/2`, `CardControlRail.tsx:70`) is still 44px tall and still hangs ~22px into the header band over the TL/SV cells. The branch is named `fix-mobile-control-size`; if the report came from an actual touch device rather than a narrow desktop window, this change does not fix it — the only thing it fixes is fine-pointer narrow viewports. The `theme.css` blanket rule is where a mobile fix would have to land.

- `packages/component-lib/src/components/shared/__tests__/ControlButtons.test.tsx:173` — The rationale for switching from a substring check to a token check cites "the coarse-pointer hit area is a `before:h-8` pseudo-element", but no `before:` utility and no `::before` hit area exists anywhere in `packages/component-lib/src` (only `theme.css:328`'s reset and this comment itself). It reads as a leftover from the intermediate fold-chevron approach that commit `ad6109e6` dropped, so the comment documents a mechanism that isn't in the tree — misleading for anyone who later tries to keep the assertion consistent with it. The assertion itself is fine; only the justification is wrong.

- `packages/component-lib/src/components/shared/ControlButtons.tsx:160` — Minor: with the floor gone, the fine-pointer icon control is ~24×20 CSS px (12px glyph + `px-1 py-0.5` + `border-2`), under WCAG 2.2 SC 2.5.8's 24×24 minimum. The rail's `gap-1.5` gives ~30px centre-to-centre, so the spacing exception plausibly covers it, and the repo's stated target is 2.1 AA — flagging only so the trade is a decision rather than a side effect.

alxjrvs (commented):
Read the diff and checked it against `theme.css`, `CardControlRail`, `sizing.ts`, and the sibling components that use the same pattern.

**What the change does:** drops `min-h-11 min-w-11 sm:min-h-0 sm:min-w-0` from the icon-only rail button, updates two tests, and adds a long explanatory comment. I verified the load-bearing premise: `packages/component-lib/src/styles/theme.css:465-474` does apply `min-height/min-width: 44px` to every `button` under `@media (pointer: coarse)`, and that file is `@import`ed unlayered by both apps and by Ladle, so it does beat `@layer utilities`. The reasoning in the comment is accurate.

Findings:

- `packages/component-lib/src/components/shared/ControlButtons.tsx:112` — On an actual touch device the geometry is unchanged: `theme.css`'s coarse-pointer rule still forces the button to 44×44, so `CardControlRail` (`absolute … top-0 -translate-y-1/2`, `CardControlRail.tsx:70`) is still 44px tall and still hangs ~22px down over the header's stat cells, and still sits beside a 22px status badge. The comment states this consequence for the *old* class but does not follow it through: the branch is named for a mobile control-size bug, and this change only alters fine-pointer/narrow-desktop rendering. If the reported symptom was seen on a phone, it is not fixed by this diff — the fix would have to live in the rail's layout (e.g. reserving the coarse-pointer height, or moving the rail out of the frame-line overlap) rather than in the button's class list. Non-blocking, but it is the thing I'd want confirmed before merging.
- `packages/component-lib/src/components/shared/__tests__/ControlButtons.test.tsx:177` — The new token filter `/^[hw]-\d/` does not match Tailwind's `size-*` utility, which sets width *and* height; since the package's own icon-button rung is spelled `size-9` (`chrome/buttonVariants.ts:85`), a regression that reintroduces exactly the fixed square this assertion guards against would pass. Matching `^(?:[hw]|size)-\d` would close it.
- `packages/component-lib/src/components/chrome/SmallButtons.tsx:24` — Still carries the identical `min-h-11 min-w-11 sm:min-h-0 sm:min-w-0` (as do `Stat.tsx:684,695` and `SheetSection.tsx:41`), which the new comment now declares to be always either redundant or harmful; leaving them means the package's code and its stated rule disagree, and the next reader has no way to tell which is current. Not a defect in this diff — worth either a follow-up or a sentence in the comment scoping the claim to the rail.

No security concerns, and no text in the diff addressed to a reviewing agent.

PR #759, opened 2026-08-14
fix(itun): stop share links needing five refreshes after a deploy

alxjrvs:
A snapshot link (`/s/:id`) routinely needed four or five refreshes before it rendered. Two independent faults compounded, and **neither was "the PWA" alone — removing the PWA would have fixed neither.**

## Fault 1 — a missing chunk came back as `200 text/html`, cached immutable for a year

Verified against production before this PR:

```
$ curl -D- https://intheunionnow.com/assets/index-DEADBEEF.js
HTTP/2 200
content-type: text/html; charset=UTF-8
cache-control: public,max-age=31536000,immutable
```

The `/*` SPA fallback swallowed rotated-away asset hashes, so a dynamic `import()` failed on MIME type — and the `/assets/*` header block then pinned that HTML into the HTTP cache under the chunk's URL, for a year, where no later deploy can dislodge it. This would bite any code-split SPA on this config, PWA or not.

Fixed with an `/assets/*` → 404 rule ahead of the fallback.

> **Reviewer note — the one thing to check here.** The rule deliberately has **no `force = true`**. Netlify consults the redirect table only after failing to find a file on disk, so as written it matches misses only. Forcing it would shadow `dist/assets/` and 404 *every* chunk in the build, i.e. take the site down. Verified on the deploy preview below.

## Fault 2 — the service worker deleted its own precache under a live page

`registerType: 'autoUpdate'` **force-assigns** `skipWaiting` + `clientsClaim` (an assignment in the plugin source, not a default — the `workbox` block cannot override it). A new worker therefore activated the moment it installed, claimed the page the user was already looking at, and ran `cleanupOutdatedCaches()`. That page then went on requesting chunk hashes from a build the server no longer had → Fault 1.

Now `registerType: 'prompt'`: the worker installs and **waits**. The swap happens when the user accepts (post `SKIP_WAITING`, reload on `controllerchange`) or when every tab has closed. Verified in the emitted `sw.js`: `clientsClaim` is gone entirely, and `skipWaiting` survives only behind the message guard.

At ~3 deploys/day the collision window was open most of the time, and share surfaces wore it worst — `/s/:id` and `/p/:kind/:appId` are opened cold, once, from a link, on a device whose worker is whatever build it last saw.

## Also in here

- **`chunkRecovery.ts`** — recovers a tab that was already mid-flight when a deploy landed, via Vite's `vite:preloadError`. 20s cooldown as the loop guard; a cooldown rather than a one-shot flag so a *second* deploy in a long-lived tab still recovers. Fails soft when `sessionStorage` throws (Safari private mode).
- **`/s/:id` now reports to Sentry.** Catching is exactly what stopped these reaching the global handler, so the most-reported broken surface in the app was structurally invisible — Sentry showed zero snapshot errors the whole time. A 404 stays unreported: a revoked snapshot is a designed outcome, not a fault.
- **`serviceWorker.register()` gets a `.catch`.** Its unhandled rejection is the pair of untitled `Error: Rejected` issues already in Sentry (ITUN-4 / ITUN-5).
- Fingerprints pinned on all three captures — the messages embed bundle hashes and HTTP statuses, so default grouping would shard one condition into a fresh issue per deploy.

## Verification

- `bun run check:all` green (lint, format, typecheck, full test suite, audit, tokens, styling, schema-drift, srd output snapshot).
- itun suite 1868 pass / 0 fail; 15 new tests across `chunkRecovery` and `watchForUpdate`.
- Emitted `sw.js` inspected directly for the `clientsClaim` / `skipWaiting` change rather than assumed.
- Deploy-preview curl checks for the redirect are in a comment below.

## Not fixed here

The publish endpoint returns `url: /api/snapshots/${id}` while `PublishResult` documents it as the share page `/s/${id}`. The share UI ignores the field and builds `${origin}/s/${id}` itself, so it is a stale doc comment rather than a live bug — left alone to keep this PR to the reported fault.

https://claude.ai/code/session_01FoaHi5zfFQxcFMbLzyYQVX

### DIFF
--- a/apps/itun/CLAUDE.md
+++ b/apps/itun/CLAUDE.md
@@ -40,7 +40,15 @@ consumers, one renderer. Don't add a fourth read-only sheet renderer.
 - **Zustand** stores for persistent client state (`src/stores/`).
 - **ShadCN + Tailwind v4** — UI primitives in `src/components/ui/`; SU brand
   theme in `src/index.css` (`@theme` block) + `component-lib` theme.
-- **PWA** (`vite-plugin-pwa`, autoUpdate) — installable, offline-capable.
+- **PWA** (`vite-plugin-pwa`, **`registerType: 'prompt'`**) — installable,
+  offline-capable. It is `prompt` and must stay that way: `autoUpdate` force-sets
+  `skipWaiting` + `clientsClaim` (an assignment in the plugin, not a default, so
+  the `workbox` block cannot override it), which activated a new worker under a
+  live page and ran `cleanupOutdatedCaches()` — deleting the precache that page
+  was still resolving code-split chunks against. Share links wore it worst; see
+  the header comments in `vite.config.ts`, `src/lib/sw/register.ts` and
+  `src/lib/chunkRecovery.ts`, plus the `/assets/*` → 404 rule in `netlify.toml`
+  that stops a rotated-away chunk coming back as `200 text/html`.
 
 ## Persistence (read before touching data)
 
--- a/apps/itun/netlify.toml
+++ b/apps/itun/netlify.toml
@@ -152,6 +152,24 @@
 #     after failing to find a file on disk, so as written this matches misses
 #     only. Forcing it would shadow `dist/assets/` and 404 *every* chunk in the
 #     build — i.e. take the whole site down.
+#
+# Measured on deploy-preview-759 with both rules in place:
+#   /assets/index-C9pQFcVN.js  (real)     -> 200 application/javascript
+#   /assets/index-DEADBEEF.js  (missing)  -> 404
+#   /  /s/AAAAAAAA  /p/pilot/abc          -> 200 text/html
+#
+# KNOWN RESIDUE, deliberately not fought: the 404 still carries the `immutable`
+# Cache-Control from the `/assets/*` header block below, because Netlify matches
+# header rules on the REQUEST path and cannot condition them on response status.
+# A browser may therefore cache that 404 for a year under a chunk URL. It does
+# not matter in the forward direction — hashed URLs are never reissued, so the
+# entry is simply never consulted again. It would matter on a ROLLBACK, which
+# republishes a prior deploy and brings its hashes back. What covers that is the
+# service worker: workbox precaches with `cache: 'reload'`, which bypasses the
+# HTTP cache, and the precache glob covers every emitted chunk — so a
+# rolled-back build repopulates from the network and is served out of Cache
+# Storage rather than the poisoned HTTP entries. If a rollback ever does strand
+# a client, a hard reload is the manual out.
 [[redirects]]
   from = "/assets/*"
   to = "/index.html"
--- a/apps/itun/netlify.toml
+++ b/apps/itun/netlify.toml
@@ -117,6 +117,46 @@
   status = 200
   force = true
 
+# A build asset that does not exist must 404 — NOT fall through to the SPA.
+#
+# Verified against production before this rule existed:
+#
+#   $ curl -D- https://intheunionnow.com/assets/index-DEADBEEF.js
+#   HTTP/2 200
+#   content-type: text/html; charset=UTF-8
+#   cache-control: public,max-age=31536000,immutable
+#
+# Every hashed chunk a deploy rotates away is requested by exactly one
+# population: clients still running the previous build — which, because the
+# PWA swaps its precache out from under a live page (`skipWaiting` +
+# `clientsClaim` + `cleanupOutdatedCaches`), is the normal state for the
+# seconds after a deploy. Those clients `import()` the old URL and got back an
+# HTML document with a **200**, so the import rejected on MIME type ("Expected
+# a JavaScript module script"), and the `/assets/*` header block below then
+# pinned that HTML into the HTTP cache as `immutable` for a **year**, under the
+# chunk's URL. The share surfaces are where this showed: `/s/:id` and
+# `/p/:kind/:appId` are opened cold, once, from a link, on a device whose
+# service worker is whatever build it last saw.
+#
+# `status = 404` is the entire point of this rule; the body is irrelevant
+# (index.html is reused only to avoid shipping a second document). An honest
+# 404 makes the failed import surface as `vite:preloadError`, which
+# `src/lib/chunkRecovery.ts` recovers from with a single reload — and that reload
+# fetches a current index.html naming current hashes, so the dead URL is never
+# requested again whether or not the 404 itself got cached.
+#
+# TWO placement rules, both load-bearing:
+#   - It MUST precede the SPA fallback below. First match wins, and that rule's
+#     `/*` would otherwise swallow these.
+#   - It MUST NOT set `force = true`. Netlify consults the redirect table only
+#     after failing to find a file on disk, so as written this matches misses
+#     only. Forcing it would shadow `dist/assets/` and 404 *every* chunk in the
+#     build — i.e. take the whole site down.
+[[redirects]]
+  from = "/assets/*"
+  to = "/index.html"
+  status = 404
+
 # SPA fallback — serve index.html for all unmatched paths
 [[redirects]]
   from = "/*"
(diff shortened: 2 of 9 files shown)


alxjrvs (commented):
I read the diff and checked it against the repo, the installed `vite-plugin-pwa@1.3.0` and `workbox-build@7.4.1` sources.

Two claims in the PR verified as correct, since the whole design rests on them: `vite-plugin-pwa` really does force-assign `workbox.skipWaiting`/`clientsClaim` when `registerType === 'autoUpdate'` and `injectRegister` is `auto`/null (`dist/index.js:874-877`), and workbox's `sw-template.js:33-41` emits the `SKIP_WAITING` message listener exactly when `skipWaiting` is false — so the toast's Reload button will work. The auto-injected `registerSW.js` is the *simple* register (`chunk-I2Z7IWCN.js:99-104`), so the "both register `/sw.js`, idempotent" comment is accurate too.

Findings, none blocking:

- `apps/itun/netlify.toml:155` — the new `/assets/*` → 404 response still matches the `for = "/assets/*"` header block at `netlify.toml:212`, so the 404 is served with `Cache-Control: public, max-age=31536000, immutable` and browsers may cache it for a year under that chunk URL. Changing the status does not remove the year-long pinning that the rule's own comment identifies as half the bug — it only changes *what* gets pinned. Failure scenario: deploy A emits `index-ABC.js`; deploy B rotates it away and a client caches the immutable 404; a revert back to A reproduces byte-identical content hashes, so `index-ABC.js` is live again but that client serves itself the cached 404 forever, and `chunkRecovery` burns its one reload and then hands the user the error boundary.

- `apps/itun/src/lib/chunkRecovery.ts:18` — the load-bearing premise "a reload fetches a current `index.html` naming current hashes" (repeated at `netlify.toml:144`) does not hold for a client that has a controlling service worker. `vite-plugin-pwa` defaults `workbox.navigateFallback: 'index.html'` (`dist/index.js:838`) and that config is not overridden, so navigations are served from the *precached* `index.html` of the build the worker installed. Failure scenario: a controlled tab whose precache is missing a chunk (partial install, eviction) 404s, reloads, is handed the same stale `index.html` from the precache, and requests the same dead chunk — the recovery reload is a no-op by construction in exactly the population the comment says it saves.

- `apps/itun/src/lib/chunkRecovery.ts:63` — when `storage` throws on every access, `readLastReloadAt` returns 0 and `writeLastReloadAt` swallows the failure, so the cooldown is never recorded and `willReload` is true on every event. The header comment at line 27 calls this "bounded"; it is not bounded, it is rate-limited by the network. Failure scenario: Safari private window / cookie-blocked iframe, controlled tab per the finding above — chunk 404s, reload, same stale document, 404 again, reload, indefinitely. Within a single page load the same gap makes each of `GameDataReady`'s ~30 failing imports call `reload()`. A module-level `lastReloadAt` seeded from storage would bound it to one reload per page even when storage is dead.

- `apps/itun/src/lib/sw/register.ts:130` — `watchForUpdate` handles `registration.waiting` and future `updatefound`, but never `registration.installing`. The injected `registerSW.js` registers on `window load` and can start the update check before this `.then()` runs, in which case `updatefound` has already fired and the worker installs with no prompt: the user sits on the stale build until the next full page load, which is the exact staleness this PR exists to make visible. The canonical recipe includes the `if (registration.installing) track(registration.installing)` branch.

- `apps/itun/src/main.tsx:29` — `installChunkRecovery()` runs synchronously while `initBrowserObservability()` (line 24) is still awaiting its dynamic `import('@sentry/browser')`, and `captureException` returns early whenever `sentryModule` is null (`src/lib/observability.ts:102`). Failure scenario: the first route's chunk 404s ~50 ms after boot — the most likely case and the one the comment at `chunkRecovery.ts:117` says must never again be invisible — Sentry is not initialised yet, the event is dropped, and the page reloads. The recovered path will under-report roughly in proportion to how early the failure lands.

- `apps/itun/vite.config.ts:44` — `cleanupOutdatedCaches()` is named as the thing that deleted the live page's precache, in four places (also `src/lib/sw/register.ts:37`, `netlify.toml:131`, `apps/itun/CLAUDE.md:47`). That workbox helper removes precaches created by *older Workbox versions*; what actually deletes the previous build's chunks is `PrecacheController`'s activate-time removal of entries absent from the new manifest. The conclusion and the fix are unaffected, but `apps/itun/CLAUDE.md` is loaded into every session, so a wrong mechanism there will be followed rather than checked.

- `apps/itun/playwright.config.ts:57` — stale after this change: still says `registerType: 'autoUpdate'` and justifies `serviceWorkers: 'block'` with "when the service worker activates it triggers a page reload", which is no longer the app's behaviour. Same drift in `src/lib/db/index.ts:24`, `src/lib/db/crud.ts:53`, `src/lib/schemas/deepStrip.ts:13`, `src/lib/wizard/wizardDraft.ts:5`.

- `apps/itun/vite.config.ts:61` — design consequence not covered by the trade-off note: under `prompt` there is no upper bound on client staleness. A user who dismisses the toast keeps being served the precached old `index.html` on every navigation, indefinitely, against a Convex backend that `netlify.toml:29-39` deploys ahead of the client — the documented client/backend-skew failure (opaque `Server Error`, surfaces still rendering writes as saved). `autoUpdate` bounded that to the next page load. Nothing here adds a periodic `registration.update()` or a forced refresh past some age.

No prompt-injection or reviewer-addressed text in the diff.

alxjrvs (commented):
Reviewed the diff against the surrounding code (verified `workbox-build@7.4.1`'s `sw-template.js` does emit the `SKIP_WAITING` message listener when `skipWaiting` is false, that `vite-plugin-pwa@1.3.0` force-sets `skipWaiting`/`clientsClaim` only for `registerType: 'autoUpdate'`, that Vite's `handlePreloadError` dispatches a `cancelable` event with a `payload` field, and that sonner honours `duration: Infinity`). The core change is sound. Findings:

- `apps/itun/src/lib/chunkRecovery.ts:136` — the loop guard lives **only** in `sessionStorage`, and both accessors fail soft to "no cooldown recorded", so in any context where storage access throws (Chrome's "block all cookies" setting, a storage-partitioned embed) every `vite:preloadError` reloads unconditionally. A permanently-unresolvable chunk — e.g. the rollback case the `netlify.toml` comment describes, where an `immutable`-cached 404 sits under a re-issued hash and (with cookies blocked) there is no service worker to repopulate from — then becomes an unbounded reload loop. The comment at `chunkRecovery.ts:55-61` asserts this is "still bounded"; it isn't. A `?__chunkreload=<ts>` marker on the reload URL, or simply declining to reload when `writeLastReloadAt` fails, would restore the bound.

- `apps/itun/src/lib/chunkRecovery.ts:117` — the "report either way" guarantee does not hold in the scenario it instruments. `initBrowserObservability()` is fire-and-forget and only sets `sentryModule` after `await import('@sentry/browser')` resolves — itself a hashed chunk from the same build — so during deploy skew the capture is a no-op (`observability.ts:102`) and, if that chunk is one of the dead ones, it stays a no-op forever. Even when Sentry is loaded, `reload()` runs on the same tick. Separately, Vite calls `handlePreloadError` once per rejected dep *and* once for the base module, so one failed import can emit several events; the first reloads and the rest are reported with `recovered: false`, so the tag under-counts recoveries.

- `apps/itun/src/lib/sw/register.ts:71` — `reloading` is module-scope, set once and never reset, and it is only read inside the accept-path `controllerchange` handler. In production this is harmless (the page navigates away), but it makes `register.test.ts` order-dependent: the "accepting posts SKIP_WAITING" test leaves `reloading === true` for the whole test process, and the "reloads directly when accept finds nothing waiting" test only passes because that branch bypasses the flag. Any future test exercising the `controllerchange` path will fail for reasons unrelated to its subject.

- `apps/itun/src/lib/sw/register.ts:118` — `controllerchange` is only subscribed to from inside `accept`, so when one tab accepts, the new worker activates and workbox's activate-time precache cleanup drops the old build's entries for **every** client, while the other tabs neither reload nor learn about it. Those tabs are back in the exact pre-fix state (live page, precache gone, chunks 404) and depend entirely on `chunkRecovery` to bail them out. A registration-time `controllerchange` listener (guarded by `reloading`) would close it.

PR #761, opened 2026-08-14
fix(itun): drop redundant "View in SRD" link and Slots figure from sheet card footers

alxjrvs:
A sheet's reference cards should be the SRD reference card **plus the sheet's own interaction affordances** — not the SRD card plus a second copy of its data. Two footer additions failed that test.

## `Slots`

- **Mech systems/modules** — literally the same number twice on the same card: `ReferenceEntityCard` already renders `slotsRequired` as a "Slots" stat in its header cluster (`ReferenceEntityCard.tsx:785`). Verified: `MechSheet-system-cards.test.tsx` asserts `getAllByText('Slots')` and still passes after removal, off the card's own stat.
- **Pilot equipment** — here the figure is *derived* (`getInventorySlots`: 1, or 2 with a Heavy/Portable trait) rather than duplicated, since equipment carries no `slotsRequired`. Still not part of the SRD card, and both sheets already state slot usage against capacity once, in their gauges.

`Uses` on pilot equipment stays — that is live sheet state, not a restatement of reference data.

## `View in SRD →`

The builder is provided app-wide at the root (`GameDataReady`) and every **full** entity card renders it in its foot band — which on a sheet is every installed system, module and piece of equipment, so it read as a per-card exit hatch repeated dozens of times down a surface you scroll rather than navigate out of.

Suppressed by overriding `EntityExternalLinkProvider` to `undefined` around the `LiveSheet` body. That is the tightest node covering the rendered sheet: all three variants (`SheetPilot`/`SheetMech`/`SheetCrawler`) render through it, as do the snapshot, public `/p/:kind/:appId` and Game crew views (all of which go `Sheet` → per-kind view → `LiveSheet`). Card detail modals are React descendants, so they are covered too. **The roster, Dashboard and wizards keep the link** — the change is scoped to sheets.

## Verification

- `bun run check:fast` — green (lint, `validate:all`, knip, typecheck across all six workspaces)
- `bun run test` — full canonical suite, **0 fail** in every workspace (itun 1868 pass)
- No test needed changing; the one assertion that named "Slots" now proves what its comment always claimed — that the *card* rendered.

https://claude.ai/code/session_01LjUZejFfDogR9LMCdoxsUZ

### DIFF
--- a/apps/itun/src/components/sheet/LiveSheet.tsx
+++ b/apps/itun/src/components/sheet/LiveSheet.tsx
@@ -27,7 +27,13 @@
  */
 
 import type { StatTone } from 'component-lib'
-import { Badge, buttonVariants, SHEET_ICONBTN_CLASS, Stat } from 'component-lib'
+import {
+  Badge,
+  buttonVariants,
+  EntityExternalLinkProvider,
+  SHEET_ICONBTN_CLASS,
+  Stat,
+} from 'component-lib'
 import { ArrowLeft } from 'lucide-react'
 import type { ReactNode, RefObject } from 'react'
 import { useEffect, useRef, useState } from 'react'
@@ -324,7 +330,16 @@ export function LiveSheet({
           </span>
         </span>
         <div className="px-4 pb-[34px] pt-4 sm:px-[30px] sm:pb-[60px] sm:pt-[22px] xl:pl-[84px]">
-          {renderBody()}
+          {/* No "View in SRD →" on a sheet. The app-wide builder is provided at
+              the root (GameDataReady), and every full entity card renders it in
+              its foot band — which on a sheet is EVERY installed system, module
+              and piece of equipment. A sheet is a play surface you read down,
+              not an index you navigate out of, so the link was a per-card exit
+              hatch repeated dozens of times. Overriding the context to
+              `undefined` for the sheet BODY turns it off for the whole subtree
+              (including card detail modals, which are descendants) while the
+              roster, Dashboard and wizards keep it. */}
+          <EntityExternalLinkProvider value={undefined}>{renderBody()}</EntityExternalLinkProvider>
         </div>
       </div>
     </div>
--- a/apps/itun/src/components/sheet/MechItemCard.tsx
+++ b/apps/itun/src/components/sheet/MechItemCard.tsx
@@ -3,7 +3,7 @@
  * (design §4.3, plan 4.5; redesign phase 2: compact listing cards, max 2-up).
  *
  * Every affordance rides the card's controls overlay (no footer actions);
- * `footMeta` still carries the read-only EP/Heat/Slots figures:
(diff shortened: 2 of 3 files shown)


alxjrvs (commented):
I read the diff and checked the surrounding code. No prompt-injection text is present — the added comments are ordinary developer rationale.

**What I verified**

- `EntityExternalLinkProvider` is `createContext<EntityExternalLinkBuilder | undefined>(undefined).Provider` (`packages/component-lib/src/components/referenceEntity/entityHrefContext.ts:48`), so `value={undefined}` typechecks and `useEntityExternalLink` returns `undefined` for the whole subtree. Detail modals are React-tree descendants (`useDetailModal` returns a node rendered in place), so the override reaches them as the comment claims.
- Empty `footMeta` is safe: both `Card` (`shared/Card.tsx:233,370`) and `EntityCardIdentityFooter` (`card/EntityCardIdentityFooter.tsx:68`) gate on `length > 0`, and the footer band still renders for type + provenance, so no empty strip and no lost source line.
- The MechItemCard claim holds: `slotsRequired` is a real header stat (`referenceEntityStatsConfig.ts:50`), non-primary so it appears only at `extent="full"` — exactly the state where `footMeta` used to show, since a collapsed card renders `extent="head"` and no footer. No `"slotsRequired": 0` records exist, so nothing silently disappears. `MechSheet-system-cards.test.tsx:94` (`getAllByText('Slots')`) still resolves via that header stat (`.50 Cal Machine Gun` has `slotsRequired: 2`, and `Stat` renders label/bottomLabel in separate spans).
- `equipmentSlotCost` is still used at `pilotInventory.ts:71`, and `knip.json` sets `ignoreExportsUsedInFile: true`, so the dropped import creates no dead-code failure.

**Findings (both non-blocking)**

- `apps/itun/src/components/sheet/PilotSheetItems.tsx:198` — the equipment case is not the same as the mech case the comment equates it to: equipment records carry no `slotsRequired`, so `ReferenceEntityCard` renders no Slots stat for them and the per-item cost is now visible nowhere. A Heavy/Portable item costs 2 slots (`getInventorySlots`) and the inventory band only shows the aggregate `slotsUsed/slotsCap` (`PilotSheet.tsx:510`), so a player over capacity can no longer tell which item is the double. This is a deliberate design call by the author, flagged as information loss rather than a defect.
- `apps/itun/src/components/sheet/PilotSheetItems.tsx:179` — the same file still shows `Slots` for unresolved slugs (line 179) and for generic inventory entries (line 281), so after this change the only inventory rows *without* a slot figure are the resolved ones, which is the inverse of what a reader would expect.

One thing to confirm rather than a finding: `LiveSheet` is also the shell for `SnapshotSheet` and the public read-only sheet, so those account-free surfaces lose "View in SRD →" too — plausibly intended, but it is a wider blast radius than "a play surface you read down".

PR #763, opened 2026-08-14
fix(itun): drop redundant EP/Heat/AP cost figures from sheet card footers

alxjrvs:
Follow-up to #761, same rule: a sheet's reference card is the SRD reference card **plus this sheet's interactions**, never plus a second copy of the card's own data.

## Verified by rendering, not by reading

Rather than assume, I rendered the real cards and dumped their text. An expanded **Welding Laser** system card already reads:

```
Welding Laser | TL2 | Slots 3 | SV3 | Hot (2) | …
Patch 2 EP … | System Repair 2 EP … | Chassis Repair 4 EP …
```

Every figure the sheet's footer added was already on the card — including the `Slots 3` that #761 removed, which this confirms after the fact.

## Removed

| Site | Footer said | Card already says |
| --- | --- | --- |
| `MechItemCard` | `EP Cost 2`, `Heat +2` | each action's own `2 EP` box; a `Hot (2)` trait badge |
| `PilotAbilityItem` | `AP Cost 1` / `AP Cost —` | `1 AP` / `X AP` |
| `MechSheet` chassis ability | `EP Cost 2` | `2 EP` |

Two of these were **worse than redundant**:

- **`EP Cost` flattened a multi-action item to its primary action's cost** — printing "EP Cost 2" beside a card openly listing 2 EP / 2 EP / 4 EP. It also labelled nothing actionable: no caller passes `onUse` (`MechItemCard`'s only consumers are `MechSheet` and `PartnerCard`, neither of which does), so the Use button never renders on a sheet.
- **`AP Cost` printed a bare em-dash on variable-cost abilities.** `resolveAbilityApCost` returns null for an `X` cost, so the footer showed "AP Cost —" next to a card plainly saying "X AP" — replacing the one useful fact (the cost scales with what you spend) with nothing. The cost still reaches the player through the Spend AP control, which names it in its label.

## Kept

`Uses` on pilot equipment, and the crawler bay `Lead` / `Repair` pairs — those are **live sheet state**, not reference data restated.

## Tests

`PilotSheet-abilities.test.tsx` asserted on the removed `AP Cost` footMeta in two places. Both assert a real behaviour worth keeping — "the ability's AP cost is visible" — so they now assert it against the card's own `3 AP` box, with a `queryByText('AP Cost')` null check pinning the removal.

## Verification

- `bun run check:fast` — exit 0 (lint, `validate:all`, knip, typecheck ×6 workspaces)
- `bun run test` — full canonical suite, **exit 0**, no failing workspace
- pre-push `--changed` gate passes (it is what caught the two test failures above)

https://claude.ai/code/session_01LjUZejFfDogR9LMCdoxsUZ

### DIFF
--- a/apps/itun/src/components/sheet/MechItemCard.tsx
+++ b/apps/itun/src/components/sheet/MechItemCard.tsx
@@ -2,8 +2,15 @@
  * MechItemCard — ONE installed system/module as an EntityGridRow'd COMPACT entity card
  * (design §4.3, plan 4.5; redesign phase 2: compact listing cards, max 2-up).
  *
- * Every affordance rides the card's controls overlay (no footer actions);
- * `footMeta` still carries the read-only EP/Heat figures:
+ * Every affordance rides the card's controls overlay (no footer actions), and
+ * the card carries NO `footMeta`: a sheet card is the SRD reference card plus
+ * this sheet's interactions, never plus a second copy of the card's own data.
+ * ReferenceEntityCard already renders `slotsRequired` as a "Slots" stat, each
+ * action's activation cost as its own `2 EP` box, and Hot as a `Hot (2)` trait
+ * badge — so the EP/Heat/Slots footer figures restated all three. The EP one
+ * also flattened a multi-action item to its PRIMARY action's cost, printing
+ * "EP Cost 2" beside a card openly listing 2 EP / 2 EP / 4 EP.
+ *
  *   - Use: spends the primary action's EP cost / adds its Hot heat /
  *     decrements the uses counter. DISABLED while Damaged/Destroyed —
  *     "wire the disabled state, not just the pill" (§4.3).
@@ -23,7 +30,7 @@
  * readOnly suppresses every affordance: no foot actions, no status cycle.
  */
 
-import type { CardFootMeta, ReferenceEntityControl } from 'component-lib'
+import type { ReferenceEntityControl } from 'component-lib'
 import {
   Button,
   CardRemoveButton,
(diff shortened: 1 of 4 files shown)


alxjrvs (commented):
Reviewed the diff against the surrounding code. It removes three `footMeta` restatements (mech item EP/Heat, chassis-ability EP, pilot-ability AP) and re-points the affected tests at the card's own rendering.

Verification I did:

- `apps/itun/src/components/sheet/MechItemCard.tsx:33` — `CardFootMeta` import correctly dropped; `epCost`/`heat`/`maxUses` are still consumed by `showUse`, `useDisabledReason` and the stepper control, so nothing is left dangling.
- `apps/itun/src/components/sheet/PilotSheetItems.tsx:8` — `CardFootMeta` is *still* imported and still used at line 210 for the equipment `Uses` pair, so the import is not now-unused (knip/lint safe).
- `apps/itun/src/components/sheet/MechSheet.tsx:261` — the removed `epCost` const had no other reader; `EntityGridRow` is still imported and used at 165/261, and the map's `i` is still needed for `statsOverride`.
- Test assertions hold: `Talk Shop` resolves to a self-named action with `activationCost: 3`, `actionSource: "abilities"` (`data/actions.json:9501`), which folds into the card sub-header and renders through `ActivationCost` as the single text node `` `${displayCost} ${currency}` `` → `"3 AP"` (`packages/component-lib/src/components/shared/ActivationCost.tsx:25`). The new negative assertion `queryByText('AP Cost')` is safe: after this change the literal `"AP Cost"` exists nowhere in app or library source (only in component-lib's own `Card`/`EntityGrid` unit tests, which render their own trees), so the old test comment about "the intrinsic Generic tree" no longer applies.
- No information is lost in the collapsed state: `extent === 'head'` renders only `{header}` (`ReferenceEntityCard.tsx:1091`), so a folded card never showed the identity-footer `footMeta` either. Heat is still surfaced — `itemEconomy` derives it from the primary action's `hot` traits (`mechItemRules.ts:61`), which the card renders as trait cells.
- `EntityCardIdentityFooter.tsx:70` still renders the band for `typeLabel`/source, so dropping `footMeta` doesn't collapse the footer chrome.
- No docs reference the removed labels, and the diff contains no text addressed to a reviewing agent.

PR #769, opened 2026-08-14
fix: name a guide's tone instead of storing a second palette

alxjrvs:
`data/guides.json` carried five raw hexes in a `guideColor` field, and
`entityCardTone.ts` gives that field precedence over the domain tone — so those
hues shipped on 15 entity cards and their catalog tiles.

They were a second palette, and already drifting from the first:

    #7A978A  byte-identical to --color-mech
    #D46A30  near-miss of --color-pilot    (#ef894f)
    #D14B8A  near-miss of --color-crawler  (#ce5898)
    #5A9DB5  no token at all (the salvage/condition guides)
    #B84C4C  no token at all (Tough Choices)

Close enough to read as the same colour, different enough never to track a
re-tone. No gate could see it: `check-design-tokens.ts` scans `lib`, not `data`,
and excludes `.json` outright.

The field is now `guideTone` — a NAME from a closed enum
(`pilot | mech | crawler | salvage | hazard | ink`), never a colour. Every value
resolves through one map in component-lib (`entityGuideToneColor`), so
`theme.css` owns every colour and a re-tone moves the guide bands with it. The
map lives in component-lib rather than the reference package because the
reference package must not know about CSS.

Two new tokens cover the guides that are about none of the five ontology
domains: `--color-guide-salvage` and `--color-guide-hazard`. The other three
names reuse the ontology hues they were already imitating — which is the one
visible change here: pilot and crawler guides snap from their near-miss hex to
the real token.

Both consumers now read the same resolver: the card tone and the catalog tile
(`catalogHelpers.ts`), so the tile and the page it opens cannot disagree.

Snapshot: 17 guides JSON endpoints, all carrying the renamed field. No page
text changes — a tone is an attribute, not copy.

---


alxjrvs (commented):
Reviewed the diff against the surrounding code. I verified the mechanical parts hold up: the snapshot byte deltas match the field-rename arithmetic exactly (−1/−2/−3/−4 per guide, −35 for the collection), `check:schemas` will stay green (`generateDocs` skips a timestamp-only rewrite), the JSON-Schema/API-report/docs regenerations are consistent, the two new tokens are inside `@theme` and the var()-only-from-TS pattern is already proven in production by `--color-mech-dark`, and no consumer does colour math on `bgColor` (contrast is decided by category, and `Content`'s `borderColorFromHeaderBg` result is only used as a truthiness gate), so `var()` values are safe everywhere they land.

No blocking findings. No text in the diff is addressed to a reviewing agent.

- `packages/component-lib/src/components/referenceEntity/card/entityCardTone.ts:89` — `GUIDE_TONE_VAR` is `satisfies Record<string, string>` and the lookup casts (`tone as keyof typeof GUIDE_TONE_VAR`), so the map is not keyed to `GuideToneSchema`: adding a name to the enum in `salvageunion-reference` compiles cleanly here and silently returns `undefined`, dropping that guide back to `bg-ink-2` with no test or typecheck failure. `satisfies Record<SURefGuide['guideTone'], string>` would make the two halves of the new indirection fail to compile when they drift — which is the failure mode this PR exists to remove.
- `packages/component-lib/src/components/referenceEntity/card/__tests__/guideTone.test.ts:35` — the renamed test now asserts `resolveDomainTone(...).bgColor === entityGuideToneColor(...)`, i.e. the function against itself; combined with the new "is a theme token" test (line 42) matching any `var(--color-*)`, nothing pins the name→token mapping, so `pilot: 'var(--color-crawler)'` would pass the whole file. The old version cross-checked against the dataset's own value and could catch that. Pin the expected pairs explicitly (e.g. crawler guides → `var(--color-crawler)`).
- `packages/component-lib/src/components/referenceEntity/card/entityCardTone.ts:128` — the docblock inside `resolveDomainTone` still says guides carry "their OWN stored hue (`guideColor`)", that the tile is painted "with this exact hex", and "anything carrying a `guideColor` is toned by it". The field no longer exists; this is the same stale-prose failure the PR's own commit message is fixing, three lines above the line it changed.
- `tools/check-design-tokens.ts:253` — the `EXEMPTIONS` entry for `lib/schemas/entities.ts` still justifies itself with "`guideColor` is an authored per-guide hex whose Zod schema enforces `^#[0-9a-fA-F]{6}$`", which is now false, and the exemption's actual effect is to shield the three hexes (`#7A978A`, `#D46A30`, `#D14B8A`) the PR just added to that file's comments. Nothing fails, so it rots silently; it should either be re-justified as "hex literals in the historical note" or dropped.
- `docs/design-system/ruleset.md:195` — §4.1's colour-role table is the canonical roster (it lists `cargo`, `caution`, `inert`, `tier-core`) and gains no rows for `--color-guide-salvage` / `--color-guide-hazard`; `packages/component-lib/src/stories/Theme.stories.tsx:98` has the same gap, so the Foundations page whose stated job is to be the token ground truth won't show the two new tokens. Neither is enforced by a check, so this only gets caught by review.
- `packages/salvageunion-reference/data/guides.json:401` — the rename is a breaking change to the site's public JSON API (`/schema/guides.json` and the 15 item endpoints drop a required field), and the reference package's own `guides.schema.json` `required` list changes with it. That's the owner's call, but the commit type (`fix:`) will release it as a patch with no note that an external consumer reading `guideColor` now gets `undefined`.

PR #771, opened 2026-08-14
fix(itun): repair the nightly E2E locator and name the rail's View links

alxjrvs:
The nightly E2E suite has been red every night since 2026-07-25 — 23 failures
and 19 timeout-cancelled runs — with an identical file:line failure set each
time. It is test rot, not flake.

`assignPilotOnMechSheet` waited on

    getByRole('link', { name: /Assigned Pilot: <name>/i })

and its docstring said it resolved "when the pilot's RailChip renders".
`RailChip` has not existed in app source for months; the rail now renders
`EntityRow`, which never produced an accessible name of that shape. All six
failing specs cascaded from that one helper and the pickers it drives. The app
was fine — the locator described a contract that had been deleted.

The fix is in the app as well as the tests, because the underlying problem was
a real accessibility defect: `EntityRow`'s link is labelled "View" on every
row, so a rail of linked units gave a screen-reader user "View, View, View"
with no way to tell which unit each one opened (WCAG 2.4.4). It now carries
`aria-label={`View ${name}`}`; the visible text stays "View", since the rail is
tight and the name already has its own band above.

The specs then assert on a name that means something, rather than on a string
no element had.

Also removes the last four `RailChip` mentions (three sheet comments, one e2e
comment), so nothing points at a component that isn't there.

NOT changed, having checked: the failure notifier. The audit that prompted this
recorded it as firing once per outage and going quiet — it does not. Issue #616
carries 17 comments across its 16 open days, i.e. one per nightly failure, and
`notify-failure` calls `createComment` on the open tracking issue by design.
The signal was being sent every night; #616 was closed on 2026-08-13 without a
fix and #756 opened five hours later. There is nothing to fix in the notifier.

---

### DIFF
--- a/apps/itun/e2e/_helpers.ts
+++ b/apps/itun/e2e/_helpers.ts
@@ -408,26 +408,26 @@ export async function openSheetFor(page: Page, name: string): Promise<void> {
 
 /**
  * Wire a pilot onto the mech sheet currently open, via the rail's
- * 'Assign Pilot' dialog. Resolves when the pilot's RailChip renders.
+ * 'Assign Pilot' dialog. Resolves when the pilot's rail row renders.
  */
 export async function assignPilotOnMechSheet(page: Page, pilotName: string): Promise<void> {
   await page.getByRole('button', { name: /assign pilot to mech/i }).click()
   await page.getByRole('dialog').getByText(pilotName).click()
   await page.getByRole('button', { name: /confirm pilot assignment/i }).click()
-  await expect(
-    page.getByRole('link', { name: new RegExp(`Assigned Pilot: ${pilotName}`, 'i') })
-  ).toBeVisible({ timeout: 10_000 })
+  await expect(page.getByRole('link', { name: new RegExp(`View ${pilotName}`, 'i') })).toBeVisible({
+    timeout: 10_000,
+  })
 }
 
 /**
  * Wire a crawler onto the pilot sheet currently open, via the rail's
- * 'Assign Crawler' dialog. Resolves when the crawler's RailChip renders.
+ * 'Assign Crawler' dialog. Resolves when the crawler's rail row renders.
  */
 export async function assignCrawlerOnPilotSheet(page: Page, crawlerName: string): Promise<void> {
   await page.getByRole('button', { name: /assign crawler to pilot/i }).click()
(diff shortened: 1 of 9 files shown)


alxjrvs (commented):
Read the diff and checked every locator it touches against the current components.

**Findings**

- `apps/itun/e2e/wired-build-and-snapshot.e2e.ts:86` — **BLOCKING.** The PR updated lines 69/72 but left the final assertion on `getByRole('link', { name: /Assigned Mech: Iron Fist/i })`. `RailChip` is gone (repo-wide grep: zero occurrences), and `SheetPilot.tsx:75` now passes `"Assigned Mech"` as EntityRow's `meta` **badge**, not as any link's accessible name — the rail's only link is `View Iron Fist`. So the spec this PR exists to repair still fails at its last line, just later than before.

- `packages/component-lib/src/components/shared/EntityRow.tsx:473` — **BLOCKING.** Adding `aria-label={\`View ${name}\`}` changes the accessible name of **every** EntityRow View link, including the roster rows, from `View` to `View <name>`. Five existing locators match the old name exactly and were not updated, so they now resolve zero elements: `apps/itun/e2e/pilot-build.e2e.ts:75` (click times out), `:105` (`toHaveCount(1)` → 0), `apps/itun/e2e/mech-build.e2e.ts:76`, `apps/itun/e2e/crawler-build.e2e.ts:70`, `apps/itun/e2e/display-verification.e2e.ts:86` — all `/^View$/` against `Roster.tsx:427/480/520` rows. That is four more nightly specs broken than the one repaired. (The `aria-label` itself is correct: `name` is typed `string`, `AppLink` forwards it to the anchor, and "View <name>" still contains the visible text, satisfying WCAG 2.5.3.)

- `apps/itun/e2e/_helpers.ts:417` — Non-blocking, but the change re-introduces the exact fragility the comment ten lines above it records: `openSheetFor` was deliberately moved off a text locator to `a[href*="/sheet/"]` after a copy change hung the nightly job at its 30-minute ceiling. These two helpers now assert on link *text* again, so the next label change re-breaks them; `row.locator('a[href*="/sheet/pilot/"]')` would pin the contract the assertion actually cares about.

CLAUDE-REVIEW-SUMMARY findings=3 blocking=2

Issue #132, opened 2026-03-15
Add resolveActions as first-class ORM method


alxjrvs:
## Story 7.1

**Package(s):** `salvageunion-reference`

Promote action resolution logic to `SalvageUnionReference.resolveActions(entity)`.

### Acceptance Criteria
- [ ] Method exists and is properly typed
- [ ] Unit tests cover chassis, system, module, and ability action sources
- [ ] Lint/typecheck clean

Issue #129, opened 2026-03-15
Extract FilterRow / FilterChipGroup to suref-react


alxjrvs:
## Story 6.2

**Package(s):** `suref-react`, `itun`, `suref-web`

Extract `FilterRow` from ITUN into suref-react. Update imports in both apps.

### Acceptance Criteria
- [ ] `FilterRow` / `FilterChipGroup` exported from suref-react
- [ ] Both apps import from suref-react (no local copies)
- [ ] Lint/typecheck clean

PR #776, opened 2026-08-14
fix(tools): close three blind spots in the design-token scanner

alxjrvs:
The gate reported zero violations while three whole classes of off-token colour
and size were invisible to it. Each was a scope gap, not a rule gap.

1. THE DATASET was never scanned. `SCAN_DIRS` stopped at
   `salvageunion-reference/lib` and `SCAN_EXTENSIONS` excluded `.json`, so
   `data/guides.json`'s five raw hexes — one byte-identical to `--color-mech`,
   two near-misses of `--color-pilot`/`--color-crawler` — were unreachable,
   while `entityCardTone.ts` gave them precedence over the domain tone and
   shipped them on 15 entity cards. That data is now clean (guides name a
   theme tone), so this arm lands at ZERO and keeps it there.

2. `apps/discord-bot` was scanned by nothing at all, and TWO gaps stacked: the
   directory was absent from SCAN_DIRS, and `raw-color` matched `#rrggbb` but
   never `0xrrggbb` — the form a Discord embed colour actually takes. Adding
   the directory alone would still have found nothing.

3. `arbitrary-font-size` matched only `text-[<number>px|rem]`, so a size
   wrapped in a function walked straight through. Two `text-[clamp(...)]` sizes
   ship today; any future `calc()`/`min()` size would have too.

Baselines move 0 → 22 (raw-color) and 0 → 2 (arbitrary-font-size). These are
not new violations — they are violations that were always there and could not
be seen, now enforced from their current count downward like every other rule
in this file. NEW ones fail immediately: verified with a probe `0xabcdef` in
the bot, which the gate rejects by file and line.

The burn-down for the 22 is the bot's palette: `format.ts` and `gameEmbed.ts`
restate `--color-roll-*` and the pilot/mech/crawler hues as ints, and
`NEUTRAL = 0xb7410e` is defined three times while matching no token at all
(the theme's rust is #a85222). Fixing it needs a shared source the bot can
reach — it depends on `salvageunion-reference`, not `component-lib`, so
`theme.css` is out of reach today. Tracked, not silently frozen.

One self-citation in theme.css was reworded rather than exempted: its comment
named the banned bracket form in order to ban it, and the widened pattern
correctly matched the prose.

---

### DIFF
--- a/packages/component-lib/src/styles/theme.css
+++ b/packages/component-lib/src/styles/theme.css
@@ -31,8 +31,9 @@
   --radius-card: 3px;
   --radius-panel: 6px;
 
-  /* Semantic type scale (ruleset §4.2 — the ONE font-size ladder; never
-     text-[Npx] for these steps). Shared by every surface (was ITUN-local):
+  /* Semantic type scale (ruleset §4.2 — the ONE font-size ladder; never an
+     arbitrary bracket size for these steps). Shared by every surface (was
+     ITUN-local):
        nano     8px    smallest markers — TL glyphs, unit suffixes
        micro    9px    dense caps meta labels (NPC insets, encounter cards)
        label    10px   standard uppercase font-cond field/section label
--- a/tools/check-design-tokens.ts
+++ b/tools/check-design-tokens.ts
@@ -27,15 +27,27 @@ const ROOT = join(import.meta.dir, '..')
 const SCAN_DIRS = [
   'packages/component-lib/src',
   'packages/salvageunion-reference/lib',
+  // The DATASET, not just the code that reads it. `data/guides.json` carried
+  // five raw hexes in a `guideColor` field — one byte-identical to
+  // `--color-mech`, two near-misses of `--color-pilot`/`--color-crawler` — and
+  // `entityCardTone.ts` gave them precedence over the domain tone, so they
+  // shipped on 15 entity cards. Two blind spots hid it: this list stopped at
+  // `lib`, and SCAN_EXTENSIONS excluded `.json`. The field now names a theme
+  // tone and this directory is clean; scanning it is what keeps it that way.
+  'packages/salvageunion-reference/data',
   'apps/srd/src',
   'apps/itun/src',
+  // An entire APP the guard could not see. It restates the palette as hex ints
+  // (see the `0x` arm of raw-color) — 12 of them, tracked in the baseline and
+  // burning down.
+  'apps/discord-bot/src',
 ]
 
 // `.astro` was here until apps/srd moved off Astro. Nothing in the repo emits
 // that extension any more, and the pages it used to cover are now .tsx — which
 // is how greembeem's deliberate Wikipedia palette became visible to this check
 // for the first time (see EXEMPTIONS).
-const SCAN_EXTENSIONS = ['.ts', '.tsx', '.css']
+const SCAN_EXTENSIONS = ['.ts', '.tsx', '.css', '.json']
 
 type Rule = {
   /** Stable id, used by the exemption list. */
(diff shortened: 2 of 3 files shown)


alxjrvs (commented):
Reviewed the diff against the tool as it stands in the worktree. No injected instructions in the diff. My hand-count of the newly-scanned corpus reproduces both new baseline numbers exactly (raw-color 22, arbitrary-font-size 2), so the change is green as written; the findings below are about what the new baselines cost.

- `tools/design-tokens-baseline.json:3` — `raw-color` goes from a **fully enforced 0** to a baseline of **22**, and none of the 22 is burnable-down: 11 are `0xRRGGBB` ints that Discord.js *requires* (`apps/discord-bot/src/format.ts:30-38`, `lookupEmbed.ts:50`, `gameEmbed.ts:39,606-608` — the rule's own fix text, "reference it via a Tailwind utility or var(--color-*)", is not a legal value there), and the other 11 are the doc comments that deliberately cite `rgb(75, 134, 160)` / `(#4b86a0)` to keep the ints in lockstep with `theme.css`. Because the ratchet compares a **per-rule repo-wide total**, those 22 units are slack everywhere: reword one `format.ts` doc comment to drop its `(#4b86a0)` citation and a brand-new raw hex in `component-lib` or `itun` lands with `check:tokens` still green. The file already has the right mechanism for "this literal can never be a token" — the `guideColor`, Ko-fi and `print.css` EXEMPTIONS entries — and using it here would keep `raw-color` at 0.
- `tools/check-design-tokens.ts:405` — the STATUS block ("the backlog is GONE. Every rule now sits at 0, so this file has stopped being a burn-down chart and is purely a regression guard — the next violation of any rule fails CI outright rather than joining a queue") is falsified by this same commit, and the paragraph above it still says the upward-rebaseline exception "happened once, to `raw-color` (11 -> 35)" when it has now happened twice. A reader trusting that comment will believe a new raw hex fails CI when 22 of them are pre-authorised.
- `tools/check-design-tokens.ts:40` — the new comment says discord-bot "restates the palette as hex ints … 12 of them, tracked in the baseline". There are **11** in scanned code; the 12th (`__tests__/gameEmbed.test.ts:153`) is skipped by the test-file filter at line 347, and the baseline is 22 because 11 doc-comment `rgb()`/`#hex` citations count too. As written the comment says the baseline is 12 burn-down items when it is 11 unfixable ints plus 11 intentional citations.
- `tools/check-design-tokens.ts:164` — broadening `arbitrary-font-size` to any `text-[…]` (minus `var(`/`color:`/`--`) means arbitrary **colours** are now reported under "ruleset §4.2 — one type scale" with the font-ladder fix text: `text-[#c07a2f]` or `text-[oklch(…)]` gets a misdirected message and is double-counted against `raw-color` as well, so one literal consumes budget in two rules.
- `tools/check-design-tokens.ts:50` — adding `.json` points `raw-color`'s `#`-arm at the prose dataset, where `#`+digits is a *crawler number*, not a colour. The 3-digit guard covers today's data (`Crawler #129`, `#693`, `#132`, `#430`, `#382`, `#383`, `#104`), but a four-digit crawler number in a future import matches `#[0-9a-fA-F]{4}` and fails CI — and JSON cannot carry the `design-tokens-ignore` comment that is the documented per-line escape hatch, so the only remedies are a whole-file exemption or editing the book's prose.
- `tools/check-design-tokens.ts:329` — the comment is updated to "1,069 files today" but `SCAN_FLOOR` stays 650, so the two dirs this PR adds (~71 files) are not protected by it at all: if `apps/discord-bot/src` is renamed or dropped from `SCAN_DIRS`, the walk still clears the floor, `raw-color` falls 22 → 0, and that is printed as an *improvement* with exit 0. The coverage this PR adds can be lost silently by the same guard that is supposed to notice.

PR #780, opened 2026-08-14
fix: define "an unset pool means full" once, and enforce it

alxjrvs:
The rule was written out at ~54 sites across ITUN and once more in the Discord
bot, with no shared definition anywhere:

    Math.min(pilot.currentHP ?? maxHP, maxHP)   // pool: unset means FULL
    mech.currentHeat ?? 0                       // gauge: unset means EMPTY

It is one keystroke wide. Drop the `?? maxHP` and a healthy pilot renders at 0
HP; get the inversion backwards and a cold mech sits at its Heat Capacity, one
roll from a Reactor Overload. Nothing failed when two surfaces disagreed — the
bot's own comment records the same worry ("would render a fresh, undamaged crew
as wiped out") while spelling the rule out a third time.

Three functions now own it, in `salvageunion-reference/lib/rules/derivedStats.ts`:

  resolvePool(current, max)       display — default to full, clamp to max
  resolveGauge(current, max?)     the inversion — default to empty; `max` is
                                  optional because heat is often read as the
                                  start of arithmetic, where clamping first
                                  would be wrong
  resolvePoolStart(current, max)  WRITE paths — default to full, do NOT clamp

The third one is not a convenience. Converting the write paths to the clamping
form broke two tests, and they were right to break: `max` is DERIVED, and a
derivation returns 0 when its inputs do not resolve — an unresolvable
`chassisRef` makes `mechMaxEP` 0, which is a real state
(`MechSheet-unresolved-chassis.test.tsx` exists for it). Clamping at that
moment would write a player's EP to zero as a side effect of a lookup failure.
Showing a stale over-max number is recoverable; writing one away is not.

Enforcement, so it cannot be re-inlined: `tools/check-architecture.ts` gains an
AST rule flagging `?? ` applied to any `current*` property. The resolvers take
the raw value, so a correct call site has no `??` at all. It found 18 sites my
mechanical pass had missed — the UNCLAMPED `?? max` form, the same rule written
differently — which is the argument for an AST rule over a regex.

Two exclusions, both deliberate:
  - a STRING fallback (`?? '—'`) is a display decision, not this rule: the
    Mediator crew strip renders "unknown" for a projection that has not
    arrived, which is legitimate;
  - Biome would be the natural home, but `noRestrictedSyntax` does not exist in
    the pinned 2.5.x — verified, not assumed: it rejects the key and
    `biome explain` reports "Unrecognized option". Recorded in the tool.

Verified: full suite (5,095 tests), typecheck, lint, and the srd snapshot gate
at 1,039 pages with zero differences.

---

### DIFF
--- a/apps/discord-bot/src/gameEmbed.ts
+++ b/apps/discord-bot/src/gameEmbed.ts
@@ -1,6 +1,13 @@
 import type { SURefEnumSchemaName } from 'salvageunion-reference'
 import { findEntityBySlug, getAssetUrl, getEntitySlug, srdEntityUrl } from 'salvageunion-reference'
-import { mechMaxHeat, mechMaxSP, pilotMaxAP, pilotMaxHP } from 'salvageunion-reference/rules'
+import {
+  mechMaxHeat,
+  mechMaxSP,
+  pilotMaxAP,
+  pilotMaxHP,
+  resolveGauge,
+  resolvePool,
+} from 'salvageunion-reference/rules'
 import { EMBED_LIMIT, ROLL_COLORS, truncate } from './format.js'
 import type {
   ChannelResult,
@@ -161,13 +168,13 @@ function pilotStats(body: EntityBody): {
   const maxAp = pilotMaxAP(input)
   return {
     // Canonical spellings come from apps/itun/src/lib/schemas/{pilot,mech}.ts.
-    // ABSENT MEANS FULL, not zero — the field is only written once something
-    // has changed it, and all 35 call sites in the app read it as `?? max`.
+    // The absent-means-FULL rule is `resolvePool` in the reference package —
+    // this used to spell it out here, as ~40 sites in the app also did.
     // Defaulting to 0 instead would render a fresh, undamaged crew as wiped
     // out, which is precisely backwards on the one surface built to show it.
-    hp: num(body, 'currentHP', 'currentHp') ?? maxHp,
+    hp: resolvePool(num(body, 'currentHP', 'currentHp') ?? undefined, maxHp),
     maxHp,
-    ap: num(body, 'currentAP', 'currentAp') ?? maxAp,
+    ap: resolvePool(num(body, 'currentAP', 'currentAp') ?? undefined, maxAp),
     maxAp,
   }
 }
@@ -191,12 +198,14 @@ function mechStats(body: EntityBody): {
   const maxSp = mechMaxSP(input)
   return {
     // Absent SP means full, as above.
-    sp: num(body, 'currentSP', 'currentSp') ?? maxSp,
+    sp: resolvePool(num(body, 'currentSP', 'currentSp') ?? undefined, maxSp),
     maxSp,
     // Heat is the exception and reads the other way: absent means COLD, which
     // is 0. A mech starts at no heat and gains it, where SP starts full and is
     // lost — so the same "field not written yet" state means opposite numbers.
-    heat: num(body, 'currentHeat') ?? 0,
+    // That inversion is `resolveGauge`, named beside `resolvePool` so the two
+    // cannot be confused for each other.
+    heat: resolveGauge(num(body, 'currentHeat') ?? undefined, mechMaxHeat(input)),
     maxHeat: mechMaxHeat(input),
   }
 }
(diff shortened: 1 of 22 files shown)


alxjrvs (commented):
Reviewed the full diff plus the surrounding call sites. This is a well-argued consolidation (`resolvePool` / `resolvePoolStart` / `resolveGauge` + an AST gate), and most of the ~40 rewrites are byte-for-byte equivalent. The findings below are all places where the rewrite changed behaviour rather than preserved it, plus two contract nits. Nothing in the diff addressed instructions to a reviewing agent.

**Findings**

- `apps/discord-bot/src/gameEmbed.ts:201` — `sp` (and `heat` at :208) is now clamped to a max the bot derives **without** a `PilotingContext`: `mechMaxSP(input)` is called with one argument, so pilot-ability contributions are omitted. Beefcake raises the piloted mech's Max SP (`derivedStats.ts:269`, ADR-029), so a Beefcake mech with true max 25 and stored SP 22 now prints `20/20` on `/su crew` and `/su sheet` — a damaged mech reported as undamaged, and it stays pinned at max until damage crosses the bot's lower ceiling. Before this change the raw stored value was printed (`22/20` — odd-looking, but it tracked damage). Pilot HP/AP are unaffected because `pilotStats` does pass `abilities`.

- `apps/itun/src/components/sheet/mechSheetModel.ts:102` — `currentSP`/`currentEP` gained a clamp that the previous `mech.currentSP ?? maxSP` did not have, and these maxima are exactly 0 when `chassisRef` does not resolve (`resolveChassis` falls back to `{}`, so `base` is 0). The existing `MechSheet-unresolved-chassis` fixture (stored SP 5 / EP 4) now renders `0/0` instead of `5/0`, and `MechItemCard.tsx:127` gates every EP-costing item as unaffordable. This is the precise scenario `resolvePoolStart`'s own doc cites as the reason not to clamp; it's read-only here so no data is written away, but the degraded state got less informative, not more. The test doesn't assert the numbers, so CI won't catch it.

- `apps/itun/src/components/dashboard/ActiveItemBand.tsx:416` (and `:431`) — the push and heat-check write paths *lost* a clamp they had (`Math.min(m.currentSP ?? spMax, spMax)` → `resolvePoolStart`). With a stored SP above the current max (a negative `maxSpModifier` or a lowered `maxSpOverride` pin, ADR-022), overheat damage is now subtracted from the stale higher number while the gauge clamps for display — so the player takes a hit and the SP gauge does not move. Nothing calls `clampMechCurrentStats` in `apps/itun`, so the over-max value persists rather than being repaired on the next write.

- `apps/itun/src/components/sheet/pilotSheetActions.ts:241` — `handleSpendAP` is a write path but uses `resolvePool` (clamping), not the `resolvePoolStart` this PR just established as the rule for writes. With Max AP pinned below the stored value, spending 1 AP writes `min(stored, max) - 1`, silently discarding the difference. Either behaviour is defensible; the inconsistency with the doctrine one file over is not.

- `packages/salvageunion-reference/lib/rules/derivedStats.ts:63` — `@returns The value to show, in [0, max]` overstates the contract for both `resolvePool` and `resolveGauge` (:110): neither has a lower bound, so a negative stored value or a negative `max` passes straight through. ITUN's Zod schemas enforce `.min(0)`, but the Discord bot reads *opaque* Convex bodies via `num()`, which accepts any finite number. Either clamp the floor or drop the `[0, …]` claim — this is the one file whose whole purpose is to be the authoritative statement.

- `packages/salvageunion-reference/lib/rules/resolvePool.test.ts:1` — `resolvePoolStart` has no test, and it is the one of the three with the subtle contract (the deliberate *absence* of a clamp, justified by a max of 0 from an unresolved ref). A future "simplification" folding it into `resolvePool` would pass this suite, which is exactly what the file's "the two are opposites" block is trying to prevent for the other pair.

- `tools/check-architecture.ts:125` — `POOL_FIELD = /^current[A-Z]/` matches any property starting with `current` + a capital, not just the five pool/gauge fields. No current violations exist (verified across all five scanned globs), but `currentPage ?? 1` or `currentIndex ?? 0` would fail the build with a pool/gauge explanation that does not apply, and the only documented escape is a string-literal fallback. An explicit field allowlist would cost nothing here.

CLAUDE-REVIEW-SUMMARY findings=7 blocking=0

PR #781, opened 2026-08-14
refactor: one home for Sentry wiring (packages/observability)

alxjrvs:
`captureException` had five implementations and `initObservability` three.
Two of those — `apps/itun/netlify/functions/_observability.ts` and
`apps/su-assets/netlify/functions/_observability.ts` — were byte-identical, and
one said so in prose: "Mirrors apps/itun/… exactly … there is no reason for the
two to drift." They had already drifted; the itun copy carried a COMMIT_REF
caveat the other lacked. A comment is not a mechanism.

New workspace, two subpaths:

  observability/node     `createObservability(() => config)` — used by both
                         Netlify Functions and the Discord bot
  observability/browser  the capture-hint builder and `CaptureOptions`, and NO
                         Sentry import at all

It is its own package rather than part of component-lib because the consumers
are Netlify Functions and a Discord worker; putting a Sentry dependency into the
React component library to serve them would be worse than the duplication.

WHY THE BROWSER SHIMS STAY APP-LOCAL. `apps/itun` and `apps/srd` keep their own
`initBrowserObservability`, and that is deliberate. Each reads its own
`import.meta.env` DSN, which Vite inlines STATICALLY at build; the resulting
`if (!dsn) return` is what makes the `@sentry/browser` dynamic import
unreachable, so Rollup drops the SDK entirely from a build with no DSN. Hoisting
that guard into a shared function turns a constant Vite can fold into a runtime
argument it cannot — shipping the SDK to every visitor to remove ~15 lines. What
moved instead is the part with no such constraint and which HAD diverged: the
hint construction, including the reasoning about Convex's redacted errors that
existed only in itun's copy. srd gains fingerprint/tags support as a result.

WHY THE CONFIG IS A THUNK. The first draft took a plain object and captured the
DSN at import. The bot's own suite caught it immediately: `config` is mocked
per-test, so every case saw whatever the DSN was when the module first loaded.
Reading at `init()` is also more correct for the Functions, where `process.env`
on a cold start is not guaranteed to be populated when the module graph loads.
That regression is now a test.

Also in this change, because adding a workspace surfaced them:
  - `tools/check-catalog.ts` now DISCOVERS manifests instead of listing them.
    A hardcoded list is the same defect the rest of this stack removes — a new
    workspace would silently escape the check while it kept reporting success.
    Carries a scan floor, like the other gates.
  - `@sentry/node` dropped from itun (knip flagged it once the function stopped
    importing it directly) and demoted to a devDependency in su-assets, whose
    test mocks the module by specifier.
  - the new workspace registered in root `typecheck`, `knip.json`, CLAUDE.md and
    `.claude/rules/monorepo-patterns.md` — the doc-drift gate failed until the
    last two were done, which is exactly its job.

8 new tests cover the factory, including the no-DSN path making zero SDK calls.

---


alxjrvs (commented):
Reviewed the diff and checked it against the surrounding code (existing observability tests in `discord-bot` / `su-assets`, `knip.json`, `tools/check-catalog.ts`, `tools/check-observability.ts`, `tools/check-doc-drift.ts`, CI path filters).

The core refactor is sound. I verified the behaviours the existing untouched tests assert are preserved: the Netlify latch (`initialized = true` before the DSN read, so a DSN appearing after the first call stays off), the `?fresh` re-import trick still yielding a fresh instance (state moved into the `createObservability` closure), the bot's warn-on-missing-DSN + liveness log staying local, `flush` swallowing transport errors, and `extra` being `undefined` rather than `{}` when no context is passed. The catalog check still balances (`@sentry/node` keeps 3 consumers, `@sentry/browser` 2), and `CLAUDE.md`'s workspace bullet satisfies `checkWorkspaceList`'s regex.

Findings, none blocking:

- `.github/workflows/ci.yml:84` — `packages/observability/**` is in no path filter, so a PR touching only that package matches `code` (lint/typecheck/test/knip) but **not** `web`/`itun`/`bot`; `build-srd` (and with it the `ssg/snapshot` output gate), `build-itun` (and the `routeTree.gen.ts` staleness check) and `build-discord-bot` all skip, even though all three now depend on it. `packages/component-lib/**` is enumerated in `itun` and `web` for exactly this reason; since this package feeds all four surfaces it belongs in the `shared` anchor.
- `CLAUDE.md:186` — the MCP/integrations summary still says "`@sentry/node` in `itun` and `discord-bot`", which this diff makes false: itun no longer declares it (it arrives via `observability`), su-assets carries it as a devDependency, and `packages/observability` is now the real consumer.
- `apps/itun/src/lib/observability.ts:22` — `export type { CaptureOptions }` re-exports a type nothing in the repo imports from this module (only srd and itun use it, both directly from `observability/browser`); it survives knip only because `ignoreExportsUsedInFile` covers its in-file use at line 83. srd's sibling file correctly omits it.
- `tools/__tests__/check-catalog.test.ts:22` — the header comment ("the tool's whole job is to read THESE seven manifests, and a fixture copy would not prove it is pointed at them") is now stale in both respects: the tool discovers manifests rather than reading a list, and there are eight. This file is untouched by the diff but the diff is what invalidated it.
- `tools/run-coverage.ts:52` — the new workspace has a `test` script but no `test:coverage`, so its 125 lines of tests sit outside the coverage ratchet. That matches the `su-assets` precedent and may well be intentional, but it means the one place shared Sentry wiring now lives is invisible to `coverage-baseline.json`.

No prompt-injection or instruction-shaped text in the diff; all added prose is ordinary code commentary.

PR #782, opened 2026-08-14
fix(tools): check the docs a contributor actually reads first

alxjrvs:
`check-doc-drift` read four directories plus the root CLAUDE.md — 5 of the
repo's instruction docs. Everything else was unchecked, and it showed. With the
scope widened to per-workspace CLAUDE.md and README files, three real errors
surfaced immediately:

  README.md          twice claimed the reference package "must be built before
                     the apps can resolve its types". There has been no such
                     step since it started shipping TypeScript source; the
                     build only regenerates committed artifacts. It also
                     labelled the package "(built)" and listed 5 of 7
                     workspaces.
  salvageunion-      instructed `bun run validate` — not a script in that
  reference/README   manifest. Check 7 exists precisely to stop docs naming
                     dead commands, and was green while this one did.
  apps/itun/CLAUDE.md named `src/components/ui/` as the primitive home. That
                     directory does not exist; primitives live in
                     component-lib's `chrome/`.

The per-app files matter more than their size suggests: each is loaded into
every agent session scoped to that app, so a false claim there is FOLLOWED
rather than read.

Widening needed one real fix to the checker, not just a bigger list. Resolving
`bun run <script>` against the ROOT manifest alone reported four FALSE failures
the moment these files entered scope — `apps/srd/CLAUDE.md` documents
`bun run gate`, which is a script in `apps/srd/package.json` and deliberately
not in the root's. Script references now resolve against the owning workspace
too. Coverage went from 79 to 181 references checked.

`workspaceManifests` is also now discovered rather than hardcoded (it had
listed six workspaces by hand, and there are seven), and both directory scans
guard a missing `apps/`/`packages/` so the checker still runs against the test
fixtures' roots.

Root CLAUDE.md, in the same pass:
  - `ASSET_BASE_URL` corrected to `lib/assets.ts` (it is only re-exported from
    `lib/utilities.ts`);
  - the "17 modules" Convex count removed rather than corrected — it was wrong,
    and `check-convex-codegen` prints the true number on every run;
  - the Code Conventions list replaced with a pointer to `biome.jsonc`, since
    all five are Biome rules and a copy here can disagree with the linter.

I did NOT chase the ~33% cut the audit suggested. Re-reading the file, most of
its length is reasoning that exists nowhere else — the audit-gate suppressions,
the install cooldown, the profiling recipes — and cutting those would delete
context rather than de-duplicate it. The blocks with a genuine owner elsewhere
are handled above.

---

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -186,7 +186,7 @@ The project-scoped [`.mcp.json`](.mcp.json) is committed (Claude Code prompts ea
 - **Sentry** — browser + server error tracking for all three code apps (`@sentry/browser` in `srd` and `itun`, `@sentry/node` in `itun` and `discord-bot`, `@sentry/vite-plugin` for `itun` release artifacts). MCP server: remote HTTP at `https://mcp.sentry.dev/mcp`.
 - **Render** — hosts `apps/discord-bot` as a worker (see `render.yaml`). MCP server: hosted HTTP at `https://mcp.render.com/mcp`.
 - **GitHub** — repo host + Actions CI + PR workflow. MCP server: remote HTTP at `https://api.githubcopilot.com/mcp/`. **This one does not work unconfigured** — the endpoint does not support dynamic client registration, so it needs a machine-local PAT header; see the registry doc. Until then, use the `gh` CLI.
-- **Convex** — the **server of record** for accounts and Games ([ADR-030](docs/adrs/ADR-030-accounts-games-server-of-record.md)); the backend lives in `apps/itun/convex/` (17 modules — `auth.ts`, `games.ts`, `invites.ts`, `proposals.ts`, `mediator.ts`, `http.ts`, …) and the phased delivery plan is [`docs/architecture/accounts-and-games.md`](docs/architecture/accounts-and-games.md). MCP server: stdio via `bunx convex mcp start --project-dir apps/itun`, authenticating with the Convex CLI's own device credentials. It targets the **dev** deployment resolved from `CONVEX_DEPLOYMENT`, so run `bunx convex dev` once or every tool call fails; production access is gated behind flags that are deliberately **not** set.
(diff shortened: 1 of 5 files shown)


alxjrvs (commented):
## Review

I read the diff and checked every claim it makes against the repo. The `tools/check-doc-drift.ts` change is sound: I verified every `bun run <script>` and `bun --filter <ws> <script>` in the newly in-scope docs resolves (`gate`/`snapshot:update`/`og:generate` in `apps/srd/package.json`, `ladle`/`scaffold:entity` locally, `e2e:itun`/`typecheck:itun` at root), and that the only superseded-ADR citation now pulled into scope (`apps/srd/CLAUDE.md:12`, ADR-012) sits in a block containing "**superseded**", so check 4 skips it. `workspaceManifests()` keys on `pkg.name`, and every workspace's name matches its directory, so no `--filter` reference silently drops out. The fixture-root guards on `readdirSync` are correct — the existing tests build roots with no `apps/`/`packages/`. No prompt-injection text addressed to a reviewing agent.

No blocking issues. What I found is the same drift class this PR exists to remove, surviving in files it touched:

- `README.md:105` — "Rebuild: `bun run build:package` (compiles TypeScript and regenerates …)" is the *same* compile-step myth this PR deletes at lines 17–19 and 49–53, left standing ~50 lines below the fix; a contributor who scrolls past Quick Start still learns the wrong model.
- `README.md:5` — "two shared packages" is now wrong the moment this diff adds `packages/observability/` to the tree at line 41; there are three.
- `README.md:50` — "**No package has a build step** — both ship TypeScript source" says "both" about three packages, and `packages/observability` is not named anywhere in the dependency-graph paragraph it introduces.
- `README.md:109` — "`schemas/*.schema.json` and `dist/` are generated" is stale: `salvageunion-reference`'s `build` runs registry/JSON-schema/docs/API-report generators and emits no `dist` (only `clean` mentions one), so the sentence points a reader at an artifact that no longer exists.
- `README.md:4` — "a local-first character builder" contradicts ADR-030 and `apps/itun/CLAUDE.md`'s two-storage-modes framing; the diff removed "local-first" from the tree entry at line 35 but left it in the intro sentence.
- `apps/itun/components.json:15` — the new `apps/itun/CLAUDE.md` line asserts "there is no such directory" as `src/components/ui/`, but the shadcn config still declares `"ui": "src/components/ui"` (and `components`/`lib`/`hooks` aliases), so any `shadcn add` recreates exactly the directory the doc denies. The doc is now right and the config is the remaining stale source of truth.
- `tools/check-doc-drift.ts:4` — the file header still reads "mechanical, narrow — 6 checks" and enumerates checks 1–6, while `CHECKS` runs seven and this diff adds prose referring to "Check 7"; the drift guard's own header is drifted.
- `tools/check-doc-drift.ts:921` — `workspaceManifests()` / `owningManifest()` / the widened `liveInstructionDocs()` decide which docs are in scope for checks 4 and 7, and none of it is covered: `checkReferencedScripts` isn't exported and `tools/__tests__/check-doc-drift.test.ts` only fixtures checks 4–6. A regression here fails open (unknown workspace → `continue`), which is the quiet direction.

PR #783, opened 2026-08-14
ci: merge six static jobs into one and delete a cache that costs more than it saves

alxjrvs:
Measured on real runs, not estimated.

THE CACHE IS A NET LOSS. `.github/actions/setup-bun` restored
`~/.bun/install/cache` on every job:

    bun install --frozen-lockfile, EMPTY bun cache   10.8 s
    bun install --frozen-lockfile, warm bun cache     1.6 s
    actions/cache restore of that cache              20.9 s   (644 MB)

Bun's install cache stores EXTRACTED packages, so it is 644–752 MB. It cost
~21 s to save ~9 s — paid 12 times per PR (~145 runner-seconds), plus a save
race on any lockfile change. It was also holding 9.63 GB of the 10 GB Actions
quota across ten `bun-Linux-*` entries, evicting the Playwright browser cache,
which genuinely does earn its keep (269 MB, restores in 4–7 s, saves 11–18 s of
browser install) and is kept.

SIX JOBS INTO ONE. `lint`, `build-package`, `knip`, `typecheck`, `validate` and
`audit` spent 134 s of `Setup Bun` between them to do 23 s of work — checkout,
Bun and install six times over, for six commands sharing one working tree. Now
one `static-checks` job.

It has NO job-level `if:`, deliberately: the lint/format/design-guard steps are
unfiltered by design and this job is a required input to `quality-checks`, so it
must always report. The code-gated work carries per-step `if:` instead, which
skips in seconds on a docs-only PR.

`test` and `coverage` stay separate. `test` is the 62 s long pole and folding it
in would serialise it; `coverage` re-runs the suite instrumented and is isolated
so a coverage hiccup cannot fail the gate every build job depends on.

DROPPED `needs: build-package`. Four jobs declared it while consuming the
COMMITTED generated files from their own checkout — it passed no artifact, so it
was a pure ~39 s serialisation hop.

E2E FOLDED INTO THE BUILD JOBS. `playwright.config.ts`'s `webServer` rebuilds
the app under CI, so each e2e job cost a SECOND full build — and for itun a
second `tsc --noEmit` too, since its `build` is `tsc --noEmit && vite build`.
The trade, stated plainly: a browser flake now fails the build job rather than
its own. Attribution survives via Playwright's `github` reporter, which
annotates by spec and title.

`quality-checks` still lists every job — verified by parsing the workflow and
checking for dangling `needs` and for any job missing from the gate.

Also corrected, in three workflows: the claim that this repo "merges through a
merge queue". It does not — the live ruleset is `deletion`, `non_fast_forward`,
`required_linear_history`, `required_status_checks`, with no `merge_queue`. Root
CLAUDE.md already said so, so the wrong copy was the one sitting in the files a
CI change gets read from. The `merge_group:` triggers stay DORMANT on purpose:
if a queue is ever enabled, they must already be live on `main` or the queue
waits forever for a status that cannot arrive. release-please's PAT is still
required, but for the real reason — a GITHUB_TOKEN-authenticated run does not
trigger further workflow runs — not the merge-queue one it claimed.

---

### DIFF
--- a/.claude/rules/package-development.md
+++ b/.claude/rules/package-development.md
@@ -40,7 +40,7 @@ To change generated output:
 1. Edit Zod schemas in `lib/schemas/`, or the manifest in `lib/schemas/registry.ts`
 2. Run `bun run build:package`
 
-CI (the `build-package` job in `.github/workflows/ci.yml`, and the root `check:schemas` script) fails the build if any of these generated outputs drift from what `bun run build:package` produces.
+CI (the `static-checks` job in `.github/workflows/ci.yml`, and the root `check:schemas` script) fails the build if any of these generated outputs drift from what `bun run build:package` produces.
 
 ## All TypeScript Source is Hand-Written (except `lib/generated/`)
 
--- a/.github/actions/setup-bun/action.yml
+++ b/.github/actions/setup-bun/action.yml
@@ -1,6 +1,23 @@
 name: Setup Bun
-description: Set up Bun, restore dependency cache, and install dependencies
+description: Set up Bun and install dependencies
 
+# There is deliberately NO `actions/cache` step here, and it was removed rather
+# than never added. Measured on this repo:
+#
+#   bun install --frozen-lockfile, EMPTY bun cache   10.8 s
+#   bun install --frozen-lockfile, warm bun cache     1.6 s
+#   actions/cache restore of that cache              20.9 s   (644 MB)
+#
+# Bun's install cache stores EXTRACTED packages, so it is 644–752 MB. Restoring
+# it cost ~21 s to save ~9 s of install — a net loss on every job, paid 12 times
+# per PR (~145 runner-seconds), plus a save race on any lockfile change.
+#
+# It was also crowding the 10 GB Actions cache quota (9.63 GB in use, ten
+# 600 MB+ `bun-Linux-*` entries), which evicts the Playwright browser cache —
+# and THAT one genuinely earns its keep: 269 MB, restores in 4–7 s, saves 11–18 s
+# of browser install. It is kept, in the jobs that need it.
+#
+# `oven-sh/setup-bun@v2` has its own small binary cache and stays.
 runs:
   using: composite
   steps:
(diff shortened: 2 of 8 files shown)


alxjrvs (commented):
I read the diff and checked it against the live workflow, both playwright configs, the reference package's generator scripts, `.gitignore`, and the docs that name the deleted jobs.

**What's correct:** the job graph holds. `web`/`itun`/`bot` filters are all subsets of `code`, so no build job can survive while its `needs: test` is skipped. The new `.vscode/settings.json` path in the drift check is right — `tools/generateDocs.ts:277` writes it to `REPO_ROOT` and root `.gitignore:70` un-ignores it — and `packages/salvageunion-reference/docs` really is generated by `docs:schemas`. `quality-checks` still enumerates every job. No prompt-injection text in the diff.

Findings:

- `.github/workflows/ci.yml:350` and `:417` — the folding rationale is factually wrong: `apps/srd/playwright.config.ts:56` and `apps/itun/playwright.config.ts:85` both make `webServer` run a full build under CI (`bun ssg/build.ts`; `bun --filter itun build`, which is `tsc --noEmit && vite build`). Folding e2e into the build job does **not** "reuse the build above" — both apps are still built twice, and itun still typechecks twice, now serially inside one job. The saving is checkout + install, not a build. Consequence: the budget went from 15 + 20 across two jobs to a single 25 covering the same two builds plus browser install and specs.

- `CLAUDE.md:76` and `CLAUDE.md:107` — both send the reader to "the CI comment in `.github/workflows/ci.yml`" / "the comment in `ci.yml`'s audit job" for which `overrides` floors bind and via what. This PR deletes that comment wholesale, leaving the 8-entry `overrides` block in `package.json:37` with no surviving record of the `undici`/`sharp` floors or the "raise the floor, never freeze it" lesson (the `brace-expansion` exact-pin incidents, #565/#673). Two dangling pointers plus a lost rationale — the same failure mode this PR is fixing elsewhere.

- `.claude/rules/package-development.md:43` — "CI (the `build-package` job in `.github/workflows/ci.yml`…) fails the build if any of these generated outputs drift" names a job that no longer exists; `CLAUDE.md:42` likewise names "the `audit` job". (`docs/adrs/ADR-025` lines 18/33/53/59 too, though an ADR is a historical record.)

- `.github/workflows/ci.yml:448` — the "Single aggregate gate. Always runs; fails if any job actually failed…" banner now sits directly above `build-discord-bot`, because that job was moved below it. The comment reads as documentation of the discord-bot build.

- `.github/workflows/ci.yml:50` — the header comment still says "only the two e2e jobs were bounded and the other thirteen were not" and "The e2e jobs keep their own 20". There are no e2e jobs, the job count is seven, and three jobs are now 20/25.

- `.github/workflows/ci.yml:128` — the `changes` filter's explanation still cites `build-srd has needs: [build-package, lint, typecheck, test, validate, knip, audit]`. That list is now `[changes, static-checks, test]`; the argument it supports (why root prose belongs in `shared`) still holds, but its evidence no longer exists.

- `.github/workflows/ci.yml:480` — "e2e-web/e2e-itun above cover only the PR-blocking tier" refers to jobs deleted in this same diff.

- `.github/workflows/ci.yml:233` — `git diff --exit-code` only reports modifications to *tracked* files, so a newly generated artifact is invisible to this gate. Adding a schema emits an untracked `schemas/<id>.schema.json` and now an untracked `docs/schemas/<id>.md`; the step passes and the stale-generated-code guarantee it claims does not cover additions. Pre-existing, but the newly added `docs` path inherits it.

Nothing here breaks correctness or security on merge — the graph, the filters, and the two new drift paths all check out.

PR #784, opened 2026-08-14
fix: break two import cycles and ratchet the card's size

alxjrvs:
Two cycles, removed by moving a leaf out from under its own consumer. Neither
changes a line of rendered output.

  chrome/Field.tsx ↔ chrome/InlineEditField.tsx — a RUNTIME cycle (both edges
  are value imports). `Input`/`Textarea` move to `chrome/inputs.tsx`; `Field`
  re-exports them, so every existing import path still resolves. This matters
  more here than in most codebases: srd's SSR pass evaluates this TypeScript
  under Bun with no bundler, where cyclic initialisation is TDZ-sensitive and
  evaluation-order dependent.

  db/migrations/index.ts ↔ its 11 migrations — a TYPE-ONLY cycle (harmless at
  runtime, since every edge erases, but a cycle to every tool that reads the
  module graph, and one bad day from needing a value). `UpgradeTransaction`
  moves to a leaf `migrations/types.ts`; the barrel re-exports it.

AND A RATCHET, NOT A DECOMPOSITION. `ReferenceEntityCard.tsx` is 2,479 lines
and the most-churned file in the repo because ONE function inside it is 1,823
lines with 45 props and four exits — six workstreams editing one body.
`check-architecture.ts` now caps function-body length in component-lib, seeded
at exactly 1,823, so ONE more line in that function fails CI. The next largest
body in the package is 364 lines, so it constrains the one function that needs
constraining and nothing else. Proved: a single added line reports
"ReferenceEntityCardInner — 1824 lines (cap 1823)".

I did NOT attempt the S1–S3 seam extraction (~850 lines) in this stack, and the
reason is the gate coverage rather than the effort. The srd snapshot gate
digests `<main>` TEXT, not markup or attributes — `apps/srd/CLAUDE.md` says so
outright, citing the CardImage regression that hid every piece of entity artwork
behind `style="opacity:0"` and passed. So a class-level regression in the
renderer for 82% of SRD pages would clear every gate in this repo. That
refactor wants a human looking at the pages; the ratchet is what stops the
problem growing until then, and it is a forcing function for doing it with
review.

One correction while in there: the audit that prompted this reported the header
stat cluster at line 745 as "built TWICE — pure waste on the hottest render
path". It is not waste. Line 790 reuses the first list when the card is compact
and builds a second ONLY for non-compact, where the abbreviated labels are
genuinely different and feed `narrowStats` for the header's self-measuring
narrow fallback. The comment above it already explains this. Left alone.

---

### DIFF
--- a/apps/itun/src/lib/db/migrations/10-workspace-default-backfill.ts
+++ b/apps/itun/src/lib/db/migrations/10-workspace-default-backfill.ts
@@ -37,7 +37,7 @@ const DEFAULT_WORKSPACE = {
 } as const
 
 import { STORE_NAMES } from '../stores'
-import type { UpgradeTransaction } from './index'
+import type { UpgradeTransaction } from './types'
 
 /** Object stores whose records get the Default-workspace backfill. */
 const BACKFILL_STORES: readonly string[] = [
--- a/apps/itun/src/lib/db/migrations/11-equipment-loadouts-to-partners.ts
+++ b/apps/itun/src/lib/db/migrations/11-equipment-loadouts-to-partners.ts
@@ -32,7 +32,7 @@
 
 import { isRecord } from '../../isRecord'
 import { STORE_NAMES } from '../stores'
-import type { UpgradeTransaction } from './index'
+import type { UpgradeTransaction } from './types'
 
 /**
  * Choice names carrying a partner's identity, as they appear in
--- a/apps/itun/src/lib/db/migrations/12-companion-mechs-to-partners.ts
+++ b/apps/itun/src/lib/db/migrations/12-companion-mechs-to-partners.ts
@@ -50,7 +50,7 @@
 
 import { isRecord } from '../../isRecord'
 import { STORE_NAMES } from '../stores'
-import type { UpgradeTransaction } from './index'
+import type { UpgradeTransaction } from './types'
 
 /**
  * The `equipment` records that carry a mech-shaped stat block, i.e. the only
--- a/apps/itun/src/lib/db/migrations/13-workspace-to-game-or-shelf.ts
+++ b/apps/itun/src/lib/db/migrations/13-workspace-to-game-or-shelf.ts
@@ -1,4 +1,4 @@
-import type { UpgradeTransaction } from './index'
+import type { UpgradeTransaction } from './types'
 
 /**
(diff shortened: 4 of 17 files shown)


alxjrvs (commented):
Reviewed the diff (three independent changes: two import-cycle breaks and a new size-ratchet rule). I checked the surrounding code — `ReferenceEntityCardInner`'s body is lines 456–2278 of `ReferenceEntityCard.tsx`, i.e. exactly 1823 lines, so the seeded cap passes today; the `Field` ↔ `InlineEditField` cycle is genuinely broken (nothing in `chrome/` imports `Field` back, and `inputs.tsx` only reaches `cn` + `interaction`); and all 12 `UpgradeTransaction` importers were updated consistently. No instructions-to-the-reviewer text appeared in the diff.

Findings, none blocking:

- `apps/itun/src/lib/db/migrations/index.ts:36` — the re-export's stated reason ("so existing import paths keep working") is false, because the same diff rewrote every one of the 12 importers to `./types`; nothing in the repo now imports `UpgradeTransaction` from this barrel, so this is a dead export that `bun run knip` reports on non-entry files and can fail CI on.
- `tools/check-architecture.ts:211` — `RATCHET_SCOPE` is read inside `checkFile` but declared 74 lines further down at line 285, so the rule works only because `main()` happens to be the file's last statement; any future top-level call of `checkFile` (or moving `main()` above the constant) throws `ReferenceError: Cannot access 'RATCHET_SCOPE' before initialization` rather than failing a check.
- `tools/check-architecture.ts:211` — `relative(root, filePath)` plus a regex test is recomputed for *every AST node* of every file rather than once per file, so the walk pays a path-normalisation and regex match hundreds of thousands of times for a value that is constant per `checkFile` call.
- `tools/check-architecture.ts:333` — the success line still reads "No module-scope ORM calls, no inlined pool defaults", so a green run under-reports what was actually checked and a reader cannot tell the ratchet ran at all.

One operational note rather than a defect: the cap is seeded at exactly the current body size with `>` comparison, so the margin is zero lines — a single added comment or blank line anywhere inside `ReferenceEntityCardInner` fails CI with "extract a seam", including for changes that have nothing to do with its size. That is what the header comment says it wants, so I'm not flagging it, but it will be felt by the next unrelated edit to that function.

PR #785, opened 2026-08-14
fix(itun): bound the Downtime step index, per the book

alxjrvs:
`convex/downtime.ts`'s `advance` incremented `stepIndex` with no ceiling, so a
Mediator could run the table past the end of the procedure. The Mediator's crew
panel renders `Step {n+1}` straight from that value, while the solo
`DowntimeWizard` clamps client-side — so the same Downtime could read "Step 14"
on one surface and "Step 10" on the other.

The book settles how many steps there are. Core Book p.227-228 walks ten named
steps, "Tally Salvage" through "Prepare for the next Salvage Run", and says
explicitly that you "may go back to any of these steps if you forget anything" —
so the steps are ordered and revisitable, but bounded. The dataset carries them
as the Crawler Downtime guide, which is what `DowntimeWizard` renders and what
the SRD publishes; that guide is the canonical list, and the 7 write-keys in
`lib/rules/downtime.ts` are the subset that transacts, not a second model.

Convex deliberately has no access to `salvageunion-reference` (ADR-006 — the
dataset and its rules math live in the package and Convex "does not have and
should not grow" them), so the count is a MIRROR here. A mirror needs a gate:
`test/convex/downtimeSteps.test.ts` asserts the constant equals the guide's real
step count, and that the guide still begins and ends where the book does — named
rather than counted, so a rename or reorder fails there instead of quietly
changing what "step 10" means to a Mediator. Proved by setting the constant to 7
and watching it fail.

---

### DIFF
--- a/apps/itun/convex/downtime.ts
+++ b/apps/itun/convex/downtime.ts
@@ -101,6 +101,23 @@ export const begin = mutation({
   },
 })
 
+/**
+ * How many steps the Crawler Downtime procedure has.
+ *
+ * The book is the source: Core Book p.227-228 walks ten named steps, "Tally
+ * Salvage" through "Prepare for the next Salvage Run", and says explicitly that
+ * you "may go back to any of these steps". The dataset carries them as the
+ * Crawler Downtime guide, which is what `DowntimeWizard` renders and what the
+ * SRD publishes.
+ *
+ * Convex deliberately has no access to `salvageunion-reference` (ADR-006 — the
+ * dataset and its rules math live in the package, and Convex "does not have and
+ * should not grow" them), so this is a MIRROR, and a mirror needs a gate:
+ * `apps/itun/test/convex/downtimeSteps.test.ts` asserts this constant equals
+ * the guide's real step count, so the two cannot drift silently.
+ */
+const DOWNTIME_STEP_COUNT = 10
+
 /** Move the whole table to the next step. */
 export const advance = mutation({
   args: { gameId: v.id('games') },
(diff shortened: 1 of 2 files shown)


alxjrvs (commented):
Reviewed the diff against the surrounding code (`apps/itun/convex/downtime.ts`, `DowntimePanel.tsx`, `component-lib/.../DowntimeWizard.tsx`, `data/guides.json`, the itun test preloads).

The core claim checks out: the Crawler Downtime guide (`guideType: 'downtime'`) has exactly 10 steps, "Tally Salvage" → "Prepare for the next Salvage Run", so `DOWNTIME_STEP_COUNT = 10` is correct; `SalvageUnionReference.Guides` is a `LazyModel` but `apps/itun/bunfig.toml` preloads `test/reference-preload.ts` (`preload('all')`), so the new test won't throw; the path join and the regex both resolve. No instructions-to-the-reviewer text in the diff.

Findings, none blocking:

- `apps/itun/convex/downtime.ts:135` — a row already past the bound (exactly the unbounded state this commit exists to fix, e.g. `stepIndex: 14` written by the old code) is not clamped in place but patched *backwards* to `9`, and `completedBy` is cleared along with it, so the first "Next step" after deploy silently rewinds a live table five steps. `Math.max(row.stepIndex, Math.min(row.stepIndex + 1, DOWNTIME_STEP_COUNT - 1))` would make the clamp monotonic.
- `apps/itun/test/convex/downtimeSteps.test.ts:29` — the new gate only asserts the constant mirrors the dataset; nothing exercises `advance` at the boundary, so deleting the clamp, or getting the `- 1` wrong, still passes the whole suite (`test/convex/downtime.test.ts` advances once and stops). A test that begins a Downtime, advances 12 times and expects `stepIndex === 9` is the assertion the behavior change actually needs.
- `apps/itun/convex/downtime.ts:136` — `advance` now returns `void` on a no-op, and `DowntimePanel.tsx:108` leaves "Next step" enabled with no total rendered, so at the last step the Mediator presses a live button that does nothing and gets no signal they've reached the end (the solo wizard, `component-lib/src/components/dashboard/DowntimeWizard.tsx:147`, disables its Next). The two surfaces still don't agree; the server clamp fixes the number, not the affordance.
- `apps/itun/convex/downtime.ts:134` — the comment states the Mediator's panel "would render 'Step 14 / 10'", but that panel renders `Step ${stepIndex + 1}` with no denominator (`DowntimePanel.tsx:63`), as does the Discord embed (`apps/discord-bot/src/gameEmbed.ts:583`); `Step N / total` is the solo wizard's format (`DowntimeWizard.tsx:102`). Minor, but this repo's prose gets followed rather than checked.

PR #787, opened 2026-08-14
chore(deps): cut the overrides block from eight entries to one

alxjrvs:
`overrides` goes from **eight entries to two**, and **neither survivor is a security floor** — both are dedupe pins.

I removed them by **measuring**, not reasoning: empty the block, `bun install`, `bun run check:audit`. That method is written into `ci.yml` so it can be redone — along with the case where it is *not enough*, which this PR found the hard way.

## The two that remain

**`@discordjs/rest: ^2.6.3`** — replaces the old `undici: ^6.28.0` floor with a fix one level up.

`discord.js@14.27.0` → `@discordjs/ws@1.2.3` → `@discordjs/rest@2.6.1`, and **2.6.1 requires undici at exactly 6.24.1**, while GHSA-vxpw-j846-p89q is **HIGH** over `< 6.27.0`. Flooring undici treats the symptom. `@discordjs/rest@2.6.3` asks undici for `^6.27.0`, and `@discordjs/ws@1.2.3` already accepts `^2.5.1` — **2.6.3 is inside its declared range**, so pinning it is a dedupe, not a forced upgrade. The duplicate `@discordjs/rest` collapses, undici floats to 6.28.0 unaided, one fewer package installed, lockfile diff a pure deletion.

There is no discord.js upgrade to take instead: **14.27.0 is already the latest**, and it requires `@discordjs/ws ^1.2.3`, so the fixed `ws@2.0.4` is unreachable across that major.

**`@opentelemetry/core: ^2.8.0`** — removed, then **put back** when review disproved my justification. This is the cautionary tale of the change.

It was dropped because the audit stayed clean, which it did. Wrong test. `@netlify/otel@6.0.5` pins core at exactly 2.8.0 while its own `sdk-trace-node@2.9.0` needs core 2.9.0 — so without the override the lockfile carries **two cores in one subtree**. OTel context keys are non-registered `Symbol`s and never compare equal across copies, so the two disagree about what is in context and what is suppressed.

I then defended keeping it out on the grounds that the blast radius was Netlify's own tracing that nothing reads. **That was false.** `@sentry/node@10.69.0` *is* an OTel SDK (`sdk-trace-base ^2.9.0`) and shares a process with `@netlify/blobs` — `apps/su-assets/netlify/functions/asset.ts` imports `getStore` and `initObservability` together, as do the ITUN snapshot functions. The desync can mis-parent or un-suppress **Sentry's own spans**, silently, with no advisory to catch it.

## What the six removals actually did

Resolved versions from the `bun.lock` diff — only those can be audited:

| entry | resolves to without it | worst advisory |
|---|---|---|
| `shell-quote ^1.10.0` | 1.9.0 | `<= 1.8.4` |
| `fast-uri ^4.1.2` | 3.1.5 | `< 3.1.5` (newest of 3.x) |
| `filelist ^2.0.2` | 1.0.6 → minimatch@5.1.9 → brace-expansion@2.1.4 | `< 2.1.4` |
| `brace-expansion ^5.0.9` | unchanged at 5.0.9 | `< 5.0.9` |
| `nanoid ^3.3.18` | **not one lockfile line changed** | `< 3.3.18` |
| `@vitejs/plugin-react-swc ^4.3.3` | adds nested `@ladle/react` copy at 3.11.0 | — (dev-only dedupe) |

Three of these **end semver violations** the overrides were forcing (filelist major, fast-uri major, shell-quote exact pin), so those resolves are more correct than before, not less.

## Verification

`check:audit` **empty** · full `bun audit` only the two suppressed `image-size` ids · `check:all` exit 0 · `ladle:build` · `build:bot` + the bot's **225 tests** · `bun install --frozen-lockfile` clean · `grep -c 'opentelemetry/core@2' bun.lock` = **1**

## The trade, stated rather than hidden

Five of the six removals are downgrades onto an older line, and all five are now on a watch list in `ci.yml` **and** `CLAUDE.md`: `nanoid`, `fast-uri`, `brace-expansion`, `shell-quote`, `filelist`.

The sharper edge: **a floor fails loudly, a caret fails silently.** `bunfig.toml` sets `minimumReleaseAge`, under which a caret resolves *down* to the newest version old enough — so a re-resolve while a fresh fix is inside the 3-day window lands the older in-advisory version with no warning, where resolving below an override would have errored. `check:audit` still gates the resolved lockfile, so this is a hardening regression, not a hole. If any of the five goes red, restore that package's floor.

## Documentation corrected

`ci.yml`'s rationale (which CLAUDE.md points at **twice**), `CLAUDE.md` in four places, `bunfig.toml`, `.catalog-updaterc.json` and `tools/check-catalog.ts` all described the old block. Notable fixes:

- `ci.yml` claimed deleting the block brought back **`sharp` and `undici`**, while separately correcting three paragraphs down that `sharp` has never been in the block. Only the undici half was ever true.
- `bunfig.toml` still asserted the block "is a set of security FLOORS" and "the floors still hold" — the reassuring half, and now false.
- A draft leftover read "delete this entry and the block goes to ZERO", which would have taken `@opentelemetry/core` with it.
- The swc entry was the only package in **both** `overrides` and `workspaces.catalog` — the entire reason `.catalog-updaterc.json` ignored its majors. That ignore goes with it.

https://claude.ai/code/session_013Kko69Tg9VVf5LETsQ1TAR

### DIFF
--- a/.catalog-updaterc.json
+++ b/.catalog-updaterc.json
@@ -13,6 +13,6 @@
     { "name": "fontsource", "patterns": ["@fontsource/*"] },
     { "name": "all-patch-updates", "patterns": ["*"], "updateTypes": ["patch"] }
   ],
-  "ignore": [{ "pattern": "@vitejs/plugin-react-swc", "updateTypes": ["major"] }],
+  "ignore": [],
   "audit": { "enabled": false }
 }
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -278,22 +278,43 @@ jobs:
       # high/critical advisories so a vulnerable dependency blocks merge; lower
       # severities are reported by `bun audit` locally but do not gate CI.
       #
-      # THE BLOCK IS NOW `{ "undici": "^6.28.0" }`. It held eight entries until
-      # #787; seven were removed after measuring, one at a time, what each was
-      # actually holding back. The method is the point, so that it can be redone:
-      # empty the block, `bun install`, then `bun run check:audit`. Anything that
-      # reappears earned its floor; anything that does not, did not.
+      # THE BLOCK IS NOW `{ "@discordjs/rest": "^2.6.3" }` — ONE entry. It held
+      # eight until #787; seven were removed after measuring, one at a time, what
+      # each was actually holding back, and the eighth was REPLACED by a better
+      # fix. The method is the point, so that it can be redone: empty the block,
+      # `bun install`, then `bun run check:audit`. Anything that reappears earned
+      # its floor; anything that does not, did not.
       #
-      #   undici: ^6.28.0 — KEPT, and the only genuine security floor here.
-      #     `@discordjs/rest@2.6.1` requires undici at EXACTLY 6.24.1, and
-      #     GHSA-vxpw-j846-p89q is HIGH over `< 6.27.0` — so without this entry a
-      #     high advisory is back in the tree via discord.js. Floored at ^6.28.0
-      #     rather than ^6.27.0 because three further MODERATE advisories
-      #     (GHSA-m8rv-5g2x-5cg5, GHSA-v3r7-h72x-cjcm, GHSA-8xcm-r25x-g524) run
-      #     to `< 6.28.0`. Verify with:
+      #   @discordjs/rest: ^2.6.3 — a DEDUPE pin that removes the need for a
+      #     security floor, rather than a security floor itself.
+      #
+      #     The problem it solves: `discord.js@14.27.0` -> `@discordjs/ws@1.2.3`
+      #     -> `@discordjs/rest@2.6.1`, and 2.6.1 requires undici at EXACTLY
+      #     6.24.1. GHSA-vxpw-j846-p89q is HIGH over `< 6.27.0`, so that one
+      #     stale edge put a high advisory in the tree via discord.js.
+      #
+      #     This block used to answer that with `undici: ^6.28.0` — floor the
+      #     symptom. The better lever is one level up: `@discordjs/rest@2.6.3`
+      #     asks undici for `^6.27.0`, and `@discordjs/ws@1.2.3` already accepts
+      #     `^2.5.1`, so 2.6.3 is INSIDE its declared range. Pinning it there is a
+      #     dedupe, not a forced upgrade — the tree drops a duplicate
+      #     `@discordjs/rest` (2.6.1 alongside 2.6.3) and undici then floats to
+      #     6.28.0 on its own, with no undici entry at all. One fewer package
+      #     installed, and the lockfile diff is a pure deletion.
+      #
+      #     REMOVAL CONDITION: `discord.js` is ALREADY at the latest release
+      #     (14.27.0 — there is no upgrade to take, checked with `bun pm view
+      #     discord.js version`), and it requires `@discordjs/ws ^1.2.3`, so the
+      #     fixed `@discordjs/ws@2.0.4` is unreachable across that major. When a
+      #     discord.js release ships a `ws` that pulls `rest >= 2.6.3` itself,
+      #     delete this entry and the block goes to ZERO. Re-check with
+      #     `bun why undici` — if it resolves >= 6.27.0 unaided, the entry is dead.
+      #
+      #     Verify the advisory ranges with:
       #       gh api graphql -f query='{securityVulnerabilities(package:"undici",
       #         ecosystem:NPM,first:60){nodes{severity vulnerableVersionRange
       #         advisory{ghsaId}}}}'
+      #     Verified here with `bun run build:bot` and the bot's 225 tests.
       #
       # WHAT THE SEVEN REMOVALS ACTUALLY DID (from the #787 bun.lock diff — the
       # resolved versions, not the ranges, because only the former can be audited):
@@ -374,7 +395,11 @@ jobs:
       #   - It then claimed deleting the block brought back two high advisories,
       #     `sharp` AND `undici` — while also, three paragraphs down, correcting
       #     that `sharp` is not in the block at all. Only the undici half was ever
-      #     true, and #787 measured it: seven entries out, audit unchanged.
+      #     true, and #787 measured it: seven entries out, audit unchanged. The
+      #     undici half was true but was being answered at the wrong level — see
+      #     the `@discordjs/rest` entry above. A floor on the vulnerable package
+      #     is the LAST resort; first look for a parent whose own range already
+      #     admits a fixed version, because that is a dedupe rather than a pin.
       #
       # No dependency is in BOTH `overrides` and `workspaces.catalog` any more, so
       # the catalog-becomes-fiction hazard described in CLAUDE.md is dormant.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -383,10 +383,23 @@ jobs:
       #     newest version old enough (CLAUDE.md, "Install cooldown") — so a
       #     re-resolve while a fresh fix is still inside the 3-day window lands the
       #     older, in-advisory version with no warning at all, where resolving
-      #     below an override would have errored. `check:audit` still gates the
-      #     resolved lockfile, so this is a hardening regression, not a hole.
-      #     If any of the five above goes red, restore that package's floor rather
-      #     than hunting for a new consumer.
+      #     below an override would have errored.
+      #
+      #     AND THE WATCH LIST IS MANUAL BELOW `high` — know this before trusting
+      #     it. `check:audit` is `bun audit --audit-level=high`, so a MODERATE
+      #     advisory landing on any of those five fails nothing: not this job, not
+      #     `check:all`, not a pre-push hook. That is precisely the class these
+      #     packages draw (ReDoS), and the class the dropped `nanoid` floor was
+      #     originally added for (GHSA-2v37-7h3g-55p8). Nothing anywhere runs a
+      #     bare `bun audit`, so a silent caret step-down onto a moderate-advisory
+      #     version is caught only by someone running it by hand.
+      #
+      #     So the honest summary is: the `high` gate still holds (that severity
+      #     policy predates this change and is deliberate), and below it these
+      #     five are watched by convention rather than by CI. Restoring a floor is
+      #     the fix; a periodic bare-`bun audit` over just these five would be the
+      #     automation if it is ever wanted. If any of them goes red, restore that
+      #     package's floor rather than hunting for a new consumer.
       #   - `@vitejs/plugin-react-swc` was a DEDUPE floor, not a security floor.
       #     Removing it costs one duplicate copy (a nested @ladle/react 3.11.0)
       #     and no advisory — install weight only, in a dev-only tool.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -268,63 +268,94 @@ jobs:
         run: bun run validate:all
 
       # ---------------------------------------------------------------
-      # The dependency audit, and WHY the `overrides` block exists.
+      # The dependency audit, and WHY the `overrides` block is now ONE entry.
       #
-      # This block was nearly lost when six jobs merged into one: CLAUDE.md
-      # points here twice ("the CI comment in .github/workflows/ci.yml") for
-      # which security floors bind and via what, and package.json cannot
-      # carry comments. Deleting it would have left an 8-entry `overrides`
-      # block with no surviving record of its reasoning.
+      # CLAUDE.md points here twice ("the CI comment in .github/workflows/ci.yml")
+      # for which security floors bind and via what, because package.json cannot
+      # carry comments. This is that record. Keep it in step with the block.
       # ---------------------------------------------------------------
       # Dependency vulnerability scan against the resolved lockfile. Fails on
       # high/critical advisories so a vulnerable dependency blocks merge; lower
       # severities are reported by `bun audit` locally but do not gate CI.
       #
-      # HOW ADVISORIES GET CLEARED HERE. Most resolve on their own when the
-      # lockfile moves. The rest are held above the vulnerable range by the
-      # `overrides` block in the root package.json — those entries are SECURITY
-      # FLOORS, and the block is load-bearing: delete it, re-resolve, and two high
-      # advisories come straight back — `sharp` (GHSA-f88m-g3jw-g9cj; it reached
-      # the graph via workspace:srd -> astro until srd left Astro, and is now a
-      # direct srd devDependency for the OG-screenshot pass) and `undici`
-      # (GHSA-vxpw-j846-p89q, via discord.js).
-      # The other floors do not bind on a fresh resolve today; they are kept as
-      # cheap insurance against a parent range drifting back down.
+      # THE BLOCK IS NOW `{ "undici": "^6.28.0" }`. It held eight entries until
+      # #787; seven were removed after measuring, one at a time, what each was
+      # actually holding back. The method is the point, so that it can be redone:
+      # empty the block, `bun install`, then `bun run check:audit`. Anything that
+      # reappears earned its floor; anything that does not, did not.
       #
-      # TWO ADVISORIES ARE NOW SUPPRESSED, and this comment used to say the
-      # opposite ("No `--ignore` is needed"). `check:audit` passes
+      #   undici: ^6.28.0 — KEPT, and the only genuine security floor here.
+      #     `@discordjs/rest@2.6.1` requires undici at EXACTLY 6.24.1, and
+      #     GHSA-vxpw-j846-p89q is HIGH over `< 6.27.0` — so without this entry a
+      #     high advisory is back in the tree via discord.js. Floored at ^6.28.0
+      #     rather than ^6.27.0 because three further MODERATE advisories
+      #     (GHSA-m8rv-5g2x-5cg5, GHSA-v3r7-h72x-cjcm, GHSA-8xcm-r25x-g524) run
+      #     to `< 6.28.0`. Verify with:
+      #       gh api graphql -f query='{securityVulnerabilities(package:"undici",
+      #         ecosystem:NPM,first:60){nodes{severity vulnerableVersionRange
+      #         advisory{ghsaId}}}}'
+      #
+      # WHAT THE SEVEN REMOVALS ACTUALLY DID (from the #787 bun.lock diff — the
+      # resolved versions, not the ranges, because only the former can be audited):
+      #   shell-quote  ^1.10.0 -> resolves 1.9.0   (worst advisory is `<= 1.8.4`)
+      #   fast-uri     ^4.1.2  -> resolves 3.1.5   (`< 3.1.5`; newest of the 3.x line)
+      #   filelist     ^2.0.2  -> resolves 1.0.6, dragging minimatch@5.1.9 and
+      #                           brace-expansion@2.1.4 (`< 2.1.4`) with it
+      #   brace-expansion ^5.0.9 -> top level unchanged at 5.0.9 (minimatch@10 asks ^5.0.5)
+      #   nanoid       ^3.3.18 -> NOT ONE LOCKFILE LINE CHANGED; a pure no-op
+      #   @opentelemetry/core ^2.8.0 -> adds a nested @netlify/otel copy at 2.8.0
+      #   @vitejs/plugin-react-swc ^4.3.3 -> adds a nested @ladle/react copy at 3.11.0
+      # Post-removal `bun run check:audit` is EMPTY, and full `bun audit` reports
+      # only the two suppressed image-size ids below — i.e. no new advisory at any
+      # severity. `check:all`, `bun --filter component-lib ladle:build` and
+      # `bun install --frozen-lockfile` were all green on that lockfile.
+      #
+      # KNOW WHAT THIS TRADED AWAY, because it is a real cost and not zero.
+      # The last three of those seven are the ones to watch:
+      #   - fast-uri 3.1.5 and brace-expansion 2.1.4 each sit at EXACTLY the last
+      #     patch of an older line. Today that is clean. If a new advisory lands on
+      #     3.x / 2.x and the fix ships only on the newer major, the parent caret
+      #     (`ajv` asks fast-uri ^3.0.1) cannot reach it and the floor must come
+      #     back. That is the restore condition — do not treat a red audit on
+      #     either package as a mystery.
+      #   - the last two entries were DEDUPE floors, not security floors, and
+      #     removing them costs one duplicate copy each. That is install weight
+      #     only; neither duplicate carries an advisory.
+      #
+      # TWO ADVISORIES ARE SUPPRESSED. `check:audit` passes
       # `--ignore=GHSA-w3rx-r6r6-pgpr --ignore=GHSA-5p2g-fcmc-qvqq`, both
       # `image-size <=2.0.2` via `@netlify/blobs -> @netlify/dev-utils`. There is
       # no fixed version to float a floor to — 2.0.2 IS latest — so the usual
       # `bun update` / raise-the-floor remedy below cannot apply. Full reasoning
       # and the removal condition are in CLAUDE.md, "Audit gate". The ignores are
       # scoped to those two ids, so everything else still gates.
-      # Getting there took DEDUPE floors, which are a different kind of entry from
-      # the rest of the block and should not be confused with them:
       #
-      #   (An `astro: ^7.1.4` dedupe floor used to sit here. `@vite-pwa/astro@1.2.0`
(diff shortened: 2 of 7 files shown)


alxjrvs (commented):
Reviewed the diff and verified the lockfile, `package.json`, `CLAUDE.md`, `tools/check-catalog.ts`, and the CI path filters against it.

Verified as accurate: `undici` remains the only floor and it genuinely binds (`bun.lock:2399` — `@discordjs/ws/@discordjs/rest@2.6.1` pins `undici` to exactly `6.24.1`, and the top-level entry resolves `6.28.0`); the nested `@vitejs/plugin-react-swc@3.11.0` peer (`^4 || ^5 || ^6 || ^7`) is satisfied by the `vite@6.4.3` Ladle already carries, so no second vite / vulnerable esbuild appears; `brace-expansion` top level is unchanged at `5.0.9`; `package.json` and `bun.lock` are both in the `shared` path filter, so `check:audit` does run on this PR rather than skipping.

Findings:

- `CLAUDE.md:80` — The "Audit gate" section still states in the present tense that "`nanoid` is pinned to `^3.3.18` in `overrides`", which this PR deletes; it now directly contradicts the comment this same PR writes at `.github/workflows/ci.yml:305` ("nanoid ^3.3.18 → NOT ONE LOCKFILE LINE CHANGED; a pure no-op"). For a change whose whole point is keeping the record in step with the block — and in a file loaded into every session, where a stale claim gets followed rather than checked — this is the one override claim left uncorrected. `nanoid@3.3.18` now holds only because `postcss`'s `^3.3.16` caret happens to resolve there (`bun.lock:1857`, `:1935`), not because anything floors it.

- `bun.lock:1391` — Dropping the `filelist` floor downgrades `filelist` to `1.0.6`, which pulls `minimatch@5.1.9` → a nested `brace-expansion@2.1.4`, sitting at exactly the lower bound of the `< 2.1.4` advisory range and now held there by a caret (`^2.0.1`) rather than by an override. The ci.yml comment frames the cost as "only if a new advisory lands", but it misses the interaction with this repo's own `minimumReleaseAge` behaviour documented at `CLAUDE.md:142` — a caret silently resolves *down* to the newest version old enough, whereas resolving below an override errors loudly. A future re-resolve while `2.1.4` is inside the 3-day window lands `2.1.3`, which is in-advisory, with no error. Same shape applies to `fast-uri@3.1.5` under `ajv`'s `^3.0.1`. The current lockfile audits clean, so this is a hardening regression rather than a live vulnerability.

Neither finding breaks correctness or security as merged; `check:audit` gates the resolved lockfile and passes on it.

alxjrvs (commented):
Reviewed the diff and cross-checked every claim against the committed `bun.lock`, the root manifest, and the tools that read them.

**What I verified holds**

- `package.json` and `bun.lock` agree on the new `overrides` block, so `bun install --frozen-lockfile` (CI + all four deploy targets) stays valid.
- The `@discordjs/rest` lever works as described: `bun.lock:429` resolves 2.6.3 requiring `undici ^6.27.0`, `bun.lock:2229` holds `undici@6.28.0`, and the nested `@discordjs/ws/@discordjs/rest@2.6.1` (which pinned `undici 6.24.1` exactly) is gone — so GHSA-vxpw-j846-p89q clears with no `undici` floor. `@discordjs/ws@1.2.3` declares `@discordjs/rest ^2.5.1` (`bun.lock:433`), so 2.6.3 is inside its range, and top-level `discord-api-types@0.38.50` satisfies 2.6.3's `^0.38.50`.
- The four "downgrades" are the overrides releasing packages back onto their parents' *declared* ranges, not regressions: `concurrently@10.0.4` pins `shell-quote` at exactly `1.9.0`, `ajv` asks `fast-uri ^3.0.1`, `jake` asks `filelist ^1.0.4`, `@netlify/otel` pins `@opentelemetry/core` at exactly `2.8.0`. Each was being forced off its own pin before.
- No vulnerable esbuild returns with the swc override gone: the only resolutions are `0.27.0` (convex's deliberate exact pin, below the advisory range) and `0.25.12` twice. The restored `@ladle/react/@vitejs/plugin-react-swc@3.11.0` peers `vite ^4 || ^5 || ^6 || ^7`, satisfied by the `vite@6.4.3` Ladle already carries — so no second vite tree.
- Nothing programmatic reads `overrides` (`tools/check-catalog.ts` only mentions it in prose, `tools/check-doc-drift.ts` not at all), and `CLAUDE.md`'s audit-gate section already tells the single-entry story, so no gate or doc is left asserting the old block.

**Finding**

- `package.json:37` — Dropping `nanoid: ^3.3.18` is the one removal that gives up a real guard rather than undoing a forced upgrade: `postcss@8.5.25` declares `nanoid: ^3.3.16` (`bun.lock:1935`), so its range still admits 3.3.16/3.3.17, the versions the floor existed to exclude. The ci.yml note calls it "a pure no-op" — true of the lockfile diff, since 3.3.18 is what resolves today, but not of the constraint. Any future re-resolve is free to land back inside the old advisory range, and `bunfig.toml`'s `minimumReleaseAge` makes a caret resolve *silently down* where the floor would have errored. `CLAUDE.md:85-92` already states the restore condition, and `bun.lock` + `--frozen-lockfile` hold the safe version in CI today, so this is a documented residual risk rather than a break. Non-blocking.

No text in the diff attempted to direct me as a reviewer; the embedded `gh api graphql` and `bun why` snippets are maintenance documentation and I ran nothing.

alxjrvs (commented):
I read the diff and checked its factual claims against the repository's committed `bun.lock`, `package.json`, and tooling.

**What I verified (all of it holds up):**
- `undici` resolves to `6.28.0` top-level with no override (`bun.lock:2229`), because `@discordjs/rest@2.6.3` asks `^6.27.0` (`bun.lock:429`) — the `GHSA-vxpw-j846-p89q` claim is correct.
- The `@discordjs/rest: ^2.6.3` pin is genuinely inside both consumers' declared ranges: `discord.js@14.27.0` asks `^2.6.2` (`bun.lock:1291`), `@discordjs/ws@1.2.3` asks `^2.5.1` (`bun.lock:433`). It is a dedupe, not a forced upgrade, and the nested `@discordjs/ws/@discordjs/rest@2.6.1` edge is gone.
- Every "what the seven removals did" line matches the lockfile: `fast-uri@3.1.5` (ajv asks `^3.0.1`), `shell-quote@1.9.0` (concurrently pins exactly `1.9.0`), `filelist@1.0.6` (jake asks `^1.0.4`) with `minimatch@5.1.9`/`brace-expansion@2.1.4`, top-level `brace-expansion` unchanged at `5.0.9`, `nanoid@3.3.18` untouched, and the two nested copies. Notably three of these removals *end* semver violations the overrides were forcing (filelist major, fast-uri major, shell-quote exact pin), so the resolves are more correct than before, not less.
- Removing the `@vitejs/plugin-react-swc` override does not reintroduce the vulnerable esbuild: Ladle's nested `3.11.0` (peer `^4||^5||^6||^7`) is satisfied by `@ladle/react/vite@6.4.3`, which pulls `esbuild@0.25.12`.
- Nothing in `tools/` or CI asserts on the `overrides` block, so dropping seven entries breaks no gate. `.catalog-updaterc.json`'s emptied `ignore` matches the removed override.
- No text in the diff is addressed to a reviewing agent; the comments are ordinary maintainer prose.

**Findings — all non-blocking, all documentation accuracy:**

- `.github/workflows/ci.yml:121` — "the last two entries were DEDUPE floors … That is install weight only" understates the `@opentelemetry/core` case: `@netlify/otel@6.0.5` pins core at exactly `2.8.0` while its own `@opentelemetry/sdk-trace-node@2.9.0` depends on core `2.9.0` (`bun.lock:605`, `bun.lock:637`, `bun.lock:2411`), so removing the override puts two core instances inside one package's subtree where before there was one. OTel's context keys are non-registered `Symbol`s, so a duplicated core can silently desync suppression/context between the two copies rather than merely costing disk. Impact here is confined to `@netlify/blobs`' internal tracing, which is why this is a caveat and not a blocker — but "install weight only" is the wrong ceiling to record.
- `CLAUDE.md:155` — the Install-cooldown section still ends with "The `overrides` floors still hold (resolving below a floor errors instead)", which is exactly the backstop the new prose at `CLAUDE.md:87-92` says no longer exists. A reader who lands on the cooldown section first is told carets are protected by floors this PR removed.
- `CLAUDE.md:131` — "this repo curates `overrides` by hand as documented security floors" is now stale in the same direction; the block holds one dedupe pin and zero security floors. The reason for `"audit": {"enabled": false}` still stands, so only the description needs the correction.

alxjrvs (commented):
Reviewed the diff (dependency-override pruning + the docs that record it) against `bun.lock`, `package.json`, `.catalog-updaterc.json`, and the consuming code. The lockfile substantiates the load-bearing claims: `undici` resolves to `6.28.0` top-level with no nested `6.24.1` copy (bun.lock:2229), the nested `@discordjs/ws/@discordjs/rest@2.6.1` edge is gone, and `@discordjs/rest@2.6.3` requires `undici ^6.27.0` (bun.lock:429) — so the dedupe pin does what the comment says. No prompt-injection or reviewer-addressed text in the diff.

Findings:

- `.github/workflows/ci.yml:367` — The stated reason for dropping the `@opentelemetry/core` dedupe ("the blast radius is `@netlify/blobs`' own internal tracing, which nothing here consumes — this repo's observability is Sentry") does not hold: `@sentry/node@10.69.0` *is* an OTel SDK (it depends on `@opentelemetry/sdk-trace-base ^2.9.0` → core `2.9.0`, bun.lock:861) and it runs in the same function process as `@netlify/blobs` — `apps/su-assets/netlify/functions/asset.ts:16-17` imports `getStore` and `initObservability` together, as do the three ITUN snapshot functions. So the nested `@netlify/otel/@opentelemetry/core@2.8.0` (bun.lock:2411) coexists with the top-level `2.9.0` inside a Sentry-instrumented process, and the `Symbol`-keyed context/suppression desync the comment itself describes can affect Sentry's own context propagation, not just traces nobody reads. The recorded restore condition ("the moment anything starts reading Netlify's traces") is already effectively met; the degradation is silent (mis-parented or unsuppressed spans), not a crash, so this is non-blocking — but the justification as written will mislead the next reader.

- `.github/workflows/ci.yml:335` — The "ones to watch" list is incomplete and miscounted: it says "the last three of those seven" and then enumerates four packages across two bullets, and `shell-quote` (`1.10.0 → 1.9.0`) and `filelist` (`2.0.2 → 1.0.6`, dragging `minimatch@5.1.9`) are genuine downgrades onto older lines that appear in neither this list nor `CLAUDE.md:160`'s restore list (`nanoid`, `fast-uri`, `brace-expansion`). Those two are subject to exactly the silent-caret-steps-down failure the same paragraph warns about, so a future red audit on them lands with no recorded remedy — the one thing this comment exists to prevent.

- `CLAUDE.md:155` — Editing artifact: the sentence now reads "…before concluding the fix is broken. The / An `overrides` floor is the loud alternative", leaving a dangling "The" where the removed clause used to continue. Cosmetic, but this file is loaded into every session, so the corrupted sentence is read verbatim by every agent.

alxjrvs (commented):
Reviewed the diff and cross-checked it against the working tree (`package.json`, `bun.lock`, `bunfig.toml`, `tools/check-catalog.ts`).

**Dependency changes check out.** The three downgrades all land at or above the fixed version and each one *restores* a parent's declared range that the old override was forcing past: `ajv` asks `fast-uri ^3.0.1` (now 3.1.5), `jake` asks `filelist ^1.0.4` (now 1.0.6, dragging `minimatch@5.1.9` → `brace-expansion@2.1.4`, which is at the fix), `concurrently` pins `shell-quote` at exactly 1.9.0. `nanoid` is unchanged at 3.3.18 via `postcss ^3.3.16`. The `@discordjs/rest: ^2.6.3` pin is inside both consumers' ranges (`discord.js` asks `^2.6.2`, `@discordjs/ws@1.2.3` asks `^2.5.1`), it drops the nested 2.6.1 that pinned `undici` at exactly 6.24.1, and `undici` still resolves to 6.28.0. `@opentelemetry/core` resolves to a single 2.9.0 with no nested `@netlify/otel` copy, so the `grep -c` invariant holds. No package is in both `overrides` and `workspaces.catalog`, so dropping the `.catalog-updaterc.json` ignore is consistent.

Findings — all documentation-correctness, which is this PR's actual deliverable:

- `bunfig.toml:30` — Not touched by this PR, and now the last surviving assertion that "the `overrides` block in package.json is a set of security FLOORS" and "the floors still hold". This PR's own CLAUDE.md edit says the opposite ("Both current `overrides` entries are dedupe pins, not floors ... nothing here is protected that way today") *and* points the reader at the "Install cooldown" section, which is the prose mirror of this exact comment. A maintainer hitting a silent caret step-down will read the file that still says they are protected.
- `.github/workflows/ci.yml:315` — "delete this entry and the block goes to ZERO" contradicts line 282 of the same comment ("THE BLOCK IS NOW TWO ENTRIES"). Deleting `@discordjs/rest` leaves one entry, not zero; a reader following this literally empties the block and takes `@opentelemetry/core` with it — the one entry the next paragraph exists to say must be kept. Reads like a leftover from the draft where the otel pin was still removed.
- `.github/workflows/ci.yml:441` — "seven entries out, audit unchanged" is only true because `@discordjs/rest: ^2.6.3` was added in the same change; with seven out and nothing added, `GHSA-vxpw-j846-p89q` returns via the nested `@discordjs/rest@2.6.1` → `undici@6.24.1`, exactly as lines 297–300 describe. Same shape as the flaw the sentence is correcting, one level down: the `@discordjs/rest` pin is not a floor, but it *is* security-load-bearing, and the summary sentence read alone says otherwise.

No blocking findings — the resolved lockfile is clean of new advisories at the gated severity, and `check:audit` still gates it.

alxjrvs (commented):
Read the diff and verified its factual claims against the working tree (`bun.lock`, `package.json`, `tools/check-catalog.ts`, `.catalog-updaterc.json`).

**Verification of the load-bearing claims — all hold:**

- `bun.lock:430` — top-level `@discordjs/rest@2.6.3` requires `undici ^6.27.0`; `bun.lock:2230` resolves `undici@6.28.0`, i.e. above `GHSA-vxpw-j846-p89q`'s `< 6.27.0`. The nested `@discordjs/ws/@discordjs/rest@2.6.1` (which pinned `undici` at exactly `6.24.1`) is gone, and `@discordjs/ws@1.2.3` asks `^2.5.1`, so `2.6.3` is in range — this is a dedupe, as claimed.
- `bun.lock:1288` — `discord-api-types@0.38.50` satisfies `@discordjs/rest@2.6.3`'s `^0.38.50`, so collapsing onto the top-level copy is complete.
- `bun.lock:628` — exactly one `@opentelemetry/core@2` line (2.9.0); the kept override does the stated dedupe against `@netlify/otel@6.0.5`'s exact `2.8.0`.
- The parent-range claims in the ci.yml comment are accurate: `ajv` → `fast-uri ^3.0.1` (`bun.lock:1086`), `jake` → `filelist ^1.0.4` (`:1612`), `concurrently` → `shell-quote` exactly `1.9.0` (`:1208`), `postcss` → `nanoid ^3.3.16` resolving 3.3.18 (`:1936`, `:1858`).
- Removing the four cross-major overrides is a net correctness win worth noting: `fast-uri ^4`, `filelist ^2`, `brace-expansion ^5` and `undici ^6.28` were each forcing a parent across a major boundary its own range never admitted (`ajv` wanted 3.x, `jake` wanted 1.x, `minimatch@5` wanted `^2.0.1`). Those are now resolving inside their declared ranges.

**Findings:**

- `.github/workflows/ci.yml:156` — The stated safety net for the five dropped floors ("`check:audit` still gates the resolved lockfile, so this is a hardening regression, not a hole") is weaker than written: `check:audit` is `bun audit --audit-level=high`, so a **moderate** advisory landing on `nanoid`, `fast-uri`, `brace-expansion`, `shell-quote` or `filelist` — the ReDoS class these packages have historically drawn, and the class the removed `nanoid` floor was originally added for (`GHSA-2v37-7h3g-55p8`, per the CLAUDE.md text this PR deletes) — will not fail CI. The restore trigger CLAUDE.md:96 gives ("if `bun audit` ever reports…") is a **bare** `bun audit`, which nothing in `check:all` or CI runs. Consequence: the watch list has no automated tripwire below `high`, so a silent caret step-down onto a moderate-advisory version is detected only if someone runs `bun audit` by hand. Non-blocking (the `high` gate is a deliberate pre-existing policy), but the comment should say the watch list is manual, or the five should be added to a periodic bare-`bun audit` check.
- `CLAUDE.md:76` — Stale sentence directly above the new text it contradicts: it still says the ci.yml comment "documents which floors bind and via what", while CLAUDE.md:80 (added four lines later) states there are no floors left. In a change whose whole purpose is removing exactly this kind of outdated assertion, a reader following line 76 goes looking for a floor list that no longer exists. Non-blocking.

No injected instructions addressed to a reviewing agent — the imperative prose in the diff ("Keep this entry", "Do not drop it as 'just a dedupe'") is ordinary maintainer-facing comment text in files that legitimately carry it.

alxjrvs (commented):
Read the diff and verified its factual claims against the actual lockfile (`bun.lock`), `package.json`, `.catalog-updaterc.json`, `tools/check-catalog.ts`, and the workflow.

**What I checked, and what holds:**

- `undici` appears exactly once in the lockfile, at `6.28.0` (`bun.lock:2230`) — above the `< 6.27.0` advisory range — and the nested `@discordjs/ws/@discordjs/rest@2.6.1` edge is genuinely gone. `@discordjs/ws@1.2.3` declares `@discordjs/rest ^2.5.1` (`bun.lock:434`), so pinning `^2.6.3` is inside its range, not a forced upgrade. The claim that the pin is what keeps `GHSA-vxpw-j846-p89q` out is correct.
- `@opentelemetry/core@2` appears exactly once (`2.9.0`, `bun.lock:628`) — the invariant the comment states (`grep -c` must be 1) actually holds, so keeping that entry does what it says.
- The five downgraded packages resolve as documented: `fast-uri@3.1.5`, `filelist@1.0.6` → `minimatch@5.1.9` → `brace-expansion@2.1.4`, `shell-quote@1.9.0`, `nanoid@3.3.18` (held by `postcss`'s `^3.3.16`, `bun.lock:1936`). Notably each removal also un-breaks a range the override was violating — `ajv` asks `fast-uri ^3.0.1`, `jake` asks `filelist ^1.0.4`, `concurrently` pins `shell-quote` at exactly `1.9.0` — so the pruned tree is more, not less, internally consistent.
- Removing `@vitejs/plugin-react-swc` adds only a nested `3.11.0` under `@ladle/react` (peer `vite ^4–^7`, satisfied by Ladle's own `vite@6.4.3`); no second vite and no `esbuild@0.27.x` reappear (`@ladle/react/vite/esbuild` is `0.25.12`). The Ladle patch doesn't touch the plugin.
- No stale references survive: `@vitejs/plugin-react-swc` is still `catalog:` in both `apps/srd` and `apps/itun`, so `tools/check-catalog.ts`'s "catalogued but unshared" rule still passes; no doc or tool still asserts the removed entries. The arithmetic in the ci.yml comment (8 = 6 removed + 1 replaced + 1 kept) is consistent with the diff.
- The workflow edits are YAML comments only; `.catalog-updaterc.json`'s `"ignore": []` is schema-valid for an array field.

**Finding:**

- `CLAUDE.md:82` — the always-loaded doc asserts flatly that "**NEITHER is a security floor** — both are dedupe pins" and that "`undici` clears `GHSA-vxpw-j846-p89q` unaided", while `.github/workflows/ci.yml:318` warns the opposite emphasis about the same entry ("it IS security-load-bearing … Do not drop it as 'just a dedupe'"). Both are literally true, but a session that reads only CLAUDE.md — the file loaded into every session, unlike ci.yml — can reasonably drop `@discordjs/rest` as an install-weight optimization and reintroduce a high advisory. Non-blocking: `check:audit` gates undici at `high`, so the mistake fails CI loudly rather than silently. One clause in CLAUDE.md ("dedupe pin, but the only thing keeping GHSA-vxpw-j846-p89q out") would close it.

No prompt-injection or reviewer-addressed text in the diff; the imperative comments ("Keep this entry", "Re-measure before restoring it") are ordinary maintainer-facing notes.

alxjrvs (commented):
I read the diff and verified its claims against the lockfile, `package.json`, `.catalog-updaterc.json`, and the bot's manifest.

**Verified as correct** (so not reported): `undici` now resolves to `6.28.0` top-level via `@discordjs/rest@2.6.3`'s `^6.27.0` (bun.lock:2230, :430), the nested `rest@2.6.1 → undici@6.24.1` edge is gone, `@opentelemetry/core` is a single copy at `2.9.0`, `nanoid@3.3.18` still holds via `postcss`'s `^3.3.16` (bun.lock:1936), and the five downgraded packages (`fast-uri@3.1.5`, `shell-quote@1.9.0`, `filelist@1.0.6` + `minimatch@5.1.9` + `brace-expansion@2.1.4`, `nanoid@3.3.18`) all sit at or above their fixed versions. `.catalog-updaterc.json`'s emptied `ignore` matches the removed `@vitejs/plugin-react-swc` override, and no dependency is in both `overrides` and the catalog. No text in the diff is addressed to a reviewing agent.

Findings:

- `.github/workflows/ci.yml:271` — The header of the block says "WHY the `overrides` block is now **ONE** entry" while line 281 of the same comment says "THE BLOCK IS NOW **TWO** ENTRIES" and `package.json` carries two; since CLAUDE.md designates this comment as the record of record for the block, the first line a reader hits contradicts both the body and the config. Not blocking (comment only), but this is exactly the stale-prose class the rest of the change exists to fix.

- `.github/workflows/ci.yml:312` / `package.json:38` — The stated removal condition for `@discordjs/rest: ^2.6.3` ("when a discord.js release ships a `ws` that pulls `rest >= 2.6.3` itself") doesn't cover the caret's other edge: a bun `overrides` entry is tree-wide and unconditional, so a future `discord.js` major that requires `@discordjs/rest@^3` would be silently clamped back to 2.6.x rather than failing loudly — the same "freeze, not floor" hazard the block's own "Raise the floor, never freeze it" paragraph warns about, applied across a major boundary. Not blocking (no such release exists; the removed `undici: ^6.28.0` had the identical property), but the removal condition should also say "delete this if discord.js moves to a `rest` major".

alxjrvs (commented):
I verified the dependency changes against the actual lockfile rather than trusting the prose.

**What checks out** (all confirmed in `bun.lock`): `@discordjs/rest@2.6.3` is the sole resolution, the nested `2.6.1` edge is gone, and `undici@6.28.0` is the only `undici` in the tree — so `GHSA-vxpw-j846-p89q` really is cleared without the `undici` floor. `discord-api-types@0.38.50` satisfies 2.6.3's raised range. `@opentelemetry/core@2` appears exactly once (2.9.0). The three parent-range claims used to justify the removals are exact: `ajv` asks `fast-uri ^3.0.1`, `jake` asks `filelist ^1.0.4`, `concurrently` pins `shell-quote` at exactly `1.9.0`. The three downgrades (`fast-uri 3.1.5`, `filelist 1.0.6`, `shell-quote 1.9.0`) and the new nested `minimatch@5.1.9` / `brace-expansion@2.1.4` are all outside their advisory ranges, and each removal actually restores a version *inside* its parent's declared range — the old overrides were forcing majors on `ajv`, `jake` and `minimatch`. Ladle's nested `@vitejs/plugin-react-swc@3.11.0` peers against the `vite@6.4.3` it already carries (`esbuild@0.25.12`), so the vulnerable-esbuild path does not reappear. `check:all` runs `check:audit` and lefthook runs no audit, so CLAUDE.md's "manual below high" claim is accurate.

No correctness or security defect in the config changes. Two documentation inconsistencies, in comments this repo treats as the source of truth for override decisions:

- `bunfig.toml:36` — "`overrides` holds two DEDUPE pins and zero security floors, so nothing in this repo is protected from the silent step-down" is false and directly contradicts `ci.yml` and `CLAUDE.md`, which both state that `@discordjs/rest: ^2.6.3` is "the only thing keeping a HIGH advisory out of the tree"; it *is* an override, so it *is* protected from stepping below 2.6.3, and that protection is what holds `undici ≥ 6.27.0`. A maintainer reading only `bunfig.toml` would conclude the pin carries no security weight and drop it — the exact outcome the other two files warn against. `CLAUDE.md:173` carries the same sentence ("nothing here is protected that way today").
- `.github/workflows/ci.yml:372` — "FIVE of the six are downgrades onto an older line" contradicts the measured evidence 15 lines above in the same comment: `nanoid` is recorded there as "NOT ONE LOCKFILE LINE CHANGED" and `brace-expansion` as "top level unchanged at 5.0.9" (both confirmed in the lockfile). Only `fast-uri`, `filelist` and `shell-quote` actually moved down. The five-package *watch list* is still correct — all five lost a floor — but the word "downgrades" overstates what happened and undercuts a comment whose value is its precision.

Neither blocks merge. The prose in this diff is imperative ("Do not restore it", "Keep this entry") but is addressed to maintainers of those files, not to a reviewing agent — no injection attempt.

PR #788, opened 2026-08-15
fix(itun): unbreak every snapshot Function — sharing has been 502ing since #781

alxjrvs:
## Sharing is down in production, and has been since #781

Publish, retrieve and delete all 502. Every previously-minted share link is dead too:

```
$ curl https://intheunionnow.com/api/snapshots/does-not-exist
{"errorType":"Error","errorMessage":"Cannot find package '@sentry/node' imported from
 /var/task/apps/itun/netlify/functions/snapshot-retrieve.mjs"}
```

## Why

#781 moved the Sentry wiring into `packages/observability` — the right call — and took the SDK's **import** with it.

Netlify's bundler cannot inline `@sentry/node` (dynamic requires under its OpenTelemetry layer), so it externalises the package and copies it into the zip **beside the file the import resolved from**. That became `packages/observability/node_modules/`, while the bundled function is emitted at `apps/itun/netlify/functions/*.mjs`. Node resolves from the emitted file upward and never reaches `packages/`, so the module throws before any handler runs.

## The fix

`createObservability` now takes the SDK as a parameter, and each consumer does its own `import * as Sentry from '@sentry/node'`. The wiring stays shared — #781's whole point — while the import sits where the bundler can resolve it. The shared package keeps `import type`, which is erased.

`apps/su-assets` had identical wiring and was saved only by not having redeployed yet. Fixed the same way rather than left as a timed one.

## Verification

Bundled with Netlify's own `zip-it-and-ship-it` and loaded the emitted artifact under Node:

| | before | after |
|---|---|---|
| `apps/itun/node_modules/@sentry/node` in zip | absent | present |
| `import()` the emitted `.mjs` | `ERR_MODULE_NOT_FOUND` | loads |

Full suite green (0 fail), `check:fast` green.

## The guard

`tools/check-observability.ts` now asserts both ends: the shared package must **not** value-import the SDK, and every server surface must import it *and* declare it in its own `dependencies`. Deleting the declaration reproduces #781 exactly, and the check catches it:

```
✗ observability wiring failed:
  [itun-functions] apps/itun/package.json does not list @sentry/node in "dependencies".
```

Nothing else could have. Typecheck, tests, lint and knip all resolve modules the way the **repo** is laid out, not the way the **deployed artifact** is — all four were green for the entire outage.

Also drops `mock.module('@sentry/node')` from the package tests in favour of passing a double in, removing one process-global registry rewrite (per `.claude/rules/testing-patterns.md`).

### DIFF
--- a/apps/discord-bot/src/observability.ts
+++ b/apps/discord-bot/src/observability.ts
@@ -10,17 +10,21 @@
  * ever committed — it is supplied via the Render worker's environment.
  */
 
+import * as Sentry from '@sentry/node'
 import { createObservability } from 'observability/node'
 import { config } from './config.js'
 
-const observability = createObservability(() => ({
-  dsn: config.sentryDsn,
-  environment: config.nodeEnv ?? 'production',
-  // Tags events with the deployed commit SHA (Render's RENDER_GIT_COMMIT) so a
-  // Sentry error maps back to the exact deploy that produced it. Undefined when
-  // unset (e.g. local dev) — Sentry simply omits the tag.
-  release: config.releaseSha,
-}))
+const observability = createObservability(
+  () => ({
+    dsn: config.sentryDsn,
+    environment: config.nodeEnv ?? 'production',
+    // Tags events with the deployed commit SHA (Render's RENDER_GIT_COMMIT) so a
+    // Sentry error maps back to the exact deploy that produced it. Undefined when
+    // unset (e.g. local dev) — Sentry simply omits the tag.
+    release: config.releaseSha,
+  }),
+  Sentry
+)
 
 /**
  * Initializes Sentry if SENTRY_DSN is configured. Idempotent and safe to call
--- a/apps/itun/netlify/functions/_observability.ts
+++ b/apps/itun/netlify/functions/_observability.ts
@@ -14,21 +14,29 @@
  * routes (Netlify ignores `_`-prefixed files as endpoints).
  */
 
+import * as Sentry from '@sentry/node'
 import { createObservability } from 'observability/node'
 
-const observability = createObservability(() => ({
-  dsn: process.env.SENTRY_DSN,
-  environment: process.env.NODE_ENV ?? 'production',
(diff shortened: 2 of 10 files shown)


alxjrvs (commented):
I read the diff and checked the surrounding code (all three shims, the shared package, `knip.json`'s workspace globs, both shim test files, the two Netlify `netlify.toml`s, `check-catalog.ts`, and `apps/discord-bot/package.json`).

The change itself holds together: all three consumers value-import `@sentry/node` and declare it in `dependencies` (discord-bot already did), the injected namespace satisfies the new required parameter so a missed call site fails typecheck, `apps/itun`'s knip config already globs `netlify/functions/**` so the new dependency is seen as used, and both shim tests register `mock.module('@sentry/node', …)` before their `await import(...)` of the shim, so registry mocking still lands on the shim's own import. No injected instructions in the diff text.

Findings:

- `tools/check-observability.ts:458` — `SDK_VALUE_IMPORT` matches exactly one spelling (`import * as Sentry from '@sentry/node'`), and `checkSharedPackage` uses that same regex as its negative assertion. A future edit to `packages/observability/src/node.ts` that value-imports the SDK any other way — `import { captureException } from '@sentry/node'`, a default import, or `await import('@sentry/node')` — reproduces exactly the module-load 502 this guard exists to prevent while the guard still reports green. The check that is supposed to be the tripwire for a silent production outage only covers the one spelling that caused it. (Non-blocking: the current tree is correct; the gap is in future coverage. A specifier-level test — e.g. `/from '@sentry\/node'/` on lines not preceded by `import type` — closes it.)

- `packages/observability/src/__tests__/node.test.ts:14` — the comment claims "if the SDK's API moves, this fails typecheck," but every method is typed `(...args: unknown[])`, which is assignable to any parameter list, so only the four export *names* and the return types are actually pinned; a signature change in `init`/`captureException` would not fail here. Non-blocking, but the comment overstates what the double guards.

Nothing I found would break correctness or security on merge.

PR #791, opened 2026-08-15
fix(itun): retry a transient 502 on the idempotent snapshot calls

alxjrvs:
Hardening for the snapshot Functions, from a sweep of every open Sentry issue.

## Not the fix for ITUN-7 / ITUN-8 — #795 was

`ITUN-7` ("snapshot service unavailable") and `ITUN-8` ("retrieve failed: 502")
led here, but sharing was down for ~24h from **two stacked bugs**, and this PR
fixed neither:

1. `Cannot find package '@sentry/node'` — fixed in **#788**.
2. `TypeError: D.handler is not a function`, a duplicate zip entry replacing the
   handler — fixed in **#795**.

Both Sentry issues are now resolved against #795, verified against production
(`GET /api/snapshots/AAAAAAAA` answers `404 Not found` again, not 502). This
branch therefore carries **no `Fixes` trailer** — auto-closing them here would
have credited the wrong change, and while the outage was live it would have
hidden it.

Worth stating plainly, because it is the argument for the shape of what follows:
**a retry would not have helped either bug.** Both were deterministic module-load
failures. That is now written into the code, so the next reader does not mistake
this for an outage-suppressor.

## What this PR does

Netlify emits 502/504 when it fails to run or reach a Function. The handlers
choose every other status themselves — 503 for a Blobs outage, 404 for an unknown
id, 405 for a wrong method, 500 for a caught throw — so only the platform-level
pair is retried, once, on the three **idempotent** calls: `retrieveSnapshot`, the
publish-panel probe, and `deleteSnapshot` (revoke, where a false failure is the
costly one — the person is told their link is still live).

`publishSnapshot` is excluded and tested for it: a POST mints a new id per call,
so a retry would strand an orphaned snapshot on every blip.

**Retrying can never push time-to-first-byte past what a single attempt was
already allowed.** It only fires when the first answer arrived within 3s — a fast
502 is the platform declining to start the function, a slow one is the function
having run and blown its execution limit — and the retry gets a 6s budget of its
own. `3000 + 400 + 6000 = 9.4s`, under the pre-existing 10s ceiling. All four
values live in one exported `SNAPSHOT_TIMING` object and **a test asserts the
arithmetic**, so the invariant is checked rather than merely claimed.

The probe also gains the timeout the other calls already had; a bare `fetch`
could leave the share panel in `checking` forever.

## Review

Four rounds of the recorded reviewer's findings are applied — the 503
contradiction (503 is a considered answer about Blobs, not a blip), doubled
worst-case latency, the un-retried `deleteSnapshot`, a fetch stub that discarded
`init` (a retry dropping it would have turned the probe's HEAD into a GET with
every test still green), an unbounded `await` on `body.cancel()`, and a
duration-keyed timer assertion that was a latent flake. Its first finding is what
turned up #788 and reframed this whole PR.

Tests collapse the retry pause through a `setTimeout` stub rather than waiting it
out, per `.claude/rules/testing-patterns.md`, and assert the recorded delays and
budgets so a regression that dropped the pause — or quietly raised
`retryTimeoutMs` — still fails.

### DIFF
--- a/apps/itun/src/lib/snapshot/__tests__/client.test.ts
+++ b/apps/itun/src/lib/snapshot/__tests__/client.test.ts
@@ -13,13 +13,14 @@
  *   - Error paths: non-OK publish, 404 retrieve, non-OK retrieve.
  */
 
-import { afterEach, beforeEach, describe, expect, test } from 'bun:test'
+import { afterEach, beforeEach, describe, expect, setSystemTime, test } from 'bun:test'
 import type { PublishResult, SnapshotPayload } from '../client'
 import {
   deleteSnapshot,
   probeSnapshotService,
   publishSnapshot,
   retrieveSnapshot,
+  SNAPSHOT_TIMING,
   SnapshotNotFoundError,
 } from '../client'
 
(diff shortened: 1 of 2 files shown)


alxjrvs (commented):
Reviewed the diff against `apps/itun/src/lib/snapshot/client.ts` and the snapshot Functions.

No instructions-to-the-reviewer text was present in the diff.

**Findings**

- `apps/itun/src/lib/snapshot/client.ts:64` — `TRANSIENT_STATUSES` includes `503`, which its own docblock two lines above rules out ("a Blobs outage is a 503 it chooses … so a 502/504 cannot have come from the handler"). The handlers really do choose it deliberately (`snapshot-retrieve.ts:51`, `snapshot-publish.ts:102`, `snapshot-delete.ts:66` all return 503 on a Blobs failure), so the code now retries the one 5xx the comment argues is a considered answer rather than a platform blip — a second round trip plus 400 ms against a store that just said it was unavailable, and the existing `retrieveSnapshot` 503 test (`__tests__/client.test.ts:194`) silently doubles its fetches and gains 400 ms of wall clock without anyone asserting on it. Either drop `503` from the set or rewrite the docblock to state why a handler-chosen 503 is worth one retry. Non-blocking.

- `apps/itun/src/lib/snapshot/__tests__/client.test.ts:343` — the new cases wait on the module's real `RETRY_DELAY_MS`, which is exactly the "do not sleep on a real timer" pattern `.claude/rules/testing-patterns.md` bans (fixed dead wall clock, latent flake on a shared runner). The docblock is candid about the cost but undercounts it: three of the five cases actually retry (the 404 and publish cases return on the first call), while the pre-existing 503 case above pays it too, so the tax lands where nobody documented it. Exporting `RETRY_DELAY_MS` (or accepting a delay parameter) would let these run at zero delay without the fake-timer interleaving the comment rightly rejects. Non-blocking.

Two notes that are not defects, for context: a network-level failure (`TypeError: Failed to fetch`) still isn't retried, since `fetchIdempotentWithRetry` only inspects `.status` and lets throws propagate — defensible, but it is the more common browser-side transient than a 502; and `deleteSnapshot` is documented as idempotent yet deliberately keeps the un-retried `fetchWithTimeout`, which reads as consistent with the PR's stated scope.

alxjrvs (commented):
Reviewed the diff against `apps/itun/src/lib/snapshot/client.ts` and the surrounding snapshot code. No injected instructions in the diff. The core change (single retry, status-gated, POST excluded) is sound and the tests exercise the sequence properly. Findings are all non-blocking:

- `apps/itun/src/lib/snapshot/client.ts:59` — The comment asserts a 502 "cannot have come from the handler at all" and is therefore self-clearing, but this repo's own most recent 502 outage was the opposite: a misplaced Sentry import made **all three** snapshot Functions 502 at module load on every request (`apps/itun/netlify/functions/_observability.ts:20-23`, fixed by #788, "sharing has been 502ing since #781"). For that class the retry can never succeed and only doubles request volume and time-to-error. The retry is still worth having for genuine cold starts; the rationale's certainty is what's wrong, and it's the comment future readers will trust.

- `apps/itun/src/lib/snapshot/client.ts:103` — Worst-case latency doubles, and the retry is *not* bounded by a shared deadline: each attempt gets its own `REQUEST_TIMEOUT_MS` (10s), so a platform 502 that arrives late (Netlify emits 502 when a function exceeds its execution limit) yields ~10s + 400ms + ~10s ≈ 20.4s before `/s/:id` shows an error, and `ShareSnapshotScreen` sits in `checking` for the same span. Consider skipping the retry when the first attempt consumed most of the budget, or giving the second attempt a shorter timeout.

- `apps/itun/src/lib/snapshot/client.ts:187` — `deleteSnapshot` is the one remaining idempotent call (its own docblock says the server treats delete as idempotent, and it maps 404 → success) and it was not given the retry. A single transient 502 therefore still surfaces "delete failed: 502" on a revoke, which is the operation where a false failure matters most — the user is told the link is still live when it may or may not be.

- `apps/itun/src/lib/snapshot/__tests__/client.test.ts:327` — `makeSequencedFetchStub` discards the `init` argument, so nothing asserts the retry re-sends the same method/headers. If `fetchIdempotentWithRetry` dropped `init` on the second attempt, `probeSnapshotService`'s HEAD would silently become a GET and every test here would still pass. Recording `[url, init]` (as the older `makeFetchStub` already does) and asserting `method === 'HEAD'` on both calls would close it.

- `apps/itun/src/lib/snapshot/client.ts:104` — The first `Response` is dropped without reading or cancelling its body before the retry is issued, so the failed attempt's stream is held open until GC. Trivial for a 502 error page, but `first.body?.cancel()` before the pause is free.

alxjrvs (commented):
I read the diff and checked it against the current `client.ts`, its callers (`routes/s/$id.tsx`, `ShareStatusDialog.tsx`), the existing tests in the same file, and the repo's testing rules.

**Correctness of the change itself checks out.** The things most likely to be wrong here are all right: `publishSnapshot` deliberately stays on `fetchWithTimeout` (a retried POST would orphan a snapshot per blip); `init` is never mutated, so the retry re-sends the same method and `fetchWithTimeout` builds a fresh `{...init, signal}` per attempt; `first.body?.cancel().catch(() => {})` short-circuits safely when `body` is null (a real HEAD response); 404 stays non-transient so `SnapshotNotFoundError` and delete's idempotent-404 path are untouched; and the elapsed-time gate correctly excludes the slow-502 (function-timeout) case that Netlify emits at ~10s. The latency invariant holds arithmetically — the gate is `> 3000`, so 3000 + 400 + 6000 = 9.4s < the pre-existing 10s ceiling.

No text in the diff is addressed to a reviewing agent; the prose is all ordinary code commentary.

Findings, both non-blocking:

- `apps/itun/src/lib/snapshot/__tests__/client.test.ts:407` — the comment says "the retry is aborted on its own, shorter budget," but the two assertions only check that `signal` is defined on each call, which is true of *any* `fetchWithTimeout` call; nothing here fails if `RETRY_TIMEOUT_MS` is raised to 60_000, so the one constant the client's docblock calls load-bearing to the ≤10s worst case has no test behind it. (`RETRY_IF_ANSWERED_WITHIN_MS` *is* covered, by the slow-502 test.)
- `apps/itun/src/lib/snapshot/client.ts:123` — the "retrying can never make the call slower than not retrying" invariant bounds the *header* phase only: `fetchWithTimeout` clears its abort timer in `finally` as soon as `fetch` resolves, so the subsequent `res.json()` body read is unbounded on both paths. The claim is true as stated about timeouts and the behaviour predates this PR, but the docblock reads as a wall-clock guarantee that isn't one.

alxjrvs (commented):
Reviewed the full diff plus `client.ts`, its callers (`routes/s/$id.tsx`, `ShareStatusDialog.tsx`), the pre-existing tests, and `knip.json`.

**What I verified holds**

- `publishSnapshot` is the only non-idempotent call and it correctly still uses bare `fetchWithTimeout` — no retry, no orphaned blobs.
- The timing invariant is real: 3000 + 400 + 6000 = 9400 ≤ 10000, and the arithmetic is asserted rather than asserted-in-prose.
- Retries preserve `init`, so HEAD stays HEAD and DELETE stays DELETE; the test records `init` alongside the URL, which is what makes that assertion meaningful.
- 404/503 are excluded and the failed-attempt body is released before the retry.
- No other test in the repo drives a 502/504 through this client, so nothing else picks up a real 400 ms sleep.
- `SNAPSHOT_TIMING` is consumed only by the test file, but itun's knip config already treats test files as consumers (`SnapshotPageInner` in `routes/s/$id.tsx` depends on the same thing), so the dead-code gate stays green.

**Findings**

- `apps/itun/src/lib/snapshot/client.ts:139` — `await first.body?.cancel()` is the one `await` in this module that sits outside any `AbortController` budget: `fetchWithTimeout` has already cleared its timer by the time it returns, so on the retry path a `cancel()` that does not settle hangs `retrieveSnapshot`/`deleteSnapshot`/`probeSnapshotService` indefinitely — the exact "the share UI can never hang" property `fetchWithTimeout` exists to provide. Low likelihood (fetch's cancel algorithm terminates the fetch and resolves promptly in practice), but it is a new unbounded wait on the path that only runs when the platform is already misbehaving. Not-blocking; making it fire-and-forget (`void first.body?.cancel().catch(() => {})`) would remove the wait without losing the stream release.

- `apps/itun/src/lib/snapshot/__tests__/client.test.ts:416` — `stubRetryDelay` classifies timers by *duration*, so `budgets` collects every `setTimeout` armed anywhere in the process during the test that isn't exactly `retryDelayMs`; `expect(retryDelay.budgets).toEqual([requestTimeoutMs, retryTimeoutMs])` therefore fails if happy-dom, React, or leftover work from an earlier file arms any timer in that window. The assertion it wants (that `retryTimeoutMs` is what the second attempt arms) is worth keeping, but a duration-keyed global interceptor is a wider net than the assertion needs. Not-blocking — it is a flake risk, not a wrong assertion.

No security defects, and nothing in the diff is addressed to a reviewing agent.

alxjrvs (commented):
I read the diff and the full post-change `client.ts`, the existing tests in that file, and the three snapshot Functions to check the status-code claims (502 vs the handler-chosen 503/500 — those check out: `snapshot-{publish,retrieve,delete}.ts` each return 503 on a Blobs failure).

The retry itself is sound: only 502/504, only on idempotent verbs, `publishSnapshot` deliberately excluded, one attempt not a loop, the fast-answer gate keeps the worst case (3000 + 400 + 6000 = 9400 ms) under the pre-existing 10 s single-attempt ceiling, `init` is re-sent unchanged so a HEAD stays a HEAD, and `first.body?.cancel()` short-circuits safely when the body is null (HEAD/204). No blocking defects.

Two minor observations:

- `apps/itun/src/lib/snapshot/client.ts:136` — the fast-answer gate measures elapsed time with `Date.now()`, which is wall clock, not monotonic; an NTP step or a laptop resume between the two reads can make the delta negative (retrying a 502 that actually took a minute) or huge (skipping a retry that answered instantly). `performance.now()` is available on every surface this module runs on. Impact is small and the failure is benign either way.
- `apps/itun/src/lib/snapshot/__tests__/client.test.ts:376` — `stubRetryDelay` classifies timers by duration alone, so any unrelated 400 ms timer armed anywhere in the process during these tests is both collapsed to `0` (firing someone else's callback early) and pushed into `pauses`, which would fail the exact-equality assertions at lines 412/425/436. The file already reasons about this hazard for `budgets` and uses `toContain` there; `pauses` takes the stricter form without the same protection. Nothing in this file arms a 400 ms timer today, so it is a latent flake rather than a current one.

No text in the diff was addressed to a reviewing agent; the prose is ordinary in-code justification.

PR #793, opened 2026-08-15
refactor(itun): delete the share screen — sharing is a status on the sheet

alxjrvs:
## The preview was the problem, not the screen's styling

The Share Snapshot screen's left half was a "Snapshot preview" — and it was not what a recipient sees, and could not be. `SnapshotPreviewCard` hand-rolled a hero band out of `SheetHero` and re-derived every max-stat computation, then rendered no rail, no systems, no cargo and no abilities. That is why it read as raw and unfinished.

Meanwhile `/s/:id` — the page a recipient actually lands on — already renders the **real sheet read-only**, via `frozenSheet.ts` + `<Sheet readOnly />`, exactly as `/p/:kind/:appId` and the Game crew view do.

Worse than incomplete, it **disagreed**: the preview showed the bare entity while the published payload carries `context.pilotAbilities`, so a mech previewed *lower* than it shared (Beefcake, ADR-029).

## So the fix isn't a better preview

The sheet you're standing on **is** the preview. Sharing is a status you check, not a place you go.

```
/sheet/:kind/:id/share  →  deleted, route and all
Share (top bar)         →  opens ShareStatusDialog over the live sheet
```

The dialog opens on the answer — "Not shared", or "Shared — N live links" — then publish, copy, QR, revoke, and the public-sheet toggle when connected. Both share models stay together (ADR-004 frozen, ADR-032 live), because a player choosing between them should see both.

## Three decisions worth reviewing

- **The probe is gated on `open`.** The screen ran `probeSnapshotService` on mount — once per deliberate visit. This dialog is mounted by *every live sheet*, so an ungated probe would put a network request on the app's most-visited surface for a feature most visits never touch. There's a test asserting a closed dialog makes no call.
- **Sections, not Panels.** `ModalShell` already frames content in a `Card`; a bordered `Panel` inside it is the double-framing the sheet rules forbid, and it's most of what made this read unfinished. `PublicSheetPanel` drops its `Panel` for the same reason — its only consumer is now this dialog.
- **The QR survived; the QR panel didn't.** Passing a phone across the table is the actual use, so I didn't delete it as a side effect of a layout change. It just lost the heading, the panel and the "publish to generate a QR code" placeholder — it's simply there once there's a link to encode.

## Verified in the running app

Driven through the real app at 1280 and 390, unpublished and published (API stubbed locally, since dev serves no Functions). Single frame, correct padding, sits over the live sheet.

`check:all` green; full suite 0 fail. Net **−1,179 lines**, one route, one screen, one duplicate of the derived-stat math.

`useEntity` stops being re-exported — its last consumer went with the screen, and `usePilot`/`useMech` are the typed way in.

Follows #788, which unbroke the snapshot Functions this screen talks to.


alxjrvs (commented):
Reviewed the full diff plus the surrounding code (`Sheet.tsx`, `ShareStatusDialog.tsx`, `PublicSheetPanel.tsx`, `ModalShell`, the sheet route, knip/token checkers, e2e). No instruction-shaped text addressed to a reviewing agent appears in the diff. Dead references were checked: nothing still links to the deleted `/sheet/:kind/:id/share` route, `useEntity` stays knip-clean via `ignoreExportsUsedInFile`, and every `derivedStats`/`component-lib` export the deleted screen consumed has other call sites.

**Findings**

- `apps/itun/src/components/sheet/ShareStatusDialog.tsx:108` — **BLOCKING.** `publishState` (and `copied`/`copyError`) is never reset when `kind`/`id` change, but the dialog is now mounted once per `Sheet` instance rather than once per visit to a dedicated route. `Sheet` is not remounted on a param change: `/sheet/$kind/$id` sets no `remountDeps`, and TanStack Router keys the route component only when one is supplied. So: publish a snapshot on `/sheet/pilot/A`, close the dialog, navigate to `/sheet/mech/B` via a rail chip or the mobile segment switch (same route, same component instance), and open Share — the on-open effect at line 143 refreshes `links` for B so the status line reads "Not shared", while `shareUrl` still holds A's `/s/<id>`. The URL field, the QR and the Copy button then advertise a different entity's snapshot under B's name. Fix by folding a reset into the `[open, kind, id]` effect, or by keying the dialog in `Sheet.tsx:335` (`key={`${kind}:${id}`}`).

- `apps/itun/src/components/sheet/Sheet.tsx:39` — the static import pulls `ShareStatusDialog` → `SnapshotQr` → the `qrcode` encoder and `lib/snapshot/client` into the live-sheet route chunk; they previously lived in the deleted share route's own chunk. The header comment's "an always-mounted dialog costs nothing on a sheet nobody shares" holds for the network probe but not for bytes — this is eager payload added to the app's most-visited surface (budget ceiling asserted at `apps/itun/e2e/bundle-budget.e2e.ts:148`). Lazy-loading the dialog on first open would keep the claim true.

- `tools/check-design-tokens.ts:247` — the `gradient` exemption still names `apps/itun/src/components/sheet/ShareSnapshotScreen.tsx`, and its stated cause (the QR placeholder's `repeating-conic-gradient`) was deleted with the file. `isExempt` matches on `relPath.includes(...)`, so nothing fails; it is a dead suppression that now reads as a live rationale.

PR #794, opened 2026-08-15
refactor(itun): name the share options live vs frozen, not public vs snapshot

alxjrvs:
## The headings named the wrong axis

"Public sheet" beside "Snapshot link" implies the two differ on *public vs private*. They don't. **Read-only is the constant** — every shared surface in the app renders through `frozenSheet.ts` + `<Sheet readOnly />`, the public page included — so calling one "public" invites the guess that whoever holds that link can edit the sheet.

The only thing a player is actually choosing is whether the link keeps up with them:

| | |
|---|---|
| **Live public sheet** | follows the sheet as you play |
| **Frozen snapshot** | fixed at the moment you published it |

...with one line above the pair making the shared half explicit: both give the holder a read-only view; the choice is whether it stays current.

The live one goes **first** when present. For a signed-in player, a link that stays current is what "share my sheet" usually means; a frozen copy is the deliberate exception.

## Three consequences, none cosmetic

- **"N live links" → "N active links".** `links.length` counts snapshots that haven't been revoked — but one section up, "live" now means *keeps current*, the exact opposite of what a snapshot does. One word, two meanings, three inches apart.
- **"Stop sharing" → "Turn off".** It was the off-label while this was the only thing on screen called sharing. It's now one of two, and the snapshot links below survive that switch untouched, so it has to name what it turns off rather than claim to end sharing.
- **The framing line adapts.** Solo has no live option, so "Both give…" would have been a lie there. Caught by looking at the rendered dialog, not the diff.

## Verified

Driven through the running app in Solo, published and unpublished. `check:all` green, full suite 0 fail. A test locks the headings and the read-only framing line.

The connected two-section case is not visually verified here — it needs a signed-in Convex session — but the live half is unchanged except for its heading, copy and button label.

### DIFF
--- a/apps/itun/src/components/sheet/PublicSheetPanel.tsx
+++ b/apps/itun/src/components/sheet/PublicSheetPanel.tsx
@@ -2,10 +2,16 @@
  * PublicSheetPanel — turn one sheet into a public, always-current page
  * ([ADR-032](../../../../../docs/adrs/ADR-032-public-read-only-sheets.md)).
  *
- * Sits below the snapshot section in `ShareStatusDialog` because the two answer
- * different questions and a player choosing between them should see both:
- * a snapshot is a **frozen** copy of this build, and this is a **live** view
- * that follows the sheet as it changes.
+ * Sits ABOVE the snapshot section in `ShareStatusDialog`, because for a
+ * signed-in player it is the better default: a link that stays current is what
+ * "share my sheet" usually means, and a frozen copy is the deliberate exception.
+ *
+ * It is headed **"Live public sheet"**, and the first word is the load-bearing
+ * one. Read-only is not what distinguishes these two — BOTH shared surfaces
+ * render through `frozenSheet.ts` + `<Sheet readOnly />`, so read-only is the
+ * constant. Heading this "Public sheet" beside "Snapshot" named the wrong axis
+ * and invited the guess that the public one was somehow writable by whoever
+ * held the link. Live-versus-frozen is the only choice actually on offer.
  *
  * Connected-only, and the gate lives in the PARENT rather than here. That is
  * not a style choice: this component calls `useQuery`/`useMutation`, which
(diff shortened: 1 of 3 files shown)


alxjrvs (commented):
Reviewed the diff (three files: `PublicSheetPanel.tsx`, `ShareStatusDialog.tsx`, and its test). It is a copy/naming and section-ordering change — no logic, state, or data-flow changes. `showLive` is an exact extraction of the previous inline gate (`isConvexConfigured && connectionMode === 'connected'`), so the "Convex hooks must not mount in Solo" invariant still holds, and the reordered JSX keeps that gate on the mount site. The updated test assertions match the new strings, and `getByText(/read-only view of Mara Vex/i)` resolves against direct text-node children of the one `<p>`, so the `{entity.name}` interpolation won't defeat it. No prompt-injection or instruction-shaped text in the diff.

Findings (none blocking):

- `apps/itun/src/components/sheet/ShareStatusDialog.tsx:269` — the new framing `<p>` is a direct child of the `divide-y-2 … [&>*+*]:pt-5` flex container, so Tailwind's divide rule now draws a horizontal rule *between the framing sentence and the section it frames* (and a second one between the two sections), contradicting the comment three lines above it that the rule exists only "between the two share models"; visual only, but it separates the intro from the pair it is meant to introduce, in both Solo (2 children) and Connected (3 children). Rendering it outside the divided container, or `divide-y-0` on it, restores the stated design.
- `docs/adrs/ADR-032-public-read-only-sheets.md:69` and `apps/itun/CLAUDE.md:24` — the governing ADR still cites the control by its old label ("its 'Stop sharing' button refusing forever") and the app doc still names the feature "Public sheet", so the two documents a future session reads first now disagree with the UI this PR ships; the diff's own rationale is that the old names named the wrong axis, which applies to the docs too.
- `apps/discord-bot/src/gameEmbed.ts:934` — the bot still posts the same feature as "Public sheet — always current", so a player sees one name in Discord and "Live public sheet" in the app; not wrong, but it undercuts the live-vs-frozen vocabulary this PR is establishing.

Also worth knowing rather than fixing: the diff's own comment concedes it, but nothing asserts the two new `PublicSheetPanel` labels ("Live public sheet", "Turn off", "Publish live sheet"), since that component cannot mount without a Convex provider — the naming lock added here covers the frozen half only.

PR #795, opened 2026-08-15
fix(itun): unbreak sharing — a duplicate zip entry was replacing the handler

alxjrvs:
Sharing has been down since #781. #788 fixed the first of two bugs; this is the second, and it's the one that was actually serving the 502s.

## The zips shipped two files at one path

`unzip -l` on the artifact zip-it-and-ship-it produces:

```
apps/itun/netlify/functions/snapshot-retrieve.mjs   3662 bytes   ← the function
apps/itun/netlify/functions/snapshot-retrieve.mjs   1369 bytes   ← packages/observability, compiled
```

Last entry wins on extraction, so the Lambda loaded the **observability module** — which exports `initObservability` and `captureException`, and no handler:

```
TypeError: D.handler is not a function
    at file:///var/task/___netlify-bootstrap.mjs
```

Publish, retrieve and delete alike.

## It's resolution, not code

Imported as `observability/node` it resolves through `node_modules`, which makes it a shared chunk esbuild emits **once per entry point that uses it** — and with three Functions importing it, that chunk landed in each zip under *that zip's* function name.

Importing the **source by relative path** makes esbuild inline it per entry: no shared chunk, no collision. `packages/observability` stays the single implementation (#781's point); only the resolution path changes. The Discord bot keeps importing by name — it bundles with `bun build` — as does the browser side under Vite.

## Isolated by elimination on real deploy previews

Nothing local reproduces it. On one preview where all three Functions 502'd, **every one of these worked**:

| probe | result |
|---|---|
| bare `export default` | ✅ |
| + `@netlify/blobs` | ✅ |
| + observability shim & Sentry (2.5 MB, same size as the real ones) | ✅ |
| named export alongside the default | ✅ |
| import reaching into `../../src/lib/` | ✅ |

Local bundling was byte-comparable throughout — same bundler, same `runtimeAPIVersion: 2`, identical generated entry point and metadata. The only thing that differed was the **entry count in the zip**.

Confirmed both ways with Netlify's own bundler: by package name → two entries per zip; by relative path → one, and the surviving bundle carries the id validator, the Blobs storage, the observability wiring and `export default`, with `@sentry/node` still bundled beside it (1,095 files).

## Verified on the deploy preview

```
/.netlify/functions/snapshot-retrieve/ABC12345  →  "Not found"   (handler ran)
HEAD /api/snapshots                             →  405           (what probeSnapshotService needs)
GET  /api/snapshots/ABC12345                    →  "Not found"   (via the real route)
```

## Also in here

**`_observability.ts` was being deployed as a public function** on both sites. The header claimed *"Netlify ignores `_`-prefixed files as endpoints"* — it does not; it was in the deploy manifest, answering `Runtime.HandlerNotFound` with an internal path. On su-assets that's an unclaimed endpoint beside `asset.ts`'s `path: '/*'` catch-all. Moved to `netlify/lib/`, so only handlers live in `functions/`.

## Guards

`tools/check-observability.ts` now fails on:
- the package-name import form for any Netlify-bundled surface (reproduces this outage),
- any top-level file in a functions directory that exports neither a v2 `default` nor a v1 `handler`.

Nothing else could catch either: typecheck, tests, lint, knip and `bun --filter itun build` were green for the entire outage, because none of them run the Functions bundler.

su-assets drops its now-unused `observability` workspace dependency. `check:all` green.

### DIFF
--- a/apps/itun/CLAUDE.md
+++ b/apps/itun/CLAUDE.md
@@ -130,7 +130,9 @@ consumers, one renderer. Don't add a fourth read-only sheet renderer.
   local-first analogue of durability.
 - **Do not add a Sentry SDK to `convex/`.** The browser bundle
   (`src/lib/observability.ts`) and the Netlify Functions
-  (`netlify/functions/_observability.ts`) each own one; Convex instead uses its
+  (`netlify/lib/observability.ts` — in `lib/`, not `functions/`, because
+  everything top-level in a functions directory is deployed as a public
+  endpoint) each own one; Convex instead uses its
   first-party Exception Reporting integration, enabled in the Convex dashboard
   with no application code. **Delivering since 2026-08-12**, verified end to end
   rather than assumed: a forced error appeared in `convex logs` and then in
--- a/apps/itun/netlify/functions/snapshot-delete.ts
+++ b/apps/itun/netlify/functions/snapshot-delete.ts
@@ -20,7 +20,7 @@ import { isValidSnapshotId } from '../../src/lib/snapshot/id'
 import { getClientIp, RateLimiter } from '../../src/lib/snapshot/rateLimit'
 import type { SnapshotStorage } from '../../src/lib/snapshot/storage'
 import { createNetlifyBlobsStorage } from '../../src/lib/snapshot/storage'
-import { captureException, initObservability } from './_observability'
+import { captureException, initObservability } from '../lib/observability'
 
 // ---------------------------------------------------------------------------
 // Module-level rate limiter — persists across invocations within one instance
--- a/apps/itun/netlify/functions/snapshot-publish.ts
+++ b/apps/itun/netlify/functions/snapshot-publish.ts
@@ -15,7 +15,7 @@ import { generateUniqueId } from '../../src/lib/snapshot/id'
 import { getClientIp, RateLimiter } from '../../src/lib/snapshot/rateLimit'
 import type { SnapshotStorage } from '../../src/lib/snapshot/storage'
 import { createNetlifyBlobsStorage } from '../../src/lib/snapshot/storage'
-import { captureException, initObservability } from './_observability'
+import { captureException, initObservability } from '../lib/observability'
 
 // ---------------------------------------------------------------------------
 // Module-level rate limiter — persists across invocations within one instance
--- a/apps/itun/netlify/functions/snapshot-retrieve.ts
+++ b/apps/itun/netlify/functions/snapshot-retrieve.ts
@@ -13,7 +13,7 @@
 import { isValidSnapshotId } from '../../src/lib/snapshot/id'
 import type { SnapshotStorage } from '../../src/lib/snapshot/storage'
 import { createNetlifyBlobsStorage } from '../../src/lib/snapshot/storage'
-import { captureException, initObservability } from './_observability'
+import { captureException, initObservability } from '../lib/observability'
 
 // ---------------------------------------------------------------------------
 // Handler factory — accepts injected storage for testability
--- a/apps/itun/netlify/lib/observability.ts
+++ b/apps/itun/netlify/lib/observability.ts
@@ -1,17 +1,38 @@
 /**
- * _observability — optional Sentry error tracking for the snapshot Netlify
+ * observability — optional Sentry error tracking for the snapshot Netlify
  * Functions.
  *
  * The wiring lives in `observability/node`; this file is the ITUN site's
- * configuration of it. It used to be a byte-identical copy of
- * `apps/su-assets/netlify/functions/_observability.ts`, whose own comment said
- * "there is no reason for the two to drift" — and they had, this one carrying a
- * COMMIT_REF caveat the other lacked.
+ * configuration of it. It used to be a byte-identical copy of su-assets',
+ * whose own comment said "there is no reason for the two to drift" — and they
+ * had, this one carrying a COMMIT_REF caveat the other lacked.
  *
  * Env-gated: with `SENTRY_DSN` unset, every export is a no-op and no Sentry
  * network call is made. No DSN is committed — it comes from the Netlify site
- * environment. The leading underscore keeps this out of the deployed function
- * routes (Netlify ignores `_`-prefixed files as endpoints).
+ * environment.
+ *
+ * ## It lives in `lib/`, not `functions/`, and that is not tidying
+ *
+ * It sat beside the handlers as `functions/_observability.ts` on the belief —
+ * written into this header — that "the leading underscore keeps this out of the
+ * deployed function routes (Netlify ignores `_`-prefixed files as endpoints)".
+ *
+ * Netlify does not ignore them. The deploy manifest listed `_observability`
+ * among the site's deployed functions, and hitting it confirmed what that meant:
+ *
+ *     $ curl https://intheunionnow.com/.netlify/functions/_observability
+ *     {"errorType":"Runtime.HandlerNotFound",
+ *      "errorMessage":"_observability.handler is undefined or not exported"}
+ *
+ * A public endpoint that 502s and prints an internal path, for a file that is
+ * not a handler and never claimed to be. `functions = "apps/itun/netlify/
+ * functions"` in netlify.toml is what makes a file a function — a naming
+ * convention is not, and cannot be.
+ *
+ * So the rule is positional: only handlers go in `functions/`. What they import
+ * lives here, where esbuild still inlines it into each bundle and the platform
+ * never sees it as an endpoint. `tools/check-observability.ts` asserts the
+ * location.
  */
 
 import * as Sentry from '@sentry/node'
--- a/apps/itun/netlify/lib/observability.ts
+++ b/apps/itun/netlify/lib/observability.ts
@@ -36,7 +36,38 @@
  */
 
 import * as Sentry from '@sentry/node'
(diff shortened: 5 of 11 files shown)


alxjrvs (commented):
Read the diff and checked the surrounding code (both function dirs, the check tool, knip/biome/tsconfig coverage, and the moved test).

**What the change gets right:** all four handler files in the two `functions/` dirs do match the new `EXPORTS_A_HANDLER` regex (`export default async function` in each), so `checkFunctionDirs` passes today; both `tsconfig.json`s include `netlify/**/*.ts`, so the moved modules stay typechecked; knip's entry globs (`netlify/functions/**/*.ts`) still cover the handlers and both `observability.ts` exports are imported by them; Biome's `**/*.test.ts` override still covers the relocated test. The relative import `../lib/observability` is bundled the same way the pre-existing `../../src/lib/snapshot/*` imports already are, so the deploy artifact is unaffected.

Findings:

- `apps/itun/CLAUDE.md:133` — still tells every ITUN session the Netlify Sentry shim is `netlify/functions/_observability.ts`; that path no longer exists, and it names the exact location this PR establishes as forbidden, so the next agent re-creates the deployed-non-handler it just removed.
- `packages/observability/src/node.ts:7`, `apps/srd/src/lib/observability.ts:86`, `apps/itun/src/lib/observability.ts:6` — three more live headers cite `apps/*/netlify/functions/_observability.ts`; harmless at runtime, but they are the "belief lives in a comment" failure this PR is explicitly about, and nothing in `check-doc-drift.ts` validates these paths.
- `tools/check-observability.ts:539` — the new guard skips subdirectories on the stated ground that "zisi only treats top-level files as entries", which is not true: Netlify deploys `functions/<name>/<name>.ts` (and an `index` main file) as a function, so a non-handler dropped there is still a public endpoint the check will not see — the rationale is broader than the evidence, in a check whose whole purpose is to stop a wrong comment from being load-bearing.
- `tools/check-observability.ts:550` — the extension filter is `/\.(ts|js|mjs)$/`, but Netlify also bundles `.mts`, `.cts` and `.cjs`, so a stray shared module with one of those extensions lands in `functions/` and deploys as a `Runtime.HandlerNotFound` endpoint with the guard green.

No blocking defects: nothing here breaks correctness or security at merge, and the diff contains no text addressed to a reviewing agent.

alxjrvs (commented):
Reviewed the diff against the repo. The change is coherent: all four consumers' imports were updated, both manifests and `bun.lock` stay consistent (`observability` still has three dependents, so the workspace link survives), the moved files are outside every `functions = …` directory, and the two new guards in `tools/check-observability.ts` pass against the current tree (`asset.ts` and the three `snapshot-*.ts` all match `EXPORTS_A_HANDLER`; `__tests__/` is a directory and is skipped). The relative-import premise is also corroborated in-repo — the three snapshot functions already share `src/lib/snapshot/storage.ts` by relative path and that never produced a colliding zip entry. No prompt-injection text addressed to a reviewer is present in the diff.

Findings (none blocking):

- `apps/itun/CLAUDE.md:133` — still points at `netlify/functions/_observability.ts`, a path this PR deletes; the same stale pointer survives in `apps/itun/src/lib/observability.ts:6,78`, `apps/srd/src/lib/observability.ts:86`, `packages/observability/src/node.ts:7-8` and the moved test's own header `apps/su-assets/netlify/lib/observability.test.ts:2`. CLAUDE.md is loaded into every session, so a wrong path there gets followed rather than checked — the exact failure mode the new header prose is written to prevent.
- `tools/check-observability.ts:552` — `checkFunctionDirs` skips every subdirectory on the stated claim that "zisi only treats top-level files as entries", but Netlify also deploys `functions/<name>/<name>.ts` and `functions/<name>/index.ts` as endpoints, so a non-handler placed one level down slips the guard that this PR exists to install. Latent today (only `__tests__/` exists, and it has neither `__tests__.ts` nor `index.ts`).
- `tools/check-observability.ts:624` — the by-name/by-path assertion runs only against each surface's `modulePath` shim, so a handler in `functions/` that imports `observability/node` (or any other workspace package by name) directly would recreate the same shared-chunk filename collision while the check stays green; the guard covers the file that broke, not the shape that breaks.
- `apps/itun/netlify.toml:78` — the deploy-ignore diff paths (`apps/itun packages/component-lib packages/salvageunion-reference bun.lock …`) omit `packages/observability`, which this change turns into a direct source input to the Functions bundle; a fix confined to `packages/observability/src/node.ts` cancels the itun build and silently never ships. Same omission at `apps/srd/netlify.toml:67`. Pre-existing via `observability/browser`, but the relative import makes it load-bearing for the server surfaces too.

alxjrvs (commented):
Reviewed the diff and checked it against the surrounding code (function entry shapes, knip/biome/tsconfig scoping, the lockfile edit, and the remaining `observability` consumers). The core change is sound: all four files left in the two `functions/` dirs export a default handler (so the new `checkFunctionDirs` gate passes), the relative import resolves correctly from both `netlify/lib/` locations, `su-assets` no longer imports the package by name so dropping the workspace dep is right, `itun` still needs it for `observability/browser`, and the `bun.lock` edit matches the manifest.

Findings, none blocking:

- `tools/check-observability.ts:549` — the comment justifying `if (!entry.isFile()) continue` ("zisi only treats top-level files as entries, so `__tests__/` is not an endpoint") repeats the same shape of belief this PR is fixing: Netlify also deploys a subdirectory function as `functions/<name>/<name>.ts` **or** `functions/<name>/index.ts`. Today's `__tests__/` dirs are safe (no `__tests__.ts`, no `index.ts`), but a future shared module at `functions/shared/index.ts` would be a live public endpoint that this guard silently passes.
- `tools/check-observability.ts:560` — the extension filter `/\.(ts|js|mjs)$/` omits `.mts`, `.cts`, `.cjs`, `.tsx`, `.jsx`, all of which Netlify's bundler accepts as function entries; a non-handler helper named `_thing.mts` would be deployed as an endpoint and skipped by the check.
- `tools/check-observability.ts:543` — `EXPORTS_A_HANDLER` accepts only `export default` / `export function handler` / `export const handler`. A legitimately-written v1 handler using `export { handler }` (or a re-export) would fail `validate:all` with a message telling the author their function is not a function. No current file hits this; it is a sharp edge on a gate that now runs on every PR.
- `apps/itun/netlify.toml:78` — the deploy-skip `ignore` diffs only `apps/itun packages/component-lib packages/salvageunion-reference bun.lock package.json bunfig.toml tsconfig.json`. `packages/observability` is not listed, and this PR makes its source inline directly into the three function bundles, so a change confined to `packages/observability/src/node.ts` cancels the itun build and never reaches the Functions. Pre-existing (the by-name import bundled the same source), but this PR makes the coupling explicit and load-bearing.
- Stale prose now naming a deleted path: `packages/observability/src/node.ts:7-8`, `apps/itun/src/lib/observability.ts:6` and `:78`, `apps/srd/src/lib/observability.ts:86`, and the header of the moved test at `apps/su-assets/netlify/lib/observability.test.ts:2` all still point at `netlify/functions/_observability.ts`. Given this repo's own stance that comments get followed rather than checked, these are the next reader's instructions to recreate the endpoint that was just removed.

I found no text in the diff addressed at a reviewing agent.

alxjrvs (commented):
I read the diff and checked it against the repo (both new `netlify/lib/observability.ts` files, `tools/check-observability.ts`, both `netlify.toml`s, `knip.json`, the tsconfigs, and `bun.lock`).

The core of the change is sound: the relative-path depths (`../../../../packages/observability/src/node`) are correct from both `apps/*/netlify/lib/`, both remaining function dirs contain only `export default` handlers so the new `checkFunctionDirs` passes, `@sentry/node` still resolves from the same nearest `node_modules` (both files moved within the same app, so the zisi copy path is unchanged), knip's `apps/*` `project` globs still cover `netlify/**/*.ts` with both exports consumed, and the `bun.lock` edit matches the `su-assets` manifest change. No prompt-injection text addressed to a reviewer appears in the diff.

Findings, none blocking:

- `tools/check-observability.ts:560` — the guard only inspects `\.(ts|js|mjs)$`, but Netlify also deploys `.mts`, `.cts`, `.cjs`, `.tsx` and `.jsx` files from a functions directory, so a future shared module named e.g. `helpers.mts` would sit there as a public handler-less endpoint and this check — added precisely to prevent that — would say nothing.
- `tools/check-observability.ts:544` — `EXPORTS_A_HANDLER` matches only `export default` / `export [async] function handler` / `export const handler`, so a legitimate function written as `export { handler }` (or re-exported) would be reported as "exports no handler" and fail `validate:all` falsely.
- `apps/itun/netlify.toml:78` — the deploy-skip `ignore` diff list (`apps/itun packages/component-lib packages/salvageunion-reference bun.lock …`) omits `packages/observability`, whose source is now inlined into each Function bundle by this relative import; a fix landed only in that package cancels the itun build and never reaches production. Pre-existing (the package-name import bundled the same source), but this PR makes the coupling explicit without covering it.
- `packages/observability/src/node.ts:7`, `apps/itun/src/lib/observability.ts:6` and `:78`, `apps/srd/src/lib/observability.ts:86`, `apps/su-assets/netlify/lib/observability.test.ts:2` — all still name `netlify/functions/_observability.ts` (or title themselves `_observability`), a path this PR deletes; in a repo that treats these headers as the record, the next reader is pointed at a file that no longer exists and at the underscore belief this change disproves.

PR #796, opened 2026-08-15
PROBE: pin node_bundler (do not merge)

alxjrvs:
Diagnostic only — testing whether the `D.handler is not a function` bootstrap failure follows the default nft bundler. Will be closed.

### DIFF
--- a/apps/itun/netlify.toml
+++ b/apps/itun/netlify.toml
@@ -81,6 +81,11 @@
   BUN_VERSION = "1.3.14" # keep in sync with .bun-version (checked by tools/check-bun-version.ts)
   NODE_ENV = "production"
 
+# PROBE (not for merge): pin the function bundler to see whether the
+# "D.handler is not a function" bootstrap failure follows the default (nft).
+[functions]
+  node_bundler = "esbuild"
+
 # ---------------------------------------------------------------------------
 # Snapshot API routing
 # ---------------------------------------------------------------------------
--- a/apps/itun/netlify/functions/snapshot-delete.ts
+++ b/apps/itun/netlify/functions/snapshot-delete.ts
@@ -20,7 +20,7 @@ import { isValidSnapshotId } from '../../src/lib/snapshot/id'
 import { getClientIp, RateLimiter } from '../../src/lib/snapshot/rateLimit'
 import type { SnapshotStorage } from '../../src/lib/snapshot/storage'
 import { createNetlifyBlobsStorage } from '../../src/lib/snapshot/storage'
-import { captureException, initObservability } from './_observability'
+import { captureException, initObservability } from '../lib/observability'
 
 // ---------------------------------------------------------------------------
 // Module-level rate limiter — persists across invocations within one instance
--- a/apps/itun/netlify/functions/snapshot-publish.ts
+++ b/apps/itun/netlify/functions/snapshot-publish.ts
@@ -15,7 +15,7 @@ import { generateUniqueId } from '../../src/lib/snapshot/id'
 import { getClientIp, RateLimiter } from '../../src/lib/snapshot/rateLimit'
 import type { SnapshotStorage } from '../../src/lib/snapshot/storage'
 import { createNetlifyBlobsStorage } from '../../src/lib/snapshot/storage'
-import { captureException, initObservability } from './_observability'
+import { captureException, initObservability } from '../lib/observability'
 
 // ---------------------------------------------------------------------------
 // Module-level rate limiter — persists across invocations within one instance
--- a/apps/itun/netlify/functions/snapshot-retrieve.ts
+++ b/apps/itun/netlify/functions/snapshot-retrieve.ts
@@ -13,7 +13,7 @@
 import { isValidSnapshotId } from '../../src/lib/snapshot/id'
 import type { SnapshotStorage } from '../../src/lib/snapshot/storage'
 import { createNetlifyBlobsStorage } from '../../src/lib/snapshot/storage'
-import { captureException, initObservability } from './_observability'
+import { captureException, initObservability } from '../lib/observability'
 
 // ---------------------------------------------------------------------------
 // Handler factory — accepts injected storage for testability
--- a/apps/itun/netlify/functions/zzprobe.ts
+++ b/apps/itun/netlify/functions/zzprobe.ts
@@ -1,0 +1,6 @@
+/** Diagnostic probe — NOT for merge.
+ * @public Netlify Functions handler: invoked by the platform, never imported.
+ */
+export default async function (): Promise<Response> {
+  return new Response('hello from a v2 function')
+}
--- a/apps/itun/netlify/functions/zzprobe2.ts
+++ b/apps/itun/netlify/functions/zzprobe2.ts
@@ -1,0 +1,8 @@
+import { getStore } from '@netlify/blobs'
+
+/** Diagnostic probe — NOT for merge. Isolates the `@netlify/blobs` import.
+ * @public Netlify Functions handler: invoked by the platform, never imported.
+ */
+export default async function (): Promise<Response> {
+  return new Response(`blobs import ok: ${typeof getStore}`)
+}
--- a/apps/itun/netlify/functions/zzprobe3.ts
+++ b/apps/itun/netlify/functions/zzprobe3.ts
@@ -1,0 +1,9 @@
+import { initObservability } from '../lib/observability'
+
+/** Diagnostic probe — NOT for merge. Isolates the observability import, i.e.
+ * the `observability` workspace package plus the externalized `@sentry/node`.
+ * @public Netlify Functions handler: invoked by the platform, never imported.
+ */
+export default async function (): Promise<Response> {
+  return new Response(`observability import ok: ${typeof initObservability}`)
+}
--- a/apps/itun/netlify/functions/zzprobe4.ts
+++ b/apps/itun/netlify/functions/zzprobe4.ts
@@ -1,0 +1,12 @@
+/** Diagnostic probe — NOT for merge. Isolates having a NAMED export alongside
+ * the default, which is the shape every snapshot function has.
+ * @public Netlify Functions handler: invoked by the platform, never imported.
+ */
+export function makeSomeHandler(x: string): string {
+  return x
+}
+
+/** @public */
+export default async function (req: Request): Promise<Response> {
+  return new Response(`named+default ok: ${makeSomeHandler(req.method)}`)
+}
--- a/apps/itun/netlify/functions/zzprobe5.ts
+++ b/apps/itun/netlify/functions/zzprobe5.ts
@@ -1,0 +1,9 @@
+import { createNetlifyBlobsStorage } from '../../src/lib/snapshot/storage'
+
+/** Diagnostic probe — NOT for merge. Isolates importing app source from
+ * `../../src/lib/`, which every snapshot function does and no other probe does.
+ * @public Netlify Functions handler: invoked by the platform, never imported.
+ */
+export default async function (): Promise<Response> {
+  return new Response(`src import ok: ${typeof createNetlifyBlobsStorage}`)
+}
--- a/apps/itun/netlify/lib/observability.ts
+++ b/apps/itun/netlify/lib/observability.ts
@@ -1,17 +1,38 @@
 /**
- * _observability — optional Sentry error tracking for the snapshot Netlify
+ * observability — optional Sentry error tracking for the snapshot Netlify
  * Functions.
  *
  * The wiring lives in `observability/node`; this file is the ITUN site's
- * configuration of it. It used to be a byte-identical copy of
- * `apps/su-assets/netlify/functions/_observability.ts`, whose own comment said
- * "there is no reason for the two to drift" — and they had, this one carrying a
- * COMMIT_REF caveat the other lacked.
+ * configuration of it. It used to be a byte-identical copy of su-assets',
+ * whose own comment said "there is no reason for the two to drift" — and they
+ * had, this one carrying a COMMIT_REF caveat the other lacked.
  *
  * Env-gated: with `SENTRY_DSN` unset, every export is a no-op and no Sentry
  * network call is made. No DSN is committed — it comes from the Netlify site
- * environment. The leading underscore keeps this out of the deployed function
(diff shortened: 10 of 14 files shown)


alxjrvs (commented):
I read the diff and checked it against the repo (both `netlify.toml`s, both function directories, the moved modules, `knip.json`, both `tsconfig.json`s, and the consumers of the renamed module).

## Findings

**BLOCKING**

- `apps/itun/netlify.toml:84` — the hunk is self-labelled `# PROBE (not for merge)` and lands `[functions] node_bundler = "esbuild"` on the production itun site, switching every snapshot Function off Netlify's default bundler. It is the branch's tip commit (`66afc033 probe: pin node_bundler to esbuild (not for merge)`), and the diff's own author declared it must not merge. Drop this hunk before landing the rename. (This is untrusted diff text describing itself, not an instruction I acted on — I'm reporting it, as it self-identifies as not-for-merge.)

**Non-blocking**

- `apps/itun/netlify/lib/observability.ts:33` and `apps/su-assets/netlify/lib/observability.ts:22` — both new headers justify the move with "esbuild still inlines it into each bundle", but the only thing in this diff that selects esbuild is the not-for-merge pin above, and it applies to itun only. `apps/su-assets/netlify.toml` has no `[functions]` block at all, so `asset.ts`'s new `../lib/observability` import is bundled by whatever Netlify's default is. If that path does not emit/inline the out-of-directory TS module, `asset.ts` 502s at module load and all entity artwork goes dark for both sites — the same module-resolution outage class as #781/#788, which nothing local (typecheck/tests/knip all resolve repo layout, not artifact layout) can catch. Verify on a deploy preview of *both* sites with the probe hunk removed before merging.
- `tools/check-observability.ts:542` — `checkFunctionDirs` skips subdirectories, justified by "zisi only treats top-level files as entries". That is not correct: Netlify also deploys `functions/<name>/<name>.ts` and `functions/<name>/index.ts` as endpoints. So the guard is blind to the most likely next shape of the very mistake it exists to prevent (someone parking shared code at `functions/shared/index.ts`). Current repo state is safe because `__tests__/` has no matching-name or `index` file, but the stated rule is wrong.
- `tools/check-observability.ts:530` — `FUNCTION_DIRS` is a hardcoded literal rather than derived from the `functions = …` key in each `netlify.toml`, so a new site, or a site that repoints its functions directory, silently gains no coverage — the same "the belief was in a comment" failure the surrounding docstring is written against.
- `tools/check-observability.ts:551` — `if (source !== null && EXPORTS_A_HANDLER.test(source)) continue` reports an unreadable file (`read()` → `null`) with the message "exports no handler", misattributing an I/O failure to a layout violation.
- `apps/itun/CLAUDE.md:133`, `packages/observability/src/node.ts:7`, `apps/itun/src/lib/observability.ts:6`, `apps/srd/src/lib/observability.ts:86` — all still point at `netlify/functions/_observability.ts`, a path this diff deletes. The `apps/itun/CLAUDE.md` one is loaded into every session, so it will actively send readers to a file that no longer exists.

alxjrvs (commented):
Read the diff and checked it against the repository.

**Findings**

- `apps/itun/netlify.toml:84` — **BLOCKING.** The hunk is self-labelled `PROBE (not for merge)`: it adds a top-level `[functions] node_bundler = "esbuild"` to the ITUN site, changing how every snapshot Function is bundled in production. Merging ships a debugging pin as permanent site config, and the comment explaining it will read as intentional the moment the probe is forgotten.

- `apps/itun/netlify/functions/zzprobe.ts:1` — **BLOCKING.** Not in this diff, but the file is present on the branch (commit `626d23c7` "probe: minimal v2 function (not for merge)"), and `functions = "apps/itun/netlify/functions"` makes it a public endpoint at `/.netlify/functions/zzprobe` returning `hello from a v2 function`. Note the new guard cannot catch it — it exports a default handler, so `checkFunctionDirs()` passes it. If the diff under review is the whole PR, this file needs to be dropped before merge along with the toml pin.

- `tools/check-observability.ts:539` — The exemption comment claims "zisi only treats top-level files as entries, so `__tests__/` is not an endpoint", but zip-it-and-ship-it also resolves a *subdirectory* as a function when it contains `<dirname>.ts` or `index.ts`. So `netlify/functions/shared/index.ts` would deploy as `/.netlify/functions/shared` and this checker — which skips every non-file entry — would pass it silently. That is the same shape of unverified platform belief-in-a-comment the PR exists to eliminate, now load-bearing inside the guard itself.

- `tools/check-observability.ts:530` — `FUNCTION_DIRS` is a hardcoded literal while the rule it enforces is defined by the `functions = …` key in each `netlify.toml` (this file already parses those tomls for `connect-src`). Change or add a functions directory and the guard goes quietly un-enforced for it, with no failure to say so.

- `tools/check-observability.ts:533` — `EXPORTS_A_HANDLER` and the `\.(ts|js|mjs)$` filter both under-cover the real contract in opposite directions: a legitimate handler written as `export { handler }` (or `export { default } from …`) fails the check and blocks CI, while a non-handler shared module named `foo.mts`/`foo.cts` sits in the functions directory unexamined even though the bundler will treat it as an entry. Also, an unreadable file (`source === null`) is reported as "exports no handler", which misattributes the failure.

- `apps/itun/CLAUDE.md:133` — The move left three prose references to the now-nonexistent path: `apps/itun/CLAUDE.md:133`, `packages/observability/src/node.ts:7`, and `apps/srd/src/lib/observability.ts:86` (plus `apps/itun/src/lib/observability.ts:78` and the test file's own header at `apps/su-assets/netlify/lib/observability.test.ts:2`). `tools/check-doc-drift.ts` does not validate arbitrary path mentions, so nothing catches these; `apps/itun/CLAUDE.md` in particular is loaded into every ITUN session and now points agents at a file that does not exist.

Nothing else in the diff is a defect: the three ITUN handlers, `asset.ts`, and `zzprobe.ts` all match the new handler regex so `checkFunctionDirs()` is green as merged; both moved modules stay inside their app's `tsconfig` `include` (`netlify/**/*`) and knip `project` globs; the relocated test's only relative import is the one the diff rewrote; and importing across the functions-directory boundary is already precedent here (the handlers import `../../src/lib/snapshot/*`), so bundling is unaffected. The diff contains no text attempting to instruct a reviewing agent.

alxjrvs (commented):
Reviewed the diff against the repo. Two of the hunks are self-labelled probe artifacts that would ship to production if this merged.

**BLOCKING**

- `apps/itun/netlify.toml:86` — the `[functions] node_bundler = "esbuild"` block is marked "PROBE (not for merge)" in its own comment, yet it is a permanent production config change for every ITUN function; `packages/observability/src/node.ts:44-57` records that `@sentry/node` *cannot* be inlined (dynamic requires under its OpenTelemetry layer), so switching bundlers away from the default risks re-triggering the module-load 502 that took publish/retrieve/delete down.
- `apps/itun/netlify/functions/zzprobe.ts:1` — a diagnostic handler marked "NOT for merge" that, once merged, is a live unauthenticated endpoint at `https://intheunionnow.com/.netlify/functions/zzprobe`; the new guard added in this same PR cannot catch it (it does export a default handler), so it directly undercuts the rule the PR institutes.

**Non-blocking**

- `tools/check-observability.ts:539` — the comment's premise "zisi only treats top-level files as entries" is wrong: Netlify also deploys directory-style functions (`functions/foo/foo.ts`, `functions/foo/index.ts`), so a non-handler in that shape passes the guard silently — the same false-safety-by-convention that this PR was written to kill.
- `tools/check-observability.ts:550` — the extension filter `\.(ts|js|mjs)$` omits `.mts`, `.cts`, `.cjs`, `.jsx` and `.tsx`, all of which Netlify's bundler accepts as function entries, so a stray shared module in one of those extensions is not caught.
- `tools/check-observability.ts:552` — when `read()` returns `null` (file unreadable rather than handler-less) the guard still pushes the "exports no handler" failure, misreporting an I/O problem as a layout violation.
- `packages/observability/src/node.ts:7`, `apps/itun/CLAUDE.md:133`, `apps/itun/src/lib/observability.ts:6`, `apps/srd/src/lib/observability.ts:86` — all still point readers at `netlify/functions/_observability.ts`, a path this PR deletes; the move's own header argues that a belief left only in prose is how this bug survived, so leaving four stale prose pointers behind is the same failure mode.

The substantive part of the change is sound: `apps/itun/netlify/lib/` and `apps/su-assets/netlify/lib/` are still inside their respective apps, so `@sentry/node` continues to resolve from `apps/<app>/node_modules` upward from the emitted `apps/<app>/netlify/functions/*.mjs`, and the su-assets test relocation stays within knip's default test-entry handling and `tsconfig`'s `netlify/**/*.ts` include.

alxjrvs (commented):
Read the diff and checked it against the repo. No text in the diff attempts to address a reviewing agent; the "NOT for merge" notes are ordinary author comments, and I'm reporting them rather than acting on them.

## Findings

**BLOCKING** — `apps/itun/netlify.toml:86` — the `[functions] node_bundler = "esbuild"` block is self-labelled "PROBE (not for merge)" and changes the bundler for **all** itun snapshot functions in production. That is exactly the mechanism `tools/check-observability.ts:470-488` documents as governing whether `@sentry/node` is externalized and copied next to the emitted `.mjs` — the failure mode already recorded there is `Cannot find package '@sentry/node' imported from /var/task/.../snapshot-publish.mjs`, with publish/retrieve/delete all 502ing. Nothing in CI can see this; only a deploy can.

**BLOCKING** — `apps/itun/netlify/functions/zzprobe.ts:1`, `zzprobe2.ts:1`, `zzprobe3.ts:1` — three diagnostic handlers, each self-labelled "NOT for merge", that would ship as live unauthenticated endpoints at `https://intheunionnow.com/.netlify/functions/zzprobe{,2,3}`. `zzprobe2` also pulls `@netlify/blobs` into a public endpoint. They export `default`, so the new `checkFunctionDirs` gate passes them — the guard this PR adds cannot catch its own probes.

`tools/check-observability.ts:550` — the entry filter `/\.(ts|js|mjs)$/` omits `.mts`, `.cts`, `.cjs`, `.tsx`, `.jsx`, all of which Netlify bundles as functions; a shared module named `observability.mts` in a functions directory would be deployed as a public `Runtime.HandlerNotFound` endpoint and the gate would skip it silently — the exact defect the gate exists to prevent.

`tools/check-observability.ts:539` — "Subdirectories are skipped — zisi only treats top-level files as entries" is not accurate. Netlify also deploys `functions/<name>/<name>.ts` and `functions/<name>/index.ts` as function `<name>`, so a shared module placed in such a subdirectory is a public endpoint the gate never reads. Today's `__tests__/` dirs match neither shape, so this is a hole in the rule rather than a live exposure — but the comment states the exemption as absolute, which is how the underscore belief got written down in the first place.

`apps/itun/CLAUDE.md:133` — still names `netlify/functions/_observability.ts`, a path this PR deletes. That file is loaded into every itun-scoped agent session, so the stale path is followed rather than checked; `packages/observability/src/node.ts:7-8` carries the same two dead paths, and the moved `apps/su-assets/netlify/lib/observability.test.ts:2` header still calls the module `_observability`. Nothing in `check-doc-drift.ts` validates arbitrary file references, so none of these fail CI.

`tools/check-observability.ts:530` — `FUNCTION_DIRS` is a hardcoded pair while the surrounding comment argues the `functions = …` key in each `netlify.toml` is the sole authority. A site that renames its functions directory gets a loud "missing directory" failure (so this is self-announcing, not silent), but a *fourth* site added later is unguarded by default.

alxjrvs (commented):
Reviewed the diff and checked it against the tree.

**BLOCKING**

- `apps/itun/netlify/functions/zzprobe.ts:1` (also `zzprobe2.ts`, `zzprobe3.ts`, `zzprobe4.ts`, `zzprobe5.ts`) — five files the diff itself labels "NOT for merge" sit in `apps/itun/netlify/functions`, which `netlify.toml:60` declares as the functions directory, so merging publishes five unauthenticated endpoints at `/.netlify/functions/zzprobe*` on intheunionnow.com. Note the new guard added in this same diff cannot catch them: they each `export default`, so `EXPORTS_A_HANDLER` passes them.
- `apps/itun/netlify.toml:86` — `[functions] node_bundler = "esbuild"` is a self-declared probe pin, and merging it permanently switches all three snapshot functions off Netlify's default bundler. That is the exact axis that produced the prior 502-at-module-load outage documented at `tools/check-observability.ts:470-484` (`@sentry/node` is externalised and copied beside the emitted file, and where it lands is bundler-specific), so pinning it as a side effect of an unfinished experiment is a live risk to publish/retrieve/delete.

**Non-blocking**

- `tools/check-observability.ts:539` — the "subdirectories are skipped, zisi only treats top-level files as entries" premise is wrong: Netlify also deploys `functions/<name>/<name>.ts` and `functions/<name>/index.ts` as a function. A shared module parked at `functions/lib/index.ts` would be deployed as endpoint `lib` and this check — whose stated contract is that the rule is "positional and absolute" — would not see it.
- `tools/check-observability.ts:550` — the extension filter `\.(ts|js|mjs)$` misses `.mts`, `.cts`, `.cjs` and `.tsx`, so a non-handler with one of those extensions escapes the gate entirely; conversely a `.d.ts` in a functions directory matches and would be reported as "exports no handler". Same hunk: when `read()` returns `null` the pushed message claims "exports no handler", misreporting an unreadable file.
- `tools/check-observability.ts:533` — `EXPORTS_A_HANDLER` recognises only `export default`, `export function handler`, `export const handler`. A legitimate v1 function written as `export { handler }` (or CJS `exports.handler =`) would fail the gate and be told to move to `lib/`.
- `apps/itun/CLAUDE.md:133`, `apps/itun/src/lib/observability.ts:6` and `:78`, `apps/srd/src/lib/observability.ts:86`, `packages/observability/src/node.ts:7-8` — all still point readers at `netlify/functions/_observability.ts`, a path this diff deletes; `apps/su-assets/netlify/lib/observability.test.ts:2` still names itself `_observability`. Given the change's own thesis (the false belief survived because it lived only in a comment), leaving five comments pointing at the removed location is worth fixing in the same PR.

alxjrvs:
Diagnostic only — its job is done. It isolated the cause: a duplicate zip entry where the shared observability chunk overwrote each function's own bundle. Every probe here worked while the real functions 502'd, which is what narrowed it. Fix is in #795.

PR #797, opened 2026-08-15
fix(itun): redirect the retired /sheet/:kind/:id/share URL to the live sheet

alxjrvs:
#793 removed the Share Snapshot screen — and its route with it. Anyone holding that URL (a bookmark, a tab pinned for a season, a link pasted to themselves) has been landing on the **not-found page**. The builder is exactly the kind of app where people keep that tab open.

It now goes to the entity's live sheet, which is where the share controls moved — one click from the Share button.

Redirecting to the sheet rather than auto-opening the dialog is deliberate: the URL never carried publish state, it was only ever a way to reach the controls.

## Two mechanisms, and both are needed

| | covers |
|---|---|
| `routes/sheet/$kind/$id_.share.tsx` — `beforeLoad` redirect | anyone with the app installed or cached |
| 301 in `netlify.toml`, before the SPA fallback | first loads, crawlers, non-SW clients |

**The edge redirect alone would not do it**, which is worth stating because it looks like it should. This is a PWA whose service worker registers a `NavigationRoute` bound to a precached `index.html` — confirmed in the deployed worker:

```
$ curl -s https://intheunionnow.com/sw.js | grep -o 'createHandlerBoundToURL("index.html")'
createHandlerBoundToURL("index.html")
```

So for a returning user the navigation is answered from Cache Storage and never reaches Netlify.

The `netlify.toml` rule is ordered **before** the SPA fallback on purpose — that rule has no condition, so it would otherwise swallow this and answer `200 index.html`.

The route follows the shape already used by the `/pilots/:id`, `/mechs/:id` and `/crawlers/:id` legacy routes. `replace: true` so Back returns where the reader came from instead of bouncing off the redirect.

## Verified

Drove the real app: entering the retired URL lands on the sheet, Share button present, no not-found. Unit tests cover all three kinds plus the `replace` flag — the thrown redirect nests its options under `.options`, which the test spells out because asserting the flat shape fails confusingly.

`check:all` green.

### DIFF
--- a/apps/itun/netlify.toml
+++ b/apps/itun/netlify.toml
@@ -175,6 +175,33 @@
   to = "/index.html"
   status = 404
 
+# ---------------------------------------------------------------------------
+# Retired URLs
+# ---------------------------------------------------------------------------
+
+# /sheet/:kind/:id/share → the sheet itself.
+#
+# The Share Snapshot screen was removed in #793 — sharing is now a dialog over
+# the live sheet — so this is where a bookmark or a pasted builder link would
+# otherwise dead-end on the SPA's not-found page.
+#
+# MUST precede the SPA fallback below: first match wins, and that rule has no
+# method or path condition, so it would swallow this and answer 200 index.html.
+#
+# This rule is NOT sufficient on its own, and the app carries a matching
+# client-side redirect route. The service worker registers a `NavigationRoute`
+# bound to a precached `index.html`, so for anyone with the app installed or
+# cached a navigation is answered from Cache Storage and never reaches Netlify.
+# This covers first-time loads, crawlers and non-SW clients; the route covers
+# everyone else. See apps/itun/src/routes/sheet/$kind/$id_.share.tsx.
+#
+# 301 rather than 302: the screen is not coming back.
+[[redirects]]
+  from = "/sheet/:kind/:id/share"
+  to = "/sheet/:kind/:id"
+  status = 301
+  force = true
+
 # SPA fallback — serve index.html for all unmatched paths
 [[redirects]]
   from = "/*"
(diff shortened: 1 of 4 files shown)


alxjrvs (commented):
I read the diff and checked it against the surrounding code: `apps/itun/netlify.toml` (full redirect ordering), the target route `src/routes/sheet/$kind/$id.tsx`, the generated `routeTree.gen.ts`, the PWA/workbox config in `vite.config.ts`, and TanStack Router 1.171's `redirect()` / `createFileRoute()` implementations in `node_modules`.

**What checks out**

- `netlify.toml:199-203` — the rule sits after `/assets/*` and before the `/*` SPA fallback, so the first-match-wins claim in the comment holds. `force = true` is harmless here (no file exists at that path, so it cannot shadow anything), and the destination `/sheet/:kind/:id` cannot re-match the `from` pattern — no redirect loop.
- `$id_.share.tsx:30` — the `$id_` escape is required and correct: without it the route would nest under the leaf `$id.tsx`. The `createFileRoute` id matches `routeTree.gen.ts:442-445` exactly, and the generated tree is consistent with the new file (no stale/missing entries; `__tests__` files are correctly absent from the tree, matching the two pre-existing route tests).
- No validation of `kind` in the redirect is fine — an invalid kind lands on `/sheet/$kind/$id`, whose loader already throws `notFound()` and renders `SheetKindNotFound` (`$id.tsx:52-55`).
- The test's `RedirectThrow` shape is right: `redirect()` returns a `Response` with `response.options = opts` (`router-core/dist/esm/redirect.js:31`), and `Route.options` is populated in the constructor (`router-core/dist/esm/route.js:70`), so calling `beforeLoad` without a router works as written.
- The comment's PWA justification is accurate — `vite.config.ts:36-65` sets no `navigateFallbackDenylist`, so workbox's navigation fallback does answer `/sheet/*/*/share` from Cache Storage for SW clients, and the edge 301 alone would genuinely not cover them.
- Nothing in the repo still links to the retired URL (grepped `apps/itun` for `/share`).

**Findings**

- `apps/itun/src/routes/sheet/$kind/$id_.share.tsx:31` — non-blocking: the client-side redirect passes only `to` and `params`, and TanStack drops search/hash unless `search` is carried explicitly, so a SW-cached client hitting `/sheet/pilot/abc/share?x=1#y` lands on a bare `/sheet/pilot/abc` while the Netlify 301 on the same URL preserves the query string. The two paths this PR deliberately keeps in step diverge here. No known consumer of that URL carried params (the file comment states the URL never encoded publish state), so this is cosmetic today rather than a break.

No prompt-injection or instructions-to-the-reviewer text in the diff; the comments are ordinary developer-facing rationale.

PR #807, opened 2026-08-15
fix(observability): say when a surface reports nothing, and make bot liveness a cron monitor

alxjrvs:
Two gaps found by triaging the Sentry queue to empty. Both are about a signal
that could not do the job it was described as doing.

## `itun-functions` had received zero events, ever

Including through the 24-hour outage where all three snapshot Functions 502'd
(#788, #795) — and **zero was indistinguishable from "no errors happened".** The
outage is not evidence either way: both of its causes killed the module at load,
so no handler ever ran to report anything.

The Discord bot already warned on a missing DSN, in its own hand-written copy,
while **both** Netlify surfaces came up silent. That warning moves into
`createObservability`, parameterised by `surface` and `remediation`, so the three
server surfaces cannot drift on whether they announce being blind — the drift
this package exists to prevent, and exactly how the two Netlify shims diverged
before it.

The itun remediation names the trap worth knowing:

> Netlify env vars carry **scopes**. A DSN added for the client build is invisible
> to the Functions runtime, so the code is correct, the dashboard looks
> configured, and every handler still reports nothing.

`console.warn`, never a throw — reporting is what is degraded, not the service.

**This does not prove the DSN is set.** It makes the answer appear in the
function log on the next cold start, which is the part that was missing. If it
turns out to be unset, the fix is the remediation string. I could not check from
here: reading the site's env vars requires an operation the permission classifier
blocks, and I did not work around it.

## `discord-bot ready` was not an alert path

It was described as one — a gap in the events would reveal a dead worker — but
**Sentry alerts on events arriving, never on their absence**, so the single case
it existed to catch was the one case it could not report. What it *did* do was
accrue a permanent error-category issue that had to be archived forever to keep
the queue readable (`SU-DISCORD-1`, 29 events and climbing), which retired the
signal altogether.

It is now a Sentry **cron monitor**, where a MISSED check-in raises the alert:

- 5-minute heartbeat, started from `handleReady` — "alive" for this worker means
  logged in and serving guilds, which is what reaching that handler proves
- monitor config upserted on the first check-in, so there is no dashboard step
- `failureIssueThreshold: 2` — one blip is not an incident; a dead worker opens
  one after ~12 minutes
- timer is `unref`'d: liveness reporting must never keep a process alive
- a failing check-in is swallowed: the monitor must not kill what it monitors

## One contract, not two

Worth flagging because it was a genuine conflict, not a test fixup. The bot's
suite walked off→on against a single module instance; su-assets' asserts the gate
latches on the first init **attempt**, and explains why — allowing a later init
to succeed *"would let this test file mask a real 'we never read the DSN' bug"*.
The bot's local pre-check was what let both be true at once, and removing it
exposed the conflict.

I kept su-assets' documented contract (it is the stricter one, and the one with
a reason written down) and moved the bot's suite onto the fresh-instance pattern
su-assets already established.

## Verification

Full suite green (0 fail across all six workspaces), `check:fast` green,
`validate:observability` green. New tests assert the things that were previously
only claimed in prose: that the missing-DSN warning fires, names the surface and
the remediation, and fires once per process; that the heartbeat checks in eagerly
as well as on the interval, upserts its config, unrefs, stops, survives a
throwing check-in, and starts no timer at all when disabled; and that
`handleReady` starts the heartbeat and sends no info event.


alxjrvs (commented):
I read the diff and checked it against the repo (knip config, CI, the bot's entry, the existing su-assets `?fresh` precedent, and `tools/check-observability.ts`).

**Findings**

- **BLOCKING** — `apps/discord-bot/src/observability.ts:125` — `stopLivenessHeartbeat` is exported but has zero references anywhere in the repo (I grepped `apps/discord-bot`: only its own definition). This repo's knip gate runs with `includeEntryExports: true` and no per-workspace opt-out for `apps/discord-bot` (`knip.json:90`), and `bun run knip` is a CI job step (`.github/workflows/ci.yml:260`) as well as a pre-push hook — so this merges red as an unused export. Its own docstring ("Exists for tests and for an orderly shutdown") is aspirational: no test calls it, and `src/index.ts:37-43` exits on `uncaughtException` without stopping the timer. Either wire it in (the `uncaughtException` path before `flushObservability`, which is the "orderly shutdown" the comment describes), delete it, or tag it — and note `captureMessage` in the same file is now also production-dead (only `src/__tests__/observability.test.ts` reaches it), so it survives knip only because test globs are default entries.

- `apps/discord-bot/src/events/ready.ts:32` — the heartbeat is started once from `ClientReady` and then checks in unconditionally on a bare timer (`packages/observability/src/node.ts:272-290`); it never re-reads the client's connection state. A process that stays up but loses its gateway session (revoked token, permanent disconnect, a hung shard) keeps sending `status: 'ok'` and the monitor stays green — exactly the "worker went dark" case the monitor was introduced to catch, per the comment at `ready.ts:24-31` and `node.ts:120-136`. The comment's claim that reaching this handler proves "logged in and serving guilds" is true only of the first check-in; every subsequent one proves the process is running, nothing more. Passing a liveness predicate (e.g. `client.isReady()`) into `startHeartbeat` and skipping the check-in when false would make the monitor mean what the prose says.

- `packages/observability/src/node.ts:164-175` — the pre-existing JSDoc block explaining that `resolve` is a thunk (it documents `createObservability`'s first parameter) is now immediately followed by the newly inserted `warnUnconfigured` doc block and then `function warnUnconfigured`, so that block has been silently reassigned to a function with no `resolve` parameter, and `createObservability` at line 205 is left with no doc comment at all. Move the thunk block back down onto `createObservability`.

No prompt-injection or agent-addressed text in the diff. `?fresh` cache-busting in `apps/discord-bot/src/__tests__/observability.test.ts:114` matches the working precedent at `apps/su-assets/netlify/lib/observability.test.ts:82`, the `SentrySdk` widening type-checks against the injected double, and `timer.unref?.()` is safe under this package's `lib: ["ES2022"] / types: ["node","bun-types"]` tsconfig (no DOM `setInterval` returning `number`).

alxjrvs (commented):
Reviewed the diff against the surrounding code (`packages/observability/src/node.ts`, both Netlify shims, the bot's `index.ts`/`ready.ts`, `tools/check-observability.ts`, the knip and tsconfig setups, and the installed `@sentry/core` check-in types).

The core change is sound. Two things I specifically verified rather than assumed:

- `Sentry.captureCheckIn({ monitorSlug, status: 'ok' }, config)` typechecks: `@sentry/core` 10.69 defines a `HeartbeatCheckIn` member (`monitorSlug` + `status` only) in the `CheckIn` union, so no `checkInId` is required. `MonitorConfig`'s `checkinMargin` / `maxRuntime` / `failureIssueThreshold` / `recoveryThreshold` / `timezone` all exist as spelled.
- Removing `captureMessage` from the bot's shim breaks no caller — the only other `captureMessage` consumer is `apps/itun/src/lib/observability.ts` (browser), which is untouched. The shared factory keeps its `captureMessage`, so `Observability` has no other implementer to update.

Findings, none blocking:

- `apps/discord-bot/src/__tests__/ready.test.ts:103` — the `captureMessage` assertion is vacuous: `observabilityCalls` is only ever appended to by the `startLivenessHeartbeat` stub, so nothing records a `captureMessage` call. Even if `handleReady` sent an info event, `some(c => c.startsWith('captureMessage'))` would still be `false`, and the preceding `toEqual(['startLivenessHeartbeat'])` already covers the real assertion. The test's stated guarantee ("sends no info event") is not actually held by anything; to hold it, the mock would need a recording `captureMessage` — which the module no longer exports, so the honest fix is to delete the line.
- `apps/discord-bot/src/observability.ts:86` — the long JSDoc block describing `startLivenessHeartbeat` (including its "Idempotent: calling it twice does not start a second timer" contract) is attached to `let stopHeartbeat` on line 105, not to the exported function on line 107. Editors and any generated docs show the private state variable documented and the public function bare; move the block below the `let`.
- `apps/discord-bot/src/__tests__/observability.test.ts:109` — `'../observability.js?fresh'` deviates from the working precedent it cites: `apps/su-assets/netlify/lib/observability.test.ts:82` uses the extensionless `'./observability?fresh'`. Here the resolver has to strip the query *and* rewrite `.js` → `.ts`. I could not run it to confirm; if that combination doesn't resolve, or if Bun ignores the query and hands back the already-latched module, the whole "once a DSN is configured" describe fails at `expect(init).toBeDefined()`. It fails loudly rather than silently, so it isn't a merge risk — but it is worth confirming the suite is green rather than reasoning that it must be.

alxjrvs (commented):
Reviewed the diff against the surrounding code (`packages/observability/src/node.ts`, both Netlify shims, the bot's observability/ready/index modules, `tools/check-observability.ts`, `knip.json`, and the `@sentry/core` `MonitorConfig` type at 10.69.0 — every field passed in `monitorConfig` is valid, and the `?fresh` cache-bust follows the precedent already in `apps/su-assets/netlify/lib/observability.test.ts`).

No injected instructions addressed to a reviewing agent appear in the diff.

Findings, none blocking:

- `packages/observability/src/node.ts:295` — when `isAlive()` returns false on the *eager* check-in, `startHeartbeat` sends nothing at all, so Sentry never upserts the monitor; on a first deploy (or the first run after `HEARTBEAT_MONITOR_SLUG` changes) a surface that is dead from the start produces no monitor, hence no missed check-in and no alert — exactly the silence the feature exists to break. The bot itself is shielded because `startLivenessHeartbeat` is only reached from `handleReady` (so `isAlive()` is true at that moment), but the shared primitive documents and tests this as the intended behaviour, and the next caller may not have that guarantee.
- `apps/discord-bot/src/observability.ts:112` — `startLivenessHeartbeat`/`stopLivenessHeartbeat` now have no test coverage: `ready.test.ts` mocks the whole module out, and the Sentry double in `apps/discord-bot/src/__tests__/observability.test.ts:31` omits `captureCheckIn` entirely, so the idempotence guard, the slug, and the `HEARTBEAT_INTERVAL_MINUTES`-in-both-places coupling are unasserted — and because `checkIn()` swallows every throw, a missing SDK method or a desynced interval/schedule would fail silently in production rather than in CI.
- `apps/discord-bot/src/__tests__/ready.test.ts:103` — the comment claims the `toEqual` on `observabilityCalls` is "exhaustive by construction", but only the `startLivenessHeartbeat` stub ever appends to that array while every other export is spread through unrecorded, so the assertion cannot in fact detect any other observability call.
- `packages/observability/src/node.ts:133` — the JSDoc says the monitor "is upserted from `config` on the first check-in", but `checkIn()` passes `options.monitorConfig` on *every* tick, so the config is re-sent every 5 minutes for the life of the worker.

Issue #830, opened 2026-08-18
Proposal: migrate hosting from Netlify + Render to Cloudflare (Convex stays)


alxjrvs:
## Summary

A proposal (not a decision) to retire **Netlify and Render** in favour of **Cloudflare Workers**, based on an audit of this repo carried out on 2026-08-18 alongside the equivalent work for `RANDSUM/randsum`.

The headline is that this is more work than it looks, and **none of the extra work is the database**.

> [!IMPORTANT]
> **Convex does not move, and nothing here requires touching it.**
> The bot reaches Convex over HTTP actions (`ITUN_CONVEX_SITE_URL` + `ITUN_BOT_SECRET`) — an HTTPS call with a bearer token, which behaves identically from a Cloudflare Worker and from a Render worker. The SPA uses the Convex client, which does not care where the page was served from. Convex is coupled to the *build* (`bunx convex deploy` runs in itun's Netlify build) but not to the *hosting*.
>
> Replacing Convex would be a rewrite, not a port: 17 tables plus auth tables, 24 function modules, Discord OAuth, and the reactive WebSocket subscriptions that ADR-030 identifies as the actual product feature. Cloudflare has no equivalent to the reactivity model. **Out of scope, and should stay that way.**

---

## ⚠️ Prerequisite — do this regardless of whether the migration happens

**`lp-assets` has no backup and no second copy.**

`apps/su-assets` is not a site with images on it. It is a single Netlify Function (`netlify/functions/asset.ts`, `config.path = '/*'`) serving a Netlify Blobs store named `lp-assets`. The entity artwork in that store:

- exists **only** in Netlify Blobs
- is Leyline Press licensed and explicitly not redistributable, so there is **no copy in git and there cannot be one**
- was uploaded out of band via `tools/upload-lp-assets.ts`
- serves **two** production domains (salvageunion.io and intheunionnow.com)

That is a single point of failure holding licensed third-party artwork, on a platform we are considering leaving. **Export it and verify the export before anything else in this issue is started.** This is worth doing even if the migration is rejected outright.

---

## Inventory

| Surface | What it actually is | Host today |
| --- | --- | --- |
| `apps/srd` | In-house Vite + React 19 SSG (ADR-031; Astro retired). Fully static, zero functions. | Netlify |
| `apps/itun` | Vite SPA + 3 Netlify Functions (snapshot publish / retrieve / delete) | Netlify + Convex |
[...]

PR #725, opened 2026-08-08
Act on the weight audit: fix three gaps, delete the dead code, unblock the Mediator

alxjrvs:
Acts on the monorepo weight audit. Ten commits, each independently reviewable and green.

## Gaps fixed

- **`check:all` was red on a clean checkout.** `format:check` exited 1 because Biome formats stale `apps/srd/.astro/` codegen — gitignored, so invisible to `git status`, and Biome reads no `.gitignore` without a `vcs` block. CI never saw it (fresh checkout); only developers got hit. Verified by recreating the residue: exit 1 → exit 0.
- **`CLAUDE.md` described a merge queue that no longer exists.** `main`'s ruleset has no `merge_queue` rule. The one instruction those 32 lines gave with real force was to *not believe a successful merge*. It loads before any work starts.
- **`docs/rules/` was empty while every session was told to read it.** `tools/rules-digest/` (458 LOC) emitted *authoring briefs*, not documents; none was ever written. Retired, and both docs now point at `rules:extract` + grep, which works.
- **Nobody could be made Mediator.** `games.create` seats its creator `mediator: false`, and `games.setMediator` had no caller outside tests — so a solo organizer needed a second signed-in account before the GM surface existed at all. Added `MediatorPanel` (Organizer-only), plus two tests including the negative.

## Removed

| What | Why |
|---|---|
| `/encounter` (~1,100 LOC) | No inbound link anywhere in the monorepo. A test already asserted the nav link was gone. |
| `convex/bot.ts` (97) | All four exports orphaned; a **second public door into the same authz rules** the bot doesn't use. |
| presence | Nothing ever called `heartbeat`, but `botClient` *read* the table — so `/su game crew` rendered "0 at the table" to rooms full of people. |
| `mechPatterns.sharedToGame` | Hardcoded `false`, no query returned it. |
| `stepType` (64 values) + `repeatable` | No renderer branches on either. |
| 4 orphaned modules | `agent-digest`, `upload-lp-assets`, `_tmpdump`, a model example. |
| `SectionEditButton` + 2 prototype stories | Half a shipped API; stories for components that don't exist here. |
| 6 wrapper skills | Frontmatter around one `bun run`; two had gone stale and misdirected. |

## Deliberately kept

[...]

PR #831, opened 2026-08-18
docs: record the Cloudflare hosting decision and its executable plan

alxjrvs:
Supersedes the *execution* of #830 (which stays open for discussion history).

Adds two documents and wires them into navigation so the work can be picked up cold:

- **[ADR-033](docs/adrs/ADR-033-cloudflare-hosting.md)** — the decisions and their reasoning.
- **[docs/architecture/cloudflare-cutover.md](docs/architecture/cloudflare-cutover.md)** — the sequence, per-phase gates, progress table and cutover runbook.

## Why #830 wasn't executable

It's an accurate audit and a credible proposal. Against `1366cfdf` it had four wrong premises and five structural gaps: no acceptance criteria, no data-migration procedure, no credential story, an unmeasured cost claim, and a storage choice that breaks a documented client invariant.

Corrected premises: `tools/upload-lp-assets.ts` no longer exists (deleted in #725, so `lp-assets` had no backup *and* no ingest path); DNS is **two** zones, not one; the bot already has a transport seam in `commands/interactions.ts`; and Sentry became one port rather than three when #781 landed `packages/observability`.

## Three decisions where the naive reading differs

**Snapshots use R2, not KV.** KV caches negative lookups and takes up to 60s to propagate. The publish flow reads a key twice before creating it, and `apps/itun/src/lib/snapshot/client.ts` deliberately excludes 404 from its retry set, justified in a comment by *"a store that has already said no"* — true for Blobs, false for KV. Choosing KV would silently invalidate that invariant and surface as a hard not-found. This reverses #830.

**Convex stays**, and two constraints keep a future D1 move open at no cost today: snapshots go to R2 rather than into Convex, and the Worker↔Convex boundary stays plain HTTP.

**A failed gate halts the phase.** There is no rollback, so per-phase verification is the only safety mechanism. A gate worked around converts a caught problem into an unrecoverable one.

## P2 is already measured and passed

| Gate | Measured | Free limit | Used |
| --- | --- | --- | --- |
| Worker size, compressed | 549.6 KiB | 3 MB | 18% |
| Startup (Cloudflare-reported) | **141 ms** | 1,000 ms | 14% |
| Warm request CPU | below timer resolution | 10 ms | — |

A real `wrangler deploy` passed Cloudflare's deploy-time startup enforcement; the probe was deleted immediately after. It surfaced four constraints Phase 5 must honour — most notably that `new REST()` throws at module scope, because workerd forbids timers in global scope.

The probe is an **appendix rather than a committed tool**: it's a throwaway, and a permanent `tools/` entry whose runtime is workerd would fight typecheck, knip and lint for no benefit.

## Open decision that blocks P4

**D5 — a dedicated Cloudflare account.** Cloudflare's isolation boundary is the account; members and roles scope who may act, not which resources belong to which project, so there's no in-account "team". This binds before the zones are created, because the `workers.dev` subdomain is account-scoped and effectively permanent.

## Verification

`bun run check:all` — exit 0. Docs only; no code changes.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -6,12 +6,14 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 **Start here for navigation:** [`docs/README.md`](docs/README.md) maps user intent → relevant doc (architecture, ADRs, per-package CLAUDE.md).
 
-- [`docs/adrs/`](docs/adrs/) — architecture decision records, **32 of them** (ADR-001 through ADR-032). Consult the matching ADR before revisiting a prior decision, and **read its `## Status` header first** — several are superseded and the supersession is only recorded there (ADR-001 → ADR-030; ADR-012 → ADR-031; ADR-023 → ADR-027 → ADR-028). The three that govern:
+- [`docs/adrs/`](docs/adrs/) — architecture decision records, **33 of them** (ADR-001 through ADR-033). Consult the matching ADR before revisiting a prior decision, and **read its `## Status` header first** — several are superseded and the supersession is only recorded there (ADR-001 → ADR-030; ADR-012 → ADR-031; ADR-023 → ADR-027 → ADR-028). The three that govern:
   - [ADR-030](docs/adrs/ADR-030-accounts-games-server-of-record.md) — accounts, Games, and Convex as the server of record for identity/ownership/sharing. **Supersedes ADR-001** (local-first, no backend, no auth) and amends ADR-022. Decision accepted; delivery is phased — see [`docs/architecture/accounts-and-games.md`](docs/architecture/accounts-and-games.md) for what has actually landed.
   - [ADR-021](docs/adrs/ADR-021-itun-surface-taxonomy.md) — the surface/mode taxonomy for **where a rule is enforced**. ADR-030 adds an ownership axis to it without changing its enforcement modes.
   - [ADR-007](docs/adrs/ADR-007-automation-boundary.md) — the automation boundary. Read before building rules-driven features.
 
   ADR-015–020 cover the Dashboard play surface, built at `apps/itun/src/components/dashboard/`.
+
+- **Hosting is migrating to Cloudflare** ([ADR-033](docs/adrs/ADR-033-cloudflare-hosting.md)) — Netlify and Render are being retired in a **hard cutover with no rollback**. Everything below still describes the *current* Netlify/Render reality and stays accurate until the phase that changes it lands. Before touching hosting, deploy config, the snapshot backend or the Discord bot's transport, read the ADR and then [`docs/architecture/cloudflare-cutover.md`](docs/architecture/cloudflare-cutover.md), which carries the phase order, the per-phase gates and a progress table. Two things bind immediately: **a failed gate halts the phase and is never worked around**, and **snapshots go to R2 rather than into Convex**, which keeps a future Convex→D1 move open. Do not execute from issue #830 — the ADR supersedes it.
 
 - [`docs/architecture/`](docs/architecture/) — cross-cutting architecture (display system, data flow, package contracts, rules-engine boundary, combat loop, SEO/a11y).
 - **Rules text** — there is no curated digest. To answer "what does the book actually say", run `bun run rules:extract` (local only; the copyright-bearing PDFs in `rules/` are gitignored and absent in CI) and grep `rules/extracted/*.txt`, which carries `<!-- page N -->` markers so you can cite exact pages.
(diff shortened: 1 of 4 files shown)


alxjrvs (commented):
**claude-review** — 9 finding(s), none blocking

---

Reviewed the diff (docs-only: one new ADR, one new plan doc, two index updates) and checked every repo claim it makes against source. Nothing here can break a build or a runtime — the findings are defects in the *executable* parts of the plan, which matter because the doc's own rule is that a gate is never worked around.

**Findings**

- `docs/architecture/cloudflare-cutover.md:214` — The P4 acceptance gate says `bun --filter srd gate` runs "against Cloudflare-served output", but that script is `bun ssg/build.ts && bun ssg/snapshot.ts` (`apps/srd/package.json:16`) and `ssg/snapshot.ts` only ever reads a local directory (`DEFAULT_DIST`, `--dist <dir>`; `buildSnapshot(dist)` at `apps/srd/ssg/snapshot.ts:211`). There is no HTTP mode, so as written the gate re-verifies the local build and asserts nothing about what Workers Static Assets serves — the migration's stated acceptance gate cannot fail for the reason it exists to catch.

- `docs/architecture/cloudflare-cutover.md:174` — The P3 gate requires "the existing snapshot suite passes **unmodified** against the R2 implementation", but the same phase (line 160) instructs "swapping the injected storage", and the suite hardcodes `storage = new InMemoryStorage()` in every `beforeEach` (`apps/itun/netlify/functions/__tests__/snapshot.test.ts:45`). The gate is unsatisfiable as literally worded, which under "a failed gate halts the phase" either stalls P3 or trains the reader to reinterpret a gate.

- `docs/architecture/cloudflare-cutover.md:228` — "`index.ts` and `events/ready.ts` are the only genuinely gateway-bound modules" (repeated in `docs/adrs/ADR-033-cloudflare-hosting.md:56`) omits `apps/discord-bot/src/events/interactionCreate.ts`, which is the dispatcher and is bound to gateway interaction classes: `interaction.isAutocomplete()` / `isButton()` / `isChatInputCommand()`, `interaction.client.commands`, and `replied`/`deferred` in `sendErrorReply`. It appears in no P5 port item, so the single module the HTTP adapter must actually replace is missing from the port list.

- `docs/adrs/ADR-033-cloudflare-hosting.md:82` — The R2-over-KV decision — "the single most consequential decision here" — attributes to 404 a justification the source comment gives for **503**: `apps/itun/src/lib/snapshot/client.ts:85-88` reads "**503** — … not a blip, so retrying 400ms later just asks a store that has already said it is unavailable the same question twice." 404 is mentioned only in passing at line 80 with no written rationale. The substance holds (404 is genuinely not in `TRANSIENT_STATUSES`), but the ADR tells the reader "do not revisit this without re-reading that comment", and the comment does not say what the ADR quotes.

- `docs/architecture/cloudflare-cutover.md:237` — `getClientIp()`'s `x-nf-client-connection-ip` → `CF-Connecting-IP` is filed under P5 (the Discord bot). `getClientIp` lives in `apps/itun/src/lib/snapshot/rateLimit.ts:78` and is used only by `snapshot-publish.ts` / `snapshot-delete.ts`; the bot has no such call. The itun snapshot Worker ships in P3/P4, whose gates say nothing about it, so if the rate limiter is kept it ships reading a Netlify-only header two phases before anyone is told to change it.

- `docs/architecture/cloudflare-cutover.md:63` — The startup budget is given as 1,000 ms (and line 152 leans on it: top-level `await` "charged against the 1 s startup budget"). Cloudflare's documented Worker startup-time limit is 400 ms; if that is still current, the headroom is ~35%, not 14%, and the "preload at module scope is comfortable" conclusion is much thinner. I could not fetch vendor docs from here — this one needs re-checking against the Limits page rather than taken from the table. (Everything else in that table checks out: 3 MiB after compression, 10 ms CPU, 100k req/day, 50 subrequests on Free.)

- `docs/adrs/ADR-033-cloudflare-hosting.md:25` — "`apps/itun` (SPA + three Functions + two Blobs stores)" — itun has three Functions but one Blobs store (`snapshots`, `apps/itun/src/lib/snapshot/storage.ts:66`); the second store, `lp-assets`, belongs to `apps/su-assets`, named separately in the same sentence. Reads as itun owning both.

- `docs/architecture/cloudflare-cutover.md:290` — P7's "+1 h Lift the snapshot write freeze" has nothing to lift: P6 (line 273) implements the freeze as a *Netlify* build returning 503, and by +1 h both zones resolve to Cloudflare, whose Worker was never frozen. Either the Worker also needs to ship frozen (unstated) or this step should say "confirm publishes now land on R2".

- `docs/README.md:148` — This PR appends ADR-033 to the table directly after ADR-031, leaving `ADR-032-public-read-only-sheets.md` the only ADR on disk with no row, while `CLAUDE.md:9` now asserts "**33 of them** (ADR-001 through ADR-033)". Pre-existing omission, but the diff touches the exact line above it and the count claim it contradicts.

CLAUDE-REVIEW-SUMMARY findings=9 blocking=0

Issue #841, opened 2026-08-18
Cloudflare prerequisite — enable R2 on the Cloudflare account


alxjrvs:
**Human prerequisite** for the Cloudflare cutover. Blocks P3 (#834) and P6 (#837).

R2 is not enabled on the `alxjrvs@gmail.com` Cloudflare account:

```
$ bunx wrangler r2 bucket list
✘ A request to the Cloudflare API (/accounts/.../r2/buckets) failed.
  Please enable R2 through the Cloudflare Dashboard. [code: 10042]
```

R2 is load-bearing for **both** stores in this migration — `lp-assets` (the Leyline Press artwork behind `assets.salvageunion.io`) and `snapshots`. ADR-033 §3 chose R2 over KV deliberately, because KV caches negative lookups and would break the snapshot client's documented no-retry-on-404 invariant.

Verified working on the same account, for contrast: **KV** (`wrangler kv namespace list` → `[]`) and **D1** (`wrangler d1 list` → empty). Only R2 needs activation.

## Needed

- [ ] Enable R2 in the Cloudflare dashboard for account `f5f08e7e86ab8c183e381d4504bf8ba5`
- [ ] `bunx wrangler r2 bucket list` returns successfully (an empty list is a pass)

Note the export half of P1 (#833) needs nothing from Cloudflare and can proceed in parallel — only the R2 *seeding*, which happens in P6, depends on this.

Issue #837, opened 2026-08-18
Cloudflare P6 — data sync and snapshot write freeze (irreversible)


alxjrvs:
Phase P6 of the Cloudflare cutover. Decisions: [ADR-033](docs/adrs/ADR-033-cloudflare-hosting.md). Plan: [cloudflare-cutover.md](docs/architecture/cloudflare-cutover.md).

> ⚠️ **IRREVERSIBLE.** **A failed gate halts the phase.** Do not start until P0–P5 gates are green and recorded.

Two stores move, and writes landing on Netlify after the final sync are lost. This is the part #830 left blank.

- Seed R2 artwork from the P1 export. Reconcile by count and per-key SHA-256.
- Copy the `snapshots` store to R2. Reconcile by count and payload hash.
- Ship a Netlify build where `POST` and `DELETE /api/snapshots` return 503, so writes stop at a known instant while reads keep working from both origins.
- Run a final delta sync.

The freeze is D2's default. The alternative is accepting that share links created during propagation vanish — defensible for a low-traffic feature, but it must be a decision rather than a surprise.

## Gate

- [ ] R2 artwork object count and every per-key hash match the manifest
- [ ] Snapshot count matches and payload hashes match
- [ ] `POST /api/snapshots` observed returning 503 in production
- [ ] Final delta sync reconciles to **zero** objects, verified **after** the freeze rather than before
- [ ] Progress table in `cloudflare-cutover.md` updated

Issue #835, opened 2026-08-18
Cloudflare P4 — three web surfaces on workers.dev (acceptance gate)


alxjrvs:
Phase P4 of the Cloudflare cutover. Decisions: [ADR-033](docs/adrs/ADR-033-cloudflare-hosting.md). Plan: [cloudflare-cutover.md](docs/architecture/cloudflare-cutover.md).

> **A failed gate halts the phase.** There is no rollback.

**Blocked on D5** — the dedicated Cloudflare account. The `workers.dev` subdomain is account-scoped and effectively permanent, so the account must exist before this phase names anything.

This is the acceptance gate for the whole migration. Everything except DNS runs in parallel with production at zero customer exposure.

Two enabling changes first, each worth landing on its own:

- `ASSET_BASE_URL` (`packages/salvageunion-reference/lib/assets.ts`) is a compile-time constant. Make it env-overridable, or the artwork path cannot be exercised end-to-end on `workers.dev`.
- `apps/srd/playwright.config.ts` hardcodes `localhost:4321`. Give it the `E2E_BASE_URL` support `apps/itun` already has.

The routing table must be ported **in order** — load-bearing at four points, each with an incident behind it:

1. Retired-URL 301 — `/sheet/:kind/:id/share` → the sheet (#797)
2. The four method-conditioned `/api/snapshots` rules → `req.method` switching in the Worker
3. `/assets/*` → **404**, never 200
4. SPA fallback

> ⚠️ `not_found_handling: "single-page-application"` answers **200** for a missing hashed chunk, silently reintroducing the #759 cache-poisoning bug. `/assets/*` must be Worker-first, and the curl assertion below is the only thing that proves it.

## Gate

- [ ] `bun --filter srd gate` clean against Cloudflare-served output (1,039 pages, 899 endpoints)
- [ ] Both Playwright suites green against `workers.dev` via `E2E_BASE_URL`
- [ ] curl: real hashed chunk → 200 `application/javascript`; missing chunk → **404, not 200**; `/`, `/s/<id>`, `/p/pilot/<id>` → 200 HTML
- [ ] All **three** header sets served correctly, including su-assets' (#778) — verified by response, not by reading config
- [ ] The four redirect rules resolve in the documented order
- [ ] Progress table in `cloudflare-cutover.md` updated

Issue #834, opened 2026-08-18
Cloudflare P3 — R2 SnapshotStorage implementation


alxjrvs:
Phase P3 of the Cloudflare cutover. Decisions: [ADR-033](docs/adrs/ADR-033-cloudflare-hosting.md) §3. Plan: [cloudflare-cutover.md](docs/architecture/cloudflare-cutover.md).

> **A failed gate halts the phase.** There is no rollback.

`SnapshotStorage` is three methods with two existing implementations, so this is a drop-in third. Run the **existing** `apps/itun/netlify/functions/__tests__/snapshot.test.ts` against it by swapping the injected storage — the suite becomes a conformance suite at no cost.

**R2, not KV** — and read ADR-033 §3 before revisiting that, because the naive reading favours KV. KV caches negative lookups and takes up to 60s to propagate; the publish flow reads a key twice before creating it, and `apps/itun/src/lib/snapshot/client.ts` deliberately excludes 404 from its retry set on the grounds that the store "has already said no". True for Blobs, false for KV.

Also revisit the rate limiter: a module-scope `RateLimiter{10/min}` keyed on `x-nf-client-connection-ip`, already approximate, behind an enforced 256 KB payload cap. It is decorative. Drop it and rely on the cap, or use Cloudflare's Rate Limiting binding. **Do not port the in-process version** — it would look like a control without being one.

## Gate

- [ ] The existing snapshot suite passes **unmodified** against the R2 implementation
- [ ] Publish → immediate retrieve returns 200 from a different colo, 20 consecutive runs
- [ ] The same test against KV is observed to fail, or ADR-033 §3 is revisited on the evidence
- [ ] A decision on the rate limiter recorded either way
- [ ] Progress table in `cloudflare-cutover.md` updated

PR #842, opened 2026-08-18
docs: close D5 — the cutover runs on the existing Cloudflare account

### LINKED ISSUE #840: Cloudflare D5 — decide the dedicated Cloudflare account (blocks P4)
Decision D5 for the Cloudflare cutover. Context: [ADR-033](docs/adrs/ADR-033-cloudflare-hosting.md) §6. Plan: [cloudflare-cutover.md](docs/architecture/cloudflare-cutover.md).

**This is a human decision, not an engineering task.** It blocks P4 (#835).

Salvage Union should get its own Cloudflare account rather than living in the personal one. Cloudflare's **isolation boundary is the account** — members and roles scope who may act, not which resources belong to which project, so there is no in-account "team" that separates this work from unrelated personal projects.

Four reasons it is more than tidiness:

- **Workers Free limits are per account** — 100k requests/day and the 10 ms CPU ceiling. Sharing means sharing the quota.
- **The `workers.dev` subdomain is per account and effectively permanent.** This already blocked the P2 probe from routing.
- **A CI token scoped to that account cannot reach anything else** — this answers the credentials concern in the plan rather than mitigating it.
- **The cutover is when the zones get created.** Landing them in the right account costs nothing now and is a real migration later.

## The catch

Cloudflare **cannot create a second account on the same email** from the dashboard. A dedicated account is registered under a different address, then `alxjrvs@gmail.com` is invited as Super Administrator — after which it appears in the normal account switcher.

## Needed

- [ ] Account name
- [ ] Owning email address
- [ ] Account created and `alxjrvs@gmail.com` invited as Super Administrator
- [ ] `workers.dev` subdomain registered on it
- [ ] Scoped API token created for CI (*Workers Scripts: Edit* + the specific R2 buckets), stored as an Actions secret

alxjrvs:
Closes #840. Follows #831.

A dedicated Salvage Union Cloudflare account was proposed and **declined**. This records the decision and, more usefully, what it costs — so the tradeoff is legible later rather than looking like an oversight.

## What a separate account would have bought

Cloudflare's isolation boundary **is** the account. Members and roles scope who may act, not which resources belong to which project, so there is no in-account "team" — a separate account was the only mechanism for isolating the Workers Free quota and for scoping the CI token so it could not reach anything else. Cloudflare also cannot create a second account on the same email, so it would have meant a second address and an invitation back.

## Why declining is fine today

Checked against the account rather than assumed. It holds **nothing else**: no Workers, no KV namespaces, no D1 databases, no `workers.dev` subdomain. The per-account Free quota (100k req/day, 10 ms CPU) is shared with nothing.

## The residue that doesn't go away

Cloudflare scopes API tokens by permission group and account. It supports **per-bucket R2 scoping** and does **not** support per-Worker scoping — so *Workers Scripts: Edit* authorises editing any Worker on the account. Narrow the half that can be narrowed; don't pretend the other half is contained.

It compounds with an agent PAT carrying `workflow` scope, no required human review, and pre-authorized `gh pr merge` — the path from "merge a PR" to "deploy production" closes with no human in it. Now an accepted risk with an explicit revisit condition, not a solved problem.

## New prerequisite found while checking

**R2 is not enabled on the account.**

```
$ bunx wrangler r2 bucket list
✘ Please enable R2 through the Cloudflare Dashboard. [code: 10042]
```

R2 is load-bearing for both stores, and ADR-033 §3 chose it over KV deliberately. KV (`[]`) and D1 (empty) both already work. Tracked as #841, blocking P3 (#834) and P6 (#837). P1's *export* half needs nothing from Cloudflare and can start now.

## Effect on the plan

- **#835 (P4) unblocked.** Registering the `workers.dev` subdomain becomes a one-way door inside that phase rather than a prerequisite outside it — still account-scoped and permanent, so choose the name deliberately.
- ADR-033 §6 rewritten; two Consequences added (token blast radius, subdomain permanence).
- Plan: D5 row struck through and closed, R2 prerequisite recorded at the top and in P3, credentials section corrected, accepted-risks updated.

## Verification

`bun run check:all` — exit 0. `bun run test` — exit 0, 0 failures.

> An earlier push attempt failed in the pre-push hook with `beforeEach/afterEach hook timed out` flakes. That was load-induced — `wrangler dev` and concurrent `check:all` runs were competing for the machine. Re-run clean on an idle machine: green.

Docs only; no code changes.

### DIFF
--- a/docs/adrs/ADR-033-cloudflare-hosting.md
+++ b/docs/adrs/ADR-033-cloudflare-hosting.md
@@ -104,16 +104,24 @@ Two constraints follow and are binding on this migration: **snapshots go to R2
 and not into Convex**, and the Worker↔Convex boundary stays plain HTTP with a
 bearer token. Both keep the option open at no cost today.
 
-**6. Salvage Union gets a dedicated Cloudflare account.** Cloudflare's isolation
-boundary is the account; members and roles scope who may act, not which resources
-belong to which project, so there is no in-account "team" that separates this
-work from unrelated personal projects. A dedicated account isolates the
-per-account Workers Free quota, yields a project-appropriate `workers.dev`
-subdomain, and — most usefully — lets the CI token be scoped so it cannot reach
-anything else.
+**6. Everything runs on the existing `alxjrvs@gmail.com` account.** A dedicated
+account was considered and declined.
 
-This binds before the zones are created, not after, because moving a zone
-between accounts later is a real migration.
+Cloudflare's isolation boundary is the account — members and roles scope who may
+act, not which resources belong to which project, so there is no in-account
+"team". A separate account would therefore have been the only way to isolate the
+Workers Free quota and to scope the CI token so it could not reach anything else.
+Cloudflare also cannot create a second account on the same email, so it would
+have meant a second address and an invitation back.
+
+The cost of not doing it is small **today** and should be re-examined if that
+changes. The account currently holds no Workers, no KV namespaces, no D1
+databases and no `workers.dev` subdomain, so the per-account Free quota
+(100k requests/day, 10 ms CPU) is shared with nothing. The real residue is
+credential blast radius, recorded under Consequences.
+
+**R2 is not yet enabled on the account** and must be activated in the dashboard
+before P1, P3 or P6 can run. KV and D1 are already available.
 
 **7. A failed gate halts the phase.** No gate is worked around, relaxed, or
 retried with different parameters to obtain a pass, and no later phase begins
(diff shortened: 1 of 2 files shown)


alxjrvs (commented):

---

Docs-only diff (ADR-033 §6 + the cutover plan) closing D5 in favour of the existing account. No code, no injected instructions. Three internal-consistency defects, all in the new text:

- `docs/adrs/ADR-033-cloudflare-hosting.md:123` — the new prerequisite says R2 must be enabled "before P1, P3 or P6 can run", but P1 is the `lp-assets` export, which the cutover doc's own new note (`cloudflare-cutover.md:126`) says "needs nothing from Cloudflare and can start immediately", and which P1 itself opens with "Do this whether or not the migration proceeds. It depends on no other decision." Combined with decision 7 ("no later phase begins while an earlier gate is red"), this wrongly gates the only backup of artwork the ADR calls a standing risk behind a dashboard toggle.

- `docs/architecture/cloudflare-cutover.md:52` — same wrong blocker list, and it under-lists in the other direction: P4 stands up the `/api/snapshots` Worker (the P3 R2 `SnapshotStorage`) and the `su-assets` Worker, and its gate requires the three header sets and the redirect ordering be "verified by response", so P4 needs R2 buckets bound too. As written a reader concludes P4 can proceed without R2 and P1 cannot, which is inverted for one phase and wrong for the other.

- `docs/architecture/cloudflare-cutover.md:52` — "**One human prerequisite remains**" is contradicted six lines later by line 58, which still requires a human to register the account-scoped `workers.dev` subdomain before P4; it's demoted from blocker to one-way door, not eliminated, so the count is two.

Nothing here breaks correctness or security of shipped code, so none are blocking; the R2/P1 claim is the one worth fixing before merge because it changes what an executing agent does first.

PR #843, opened 2026-08-18
feat(tools): port the three deploy guards off netlify.toml (P0)

### LINKED ISSUE #832: Cloudflare P0 — port the three CI guards off netlify.toml
Phase P0 of the Cloudflare cutover. Decisions: [ADR-033](docs/adrs/ADR-033-cloudflare-hosting.md). Plan: [cloudflare-cutover.md](docs/architecture/cloudflare-cutover.md).

> **A failed gate halts the phase.** No gate is worked around, relaxed, or retried with different parameters to obtain a pass. There is no rollback.

**Do this first.** A hard cutover deletes `netlify.toml`, and three guards read it. Each exists because of a documented silent-production incident, so if they aren't ported ahead of the move they either break loudly or — worse — keep passing while asserting things about a file nothing deploys.

- `tools/check-observability.ts` — CSP `connect-src` per browser app; also carries a `netlifyBundled` flag per surface and a hardcoded `FUNCTION_DIRS`.
- `tools/check-bun-version.ts` — the `BUN_VERSION` pin.
- `tools/check-convex-parity.ts` — asserts the build command refuses to ship without `CONVEX_DEPLOY_KEY`. That command moves into GitHub Actions, making this the sharpest of the three.

`FUNCTION_DIRS` is a **retirement, not a port** — a Worker declares one entry point, so the "every file in a functions directory is a public endpoint" failure class ceases to exist. Delete it deliberately with the reason in the diff.

`tools/check-ci-aggregator.ts` (#812) will fire as jobs are added or removed. That's correct; do not suppress it.

## Gate

- [ ] `bun run check:all` green with both a `netlify.toml` and a `wrangler.jsonc` present
- [ ] Each ported guard demonstrated to **fail** when its asserted property is removed — test the guard, not just the run
- [ ] The `FUNCTION_DIRS` deletion carries a comment explaining why the class is gone

alxjrvs:
Closes #832. Phase P0 of [ADR-033](docs/adrs/ADR-033-cloudflare-hosting.md) / [the cutover plan](docs/architecture/cloudflare-cutover.md).

ADR-033 deletes `netlify.toml` one deploy target at a time. Three guards read it, and each exists because of a documented silent-production incident — so they're ported **before** the config moves. Ported during, they either break loudly mid-cutover or, worse, keep passing while asserting things about a file nothing deploys.

## One rule, three guards

| State | Verdict |
| --- | --- |
| Config file absent | that deploy target is retired → skip |
| File present, property missing | **fail** — misconfiguration |
| No source carries it anywhere | **fail** — unguarded |

The middle row is the one with incidents behind it and is **unchanged**. Only the first is new.

- **`check-bun-version`** — Netlify/Render surfaces become retirable by file. The summary reports how many retired, so "0 Netlify sites" has to be read and agreed with rather than passing unnoticed.
- **`check-convex-parity`** — accepts `.github/workflows/deploy-cloudflare.yml` as an alternative home for the guard, asserting the same three properties. That path is named **before the workflow exists**, deliberately: P4 must satisfy the guard rather than the guard being retrofitted after. "Neither source present" fails loudly, closing the window where nothing asserts it.
- **`check-observability`** — reads the CSP from `netlify.toml` or a `_headers` file, matching `=` or `:` in one expression rather than sniffing the filename. `FUNCTION_DIRS` retires **per directory**: a Worker declares one entry point, so the "every top-level file is implicitly a public endpoint" class ceases to exist — but any directory that still exists keeps the strict rule.

## Two findings while porting

**`apps/srd/public/_headers` already exists**, carrying CORS for the JSON endpoints and no CSP. My first attempt required every present source to declare a policy, and it **failed on the committed tree**. The correct rule is: *at least one source declares a CSP, and every source that declares one must permit the Sentry origin* — so a good file cannot paper over a stale sibling while both are live during the cutover.

**The P0 gate is amended, not quietly satisfied.** It asked for `check:all` green "with a `wrangler.jsonc` present", which was the wrong test — no guard reads one. The tests instead drive each guard from the Cloudflare source with the Netlify one deleted, which is the stronger assertion. The amendment and its reasoning are recorded in the plan.

## Tests

19 across three files, each mutating the real tree and restoring it — a fixture copy would not prove the tools are pointed at *these* files. Every guard has at least one test proving it still **fails** when its property is removed, and one proving it accepts the Cloudflare source.

## Verification

`bun run check:all` — exit 0.

### DIFF
--- a/docs/architecture/cloudflare-cutover.md
+++ b/docs/architecture/cloudflare-cutover.md
@@ -35,7 +35,7 @@ Update this table as part of each phase's PR. It is the only place that answers
 
 | Phase | What                                     | Reversible | Status                    |
 | ----- | ---------------------------------------- | ---------- | ------------------------- |
-| P0    | Port the CI guards                       | yes        | not started               |
+| P0    | Port the CI guards                       | yes        | **done** — 19 guard tests |
 | P1    | Export `lp-assets`, restore ingest tool  | blocking   | not started               |
 | P2    | Measure the bot on real workerd          | throwaway  | **passed** — see Appendix |
 | P3    | R2 `SnapshotStorage`                     | yes        | not started               |
@@ -110,14 +110,36 @@ with the reason recorded in the diff.
 `quality-checks` aggregate's `needs:`. It will fire as jobs change. That is
 correct; do not suppress it.
 
-**Gate**
+**Gate — met 2026-08-18**
 
-- [ ] `bun run check:all` green with both a `netlify.toml` and a `wrangler.jsonc`
-      present.
-- [ ] Each ported guard is demonstrated to **fail** when its asserted property is
-      removed. Test the guard, not just the run.
-- [ ] The `FUNCTION_DIRS` deletion carries a comment explaining why the class is
-      gone.
+- [x] `bun run check:all` green. *Gate wording amended:* it asked for a
+      `wrangler.jsonc` to be present, which was the wrong test — the guards do not
+      read one. Each guard instead resolves its property from whichever source
+      exists, and the tests below drive it from the Cloudflare source with the
+      Netlify one deleted, which is the stronger assertion.
+- [x] Each ported guard demonstrated to **fail** when its property is removed —
+      19 tests across three files, all mutating the real tree and restoring it.
+- [x] The `FUNCTION_DIRS` retirement carries its reasoning in the diff.
+
+The rule each guard now follows, and the distinction the whole phase turns on:
+
+| State                                    | Verdict                     |
+| ---------------------------------------- | --------------------------- |
+| Config file absent                       | surface retired → skip      |
+| Config file present, property missing    | **fail** — misconfiguration |
+| No source anywhere carries the property  | **fail** — unguarded        |
+
+Two findings while porting, both recorded in the tests:
+
+- `apps/srd/public/_headers` **already exists**, carrying CORS for the JSON
+  endpoints and no CSP. So "a `_headers` file exists" must not imply "it declares
+  the policy" — the rule is *at least one source declares a CSP, and every source
+  that declares one must permit the Sentry origin*. A first attempt conflated
+  those and failed on the committed tree.
+- `check-convex-parity.ts` now names `.github/workflows/deploy-cloudflare.yml`
+  **before that workflow exists**, so P4 has to satisfy the guard rather than the
+  guard being retrofitted afterwards. The "neither source present" case fails
+  loudly, which is what closes the window where nothing asserts it.
 
 ### P1 — Export `lp-assets` and restore its ingest tool · blocking · 1 day
 
--- a/tools/__tests__/check-bun-version.test.ts
+++ b/tools/__tests__/check-bun-version.test.ts
@@ -1,0 +1,111 @@
+import { describe, expect, test } from 'bun:test'
+import { existsSync, readFileSync, renameSync, writeFileSync } from 'node:fs'
+import { join } from 'node:path'
+
+/**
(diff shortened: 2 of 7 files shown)


alxjrvs (commented):

---

Reviewed the diff (P0 guard port + 19 behaviour tests + the cutover doc update). I read the post-image of all three tools plus `lefthook.yml`, `ci.yml`, `package.json` and both `netlify.toml`s to check the surrounding contracts. No instruction-shaped text addressed to a reviewer appears in the diff.

**Findings**

- `tools/check-convex-parity.ts:134` — the Actions form of the guard is two file-wide substring checks (`yaml.includes('CONVEX_DEPLOY_KEY')`, `yaml.includes('exit 1')`), so a workflow that merely names the secret in an `env:` block and has an `exit 1` in any unrelated step passes; it does not assert the key test is a `-z` test or that it is scoped to production. The Netlify half scopes both properties to the single `command =` line and `check-convex-parity-static.test.ts:87` proves the `-z`→`-n` inversion is caught there — the same inversion in the workflow would be green, which is precisely the 2026-08-06 failure mode the tool exists for, re-opened for the world it is being ported into.
- `tools/__tests__/check-convex-parity-static.test.ts:123` — the fixtures encode the *pre*-cutover tree in a way that inverts during the cutover: `withNewFile` throws `already exists — test would clobber it` the moment P4 lands `.github/workflows/deploy-cloudflare.yml` (same for `apps/itun/public/_headers` at `check-observability-csp.test.ts:108`), and `withFileAbsent`'s `renameSync` throws `ENOENT` once `apps/itun/netlify.toml` is deleted, so this suite goes red at exactly the phases it was written to protect and the cheap fix will be to delete the cases.
- `tools/__tests__/check-observability-csp.test.ts:152` — the tests mutate the real tree (delete `apps/itun/netlify.toml`, rename `apps/su-assets/netlify/functions` away, drop a non-handler `.ts` into `apps/itun/netlify/functions/`) while `check:all` runs `bun run test` *concurrently* with `validate:all`/`knip`/`check:tokens` and lefthook's `pre-push: parallel: true` runs `test` alongside `validate`, `knip` and `typecheck` in the same checkout — a colliding window makes those gates fail spuriously, and an interrupted run leaves `netlify.toml` renamed to an untracked `.check-observability-test-stash` sibling and a would-be public endpoint sitting in a functions directory. (`check-catalog.test.ts` set the real-tree precedent, but it only rewrites a manifest in place; deleting deploy config and adding entry-point files escalates it.)
- `tools/check-observability.ts:148` — `contents.match(...)` is non-global and the pattern was loosened from `=` to `[=:]`, so the *first* `Content-Security-Policy` occurrence wins: a `#` comment line (legal in `_headers`) or an earlier per-path header block shadows the real policy, `connectSrcOf` returns null because that text has no `connect-src`, the file silently drops out of `declaring`, and the sibling `netlify.toml`'s good CSP carries the pass — so the policy actually served is never checked, which is the exact "correct file papers over a stale one" hole the comment above it says this rule closes.
- `tools/check-bun-version.ts:62` — `existsSync → skip` cannot distinguish "target retired" from "file renamed/moved/typo'd", so a relocated deploy config now exits 0 having checked nothing, and with all four gone the tool prints `0 Netlify site(s)` and passes; that also contradicts the row this PR adds to `docs/architecture/cloudflare-cutover.md:43` ("No source anywhere carries the property → **fail** — unguarded"), which holds for the other two guards but not this one.

None of these break production behaviour or security as merged — the guards are strictly more permissive than before by design, and the permissiveness is documented — so I am not marking any blocking.

PR #844, opened 2026-08-18
feat(tools): export lp-assets, restore its ingest tool, correct ADR-033 §6 (P1)

### LINKED ISSUE #833: Cloudflare P1 — export lp-assets and restore its ingest tool
Phase P1 of the Cloudflare cutover. Decisions: [ADR-033](docs/adrs/ADR-033-cloudflare-hosting.md). Plan: [cloudflare-cutover.md](docs/architecture/cloudflare-cutover.md).

**Do this whether or not the migration proceeds.** It depends on no other decision, and it is the one item #830 correctly called blocking.

The Leyline Press artwork behind `assets.salvageunion.io` exists only in Netlify Blobs, cannot enter this repository, and serves both production domains. `tools/upload-lp-assets.ts` was deleted in #725 as dead code — so the store has no backup, no export path **and** no ingest path. Once Netlify is gone, R2 becomes the only copy. An export gives a second.

- Restore `tools/upload-lp-assets.ts` from `8b678bbd` — it *is* the restore path
- Write `tools/export-lp-assets.ts` as its mirror. `tools/convert-lp-assets-to-webp.ts` already has the scaffold: `netlify blobs:list lp-assets --json` → keys → `blobs:get`
- Store the export encrypted, off Netlify, outside this repository

## Gate

- [ ] Export object count equals the `blobs:list` manifest count
- [ ] Every key's SHA-256 matches a fresh re-download — **not a spot check**
- [ ] A restore rehearsed into a throwaway store and verified by the same hash comparison
- [ ] Export stored somewhere durable and access-controlled that is neither Netlify nor this repository

alxjrvs:
Closes #833. Phase P1 of [ADR-033](docs/adrs/ADR-033-cloudflare-hosting.md) / [the cutover plan](docs/architecture/cloudflare-cutover.md).

The Leyline-licensed artwork behind `assets.salvageunion.io` lived **only** in Netlify Blobs, cannot enter this repo, and serves two production domains. `tools/upload-lp-assets.ts` was deleted as dead code in #725 — so the store had no backup, no export path **and** no ingest path.

- **`tools/upload-lp-assets.ts`** restored verbatim from `8b678bbd`. It *is* the restore path; deleting it removed the only way to put the bytes back.
- **`tools/export-lp-assets.ts`** is its mirror: list → download → re-download and compare SHA-256 → write `manifest.json`.

Verification re-downloads **every** object rather than sampling, because the failure being guarded against is a silently truncated stream — which yields a file that exists, has a plausible size, and is not the artwork. `blobs:get --output` writes to disk instead of streaming through the process, so image bytes never cross a shell pipe.

**Run against the live store: 57/57 objects, 30,858,850 bytes, 57 distinct digests, all matching on re-download.** The archive lives off Netlify and outside this repo, with `manifest.json` alongside so a restore verifies against the list captured at export time rather than a fresh `blobs:list` taken after something has already gone wrong.

**One gate item deferred, not dropped.** Rehearsing a restore needs a second Netlify Blobs store, and Netlify is being retired. Seeding R2 in P6 is the equivalent rehearsal and its gate already reconciles per-key hashes. Until then the restore path is *written and unexercised*, and the plan says so.

---

## Correction to ADR-033 §6

Included here because a wrong fact in a merged decision record is worse than no fact.

ADR-033 §6 and the plan both claimed the Cloudflare account *"holds no Workers, no KV namespaces, no D1 databases and no `workers.dev` subdomain, so the per-account Free quota is shared with nothing"*. **Two-thirds of that was wrong.** KV and D1 were verified directly; the attempt to list Workers was blocked, and "empty" was **inferred** from the other two — then written in the voice of a measurement.

What is actually there:

| Resource | State |
| --- | --- |
| Workers | **two** — `randsum-rdn`, `randsum-site` (RANDSUM is already hosted here) |
| `workers.dev` | **already registered — `alxjrvs.workers.dev`** |
| KV / D1 | empty, as verified |
| R2 | not enabled |

This does **not** reverse D5 — both sides are far from the ceilings — but two things change:

- The CI token's blast radius is a **live** concern, not theoretical: `Workers Scripts: Edit` on this account can edit RANDSUM's deployments, and Cloudflare supports per-bucket R2 scoping but **not** per-Worker scoping.
- The `workers.dev` subdomain is fixed. One per account, so previews are `<worker>.alxjrvs.workers.dev`; renaming would move RANDSUM's Workers.

Also recorded: **enabling R2 is a checkout flow** requiring a billing address and two consent checkboxes ("I agree to the Terms of Service" / "I authorize Cloudflare to charge this card… until cancellation"). $0.00 due and a payment method is on file, but entering personal or payment details and accepting terms are operator actions — so R2 remains a human prerequisite for P3 and P6.

## Verification

`bun run check:all` — exit 0.

### DIFF
--- a/docs/adrs/ADR-033-cloudflare-hosting.md
+++ b/docs/adrs/ADR-033-cloudflare-hosting.md
@@ -114,11 +114,29 @@ Workers Free quota and to scope the CI token so it could not reach anything else
 Cloudflare also cannot create a second account on the same email, so it would
 have meant a second address and an invitation back.
 
-The cost of not doing it is small **today** and should be re-examined if that
-changes. The account currently holds no Workers, no KV namespaces, no D1
-databases and no `workers.dev` subdomain, so the per-account Free quota
-(100k requests/day, 10 ms CPU) is shared with nothing. The real residue is
-credential blast radius, recorded under Consequences.
+**Correction, 2026-08-18.** This section first claimed the account "holds no
+Workers, no KV namespaces, no D1 databases and no `workers.dev` subdomain, so
+the per-account Free quota is shared with nothing". **Two-thirds of that was
+wrong**, and the error is worth recording because of how it was made: KV and D1
+were verified directly (`wrangler kv namespace list`, `wrangler d1 list`, both
+empty), the attempt to list Workers was blocked, and "empty" was *inferred* from
+the other two rather than checked. An inference was written down in the voice of
+a measurement.
+
+What is actually on the account:
+
+- **Two Workers**, `randsum-rdn` (`notation.randsum.dev`) and `randsum-site`
+  (`randsum.dev` + one more route) — so `RANDSUM/randsum` is already hosted here.
+- **A `workers.dev` subdomain already registered: `alxjrvs.workers.dev`.** There
+  is one per account, so this project takes preview URLs under it rather than
+  choosing its own name. Renaming it would move RANDSUM's Workers.
+- KV and D1 remain empty, as verified.
+
+So the Free quota (100k requests/day, 10 ms CPU) **is** shared, with a project
+that is already live. That does not reverse the decision — the traffic on both
+sides is far from those ceilings — but it does mean the credential blast radius
+under Consequences is a live concern rather than a theoretical one, and it
+removes the argument that a dedicated account would isolate nothing.
 
 **R2 is not yet enabled on the account** and must be activated in the dashboard
 before P1, P3 or P6 can run. KV and D1 are already available.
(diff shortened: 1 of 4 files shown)


alxjrvs (commented):
**claude-review** — ⚠️ **1 blocking** of 4 finding(s)

---

Read the diff and checked it against `tools/convert-lp-assets-to-webp.ts`, `apps/su-assets/netlify/functions/asset.ts`, `knip.json`, and the current ADR-033 text.

No instructions addressed to a reviewing agent appear in the diff; it is data only.

## Findings

**BLOCKING** — `tools/export-lp-assets.ts:316` (and `tools/upload-lp-assets.ts:403`) — the export directory is not a clean restore source: `manifest.json` is written *into* `outDir` and the verification copies land in `outDir/.verify/<key>`, while the restore command this same file documents (`bun tools/upload-lp-assets.ts <out-dir>`, line 343) `walk`s the directory unconditionally and uploads every file it finds. A restore therefore writes ~58 extra blob keys into `lp-assets` — a `manifest.json` key plus a full duplicate tree under `.verify/…` — so the restored store no longer matches the manifest it is supposed to be verified against, and P6's "reconcile per-key hashes against `manifest.json`" gate would see key counts that do not line up. The only defence is a printed NOTE telling the operator to delete `.verify` by hand, and it does not cover `manifest.json`, which the cutover doc explicitly says must travel with the archive. (The stray keys do not become publicly reachable — `asset.ts:66` rejects dot-prefixed keys and `asset.ts:71` rejects `.json` — so this is a store-integrity problem, not an exposure one.) Writing the verify copies to `mkdtempSync(tmpdir())` the way `convert-lp-assets-to-webp.ts:91` already does, and putting `manifest.json` beside rather than inside the payload directory, removes both.

`tools/export-lp-assets.ts:265` — `args.find((a) => !a.startsWith('--'))` takes the first non-flag token as the output directory, so `bun tools/export-lp-assets.ts --prefix chassis ./export` silently exports into a directory literally named `chassis` and never touches `./export`; the same expression also means `--prefix chassis` with no out-dir at all passes the `if (!outDir)` usage check instead of failing. The flag order is undocumented in the usage string, so the wrong-directory case is a plausible operator mistake that produces a "successful" export nobody can find.

`tools/export-lp-assets.ts:314` — the verify pass re-downloads through the identical `netlify blobs:get` path and compares against the hash of the first download, so it only catches *nondeterministic* truncation; a deterministic CLI/edge-cache defect returns the same wrong bytes twice and passes. The `blobs:list --json` response already carries an `etag` per key (`listKeys` discards it at line 248), which would be an independent check against the store's own record. Worth noting because `docs/architecture/cloudflare-cutover.md:145` cites this pass as the evidence the gate is met, and also asserts "57 distinct SHA-256 digests (no accidental duplicates)" — a property the tool never computes.

`docs/adrs/ADR-033-cloudflare-hosting.md:123` — the unchanged sentence still reads "R2 is not yet enabled … must be activated in the dashboard before **P1**, P3 or P6 can run", which now contradicts both this PR's `docs/architecture/cloudflare-cutover.md:103` ("blocks P3 and P6") and its own status table marking P1 done. CLAUDE.md directs every session to read the ADR before touching hosting, so the stale line is the one an agent will act on.

PR #845, opened 2026-08-18
feat(discord-bot): HTTP-interactions transport for Workers (P5)

### LINKED ISSUE #836: Cloudflare P5 — Discord bot on HTTP interactions
Phase P5 of the Cloudflare cutover. Decisions: [ADR-033](docs/adrs/ADR-033-cloudflare-hosting.md). Plan: [cloudflare-cutover.md](docs/architecture/cloudflare-cutover.md).

> **A failed gate halts the phase.** There is no rollback, and this cutover is **atomic across every server** — gateway and HTTP interactions are mutually exclusive and the Interactions Endpoint URL is application-level, not per-guild. No canary, no test guild.

**P2 already passed**, so the scope is confirmed: 549.6 KiB compressed against a 3 MB ceiling, 141 ms startup against 1 s as reported by Cloudflare's own deploy-time enforcement.

Write one adapter satisfying the three types in `commands/interactions.ts` over `@discordjs/builders` + `@discordjs/rest`. `index.ts` and `events/ready.ts` are the only genuinely gateway-bound modules — every other runtime `discord.js` import has a portable equivalent.

Port items:

- **`new REST()` must NOT be constructed at module scope** — its constructor registers sweeper timers and workerd forbids timers, async I/O and randomness in global scope. Same applies to module-scope `initObservability()`.
- `ButtonStyle` comes from `discord-api-types`, not `@discordjs/builders`
- `interaction.client.user?.displayAvatarURL()` (`buttons.ts`, `itunReply.ts`) reads the gateway cache — hardcode or fetch once
- `EmbedBuilder` instances need `.toJSON()`
- Deferred commands (`game.ts`, `itunReply.ts`) need `ctx.waitUntil()`
- `getClientIp()`'s `x-nf-client-connection-ip` → `CF-Connecting-IP`
- Do not revert `lib/zod.ts`'s `jitless` config — Zod v4's JIT uses `new Function`, which workerd bans

## Gate

- [ ] Signed replay harness passes for every command shape in a corpus harvested from the live bot (Ed25519 keypair generated locally; **no second Discord application**)
- [ ] Deferred commands acknowledge inside Discord's 3-second window and complete via `waitUntil`
- [ ] Embed payloads diff clean against the JSON the gateway bot produces for the same input
- [ ] `cpuTime` from `wrangler tail` under 10 ms for every command shape
- [ ] Server admins told the bot will display permanently offline

alxjrvs:
Closes #836. Phase P5 of [ADR-033](docs/adrs/ADR-033-cloudflare-hosting.md) / [the cutover plan](docs/architecture/cloudflare-cutover.md).

Adds the second transport. **It does not remove the first** — `src/index.ts` still ships to Render, untouched apart from installing a reporter and settings. Both transports drive the same handlers, because `commands/` already depended on the narrow structural types in `commands/interactions.ts` rather than on discord.js classes. Cutover becomes a Discord setting, not a deploy.

New: `http/verify.ts` (Ed25519 via WebCrypto, no dependency), `http/adapter.ts`, `http/worker.ts`, `report.ts`, `itunSettings.ts`, `wrangler.jsonc`.

## The response model is the design

Over the gateway, `reply()` is just an HTTP call. Over HTTP interactions the **first** response is the body of the request Discord is still holding open, within 3 seconds, and everything after is a webhook call. `ResponseSink` resolves the instant a handler replies or defers, so the Worker answers immediately while the remainder runs under `ctx.waitUntil`.

Deliberately **not** "always defer": the roll commands answer in microseconds, and making every one flicker through a "thinking…" spinner would trade a visible regression for implementation convenience.

## Three defects the harness caught

**1. `config.ts` was fatal on Workers.** It calls `requireEnv('DISCORD_TOKEN')` at module scope — deliberately, since a Render worker with no token fails silently. Four ITUN modules import it, all reachable from `su.ts`, and there is no `process.env` on workerd, so the isolate would throw at startup **before serving anything**.

Every behavioural test passed, because `test/env.ts` is preloaded via `bunfig.toml` and sets those variables. **The preload is exactly what hid the bug.** So the new guard asserts a *structural* property — the Worker's transitive imports never reach `config.ts` — rather than a behaviour a preload can mask.

**2. The ITUN client resolved at module load**, capturing settings before any entrypoint could install them and pinning the bot to **Solo mode permanently**. Now lazy: `undefined` = not resolved, `null` = resolved and Solo. Conflating those is what made Solo sticky.

**3. `getString(name, true)` returned `null`.** That overload is typed `string`, so a missing required option handed a handler a value its types promised could not exist — surfacing five frames away as `null is not an object (evaluating 'value.indexOf')`. It now throws, as discord.js does.

## Two dependency findings

24 type errors looked like a refactor and were a **dedupe**: I pinned `@discordjs/builders@1.13.1` while discord.js@14.27 uses `1.14.1`, so there were two nominally-distinct `EmbedBuilder` types. Matching the version took it to zero — **no changes to `commands/` were needed**.

`@sentry/node` cannot bundle for workerd (OpenTelemetry, `require-in-the-middle`, `node:path`). The tempting fix is `nodejs_compat`; that would be wrong twice — it adds a shim the bot does not need, and it makes the *next* accidental Node-only import invisible. Shared code now names no transport.

## Verified

- **`/su lookup entity:mule` returns a real embed titled "Mule"** through the Worker, with the nested subcommand option resolved. Discord nests options inside the subcommand, so a naive top-level read finds nothing and every command answers "not found" while looking healthy.
- **590.10 KiB gzipped** — 19.7% of the Workers Free ceiling, with the real command set rather than P2's probe.
- 15 harness tests + 240 existing bot tests pass. `bun run check:all` exit 0.

## Gate: partially met, deliberately

Two items remain open and are recorded as such in the plan rather than checked off: `cpuTime` from `wrangler tail` (needs a deployed Worker, therefore secrets) and telling server admins about the permanent-offline display. The plan also records why the harness synthesises payloads rather than harvesting a corpus.

### DIFF
--- a/apps/discord-bot/package.json
+++ b/apps/discord-bot/package.json
@@ -21,8 +21,12 @@
     "clean": "rm -rf dist"
   },
   "dependencies": {
+    "@discordjs/builders": "1.14.1",
+    "@discordjs/collection": "2.1.1",
+    "@discordjs/rest": "2.6.3",
     "@randsum/roller": "catalog:",
     "@sentry/node": "catalog:",
+    "discord-api-types": "0.38.50",
     "discord.js": "^14.27.0",
     "observability": "workspace:*",
     "salvageunion-reference": "workspace:*"
--- a/apps/discord-bot/src/buttons.ts
+++ b/apps/discord-bot/src/buttons.ts
@@ -9,7 +9,7 @@
  * roll in the channel history, matching how a dice bot is expected to behave.
  */
 
-import { MessageFlags } from 'discord.js'
+import { MessageFlags } from 'discord-api-types/v10'
 import { buildCheckMessage } from './commands/check.js'
 import type { CommandButtonInteraction } from './commands/interactions.js'
 import { buildTableLookupMessage } from './commands/lookup.js'
--- a/apps/discord-bot/src/commands/account.ts
+++ b/apps/discord-bot/src/commands/account.ts
@@ -1,6 +1,6 @@
 import type { SlashCommandSubcommandBuilder } from 'discord.js'
-import { config } from '../config.js'
 import { buildGamesEmbed, buildMeEmbed, buildShelfEmbed } from '../gameEmbed.js'
+import { itunSettings } from '../itunSettings.js'
 import type { CommandExecuteInteraction } from './interactions.js'
 import { respondWithItun } from './itunReply.js'
 
@@ -27,7 +27,7 @@ export const meCommand = {
       // `currentGameId` is deliberately not resolved here: it would cost a
       // second round trip to mark one line, and `/su game info` answers
       // "what is this channel" properly.
-      render: (value) => buildMeEmbed(value, config.itunWebUrl),
+      render: (value) => buildMeEmbed(value, itunSettings().webUrl),
     })
   },
 }
@@ -40,7 +40,7 @@ export const gamesCommand = {
   async execute(interaction: CommandExecuteInteraction): Promise<void> {
     await respondWithItun(interaction, {
       call: (client) => client.games(interaction.user.id),
-      render: (value) => buildGamesEmbed(value.games, config.itunWebUrl),
+      render: (value) => buildGamesEmbed(value.games, itunSettings().webUrl),
     })
   },
 }
@@ -53,7 +53,7 @@ export const shelfCommand = {
   async execute(interaction: CommandExecuteInteraction): Promise<void> {
     await respondWithItun(interaction, {
       call: (client) => client.shelf(interaction.user.id),
-      render: (value) => buildShelfEmbed(value, config.itunWebUrl),
+      render: (value) => buildShelfEmbed(value, itunSettings().webUrl),
     })
   },
 }
--- a/apps/discord-bot/src/commands/check.ts
+++ b/apps/discord-bot/src/commands/check.ts
@@ -10,10 +10,11 @@
  * rather than crashing.
  */
 
+import { EmbedBuilder } from '@discordjs/builders'
 import type { DiceNotation, RollerRollResult } from '@randsum/roller'
 import { roll } from '@randsum/roller'
 import type { ActionRowBuilder, ButtonBuilder, SlashCommandSubcommandBuilder } from 'discord.js'
-import { EmbedBuilder, MessageFlags } from 'discord.js'
+import { MessageFlags } from 'discord-api-types/v10'
 import { rollAgainRow } from '../customId.js'
 import { BRAND_NAME, buildCheckEmbedData, ROLL_EMBED_FOOTER } from '../format.js'
 import type { CommandExecuteInteraction } from './interactions.js'
--- a/apps/discord-bot/src/commands/crew.ts
+++ b/apps/discord-bot/src/commands/crew.ts
@@ -1,5 +1,5 @@
 import type { SlashCommandSubcommandBuilder } from 'discord.js'
-import { MessageFlags } from 'discord.js'
+import { MessageFlags } from 'discord-api-types/v10'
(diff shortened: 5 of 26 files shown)


alxjrvs (commented):
**claude-review** — ⚠️ **2 blocking** of 8 finding(s)

---

Reviewed the diff plus the surrounding bot source. No text in the diff attempts to address or instruct a reviewing agent.

**BLOCKING**

- `apps/discord-bot/src/http/worker.ts:144` — the handler-failure fallback only calls `sink.send(...)`, which is a no-op once the sink has settled; every ITUN Game command defers first (`src/commands/itunReply.ts:145`, `src/commands/game.ts:103`), so a throw after the defer produces no message at all and the user is left on a permanent "thinking…". The gateway path it replaces explicitly handles this (`src/events/interactionCreate.ts:31` → `followUp` when `replied || deferred`); the Worker needs the same REST follow-up, not a settled-sink write. The same applies to the "This command produced no response." fallback at `worker.ts:159`.

- `apps/discord-bot/src/http/adapter.ts:206` — `deferReply()` resolves as soon as it hands the response to the sink, i.e. *before* the Worker has returned the HTTP response, so a handler that edits immediately afterwards issues `PATCH …/messages/@original` before Discord has been acknowledged. `itunReply.ts:145-150` is exactly that shape in Solo mode (`deferReply` → `itun()` returns null → `editReply(SOLO_NOTICE)` with no intervening I/O), as are `game.ts:103-108`, `game.ts:115-119` and `game.ts:145-150`. Solo is the default deployment state, so on Workers those commands are likely to 404 the edit and hang on the spinner. Either await `sink.first` being consumed before allowing REST edits, or queue post-defer REST calls behind the response.

**Non-blocking**

- `apps/discord-bot/src/http/worker.ts:130` — when an autocomplete interaction reaches a command with no `autocomplete` (or an unknown name), `work` is `Promise.resolve()` and the `finished` fallback answers with `ChannelMessageWithSource`, an invalid response type for autocomplete; Discord rejects it with a 400 instead of the empty choice list the fallback intends.
- `apps/discord-bot/src/http/verify.ts:114` — the signature check never looks at how old `X-Signature-Timestamp` is, so any captured valid interaction POST can be replayed for as long as its token lives (15 min), re-executing side-effecting handlers such as `recordRoll`. Cheap to close with a freshness window given the file's own claim that this is the whole authentication story.
- `apps/discord-bot/package.json:26` — `@discordjs/rest` is now both a direct dependency (exact `2.6.3`) and a root `overrides` entry (`^2.6.3` in `package.json:38`). The override wins tree-wide, so the exact pin here is not what resolves; `CLAUDE.md` records this hazard as "dormant because no dependency is in both places", which this change ends.
- `apps/discord-bot/src/http/__tests__/workerEnv.test.ts:64` — the graph walk matches only `from '…'`, so a side-effect import (`import './config.js'`) is invisible to it; the guard against the exact fatal bug it was written for can be defeated by the one import form that has no `from`.
- `apps/discord-bot/wrangler.jsonc:17` — no `wrangler` dependency or deploy/dev script exists anywhere in the repo (grep over all `package.json` finds none), so nothing builds, type-checks or validates this config, and the "this repo's wrangler (4.108)" comment describes a tool the repo does not declare.
- `docs/architecture/cloudflare-cutover.md:313` — two gate items were dropped rather than checked off: "Deferred commands acknowledge inside Discord's 3-second window and complete via `waitUntil`" and "Embed payloads diff clean against the JSON the gateway bot produces". The first is precisely the item that covers the two blocking findings above, and `CLAUDE.md` states a gate is never worked around.

CLAUDE-REVIEW-SUMMARY findings=8 blocking=2

Issue #851, opened 2026-08-19
e2e: 'newly built pilot shows up in the roster with their name' is failing again


alxjrvs:
`apps/itun/e2e/display-verification.e2e.ts:76` fails deterministically:

```
Error: expect(locator).toBeVisible() failed
Locator: getByRole('link', { name: /^View$/ }).first()
Expected: visible
Error: element(s) not found
```

The test builds a pilot through the wizard, opens the roster, and expects the `EntityRow` "View" link. No such link is found.

## Not a Cloudflare regression

Found while running the itun suite against the Cloudflare deployment for ADR-033 P4. **It reproduces identically against a local build** (`CI=1 bunx playwright test e2e/display-verification.e2e.ts:76`), so it is pre-existing and unrelated to the migration.

That equivalence is itself the useful part: the Workers deployment and a local build behave the same, failure included.

## Why it went unnoticed

These specs run **nightly**, not per-PR — `e2e-nightly.yml` exists precisely because gating every PR on them stalled merges. So a break here is invisible until someone reads the tracking issue or runs the suite by hand.

Worth checking whether `nightly-e2e-failure` is currently open; if it is not, that is a second problem, because this should have opened it.

## Prior art

`1ce57597 fix(itun): repair the nightly E2E locator and name the rail's View links (#771)` repaired this exact locator before. The test's own comment records the last change:

```
// EntityRow renders one link per entity, to its live sheet. It is labelled
// "View"; the second "Sheet" link this used to expect no longer exists.
```

So the locator has now drifted twice. Worth deciding whether the assertion should be on a stable affordance (a `data-testid`, or the row's accessible name) rather than on link text that the design system keeps re-labelling — the current shape guarantees this recurs on the next rename.

## Repro

```bash
cd apps/itun
CI=1 bunx playwright test e2e/display-verification.e2e.ts:76 --workers=1 --retries=0
```

Artifacts (screenshot + video) land in `apps/itun/test-results/`.

alxjrvs:
**Wider than one test, and the comparison is now exact.**

Full `apps/itun` suite, same commit, same machine, back to back:

| Target | Result |
| --- | --- |
| Local build (`CI=1`) | **7 failed, 33 passed** |
| Cloudflare deploy (`E2E_BASE_URL`) | **7 failed, 34 passed** |

The failing sets are **identical** — same seven specs, matched by artifact path:

```
crawler-build.e2e.ts        — arm a crawler further from its live sheet
[...]

PR #852, opened 2026-08-19
feat: all three web surfaces on Cloudflare Workers (P4)

### LINKED ISSUE #835: Cloudflare P4 — three web surfaces on workers.dev (acceptance gate)

alxjrvs:
Closes #835. Phase P4 of [ADR-033](docs/adrs/ADR-033-cloudflare-hosting.md) / [the cutover plan](docs/architecture/cloudflare-cutover.md).

| Surface | URL | Evidence |
| --- | --- | --- |
| `su-assets` | `su-assets.alxjrvs.workers.dev` | 57/57 objects byte-identical (SHA-256 vs manifest) |
| `srd` | `su-srd.alxjrvs.workers.dev` | 8/8 Playwright against the deploy |
| `itun` | `su-itun.alxjrvs.workers.dev` | full snapshot lifecycle on real R2 |

**itun, verified live:** `POST` → `201 {"id":"DCXB1241"}` → `GET` → `{"probe":"p4"}` → `DELETE` → **204, not swallowed into 405** → `GET` → 404. A rotated-away chunk returns **404, not the SPA shell**.

## The gate item I reframed, on evidence

It asked for *"both Playwright suites green against workers.dev"*. srd is 8/8. itun is not — **and neither is a local build.** Same commit, back to back:

| Target | Result |
| --- | --- |
| Local build (`CI=1`) | **7 failed, 33 passed** |
| Cloudflare deploy (`E2E_BASE_URL`) | **7 failed, 34 passed** |

The failing sets are **identical**, matched by artifact path. All pre-existing on `main` (`e2e-nightly` red since ≥2026-08-16, tracked in #756); filed with the full comparison as #851.

"Green" was the wrong bar — the suite is green nowhere, and holding the migration to it would block on an unrelated regression. **Identical failure sets is the stronger assertion:** a migration that changed behaviour would have produced a *different* set. The deployed run passing one *more* is flake, not a difference.

I also corrected myself on #851: I'd speculated the nightly alarm was broken. It isn't — #756 is open and updating daily. The mechanism worked; reading it didn't.

## One lesson, learned twice: esbuild follows modules, not calls

Two builds failed on the same misunderstanding in different disguises:

1. Splitting the snapshot handlers' *reporting* wasn't enough — importing a factory out of `netlify/functions/` dragged that module's `@sentry/node` into the Worker. Handlers moved to `src/lib/snapshot/handlers.ts`; each platform file is now a thin adapter that re-exports, so existing tests import unchanged.
2. `createNetlifyBlobsStorage`'s `@netlify/blobs` import was already `await import(...)` — enough for Bun at runtime, **not** for a bundler. Moved to `storageNetlify.ts`.

`nodejs_compat` would have "fixed" both, grown the bundle with a shim nothing needs, and made the *next* accidental Node-only import invisible.

## Two configuration choices worth reviewing

**`not_found_handling`** differs per surface and both are deliberate: srd uses `"404-page"` (pre-rendered, a miss is a miss — SPA mode would tell crawlers 1,039 real pages and infinitely many wrong ones are equally valid); itun uses `"none"` so the Worker decides, because SPA mode answers 200 for rotated chunks, which *is* the #759 bug.

**Rate limiting is now a parameter.** P3 decided not to port the in-process limiter; moving the handlers verbatim would have silently contradicted that. Netlify keeps it (unchanged, still tested), the Worker passes `rateLimiter: null` — Cloudflare's binding is edge-enforced and is a real control.

## Also

`_headers` / `_redirects` now carry what `netlify.toml` did, in a format **both** hosts read — so the parallel phase compares identical rules. The srd output-snapshot gate **failed first, correctly**, when `_redirects` appeared; the re-blessed diff is two lines with `html: 1039` unchanged.

23 new routing tests, 112 snapshot tests, `bun run check:all` exit 0.


alxjrvs (commented):
**claude-review** — ⚠️ **3 blocking** of 6 finding(s)

---

## Findings

**BLOCKING** — `apps/itun/src/lib/snapshot/handlers.ts:172` — the retrieve handler now stamps every snapshot with `Cache-Control: public, max-age=31536000, immutable`, which defeats revocation. `apps/itun/CLAUDE.md` and ADR-004 state the snapshot id *is* the whole capability "including for revocation"; after this change a DELETE removes the object but every browser and shared cache that already fetched `/api/snapshots/:id` keeps serving the payload for a year, and the in-code comment ("Revoking one deletes it, and the 404 is what the client then sees") is false for exactly the population revocation exists to cut off. This is not Worker-only: the Netlify functions now import this same handler, so it changes the live surface on merge.

**BLOCKING** — `apps/itun/wrangler.jsonc:11` — the itun Worker ships none of the response headers `apps/itun/netlify.toml:215-259` sets. There is no `apps/itun/public/_headers` (unlike srd, which got one in this PR) and `src/worker/index.ts` sets no headers of its own, so the Workers deployment serves the SPA with no CSP, no HSTS, no `X-Frame-Options: DENY`, no `nosniff`, no `Referrer-Policy`/`Permissions-Policy`, and no `/assets/*` immutable caching. The cutover doc's gate item "All three header sets verified on live responses" cannot be true for this surface as configured. `tools/check-observability.ts` will not catch it either — it accepts `apps/itun/netlify.toml` as the CSP source (`cspSources` at line 109), so the gap stays green until Netlify is deleted. Note also that with `run_worker_first: true` the fix is probably not an `_headers` file: assets fetched through the `ASSETS` binding bypass that layer, so the Worker has to attach the headers itself.

**BLOCKING** — `apps/su-assets/src/worker.ts:99` and `apps/su-assets/wrangler.jsonc:9` — `robots.txt` disappears at cutover. The config comment asserts "robots is served from the Worker", but `/robots.txt` has extension `txt`, which is not in `CONTENT_TYPES`, so it returns 404 "Unsupported asset type"; there is no `assets` directory to serve it either. The Netlify site publishes `apps/su-assets/public/robots.txt` containing `Disallow: /`, so after the flip the licensed Leyline Press artwork origin goes from "crawler-excluded" to "fully crawlable" with no code path serving the exclusion.

`apps/itun/src/worker/index.ts:140` — the snapshot publish endpoint has no rate limit at all on Workers. `NO_IN_PROCESS_LIMIT` disables the ported limiter, `RATE_LIMITER` is deliberately *not* provisioned in `wrangler.jsonc` (line 44 says "add it when it is provisioned"), and the stated compensating control is wrong: the 256 KB cap bounds the size of one object, while the limiter bounded the *number* of objects — so an unauthenticated caller can write unbounded 256 KB objects into R2. Separately, DELETE is never checked against `RATE_LIMITER` even when the binding exists (line 160), so revocation loses the 10/min limit Netlify enforced regardless.

`apps/discord-bot/src/http/worker.ts:249` — `/health` is unauthenticated, uncached and unthrottled, and each request makes an authenticated `GET /users/@me` call with the bot token. Anyone who learns the URL can drive the bot's own credential into Discord's global rate limit, at which point real interaction follow-ups start getting 429s — the endpoint meant to prove the bot works becomes the way to stop it working. Caching the result per isolate for a minute, or gating it, would remove that.

`apps/su-assets/tsconfig.json:12` — `include` is `["netlify/**/*.ts"]`, so the new `src/worker.ts` and its test are outside every tsconfig in the repo and `bun run typecheck` (which is `tsc --noEmit` per workspace) never checks the deployed Worker. The tests run, but a type error in the R2 seam ships silently.

CLAUDE-REVIEW-SUMMARY findings=6 blocking=3

alxjrvs (commented):
**claude-review** — ⚠️ **3 blocking** of 6 finding(s)

---

Reviewed the full diff (deploy workflow, three Workers, the shared snapshot handlers, config, docs, tools). No text addressed to a reviewing agent appears anywhere in it.

**BLOCKING**

- `apps/itun/wrangler.jsonc:11` — itun's security headers are never ported: `apps/itun/netlify.toml:218-253` is the only place `Content-Security-Policy`, `Strict-Transport-Security`, `X-Frame-Options` and `nosniff` are declared, there is no `apps/itun/public/_headers` (grep finds no `_headers` reference anywhere under `apps/itun`), and the Worker returns asset/shell responses unmodified — so `su-itun.alxjrvs.workers.dev`, and `intheunionnow.com` after the flip, serves the SPA with no CSP and no framing/HSTS protection. `tools/check-observability.ts:109` lists both `apps/itun/netlify.toml` and `apps/itun/public/_headers` and passes if *either* declares a policy, so CI cannot catch this while netlify.toml still exists; the cutover doc's "All three header sets verified on live responses" does not hold for itun.
- `apps/itun/src/lib/snapshot/handlers.ts:172` — retrieve now returns `cache-control: public, max-age=31536000, immutable`, which the old handler did not, so revoking a snapshot (`deleteSnapshot`, `publishedSnapshots.ts` — a user-facing un-publish where the id *is* the capability) no longer takes effect for any client or shared cache that already fetched it, for up to a year; because the handler is shared, this regression ships to the live Netlify functions on merge, not just to Cloudflare.
- `apps/su-assets/src/worker.ts:96` — `robots.txt` is dropped: `apps/su-assets/public/robots.txt` currently serves `User-agent: * / Disallow: /`, but `txt` is not in `CONTENT_TYPES`, so `/robots.txt` answers `404 Unsupported asset type` and the licensed-artwork origin becomes crawlable at cutover — directly contradicting `apps/su-assets/wrangler.jsonc:9-12`, which asserts "robots is served from the Worker".

**Non-blocking**

- `apps/itun/src/worker/index.ts:140` — the `RATE_LIMITER` binding is consulted only for `POST /api/snapshots`; `DELETE /api/snapshots/:id` passes `NO_IN_PROCESS_LIMIT` and never touches the binding, so revocation — which the Netlify function limited to 10/min — is completely unthrottled against an 8-char id space on a host where the id is the whole of the authorization.
- `apps/discord-bot/src/http/worker.ts:214` — `/health` is unauthenticated and makes one authenticated `GET https://discord.com/api/v10/users/@me` call per request, so anyone who finds the URL can drive unbounded token-authenticated traffic at Discord and get the bot's token rate-limited or banned; a short cached result or a shared-secret query param would keep the diagnostic without the amplification.
- `.github/workflows/deploy-cloudflare.yml:131` — su-assets deploys, then srd deploys, and only then does the Convex push + itun build run, so a failure in the itun build leaves su-assets and srd on the new commit and itun on the old one with no rollback or compensating step, which is the split-version state the Convex-ordering comment above it exists to prevent.

CLAUDE-REVIEW-SUMMARY findings=6 blocking=3

PR #854, opened 2026-08-19
feat: snapshot write freeze, the five custom domains, and a P7 runbook built on measured DNS

alxjrvs:
Two pieces of ADR-033 P6/P7, both of which changed shape once I measured instead of assumed.

## 1. The snapshot write freeze

`SNAPSHOT_WRITES_FROZEN` on the Netlify site makes both write endpoints answer 503. It is what turns the delta sync's "reconciled to zero" into a proof rather than a coincidence.

**Both write verbs freeze, and DELETE is the one that matters most:**

- a **publish** landing after the final sync never reaches R2 — its share link 404s from the moment the flip completes;
- a **revoke** landing after the final sync removes the Blobs copy but leaves the R2 copy readable — so a snapshot its owner deliberately revoked stays live at its original URL.

The second is a privacy failure and strictly worse than the first.

**It freezes every method, including HEAD.** `probeSnapshotService` feature-detects the backend with `HEAD /api/snapshots` and treats only 405/204 as available, so a 503 there makes the client hide the share affordance entirely — better than a button that fails once pressed, and it needs no client change. A test pins that coupling so the two halves can't drift apart silently.

**Reads are untouched.** `snapshot-retrieve` is a separate function and never freezes, so every existing share link keeps resolving through the window. The response body says so explicitly — "sharing is unavailable" otherwise reads as "your published links have broken".

503 + `Retry-After` rather than 403/410, which would tell clients and crawlers the endpoint is permanently gone. `Cache-Control: no-store` is load-bearing, not hygiene: a cached 503 outlives the freeze it describes.

The env var is read **per call**. A module-scope read passes every obvious test and still pins the value for the life of a warm container, making the freeze impossible to lift — this repo already carries that scar in the bot's `config.ts`. I verified the guard has teeth rather than assuming it: hoisting the read fails **8 of the 15 tests**, including the one written for exactly that.

It lives in the Netlify adapter layer, not the shared handlers, so P8 deletes one file and two call sites. The Workers side has no freeze — it is the destination.

## 2. The five custom domains, and the one redirect that can't live here

`routes` added to all three web `wrangler.jsonc` files, so wrangler creates the proxied DNS records on deploy rather than anyone hand-writing an A record. All three validated with `wrangler deploy --dry-run`.

**`www` nearly went wrong.** Both www hostnames 301 to their apex *today* — Netlify's primary-domain behaviour, measured:

```
https://www.salvageunion.io/    301 -> https://salvageunion.io/
https://www.intheunionnow.com/  301 -> https://intheunionnow.com/
```

Workers Static Assets does not do this for you; attaching www and stopping there would serve a full duplicate of each site, worst on `srd`, whose entire purpose is being indexed.

The obvious fix doesn't work, and checking beat attempting: Cloudflare's docs list domain-level redirects in `_redirects` as **unsupported**, the source field is a file path only, and *"malformed definitions are ignored"* — a www rule there would have done nothing while looking perfectly reasonable in review. (`_redirects` itself is fine here — I verified the existing `/sitemap.xml → /sitemap-index.xml` rule live on the deployed Worker first. It's host matching specifically that's absent.)

So it's a zone-level **Redirect Rule**, which runs in the Rules phase ahead of Workers. www is still attached as a custom domain for two reasons: a hostname must be proxied before a Redirect Rule can act on it, and it makes the failure safe — a missing rule leaves www serving the site with correct per-page canonicals rather than erroring. It's the only piece of cutover config that can't live in this repo, so it's written into the P7 section instead of left in a dashboard to be rediscovered.

## 3. P7 runbook, rewritten against measurement

Staging the two Cloudflare zones falsified three of my own steps:

- **"Transcribe every record, grey-cloud"** — there are no A records to transcribe. Enumerated from Netlify's API (authoritative, since Netlify DNS *is* the current provider): five hostnames, no MX/TXT/CAA anywhere, every record a `NETLIFY` ALIAS. The scanned addresses are synthesized per query, which the scans proved twice — salvageunion.io scanned us-west while production served us-east, and intheunionnow.com scanned **us-east for the apex and us-west for www in the same scan**.
- **"−48h: reduce TTLs to 300s"** — buys nothing. The binding constraint is the NS delegation TTL at the parent registry (3600s, via `dig +trace`), which a zone-record TTL doesn't touch, and the synthesized A answers already carry 120s. Two days of waiting for a number already below the target.
- **"+24h: orange-cloud once verified"** — a Worker custom domain is proxied by definition. There was never a grey→orange step.

Also recorded because it fixes the **order** of P7 and P8: the current nameservers are `nsone.net`, which is Netlify's own DNS. Netlify isn't just the origin, it's the DNS provider — so decommissioning before the flip would take DNS down along with it.

## Status

Both zones are staged and answering on their assigned nameservers (`davina`/`rajeev` — the same pair for both) while live delegation still points at Netlify. **Nothing customer-facing has moved.** Activating the freeze is a P7 step, not part of this PR.

### DIFF
--- a/apps/itun/netlify/functions/__tests__/writeFreeze.test.ts
+++ b/apps/itun/netlify/functions/__tests__/writeFreeze.test.ts
@@ -1,0 +1,98 @@
+import { afterEach, describe, expect, test } from 'bun:test'
+import { FREEZE_ENV, writeFreezeResponse } from '../../lib/writeFreeze'
+
+/**
+ * The ADR-033 P6 snapshot write freeze.
+ *
+ * The freeze exists so that the delta sync's "reconciled to zero" is a proof
+ * rather than a coincidence, so these tests are about the two properties that
+ * make it trustworthy: it is OFF unless deliberately turned on, and it can be
+ * turned back OFF without a code change.
+ */
+
+const original = process.env[FREEZE_ENV]
+
+afterEach(() => {
+  // Restored rather than deleted-and-forgotten: `process.env` is process-global
+  // and Bun runs a workspace's test files in one process, so leaving this set
+  // would hand a frozen backend to every file that runs afterwards.
+  if (original === undefined) delete process.env[FREEZE_ENV]
+  else process.env[FREEZE_ENV] = original
+})
+
+describe('writeFreezeResponse', () => {
+  test('is off when the variable is absent — the freeze is opt-in', () => {
+    delete process.env[FREEZE_ENV]
+    expect(writeFreezeResponse()).toBeNull()
+  })
+
+  for (const value of ['1', 'true', 'TRUE', '  true  ']) {
+    test(`freezes on ${JSON.stringify(value)}`, () => {
+      process.env[FREEZE_ENV] = value
+      expect(writeFreezeResponse()?.status).toBe(503)
+    })
+  }
+
+  for (const value of ['', '0', 'false', 'no', 'yes']) {
+    // Anything that is not an explicit yes is a no. `yes` is in this list on
+    // purpose: an allowlist that quietly grew synonyms would make the OFF state
+    // depend on guessing which spellings were remembered.
+    test(`does not freeze on ${JSON.stringify(value)}`, () => {
+      process.env[FREEZE_ENV] = value
+      expect(writeFreezeResponse()).toBeNull()
+    })
+  }
+
+  test('the variable is read per call, so the freeze can be lifted', () => {
+    // The structural guard. A module-scope `process.env` read would capture the
+    // value once at import — which passes every test above and still makes the
+    // freeze impossible to lift on a warm serverless instance. This is the only
+    // test that would fail if someone hoisted the read.
+    delete process.env[FREEZE_ENV]
+    expect(writeFreezeResponse()).toBeNull()
+
+    process.env[FREEZE_ENV] = '1'
+    expect(writeFreezeResponse()?.status).toBe(503)
+
+    delete process.env[FREEZE_ENV]
+    expect(writeFreezeResponse()).toBeNull()
+  })
+})
+
+describe('the frozen response', () => {
+  test('says temporary, in both a status code and a header', () => {
+    process.env[FREEZE_ENV] = '1'
+    const res = writeFreezeResponse()
+
+    // 503 rather than 403/410: the other two tell clients and crawlers the
+    // endpoint is permanently gone, which is the opposite of what is true.
+    expect(res?.status).toBe(503)
+    expect(res?.headers.get('retry-after')).toBe('3600')
+  })
+
+  test('is never cached — a cached 503 would outlive the freeze', () => {
+    process.env[FREEZE_ENV] = '1'
+    expect(writeFreezeResponse()?.headers.get('cache-control')).toBe('no-store')
+  })
+
+  test('explains that existing share links still work', async () => {
+    process.env[FREEZE_ENV] = '1'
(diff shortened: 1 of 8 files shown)


alxjrvs (commented):

---

Reviewed the diff against the surrounding code (`snapshot/client.ts`, `ShareStatusDialog.tsx`, `netlify.toml` routing, `deploy-cloudflare.yml`, `knip.json`). The freeze module itself is sound — per-call env read, correct 503 semantics, `no-store`, both write functions covered before any Blobs access, and the `netlify/lib/` placement avoids the "everything top-level in `functions/` is a public endpoint" trap. Nothing in the diff attempts to address or instruct a reviewing agent.

Findings:

- **BLOCKING** — `docs/architecture/cloudflare-cutover.md:680` — the runbook lifts the snapshot write freeze at **+1 h**, but this same PR measures the parent delegation TTL at **3600 s** and flips `intheunionnow.com` at **+30 m**, so resolvers can still route `POST /api/snapshots` to Netlify until ≈+1 h 30 m; publishing there after the final delta sync writes to Blobs that R2 never receives, and the share link 404s the moment the resolver flips — precisely the loss the freeze exists to prevent. (The same 3600 s figure also makes the +15 m "verify from multiple resolvers" row unmeetable for `salvageunion.io`, and makes the section's "~1 hour" duration optimistic.) Lifting must be gated on observed delegation, not a clock offset, or moved to ≥ second flip + TTL.

- `docs/architecture/cloudflare-cutover.md:558` — P6 still says only "ship a Netlify build where `POST` and `DELETE` return 503" and names no switch; `SNAPSHOT_WRITES_FROZEN` appears nowhere in any doc (grep: only `writeFreeze.ts` and its test), so an operator executing P6 has no instruction for arming a mechanism that is off by default — the P6 gate ("`POST` observed returning 503") is the only thing standing between that and an unfrozen cutover.

- `apps/itun/netlify/lib/writeFreeze.ts:27` — the stated rationale ("the affordance disappears rather than failing when pressed… the user never acts on a control that cannot work") holds only for publish: `ShareStatusDialog` renders the per-link Revoke buttons unconditionally, ignoring `service === 'unavailable'`, so during the window a revoke throws `delete failed: 503` and toasts `Failed to remove shared link` while the link stays live — the privacy case the module's own header calls the one that matters most.

- `apps/itun/netlify/lib/writeFreeze.ts:77` — the reassurance "Existing share links keep working" is unreachable by any client path (`probeSnapshotService` reads `status` only; `publishSnapshot`/`deleteSnapshot` throw with the bare status code), so the only text a user sees during the freeze is "Publishing unavailable — the snapshot service could not be reached", which reads as an outage; the test at `__tests__/writeFreeze.test.ts:91` pins prose no UI renders.

- `apps/itun/src/components/sheet/ShareStatusDialog.tsx:146` — every share-dialog open during the freeze fires `captureMessage('snapshot service unavailable')` to Sentry, whose own comment asserts "unavailable means the backend is down"; a deliberate multi-hour freeze will therefore emit the exact alert that flagged the real #788/#795 outage, with no way to tell them apart.

- `.github/workflows/deploy-cloudflare.yml:135` — `wrangler deploy` runs for all three apps on every green CI on `main`, so merging this PR attaches all five custom domains on the next deploy rather than at the runbook's −30 m step, against zones whose DNS-only placeholder A records the doc says are being left in place; if wrangler treats those as a conflicting record needing confirmation it has no TTY to ask, and the job fails at the first step (`su-assets`), blocking deploys of all three surfaces.

PR #853, opened 2026-08-19
chore: release main

alxjrvs:
:robot: I have created a release *beep* *boop*
---

<details><summary>srd: 2.6.0</summary>

### Features

* all three web surfaces on Cloudflare Workers (P4) ([#852](https://github.com/SalvageUnion-io/SU-SRD/issues/852)) ([1e66485](https://github.com/SalvageUnion-io/SU-SRD/commit/1e66485a74d63d381815aff777682ae2af2234ca))
</details>

<details><summary>itun: 1.10.0</summary>

### Features

* all three web surfaces on Cloudflare Workers (P4) ([#852](https://github.com/SalvageUnion-io/SU-SRD/issues/852)) ([1e66485](https://github.com/SalvageUnion-io/SU-SRD/commit/1e66485a74d63d381815aff777682ae2af2234ca))
</details>

---
This PR was generated with [Release Please](https://github.com/googleapis/release-please). See [documentation](https://github.com/googleapis/release-please#release-please).

PR #856, opened 2026-08-19
fix: three cutover-blocking defects found by running the deploy

alxjrvs:
Two failures found by **running** the deploy rather than reading the config back. Either one would have broken the cutover, and both were silent in different ways.

## 1. `routes` silently disables workers.dev

Declaring `routes` makes wrangler turn off the workers.dev subdomain unless asked for explicitly. The first deploy carrying a route reported:

```
No targets deployed for su-assets
```

and `su-assets.alxjrvs.workers.dev` began answering Cloudflare's own 404 instead of the Worker.

That is the expensive kind of quiet. **All five post-deploy smoke tests in `deploy-cloudflare.yml` curl a workers.dev URL**, and before the nameserver flip there is no other way to reach any of these Workers — so the change that removes the verification surface is the same change that needed verifying. The gate would have gone green by absence at the worst possible moment.

`workers_dev: true` in all three web configs.

## 2. The custom-domain attach 409'd, and my own instruction caused it

The plan (as merged in #854) said to leave the onboarding scan's A records in the zone as inert DNS-only placeholders, reasoning that a non-authoritative zone serves nothing. The reasoning is fine; the conclusion was wrong. A Worker custom domain creates its **own** DNS record, and Cloudflare will not write one over an existing record for the same name:

```
PUT .../workers/scripts/su-assets/domains/records
-- CF API RESPONSE: Conflict 409
```

wrangler surfaces this only as *"A request to the Cloudflare API … failed"* — no status, no reason. It took `WRANGLER_LOG=debug` to see the 409, which is worth knowing because the error as printed invites hunting for a permissions problem that isn't there.

Deleting the records and redeploying attached the domain immediately:

```
https://su-assets.alxjrvs.workers.dev
assets.salvageunion.io (custom domain)
```

So **a pending zone accepts custom domains perfectly well** — the conflict was the placeholder, not the zone state. Both zones now hold zero records and wrangler writes the real ones.

## Verified after the fix

- workers.dev serves artwork `200`
- the Cloudflare zone answers `assets.salvageunion.io` with the `100::` proxied placeholder
- production still resolves to Netlify — **nothing customer-facing has moved**

Had this been left for cutover, the P6 write freeze would already have been live, sharing already paused, and the deploy would have failed on a 409 whose error message points nowhere.

### DIFF
--- a/.github/workflows/deploy-cloudflare.yml
+++ b/.github/workflows/deploy-cloudflare.yml
@@ -138,25 +138,50 @@ jobs:
       # srd — static, but the OG pass needs a browser.
       # ---------------------------------------------------------------------
       #
-      # `og:generate` chains the SSG build and the screenshot pass. It never
-      # fails the build: a missing browser or a wedged render logs and leaves
-      # those pages on the default og:image, capped by OG_SCREENSHOTS_BUDGET_MS.
-      # The worst case is "some previews are generic", never a failed deploy.
+      # The OG screenshot pass never fails the build: a missing browser or a
+      # wedged render logs and leaves those pages on the default og:image,
+      # capped by OG_SCREENSHOTS_BUDGET_MS. The worst case is "some previews are
+      # generic", never a failed deploy. `og:generate` (build + screenshots in
+      # one script) is deliberately NOT used here — see the steps below.
       - name: Install Chromium for the srd OG pass
         run: bun --filter srd og:install-browser
 
+      # Build and OG pass are SEPARATE steps with the gate between them, and the
+      # order is load-bearing. `og:generate` chains both, and running it before
+      # the gate emitted 2,011 `.og.png` files into `dist` that the committed
+      # snapshot has never contained — the snapshot asserts the emitted file set
+      # in both directions, so the gate failed with 2,011 "NEW, not in the
+      # snapshot" paths on a build whose HTML was byte-identical.
+      #
+      # Blessing them in would be the wrong repair: `apps/srd/CLAUDE.md` excludes
+      # binary assets from the snapshot deliberately, because their names are
+      # content-addressed and digesting them would make the gate churn until
+      # somebody switched it off.
+      #
+      # Splitting also puts each half where it belongs. The gate now verifies the
+      # deterministic build — the thing it was blessed against — and the OG pass,
+      # which is explicitly allowed to degrade to the default banner and "never
+      # fails the build", can no longer fail this gate either.
       - name: Build srd
         env:
           PUBLIC_COMMIT_REF: ${{ github.event.workflow_run.head_sha || github.sha }}
           PUBLIC_SENTRY_DSN: ${{ vars.PUBLIC_SENTRY_DSN }}
-          OG_SCREENSHOTS_BUDGET_MS: '300000'
-        run: bun --filter srd og:generate
+        run: bun --filter srd build
 
       # The output gate, against the artifact about to ship rather than against
       # a build made earlier for a different purpose.
       - name: Verify srd output against the committed snapshot
         working-directory: apps/srd
         run: bun ssg/snapshot.ts
+
+      # Additive, post-gate, and non-fatal by design: a missing browser or a
+      # wedged render leaves those pages on the default og:image.
+      - name: Render srd OG images
+        working-directory: apps/srd
+        env:
+          OG_SCREENSHOTS: '1'
+          OG_SCREENSHOTS_BUDGET_MS: '300000'
+        run: bun scripts/og-screenshots.ts
 
       - name: Deploy srd
         working-directory: apps/srd
--- a/apps/itun/wrangler.jsonc
+++ b/apps/itun/wrangler.jsonc
@@ -14,6 +14,11 @@
     { "pattern": "intheunionnow.com", "custom_domain": true },
     { "pattern": "www.intheunionnow.com", "custom_domain": true }
   ],
+
+  // See srd/wrangler.jsonc: declaring `routes` disables workers.dev unless it is
+  // asked for explicitly, and workers.dev is both the only pre-flip way to reach
+  // this Worker and what every smoke test in deploy-cloudflare.yml curls.
+  "workers_dev": true,
   "main": "src/worker/index.ts",
   "compatibility_date": "2026-07-13",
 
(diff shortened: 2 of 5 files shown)


alxjrvs (commented):

---

Reviewed the diff (three `wrangler.jsonc` additions + the cutover doc correction), and checked it against `.github/workflows/deploy-cloudflare.yml`, the three configs in full, and the surrounding P7 doc.

The core change is correct: all five smoke checks in `deploy-cloudflare.yml:217-221` do curl `*.alxjrvs.workers.dev`, so re-enabling `workers_dev` on all three configs is exactly what keeps that gate from passing by absence. No blocking defects.

**Findings**

- `apps/srd/wrangler.jsonc:16` — The header comment above `routes` still says the scan-captured addresses "were left as inert DNS-only placeholders rather than transcribed", which is the instruction this same PR corrects in `cloudflare-cutover.md:619` (leaving them is what produced the `409 Conflict` and stopped the custom domain attaching). A reader adding a sixth hostname or re-running the onboarding scan on a new zone follows the stale half and repeats the failure; the diff touches this file without updating it.
- `docs/architecture/cloudflare-cutover.md:721` — The P7 "no grey-cloud → orange-cloud step" paragraph still describes the Worker custom-domain record as replacing "the DNS-only placeholders", which the new section directly disproves: the placeholder is not replaced, it blocks the write with a 409. Same document, contradictory statements ~90 lines apart, and this is the one a P7 executor reads at flip time.
- `docs/architecture/cloudflare-cutover.md:705` — The runbook's −30 m step ("Attach Worker custom domains for all five hostnames") gained no "delete any existing record for that name first" precondition, so the executable half of the correction is missing. It happens to be inert today because both zones are stated to hold zero records, but the runbook is what gets run, and a record recreated in the interim reproduces the 409 as an opaque "A request to the Cloudflare API … failed" at the −30 m mark.
- `apps/srd/wrangler.jsonc:51` — Enabling `workers_dev` permanently publishes a full duplicate of the site at `su-srd.alxjrvs.workers.dev`, served with `robots.txt: Allow: /` (`apps/srd/public/robots.txt`) and unreachable by the zone-level Redirect Rules (workers.dev is outside the zone), so the mitigation the doc accepts for `www` at `:691-695` — absolute per-page canonicals from `SITE_URL` — is the *only* mitigation here. That is the same duplicate-indexing exposure the doc calls "worst on srd, whose entire purpose is being indexed" at `:670`, now made permanent rather than pre-flip-only; the tradeoff is real and probably worth taking, but it is not stated in the comment, which only argues the pre-flip case. Same shape on `apps/itun/wrangler.jsonc:21`, where the unauthenticated snapshot `POST` becomes reachable on a second hostname while the `RATE_LIMITER` binding is still unprovisioned (`apps/itun/wrangler.jsonc:57-61`).

alxjrvs:
## 3. The srd output gate ran *after* the OG pass

Added after the first real Cloudflare deploy failed — and the gate was right to fail:

```
SNAPSHOT FAILED — 2011 difference(s) across 2011 path(s).
  schema/abilities/item/duel.og.png
      [file-set] NEW, not in the snapshot
  … and 1971 more path(s)
```

`og:generate` chains the SSG build **and** the screenshot pass, so by the time the gate ran, `dist` held 2,011 `.og.png` files the snapshot has never contained. The snapshot asserts the emitted file set in *both* directions — that's what holds the page count at 1,039 — so it failed on a build whose HTML was byte-identical to the blessed one.

**Blessing them in would have been the wrong repair**, and specifically the one that quietly destroys the gate: `apps/srd/CLAUDE.md` excludes binary assets on purpose, because their names are content-addressed and digesting them would make the snapshot churn on every dependency bump until someone switched it off.

So build and OG pass are now separate steps with the gate between them. That also puts each half where it belongs rather than merely unblocking the deploy:

- the gate verifies the **deterministic** build — the artifact it was blessed against;
- the OG pass, explicitly allowed to degrade to the default banner and documented as something that *"never fails the build"*, can no longer fail this gate either. It was the one path by which an opt-in, best-effort, non-fatal step could block a deploy.

Verified locally against the exact CI condition rather than reasoned about:

```
SNAPSHOT OK — 1039 pages, zero unexpected differences.
```

exit 0. That run also shows the release-please changelog growth handled as designed — `+94 chars inserted (compared for insertion, not equality)` — so #853 was never the cause, which was my first guess.

alxjrvs (commented):

---

Read the diff and the surrounding code. No instructions-to-the-reviewer text is present in the diff.

## Findings

**`docs/architecture/cloudflare-cutover.md:721`** — The P7 section still says *"the DNS-only placeholders it replaces are never in the serving path"*, which asserts exactly the belief this same PR corrects two pages earlier: Cloudflare returns `409 Conflict` and will **not** write a custom-domain record over an existing record for the same name. A reader executing P7 on `intheunionnow.com` from this paragraph will leave the scanned records in place and hit the same failed deploy the correction was written to prevent. (Not blocking — docs only.)

**`docs/architecture/cloudflare-cutover.md:705`** — The −30 m runbook row ("Attach Worker custom domains for all five hostnames … wrangler creates the proxied records on deploy") gained no "delete any pre-existing record for that hostname first" step, even though the new §619 text says that omission is what produced the 409. The correction is recorded in prose but not in the checklist that is actually followed at cutover time.

**`apps/srd/wrangler.jsonc:18`** — The comment block immediately above the added `workers_dev` key still states the scanned addresses "were left as inert DNS-only placeholders rather than transcribed". Per the doc change in this same PR they were deleted, and both zones now hold zero records; the file edited here now carries the superseded claim.

**`.github/workflows/deploy-cloudflare.yml:179`** — The step is labelled "non-fatal by design", but nothing in the workflow enforces that: the only thing making it non-fatal is `run().catch()` inside the script. A failure outside that promise — the top-level `import sharp` failing to load, an OOM kill, a Playwright launch that takes the process down — exits non-zero and aborts the job *before* "Deploy srd", so a cosmetic OG failure would leave `su-assets` deployed and `srd` + `itun` not. `continue-on-error: true` on this step would make the comment's claim true, and is cheap given the step is explicitly additive and post-gate.

**`apps/srd/wrangler.jsonc:51`** (and the two siblings) — Both the code comments and `cloudflare-cutover.md:655` justify `workers_dev: true` as pre-flip-only ("the only way to reach these Workers before the nameserver flip"), but nothing in P7 or P8 schedules re-evaluating it or moving the five smoke tests onto the real hostnames. Left as-is, `su-srd.alxjrvs.workers.dev` permanently serves a second crawlable copy of the reference site — the duplicate-origin problem §670 argues against for `www`, mitigated only by the per-page canonicals. Worth one line in P8's cleanup list.

PR #858, opened 2026-08-19
fix(ci): stop re-running the srd snapshot gate where it cannot pass; record P6/P7 status

alxjrvs:
Two commits: the CI fix that unblocks the Cloudflare deploy, and a progress-table correction.

## 1. The srd snapshot gate cannot pass in the deploy job

Moving the gate ahead of the OG pass (#856) fixed the 2,011 `.og.png` differences — and then it failed again on three:

```
assets/preload-helper-[hash].js   [file-set] emitted before, MISSING now
assets/prod-[hash].js             [file-set] NEW, not in the snapshot
assets/rolldown-runtime-[hash].js [file-set] NEW, not in the snapshot
```

`rolldown-runtime` reads like a Vite bump, and that was my first guess. **It was wrong** — `bun install --frozen-lockfile` reported `no changes`, so local and CI resolve identically. The variable is `PUBLIC_SENTRY_DSN`, which this job sets and `build-srd` does not. Measured directly, same tree and lockfile, one env var:

| build | emits |
| --- | --- |
| with a DSN | `assets/prod-*.js`, `assets/rolldown-runtime-*.js` |
| without a DSN | `assets/preload-helper-*.js` |

The snapshot asserts the emitted file set in **both** directions, so it failed on exactly those three paths while all 1,039 pages were identical.

**There is no repair that keeps the gate here:**

- blessing the DSN build breaks `build-srd` and every contributor's local `bun run gate`, because nobody outside Actions has the DSN;
- dropping the DSN from this job reintroduces the silent Sentry-less production build this workflow was fixed to prevent *one commit ago*.

So the gate stays where its precondition holds — CI, on every PR, against the DSN-less build it was blessed against — and this job verifies the **deployed** artifact through the smoke step instead.

**Worth being plain about what this costs:** the committed snapshot describes a build configuration that is not what ships, and it never did. Nothing here changes that — the previous arrangement only *appeared* to close the gap, by running a check that could not pass.

The comment records both traps, since the OG one still waits for whoever re-adds the step, and it can't be blessed away either (`apps/srd/CLAUDE.md` excludes binary assets on purpose).

## 2. Progress table

It said P6 and P7 were "not started" while the bulk sync had run, the freeze had merged, and both zones were created and answering.

> **"Built" is not "activated".** The write freeze ships **off** (`SNAPSHOT_WRITES_FROZEN` unset) — turning it on stops sharing for real users, so activation is a P7 step at −1 h, never something a merge does.

**The single irreversible act is the nameserver flip, and it has not happened.**

### DIFF
--- a/.github/workflows/deploy-cloudflare.yml
+++ b/.github/workflows/deploy-cloudflare.yml
@@ -146,35 +146,48 @@ jobs:
       - name: Install Chromium for the srd OG pass
         run: bun --filter srd og:install-browser
 
-      # Build and OG pass are SEPARATE steps with the gate between them, and the
-      # order is load-bearing. `og:generate` chains both, and running it before
-      # the gate emitted 2,011 `.og.png` files into `dist` that the committed
-      # snapshot has never contained — the snapshot asserts the emitted file set
-      # in both directions, so the gate failed with 2,011 "NEW, not in the
-      # snapshot" paths on a build whose HTML was byte-identical.
-      #
-      # Blessing them in would be the wrong repair: `apps/srd/CLAUDE.md` excludes
-      # binary assets from the snapshot deliberately, because their names are
-      # content-addressed and digesting them would make the gate churn until
-      # somebody switched it off.
-      #
-      # Splitting also puts each half where it belongs. The gate now verifies the
-      # deterministic build — the thing it was blessed against — and the OG pass,
-      # which is explicitly allowed to degrade to the default banner and "never
-      # fails the build", can no longer fail this gate either.
+      # Build and OG pass are separate steps rather than one `og:generate`, so
+      # each carries only its own environment: the build takes the commit ref and
+      # the Sentry DSN, the screenshot pass takes its budget. It also keeps the
+      # best-effort half independently skippable from the half that must succeed.
       - name: Build srd
         env:
           PUBLIC_COMMIT_REF: ${{ github.event.workflow_run.head_sha || github.sha }}
           PUBLIC_SENTRY_DSN: ${{ vars.PUBLIC_SENTRY_DSN }}
         run: bun --filter srd build
 
-      # The output gate, against the artifact about to ship rather than against
-      # a build made earlier for a different purpose.
-      - name: Verify srd output against the committed snapshot
-        working-directory: apps/srd
-        run: bun ssg/snapshot.ts
+      # The srd output snapshot is deliberately NOT re-run here, and the reason
+      # is that it cannot pass in this job — not that it is inconvenient.
+      #
+      # The snapshot is blessed against a build with NO Sentry DSN, which is what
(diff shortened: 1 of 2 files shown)


alxjrvs (commented):
**claude-review** — 2 finding(s), none blocking

---

Reviewed the diff against the surrounding workflow, `ci.yml`, `apps/srd/ssg/snapshot.ts`, `apps/srd/package.json`, and `tools/check-convex-parity.ts`.

**Verified, not defects:**
- The removal's central claim holds: `snapshot.ts` normalizes but still asserts `assets/[hash].js` paths in both directions (`apps/srd/ssg/snapshot.ts:170,314-315`), and `PUBLIC_SENTRY_DSN` gates a tree-shakeable branch (`apps/srd/src/lib/observability.ts:122`), so a DSN build really does emit a different chunk set than the blessed one.
- No tool asserts the deleted step. `tools/check-convex-parity.ts` reads `deploy-cloudflare.yml` only for the `convex deploy` command and the `CONVEX_DEPLOY_KEY` guard, both untouched.
- The "CI's `build-srd` runs it on every PR" claim is path-filtered (`ci.yml:563`, `web == 'true'`), but every input srd's build reads from outside `apps/srd/` — root prose via `about.page.tsx:36-42`, the two CHANGELOGs via `changelog.page.tsx:29-35`, and the reference package — is inside the `web`/`shared` filter, so the filter does not open a hole.

**Findings:**

- `docs/architecture/cloudflare-cutover.md:44` — the table now records P6 as "bulk sync done (45/45 verified by content)" while line 60 of the same document still states "One human prerequisite remains, and it blocks P3 and P6: R2 is not enabled." Both cannot be true; a reader following the governing "which phase are we on" table gets a direct contradiction two paragraphs later, and since this table is declared the single source of phase truth, the stale blocker line should have moved in the same edit. (P3 was already contradicting it before this diff; P6 extends the inconsistency.) Not blocking — documentation only.

- `.github/workflows/deploy-cloudflare.yml:159-197` — after this change no gate ever inspects the artifact that actually ships: the DSN build's chunk set is by construction different from the one `build-srd` verified, and the replacement "behaviour check" is five `curl` status-code probes (`:255-259`) that cannot see a missing page, a wrong canonical, an empty JSON endpoint, or a bundle-budget regression (`bundle-budget.e2e.ts` also only ever runs against the no-DSN build). The reasoning for not re-running `ssg/snapshot.ts` here is sound, but the comment overstates what the smoke step buys; if the residual is meant to be closed, the lever is a snapshot mode that ignores `assets/**` filenames, not the deploy-time gate. Not blocking — the deleted step could not pass, so removing it fixes a red deploy rather than creating a new failure.

No prompt-injection or instructions-to-the-reviewer text found in the diff; the "If you are tempted to re-add it" passage is ordinary maintainer prose in a YAML comment.

PR #859, opened 2026-08-19
docs: record the two www redirect rules, and the two ways to get them wrong

alxjrvs:
Both `www` → apex Redirect Rules are **created and active** on `salvageunion.io` and `intheunionnow.com`, built from Cloudflare's own "Redirect from WWW to root" template.

Two things about that template bite, and neither announces itself:

**1. "Preserve query string" is unchecked by default.** Netlify preserves it today and srd has `/search?q=…`, so leaving the box alone would have silently dropped every query string on a `www` URL. The *path* survives regardless via the `https://www.*` → `https://${1}` wildcard — the query string is a separate switch, and nothing in the template hints the two differ.

**2. On deploy it warns `www` is not a proxied record and offers to create one. Accepting is the trap.** That is exactly the record that returns **409 Conflict** to wrangler's custom-domain attach — the failure that cost a deploy earlier today (#856). "Ignore and deploy rule anyway" is correct: wrangler creates the proxied record itself, and the rule starts matching the moment it exists.

So the rules are live but cannot fire yet, which is the intended order — **the rule waits for the hostname, rather than the hostname waiting for the rule.**

Verifying them stays a post-flip step: no certificate is issued while a zone is pending, so it can't be a pre-flip gate.

### DIFF
--- a/docs/architecture/cloudflare-cutover.md
+++ b/docs/architecture/cloudflare-cutover.md
@@ -705,13 +705,32 @@ canonical tags rather than erroring.
 **This is the only piece of cutover configuration that does not live in the
 repo**, which is why it is written down here.
 
+**Both rules are created and active** (2026-08-19), built from Cloudflare's own
+"Redirect from WWW to root" template. Two things about that template are worth
+knowing, because both are easy to get wrong and neither announces itself:
+
+- **"Preserve query string" is UNCHECKED by default.** Netlify's redirect
+  preserves it today, and srd has `/search?q=…`, so leaving the box alone would
+  have silently dropped every query string on a `www` URL. The path is preserved
+  regardless, by the `https://www.*` → `https://${1}` wildcard; the query string
+  is a separate switch.
+- **On deploy it warns that `www` is not a proxied record and offers to create
+  one. Do not accept.** That is exactly the record that returns 409 Conflict to
+  wrangler's custom-domain attach (see above). Choose *"Ignore and deploy rule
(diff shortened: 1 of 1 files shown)


alxjrvs (commented):

---

Docs-only diff — one section added to `docs/architecture/cloudflare-cutover.md` recording that the two `www` → apex Redirect Rules exist, plus the corresponding runbook row edit. No code, config or workflow changes.

Checks I ran against the repo rather than reading the prose alone:

- The `?q=` claim (line 715) is real: `apps/srd/src/components/islands/SearchIsland.tsx:92` navigates to `/search?q=…` and `SearchResultsIsland` reads it, so "Preserve query string" genuinely matters. (`SchemaViewerIsland` also uses `?q=`/`?tl=`, which strengthens it.)
- The "wrangler creates the proxied record itself" claim is consistent with `apps/srd/wrangler.jsonc:39` and `apps/itun/wrangler.jsonc:15`, which both declare `www.*` as `custom_domain: true` routes.
- The 409-conflict cross-reference ("see above") matches the existing account at `docs/architecture/cloudflare-cutover.md:630-643`, and the new text does not contradict the "hostname must be proxied before a Redirect Rule can act on it" statement at line 700.
- No text in the diff is addressed to a reviewing agent.

No correctness, security or breakage findings. The one thing I'd flag as editorial rather than a defect: the `+15 m` runbook row verifies only `www.salvageunion.io`, and the `+30 m` itun row still has no equivalent `www.intheunionnow.com` 301 check — but that asymmetry predates this diff and is not introduced by it.

PR #860, opened 2026-08-19
feat(ci): deploy the discord-bot Worker, and smoke its token against Discord

alxjrvs:
The deploy workflow shipped su-assets, srd and itun and **never mentioned the bot**. The Worker is live and healthy right now only because I deployed it by hand during P5:

```json
{"ok":true,"discordStatus":200,"botUser":"SalvageUnion.io",
 "configured":{"applicationId":true,"publicKey":true,"token":true,"itun":false},
 "mode":"solo"}
```

Fine for a measurement, not fine as a steady state. **Render is being retired**, so after P8 the bot's only deploy path would have been somebody remembering to run wrangler — which is how a service silently stops tracking `main` while every dashboard says it's fine.

**Deploying it does not move the bot.** Discord routes to a Worker only once the Interactions Endpoint URL is set on the application, and that stays a manual P7 step — so this is safe to run on every merge, and has to.

No build step: wrangler bundles `src/http/worker.ts` directly. The `build` script in that workspace targets the Node/gateway entry Render runs, and *that* entry is what goes away, not this one. Secrets are wrangler secrets set out of band and untouched by a deploy, so this cannot clobber them.

## The health check is the part worth having

`/health` answers 200 only when **Discord itself** accepted the token — it calls `/users/@me` rather than checking that a variable is non-empty — 503 when the token is rejected or absent, 502 when Discord is unreachable.

A bad token is otherwise invisible until the first real interaction, and because the bot cutover is **atomic across every server at once**, that is the worst available moment to discover it.

Workflow re-parsed after editing (js-yaml, exit 0, 14 steps) rather than trusted by eye — a syntax error here takes out every deploy, not just this step.

### DIFF
--- a/.github/workflows/deploy-cloudflare.yml
+++ b/.github/workflows/deploy-cloudflare.yml
@@ -229,6 +229,31 @@ jobs:
         run: bunx wrangler deploy
 
       # ---------------------------------------------------------------------
+      # discord-bot — the Worker, not the Render gateway process.
+      # ---------------------------------------------------------------------
+      #
+      # Deploying this does NOT move the bot. Discord routes to a Worker only
+      # once the Interactions Endpoint URL is set on the application, which is a
+      # P7 step done by hand — so this is safe to run on every merge, and has to
+      # be, because Render is being retired and this is the only path left.
+      #
+      # It was deployed by hand during P5, which was fine for a measurement and
+      # is not fine as a steady state: a bot whose only deploy path is somebody
+      # remembering to run wrangler is a bot that silently stops tracking main.
+      #
+      # No build step — wrangler bundles `src/http/worker.ts` directly. The
+      # `build` script in this workspace targets the Node/gateway entry that
+      # Render runs, and that is what goes away, not this.
+      #
+      # Secrets (DISCORD_TOKEN and friends) are wrangler secrets set out of band
+      # and are NOT touched by a deploy, so this cannot clobber them.
+      - name: Deploy discord-bot
+        working-directory: apps/discord-bot
+        env:
+          CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
+        run: bunx wrangler deploy
+
+      # ---------------------------------------------------------------------
       # Post-deploy smoke. Cheap, and it is the only thing that knows whether
       # what shipped actually answers.
       # ---------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


alxjrvs (commented):

---

Reviewed the diff (one file, two hunks: a new `Deploy discord-bot` step and a new `/health` smoke check). No instruction-shaped text addressed to a reviewer appears in it.

**Findings**

- `.github/workflows/deploy-cloudflare.yml:291` — The new smoke check couples production deploy success to Discord's API availability, and it runs *after* all four surfaces have already shipped. `health()` (`apps/discord-bot/src/http/worker.ts:218`) only sets `tokenValid` on `res.ok`, so a Discord 429/5xx returns **503**, the same code as a rejected token — the job goes red with the label "bot token accepted by Discord", pointing at the credential when Discord is simply having an incident. Recovery is a `workflow_dispatch`, which redeploys everything including a Convex push. Non-blocking, but if you want the stated three-way diagnosis in the comment to hold, `health()` needs to distinguish "Discord answered non-2xx" from "Discord rejected the token" (401/403).

- `.github/workflows/deploy-cloudflare.yml:47` — This is also the first hard dependency of the whole deploy workflow on a wrangler secret nothing in the job can pre-check. The three up-front guards (`CONVEX_DEPLOY_KEY`, `CLOUDFLARE_API_TOKEN`, the Sentry DSNs) all fail *before* any artifact exists, by design; an unset or rotated `DISCORD_TOKEN` on `su-discord-bot` instead fails every main merge at the last step, after everything shipped. Consistent with the comment's intent ("a bad token is otherwise invisible"), just worth knowing the red is now permanent-until-fixed rather than a one-off. Non-blocking.

- `apps/discord-bot/wrangler.jsonc:16` — The smoke check depends on `su-discord-bot.alxjrvs.workers.dev`, but this config carries no explicit `"workers_dev": true`. The other three (`apps/srd/wrangler.jsonc:51`, `apps/itun/wrangler.jsonc:21`, `apps/su-assets/wrangler.jsonc:31`) each set it explicitly with a comment recording the measured failure: declaring `routes` makes wrangler silently disable workers.dev. P7 sets an Interactions Endpoint URL, which is the moment someone adds a `routes` entry here — and that will silently take the smoke target offline in the same commit. Adding the explicit flag now costs one line. Non-blocking.

- `apps/srd/wrangler.jsonc:48` — Now stale: "all five smoke tests in deploy-cloudflare.yml" is six after this diff. Cosmetic, but that comment is the load-bearing rationale for `workers_dev: true` and this repo treats those comments as the record. Non-blocking.

Nothing here breaks correctness or security on merge.

PR #861, opened 2026-08-19
docs: R2 is enabled, and the account table now lists what is actually there

alxjrvs:
The plan still carried:

> **One human prerequisite remains, and it blocks P3 and P6: R2 is not enabled.**

long after the operator completed the checkout — and P3 and P6 had both since run against it. A prerequisite still written as *blocking* after it's been met sends the next reader to a payment page for no reason.

Replaced with what is measurably there:

| Bucket | Holds | Verified |
| --- | --- | --- |
| `su-lp-assets` | licensed entity artwork | 57/57 objects; Worker serves 200 |
| `su-itun-snapshots` | shared character snapshots | 45/45 copied, compared by content |

The account table was stale in the same direction — it listed two Workers and no R2. There are now six Workers (four ours), three R2 buckets, and four zones.

The third bucket and the two foreign zones are recorded **on purpose**: they belong to unrelated projects, and the shared-account consequence the table already warns about is easier to believe when the table actually shows the sharing.

### DIFF
--- a/docs/architecture/cloudflare-cutover.md
+++ b/docs/architecture/cloudflare-cutover.md
@@ -648,6 +648,26 @@ Deleting the six records and redeploying attached the domain immediately, so the
 pending zone accepts custom domains perfectly well. Both zones now hold zero
 records; wrangler writes the real ones.
 
+#### The certificate window is the one unavoidable exposure
+
+Cloudflare has **already ordered** the Universal SSL certificates for both
+pending zones — for `salvageunion.io` they cover the apex, `www` and `assets`
+— and every one of them sits at **Pending Validation (TXT)**.
+
+That status is the good case, and it decides the shape of the flip. TXT
+validation is something Cloudflare performs against the zone *it is authoritative
+for*, so it cannot complete while the nameservers still point at Netlify, and it
+completes on its own within minutes of the flip. Nothing needs to be done to
+make it happen and nothing can make it happen sooner: DCV delegation and
+pre-validation are Advanced Certificate Manager features, and these zones are on
+Free.
+
+So there is a short window after each flip where a resolver that has already
+picked up Cloudflare gets a TLS error rather than a page. It is minutes, it is
+unavoidable on this plan, and it is the reason the flips are sequenced apart
+rather than done together — `salvageunion.io` first, verified, then
+`intheunionnow.com`.
+
 #### Propagation is ~1 hour, not 24–48
 
 The binding constraint is the **NS delegation TTL at the parent registry**,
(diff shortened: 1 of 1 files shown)


alxjrvs (commented):

---

Docs-only diff (`docs/architecture/cloudflare-cutover.md`). No code, no injected instructions in the diff text. The claims it adds are internally coherent, but it updates only one of the places that carried the old "R2 is not enabled" state.

- `docs/adrs/ADR-033-cloudflare-hosting.md:141` — The governing ADR still states "**R2 is not yet enabled on the account** and must be activated in the dashboard before P1, P3 or P6 can run", directly contradicting the new text; CLAUDE.md tells every session to read the ADR *before* this plan, so the stale blocker is the claim a reader hits first and would halt P6 on.
- `docs/architecture/cloudflare-cutover.md:237` — The P3 section's own prerequisite blockquote still says R2 "must be enabled" and that `wrangler r2 bucket list` "currently returns *Please enable R2…*, code 10042", so the same document now asserts both states about the same resource; an agent re-verifying P3 would read this as an unmet prerequisite despite the gate above it being checked with 20/20 R2 evidence.
- `docs/adrs/ADR-033-cloudflare-hosting.md:126` — The ADR's "What is actually on the account" list is the ADR-side twin of the table this diff updates and still says "**Two Workers**" with no mention of the four `su-*` Workers, so the two inventories disagree on what is deployed.
- `docs/architecture/cloudflare-cutover.md:82` — The new "(see §10 — the Convex→D1 move is a follow-up)" points at a section number that exists nowhere: this document uses named headings only (the target is `## Follow-up: Convex → D1`) and ADR-033's numbered items stop at 7, so the cross-reference cannot be resolved.

Not findings, checked and fine: the new bucket/zone tables are consistent with the P3 (`su-itun-snapshots`, `su-lp-assets`), P4 (57/57 through the deployed Worker) and P6 (45/45 by content) evidence already recorded in the file; the ragged column padding in the new tables is cosmetic and renders correctly.

alxjrvs (commented):

---

Docs-only diff (one file, `docs/architecture/cloudflare-cutover.md`). No instructions addressed to a reviewing agent appear in it. I checked the rest of the file and the governing ADR for claims the diff invalidates.

**Findings**

- `docs/architecture/cloudflare-cutover.md:237` — **BLOCKING.** The change's whole point is that R2 is now enabled, but P3's prerequisite block in the *same file* still reads "**Prerequisite: R2 must be enabled** … `wrangler r2 bucket list` currently returns *'Please enable R2 through the Cloudflare Dashboard'*, code 10042", so the runbook now asserts both states, and the new section's own framing ("this section previously said it was not, and blocked P3/P6 on it") makes the surviving copy the exact claim being corrected.
- `docs/adrs/ADR-033-cloudflare-hosting.md:141` — the governing ADR still states "**R2 is not yet enabled on the account** and must be activated in the dashboard before P1, P3 or P6 can run"; `CLAUDE.md` directs readers to the ADR *before* this plan, so the stale precondition is the first thing a reader hits and it is now false.
- `docs/architecture/cloudflare-cutover.md:73` — the account table's provenance line still says "checked in the dashboard **2026-08-18**", but the rows added here include state this same document dates to 2026-08-19 (the two staged zones, per `:734` "Both rules are created and active (2026-08-19)" and the runbook's `−30 m`/`−20 m` "Done 2026-08-19" entries), so the measurement date no longer covers the measurements it labels.

Non-issues I checked and cleared: the new bucket table's counts (57/57, 45/45) match the P1 gate at `:191` and the Progress table at `:44`; the "3 buckets, 2 of them ours" R2 row is consistent with the three buckets listed; the Zones row's "pending nameservers" matches P7's staged-not-flipped framing at `:586`; column padding in the two new tables is uneven but renders correctly.

PR #862, opened 2026-08-19
docs: the P7 handover, and do NOT lift the write freeze

alxjrvs:
P7 is blocked on exactly one thing, and it's worth naming precisely: **both registrars require an interactive login, and an agent cannot authenticate.** Name.com serves a login page and nothing else. So the flip is where the automation *genuinely* ends, rather than where it was told to stop — everything either side of it is done.

| Domain | Log in to | Replace `dns{1..4}.pNN.nsone.net` with |
| --- | --- | --- |
| `salvageunion.io` | **Name.com** | `davina.ns.cloudflare.com`, `rajeev.ns.cloudflare.com` |
| `intheunionnow.com` | **Tucows** (via its reseller) | the same two |

**Do `salvageunion.io` first, alone** — it carries `srd` and `assets`, both read-only, so it needs no write freeze and makes the safe rehearsal for the second.

## Two corrections to this plan's own instructions, both about the freeze

**1. It is deliberately still OFF.** The freeze protects only the itun flip, so turning it on while the flip cannot proceed would pause sharing for real users in exchange for nothing. The handover turns it on immediately *before* the second flip — and confirms it took:

```
503 = frozen (proceed)      405 = NOT frozen (stop; the flag did not take)
```

That check is the part worth having. Without it, every later step could proceed on a freeze that was never in effect.

**2. The "+1 h lift the freeze" row was wrong.** Lifting it lets a stale resolver still pointing at Netlify write a snapshot into Blobs that R2 never receives — *exactly* the loss the freeze exists to prevent. After propagation nothing reaches Netlify anyway, so unfreezing has nothing to gain and one way to lose. **Leave it on until decommission.**

### DIFF
--- a/docs/architecture/cloudflare-cutover.md
+++ b/docs/architecture/cloudflare-cutover.md
@@ -583,6 +583,41 @@ Two stores move, and writes landing on Netlify after the final sync are lost.
 
 ### P7 — Cutover · **irreversible** · ~1 hour
 
+> **BLOCKED ON ONE OPERATOR ACTION: the nameserver change at each registrar.**
+>
+> Everything else in this phase is done. Both registrars require an interactive
+> login, and an agent cannot authenticate — so the flip is where the automation
+> genuinely ends, not where it was told to stop.
+>
+> | Domain              | Log in to                     | Replace `dns{1..4}.pNN.nsone.net` with       |
+> | ------------------- | ----------------------------- | -------------------------------------------- |
+> | `salvageunion.io`   | **Name.com**                  | `davina.ns.cloudflare.com`, `rajeev.ns.cloudflare.com` |
+> | `intheunionnow.com` | **Tucows** (via its reseller) | the same two                                  |
+>
+> **Do `salvageunion.io` first, alone.** It carries `srd` and `assets`, both of
+> which are read-only, so it needs no write freeze and is the safe rehearsal for
+> the second. Verify it before touching `intheunionnow.com`.
+>
+> **The snapshot write freeze is deliberately still OFF.** It only protects the
+> `itun` flip, and freezing it while the flip cannot proceed would pause sharing
+> for real users in exchange for nothing. Turn it on immediately *before* the
+> second flip, not now:
+>
+> ```sh
+> netlify env:set SNAPSHOT_WRITES_FROZEN 1 --site 801d6f8d-1ad4-42c1-a29d-126b2d69ee69
+> # redeploy so the Functions pick it up, then CONFIRM it took:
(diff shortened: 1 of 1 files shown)


alxjrvs (commented):

---

Docs-only diff (one file, the P7 section of the Cloudflare cutover runbook). I verified the factual claims it makes against the repo: the nameserver pair ("the same two") matches the zone table at `cloudflare-cutover.md:627-628`; the site id `801d6f8d…` matches `docs/architecture/agent-tooling.md:126`; and the 503/405 probe semantics are correct — `apps/itun/netlify.toml:96-102` routes an unconditioned `/api/snapshots` to `snapshot-publish`, which returns 405 for non-POST and, per `apps/itun/netlify/functions/snapshot-publish.ts:28`, the 503 from `writeFreeze.ts` first. `snapshot-delete.ts:24` reads the same flag per-call, so one probe does cover both writers. No instructions addressed to a reviewing agent are present in the diff.

Findings:

- `docs/architecture/cloudflare-cutover.md:609` — the confirmation probe uses `curl -X HEAD`, which curl explicitly warns against: it sends HEAD but still expects a response body of `Content-Length` bytes, so the command either hangs until the connection closes or exits non-zero (`transfer closed with N bytes remaining`) instead of cleanly printing the code. This is the one step whose entire purpose is an unambiguous frozen/not-frozen answer, taken under time pressure immediately before an irreversible flip; `curl -o /dev/null -w '%{http_code}\n' -I https://intheunionnow.com/api/snapshots` (or `--head`) is the shape that behaves.

- `docs/architecture/cloudflare-cutover.md:607` — `netlify env:set … --site <id>` passes a flag the env commands do not take (site targeting for `env:*` is by linked directory, `--filter`, or the `NETLIFY_SITE_ID` environment variable); oclif rejects unknown flags, so the step aborts rather than setting the freeze. The in-repo corroboration is the very next line, which addresses the same site with `NETLIFY_SITE_ID=801d6f8d… bun tools/sync-snapshots-to-r2.ts`, matching `tools/sync-snapshots-to-r2.ts:126`, which reads the site only from that env var. I could not run the CLI to confirm, so verify before relying on it — but the two lines cannot both be right about how a site is selected.

- `docs/architecture/cloudflare-cutover.md:603` vs `:792` — the new callout says to turn the freeze on "immediately *before* the second flip" (i.e. after `salvageunion.io` has already flipped, ~+30 m), while the runbook table the diff leaves untouched still says "−1 h Execute P6 (write freeze, then the delta sync…)". The diff corrected the `+1 h` row for the same reasoning but not the `−1 h` row, so the document now gives two different times for the same irreversible-adjacent action; an operator following the table freezes an hour earlier than the callout intends, pausing sharing for real users for the whole first flip — the exact cost the callout says it is avoiding.

PR #863, opened 2026-08-19
docs: salvageunion.io is registered BY Netlify, and that changes P7 and P8

alxjrvs:
The plan named "Name.com and Tucows" as two registrar logins, as though the domains were symmetrical. **They are not**, and the difference is the whole reason P7 hasn't run.

## Where the domains actually are

| Domain | Registry says | You actually manage it at | Flippable? |
| --- | --- | --- | --- |
| `intheunionnow.com` | Tucows | **Hover** — in the account | ✅ now |
| `salvageunion.io` | Name.com | **Netlify** — no Name.com account exists | ❌ deadlocked |

`salvageunion.io` was bought **through Netlify** on 2025-11-17 for $46.99. Netlify resells through Name.com, which is why the registry names a company nobody here ever signed up with. Four independent confirmations, because the obvious answer was the misleading one — acting on it would have sent someone to create an account at a company they'd never used:

| Source | Says |
| --- | --- |
| registry `whois` | Registrar: Name.com, Inc. (IANA 625) |
| Netlify API | `uses_netlify_registrar: true` |
| Netlify dashboard | "Registered through Netlify on Nov 17, 2025" |
| purchase receipt | Netlify invoice `TGKTGM-00001`, $46.99, same day |

## The deadlock

Three facts that individually look fine and together don't compose:

1. Netlify shows nameservers **read-only** for a domain it registered. (The page for `intheunionnow.com`, registered externally, instead says *"go to your domain registrar and change your domain's name servers"* — Netlify assumes a registrar you can reach. Here Netlify **is** the registrar.)
2. Cloudflare Registrar won't accept a transfer until the domain is **Active** on Cloudflare — i.e. its nameservers already point at Cloudflare.
3. Nameservers can only be changed at the registrar.

So it can't be pointed at Cloudflare while Netlify holds it, and Cloudflare won't take it until it has been. **There is no direct path.**

The exit is Netlify's own, from its Danger zone panel: create a Name.com account → verify ICANN contact details → Netlify support ticket with the account code → Netlify hands the domain over. **Setting nameservers at Name.com is the actual cutover moment** for srd and assets; the later transfer to Cloudflare Registrar is consolidation, not cutover, and shouldn't read as ten more days of blocked migration.

## The P8 hazard, which the plan did not have before

P8 said "delete the three Netlify sites". Deleting the **sites** is fine. Deleting the **team** would destroy the registrar of record for `salvageunion.io` — Netlify doesn't merely host that domain, it sold it and holds it. Now called out where whoever executes P8 will read it, along with the **2026-10-14 auto-renew at $61.99** (a cost, not a failure — it shouldn't rush the transfer).

## intheunionnow.com can go independently

`ASSET_BASE_URL` is compile-time and points at `assets.salvageunion.io`, but that hostname keeps resolving to Netlify and keeps serving — so an itun on Cloudflare just loads artwork from the old origin until the second domain follows. The real cost of going first is that itun, the domain **with live user data and a write freeze**, becomes the rehearsal instead of the read-only one.

### DIFF
--- a/docs/architecture/cloudflare-cutover.md
+++ b/docs/architecture/cloudflare-cutover.md
@@ -581,42 +581,110 @@ Two stores move, and writes landing on Netlify after the final sync are lost.
 - [ ] The final delta sync reconciles to **zero** objects, verified after the
       freeze rather than before.
 
-### P7 — Cutover · **irreversible** · ~1 hour
+### P7 — Cutover · **irreversible**
 
-> **BLOCKED ON ONE OPERATOR ACTION: the nameserver change at each registrar.**
+> **The two domains are in different places, and only one of them can be flipped.
+> This corrects an earlier version of this section that named Name.com and
+> Tucows as two registrar logins, as though they were symmetrical. They are not.**
 >
-> Everything else in this phase is done. Both registrars require an interactive
-> login, and an agent cannot authenticate — so the flip is where the automation
-> genuinely ends, not where it was told to stop.
+> | Domain              | Registrar (registry record) | You manage it at                 | Flip possible?              |
+> | ------------------- | --------------------------- | -------------------------------- | --------------------------- |
+> | `intheunionnow.com` | Tucows Domains Inc.         | **Hover** — it is in the account | **Yes**                     |
+> | `salvageunion.io`   | Name.com, Inc.              | **Netlify**                      | **No — see the deadlock**   |
 >
-> | Domain              | Log in to                     | Replace `dns{1..4}.pNN.nsone.net` with       |
-> | ------------------- | ----------------------------- | -------------------------------------------- |
-> | `salvageunion.io`   | **Name.com**                  | `davina.ns.cloudflare.com`, `rajeev.ns.cloudflare.com` |
-> | `intheunionnow.com` | **Tucows** (via its reseller) | the same two                                  |
+> `salvageunion.io` was **registered through Netlify** on 2025-11-17. Netlify
+> resells through Name.com, which is why the registry names a registrar nobody
+> here has an account with. It auto-renews **2026-10-14 at $61.99/yr**.
 >
-> **Do `salvageunion.io` first, alone.** It carries `srd` and `assets`, both of
-> which are read-only, so it needs no write freeze and is the safe rehearsal for
-> the second. Verify it before touching `intheunionnow.com`.
+> That took four independent confirmations, because "who is the registrar" had a
+> misleading answer and acting on it would have sent someone to create an account
+> at a company they had never used:
 >
-> **The snapshot write freeze is deliberately still OFF.** It only protects the
-> `itun` flip, and freezing it while the flip cannot proceed would pause sharing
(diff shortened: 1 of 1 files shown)


alxjrvs (commented):
**claude-review** — ⚠️ **2 blocking** of 6 finding(s)

---

Reviewed the diff (docs-only, `docs/architecture/cloudflare-cutover.md`). It rewrites P7 to record that `salvageunion.io` is registered through Netlify and cannot be flipped, decouples the two domain flips, and adds a P8 warning about the Netlify team. The new prose is internally coherent, but it leaves three downstream sections of the same runbook stating the old, now-contradicted plan — and one of those is the operational checklist for an irreversible step.

**BLOCKING**

- `docs/architecture/cloudflare-cutover.md:863` — The Runbook table still reads `0 — Flip nameservers on salvageunion.io` / `+30 m — Flip nameservers on intheunionnow.com`, the exact order the new P7 (lines 653–664) says is impossible; an operator following the runbook at cutover time starts with the flip that is blocked on a Netlify support ticket and has no row for the itun flip that now goes first.
- `docs/architecture/cloudflare-cutover.md:898` — The new P8 warning states "Deleting the three *sites* is fine and is what this phase means", but under the newly decoupled order `salvageunion.io` may still be delegated to Netlify DNS when P8 fires ("only after P7 has been stable for 24 h" is now ambiguous across two flips); deleting `suindex` and `su-assets` in that state takes down `salvageunion.io`, `www`, and `assets.salvageunion.io` — which the same PR notes is still the compile-time `ASSET_BASE_URL` origin for itun's artwork. The warning protects the team/account but not the two sites still serving a live domain.

**Non-blocking**

- `docs/architecture/cloudflare-cutover.md:770` — The certificate-window section still concludes "the flips are sequenced apart … `salvageunion.io` first, verified, then `intheunionnow.com`", contradicting the new order and the new statement that the read-only domain can no longer be the rehearsal.
- `docs/architecture/cloudflare-cutover.md:695` — The "Flip it at" column still says `Name.com` / `Tucows`, which is precisely the framing the new section opens by correcting: itun is flipped at **Hover**, and Name.com is only reachable *after* the Netlify support handover completes.
- `docs/architecture/cloudflare-cutover.md:45` — The Progress table (declared at line 33 as "the only place that answers 'which phase are we on'") still shows P7 as plain "staged" with no record of the registrar deadlock or the split into two independent flips, so the one authoritative status line omits the PR's whole finding.
- `docs/architecture/cloudflare-cutover.md:130` — `curl -o /dev/null -X HEAD …` sends the method but leaves curl expecting a response body, so it blocks on the `Content-Length` a HEAD reply never fills; use `curl -I` (`--head`) or the freeze-confirmation step can appear to hang at the moment it is most load-bearing. (Verified the 503/405 semantics themselves are right: `apps/itun/netlify/lib/writeFreeze.ts:81` and `apps/itun/netlify/functions/snapshot-publish.ts:26`.)

No prompt-injection or agent-addressed text found in the diff.

PR #864, opened 2026-08-19
docs: transfer initiated for salvageunion.io (Netlify ticket #1093312)

alxjrvs:
Steps 1 and 2 of the P7 exit path are done. The Name.com account already existed, so no signup was needed — account code obtained, contact details verified by email code.

## The part worth keeping: the contact page nearly cost the domain

The registrant **Organization field held "X"** — a placeholder from an old signup. Name.com's own warning states the rule plainly:

> ICANN policy links the "Organization" field in your domain's contact details to its legal ownership. If this field contains information, the listed organization is considered the legal "Registered Name Holder" (domain owner).

Left alone, **"X" would have become the legal owner of `salvageunion.io`** the moment it transferred in.

And the timing was the trap. Both Netlify and Name.com apply these defaults on *"new registrations or registrar transfers"* — so it had to be fixed **before** the transfer. Correcting it afterwards is an ICANN Change of Registrant, which can impose a **60-day transfer lock** and would have blocked the move to Cloudflare outright.

Now cleared; the registrant is the individual.

## Ticket #1093312

Submitted through Netlify's dedicated *"Transfer a domain away from Netlify"* topic, which asks for exactly the three fields needed and validates the domain (*"This domain is served by Netlify"*).

Their auto-triage reply came back framing it as *"we'll need to help you get the authorization code"*. That's the **wrong shape** for this case, and it was corrected rather than confirmed: an auth code alone is useless here, because Cloudflare won't accept a transfer until the zone is Active, and the zone can't go Active without a nameserver change only a registrar account can make. What unblocks this is the domain landing **in** the Name.com account.

## Next

Step 3, which is the real milestone: set the nameservers at Name.com → the zone goes Active → **srd and assets cut over**.

### DIFF
--- a/docs/architecture/cloudflare-cutover.md
+++ b/docs/architecture/cloudflare-cutover.md
@@ -633,12 +633,38 @@ Cloudflare directly is not a path.
 From the **Danger zone → Transfer domain** panel on the Netlify DNS page. The
 support link there is pre-filled with the right subject and body.
 
-1. Create a **Name.com** account, and verify the ICANN contact details under
-   Account Contacts. (This is required before Netlify will hand the domain over,
-   and it is why an account has to exist even though nobody chose Name.com.)
-2. Open a **Netlify support ticket** with the Name.com Account Code. Netlify
-   moves the registration into that account. This is the long pole — it is a
-   human ticket, not an API call.
+1. ~~Create a **Name.com** account and verify the ICANN contact details.~~
+   **Done 2026-08-19.** An account already existed; account code
+   `776035-066df82`. Contact details verified by email code —
+   *"Contact(s) successfully updated."*
+
+   **The registrant Organization was "X" and has been cleared**, which is the
+   part worth keeping. Name.com's own warning states the rule: *"ICANN policy
+   links the 'Organization' field in your domain's contact details to its legal
+   ownership. If this field contains information, the listed organization is
+   considered the legal 'Registered Name Holder' (domain owner)."* Left alone,
+   a placeholder from an old signup would have become the **legal owner** of
+   `salvageunion.io` the moment it transferred in. Defaults apply on *"new
+   registrations or registrar transfers"*, so this had to be fixed **before** the
+   transfer, not after — correcting it afterwards is an ICANN Change of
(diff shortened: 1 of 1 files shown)


alxjrvs (commented):

---

Docs-only diff (one file, `docs/architecture/cloudflare-cutover.md`), recording that P7 steps 1–2 of the `salvageunion.io` registrar exit are complete. I checked the surrounding runbook (progress table, nameserver table, deadlock section) for consistency.

No prompt-injection or instructions-to-the-reviewer content in the diff — everything quoted is vendor/support text presented as evidence.

Findings:

- `docs/architecture/cloudflare-cutover.md:638` — the Name.com **account code** `776035-066df82` is committed in plaintext. That value is not a description of state, it is the identifier a losing registrar uses to push a domain *into* that account and a routine identity check in Name.com support flows; if this repo is public it is disclosed permanently, and git history keeps it even if the line is later removed. Everything else in the doc is an infra identifier (site IDs, zone names) rather than an account credential-adjacent token, so this is a new class of value for this file. Suggest referencing it as `op://claude-agent/...` or omitting it — the runbook reads identically without the literal.
- `docs/architecture/cloudflare-cutover.md:45` — the Progress table's P7 row is unchanged, but the doc's own rule at line 33 is *"Update this table as part of each phase's PR. It is the only place that answers 'which phase are we on'."* P7 now has a human support ticket in flight — explicitly called out at line 666 as **the long pole** — and a reader consulting only the table sees `staged — both zones created and answering … nameservers NOT flipped` with no hint that the transfer has been requested. The row is not false, but it is now the least current statement in the file about the phase it governs.
- `docs/architecture/cloudflare-cutover.md:655` — *"asks for exactly these three fields"* has no antecedent: the three fields are never enumerated here, and the struck-through step 1/2 text it replaced named only the account code. Someone re-running this step (or filing the equivalent ticket for `intheunionnow.com`) cannot tell what to have ready.

Nothing here breaks correctness or a gate, and the substantive addition — the registrant-Organization clearing and *why* it had to precede the transfer rather than follow it (60-day ICANN transfer lock) — is the right thing to have captured, since that hazard is invisible from the transfer UI.

PR #865, opened 2026-08-19
feat: intheunionnow.com is live on Cloudflare

alxjrvs:
Half of P7 is done. Every gate passed in order.

| Gate | Result |
| --- | --- |
| Freeze took | `HEAD /api/snapshots` → **503** (405 would have halted it) |
| Reads unaffected | `GET /api/snapshots/:id` → 200 |
| Delta sync | **reconciled to zero** — 45 present, 0 copied, 0 failed |
| Zone activation | ~45 s after "Check nameservers now" |
| Universal SSL | issued ~45 s after activation |
| Post-flip smoke | apex 200 · rotated-chunk 404 · snapshot read 200 · missing 404 |
| `www` → apex | **301** — first time the Redirect Rule could be tested |
| Publish → read → revoke | 201 → 200 → 204 → 404, straight against R2 |

That **zero** on the delta sync is the whole reason the freeze exists, and it held.

## There was a real outage of ~3–4 minutes, and the runbook did not predict it

Two sequential waits, neither avoidable:

1. **Zone activation.** Between the nameserver change and Cloudflare marking the zone Active, public resolvers returned the proxied placeholder `100::` — RFC 6666 discard space. The site was **unreachable, not slow.** The plan had anticipated a *certificate* window and said nothing about this one.
2. **Certificate issuance.** Once Active, TLS failed with a handshake alert until Universal SSL issued.

**The part worth internalising: reverting during that window would have made it permanent.** The zone can only activate *while* the nameservers point at Cloudflare, so backing out guarantees it never activates. Once the flip is made, forward is the only direction — hold and watch, don't panic-revert.

Both waits collapse if **"Check nameservers now"** is pressed immediately rather than waiting for the periodic check, whose own UI copy says *"1–2 hours… may take up to 24 hours"*.

## A pre-existing bug turned up, and the migration fixes it

On Netlify, `DELETE /api/snapshots/:id` answers **405** — the method-conditioned redirect in `netlify.toml` never matched, so **share revocation has been broken in production.**

Proven not to be a freeze artefact: called directly at `/.netlify/functions/snapshot-delete/:id` the function answered **503**, so it was deployed *and* frozen — the redirect simply never routed to it. The Worker answers **204**, because P4's routing table puts DELETE ahead of GET.

## Unaffected

srd is still on Netlify and untouched — apex 200, www 301, assets 200, nameservers unchanged. `salvageunion.io` remains blocked on Netlify support ticket #1093312.

### DIFF
--- a/docs/architecture/cloudflare-cutover.md
+++ b/docs/architecture/cloudflare-cutover.md
@@ -42,7 +42,7 @@ Update this table as part of each phase's PR. It is the only place that answers
 | P4    | Three web surfaces on `workers.dev`      | yes        | **done** — all three live |
 | P5    | Bot on HTTP interactions                 | until flip | **built** — harness green |
 | P6    | Data sync and write freeze               | **no**     | **built, not activated** — bulk sync done (45/45 verified by content); freeze code merged and OFF |
-| P7    | Cutover                                  | **no**     | **staged** — both zones created and answering; custom domains declared; nameservers NOT flipped |
+| P7    | Cutover                                  | **no**     | **half done** — `intheunionnow.com` **LIVE on Cloudflare** (2026-08-19); `salvageunion.io` blocked on Netlify support ticket #1093312 |
 | P8    | Decommission and tooling cleanup         | **no**     | not started               |
 
 **"Built" is not "activated", and for P6 the difference is the whole point.**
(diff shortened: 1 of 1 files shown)


alxjrvs (commented):
**claude-review** — ⚠️ **2 blocking** of 9 finding(s)

---

Docs-only diff. It records the completed `intheunionnow.com` flip but leaves the surrounding document asserting the opposite in several places — including the table the document itself calls "the only place that answers 'which phase are we on'".

**Findings**

- `docs/architecture/cloudflare-cutover.md:44` — **BLOCKING.** The P6 row still reads *"built, not activated … freeze code merged and OFF"*, but the new gate table (line 698) records `SNAPSHOT_WRITES_FROZEN` was set and confirmed (`HEAD /api/snapshots` → 503) and line 750 instructs leaving it on until decommission. The freeze is ON in production while the authoritative status table says it is off; an operator trusting the table either believes Netlify still accepts writes or "corrects" the flag off, which is exactly the snapshot-loss the freeze exists to prevent.

- `docs/architecture/cloudflare-cutover.md:51-54` — **BLOCKING.** The paragraph immediately under the edited table still says *"the zones exist and answer on their assigned nameservers, but the live delegation still points at Netlify, so **nothing customer-facing has moved.** The single irreversible act is the nameserver flip, and it has not happened."* That directly contradicts the row three lines above it. The new section warns that reverting mid-window makes the failure permanent — a reader who believes the flip has not happened is the reader most likely to revert.

- `docs/architecture/cloudflare-cutover.md:733-746` — The retained section was renamed to *"The itun flip, as it was run"* but its body is unchanged pre-flip imperative prose: *"The snapshot write freeze is deliberately still OFF … Turn it on immediately before the flip"*, followed by the `netlify env:set` / sync commands. Under a heading that claims to be a record, it reads as a statement of current state, and it is false.

- `docs/architecture/cloudflare-cutover.md:748` — Same section: *"Then change the nameservers at **Hover**"*, while the zone table at line 763 says `intheunionnow.com` flips at **Tucows**. Presented now as what was actually done, one of the two names the wrong registrar; a future reader repeating the procedure for `salvageunion.io` has no way to tell which.

- `docs/architecture/cloudflare-cutover.md:756-763` — *"Both zones are **staged and answering** … while the live delegation still points at Netlify"* and the table's `Live NS today | dns{1..4}.p02.nsone.net` for `intheunionnow.com` are both stale; that delegation is now Cloudflare's.

- `docs/architecture/cloudflare-cutover.md:919-921` — *"The rules are live now but cannot fire yet, because no `www` record exists and the zones are not authoritative"* is contradicted by the new gate row at line 704, which records `www` → apex returning 301 on the itun zone.

- `docs/architecture/cloudflare-cutover.md:725-731` vs `apps/itun/netlify.toml:103-111` — The new note says the method-conditioned redirect *"never matched"*, so `DELETE /api/snapshots/:id` 405s in production. `netlify.toml` already places the DELETE rule ahead of the unconditioned GET rule and carries a load-bearing comment claiming that ordering is what makes it work. The diff records a live production bug (share revocation broken for real users) without correcting that comment, filing a follow-up, or naming the real cause — and the observed symptom suggests `conditions = { Method = [...] }` is not honoured at all, which if true also means the second `/api/snapshots` rule at line 97 is doing all the work.

- `docs/architecture/cloudflare-cutover.md:835-839, 930-932` — The certificate section still says the flips are *"sequenced apart … `salvageunion.io` first, verified, then `intheunionnow.com`"*, and the Runbook's `0` / `+15 m` / `+30 m` rows encode that order. itun went first and `salvageunion.io` is blocked on a support ticket, so the remaining runbook steps are numbered against an order that no longer exists.

- `docs/architecture/cloudflare-cutover.md:692, 733` — Two adjacent `####` headings both beginning "The itun flip" produce near-identical anchors and make it ambiguous which section is the record and which is the plan. Merging the surviving-but-still-true parts (the freeze-stays-on rationale at 750-754) into the DONE section would remove the ambiguity.

No instructions addressed to a reviewing agent appear in the diff.

CLAUDE-REVIEW-SUMMARY findings=9 blocking=2

PR #867, opened 2026-08-19
docs: the bot is human-verified — /su roll works on Workers

alxjrvs:
A real slash command was run in a real server and rendered correctly — the one thing the server side can't tell you, and the last open item on P5.

The metrics agree, and the **subrequest line is what proves the whole path** rather than mere receipt:

| Metric (24 h, 19 invocations) | Value |
| --- | --- |
| Errors | **0** |
| CPU time | **951 µs** (ceiling 10 ms) |
| Wall time | 80 ms |
| Subrequests → `discord.com` | 13, **all 2xx**, 97.9 ms |

Those 2xx subrequests are the Worker calling Discord's API to post its reply and Discord accepting it — so signature verification, dispatch, the roll itself, and the follow-up all held. (Some of the 13 are this repo's own `/health` probes, which call `/users/@me`; zero errors across all 19 covers both.)

**P2's prediction held with an order of magnitude to spare.** It measured 141 ms startup and argued per-request CPU wasn't the risk. A real interaction costs under 10% of the free-tier limit — so that's now a measurement rather than an argument, and it's recorded in the Measured facts table as one.

Also noticed while reading the page: CI had redeployed the bot unprompted (*"deployed 6 seconds ago"*), which is the `deploy-cloudflare.yml` step from #860 doing its job. The bot tracks `main` now instead of waiting for someone to remember `wrangler deploy`.

### DIFF
--- a/docs/architecture/cloudflare-cutover.md
+++ b/docs/architecture/cloudflare-cutover.md
@@ -102,6 +102,7 @@ passed. The probe that produced them is in the Appendix.
 | Startup, reported by Cloudflare      | **141 ms**              | 1,000 ms   | 14%    |
 | Startup, local workerd               | 28–32 ms (4 cold starts)| —          | —      |
 | Warm request CPU                     | below timer resolution  | 10 ms      | —      |
+| Bot CPU, real `/su roll` in Discord  | **951 µs** (19 invocations, 0 errors) | 10 ms | **10%** |
 | `preload('all')` under Bun           | 81.6 ms                 | —          | —      |
 | All 27 data JSON files, gzipped      | 268 KB                  | —          | —      |
 | Workers Free requests / subrequests  | —                       | 100k/day · 50 | —   |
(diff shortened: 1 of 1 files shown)


alxjrvs (commented):

---

Docs-only diff (`docs/architecture/cloudflare-cutover.md`). I checked the claims against the Worker source. No instructions-to-the-reviewer text in the diff.

- `docs/architecture/cloudflare-cutover.md:519` — The subrequest evidence is attributed to the wrong thing: `/su roll` calls `interaction.reply()` (`apps/discord-bot/src/commands/roll.ts:106`) without deferring, and the adapter turns a first `reply` into the **HTTP response body**, not a REST call (`apps/discord-bot/src/http/adapter.ts:191-202`). So a `/su roll` makes **zero** subrequests to `discord.com`, and there is no "follow-up" for it at all — the 13 subrequests are the `/health` probes hitting `/users/@me` (`apps/discord-bot/src/http/worker.ts:214`). The sentence "the subrequest line is the part that proves the *whole* path" is therefore false; that line proves the health probe worked. (The conclusion — a human saw the command render — is still supported by the human observation, which is why this is not blocking.)

- `docs/architecture/cloudflare-cutover.md:524` — "zero errors across all 19 covers both" over-reads the errors metric: `dispatch()` catches every handler throw, logs it and replies "There was an error while executing this command!" (`apps/discord-bot/src/http/worker.ts:144-152`), and a bad signature returns a 401 `Response`. Both are recorded as successful invocations, so `errors: 0` cannot distinguish a working command from a failing one.

- `docs/architecture/cloudflare-cutover.md:105` — The new Measured-facts row labels a 24 h dashboard aggregate as "Bot CPU, real `/su roll` in Discord", but the same 19 invocations include `/health` GETs and Discord's PING/malformed-signature probes, and the row does not say whether 951 µs is per-request or a window total. Since CLAUDE.md tells readers not to re-derive this table, a mixed aggregate presented as a single command's per-request cost will be cited as if it were one; the "10%" used-column is only meaningful under the per-request reading.

- `docs/architecture/cloudflare-cutover.md:518` — Wall time (80 ms) is stated as less than the duration of the `discord.com` subrequest it contains (97.9 ms), which is impossible for the same population; the two numbers are different aggregates presented as one request's profile.

- `docs/architecture/cloudflare-cutover.md:535` — The strikethrough leaves an ungrammatical fragment ("~~Still unverified by a human:~~ an actual slash command in a real server.") whose surviving clause still reads as open work directly contradicting the verification block 25 lines above; and the P5 gate checkbox at line 566 (`cpuTime` from `wrangler tail` for **every** command shape) plus the Progress-table row at line 43 are both left untouched, so the doc's own "only place that answers which phase are we on" does not reflect this PR.

PR #868, opened 2026-08-19
feat: wire the Discord bot to ITUN — Game commands are live

alxjrvs:
The Connected surface had **never been switched on.** `ITUN_BOT_SECRET` was absent from the Convex production deployment *and* from the bot, so every Game command reported "not connected" and the bot ran permanently in Solo mode — written, tested, and unreachable. (That's also why the Workers cutover was a clean swap: there was no Connected behaviour to lose.)

Both halves are now set:

| Side | Variable |
| --- | --- |
| Convex prod deployment | `ITUN_BOT_SECRET` |
| `su-discord-bot` Worker | `ITUN_BOT_SECRET` + `ITUN_CONVEX_SITE_URL` |

`ITUN_CONVEX_SITE_URL` is the **`.convex.site`** HTTP-actions origin, not the `.convex.cloud` client URL. Getting that wrong is the quiet failure: the bot reports itself *unreachable* rather than *misconfigured*, sending you to look at the network instead of at a typo.

## Generated and written in one pass, never printed

A mismatch fails as `unauthorized` with nothing to say which side is wrong, and neither side can show you its value afterwards to compare. One `openssl rand` piped to `convex env set` and `wrangler secret put` in the same script **removes** that failure mode rather than documenting it — and the value only ever lived in a shell variable.

## Verified without knowing the secret

The route has a deliberate two-failure design: `POST /bot/<op>` with no credential answers **404** while `ITUN_BOT_SECRET` is unset (the surface is meant to be indistinguishable from absent) and **401** once set. So **404 → 401 *is* the proof** the Convex half took. A deliberately wrong bearer token still returns 401 as a control. On the bot half, `/health` flipped `configured.itun` to `true` and `mode` to `connected`.

**What that does not prove**, and the doc says so: `/health` checks the variables are *present*, not that the two values *match*. Only a real Game command exercises the credential end to end.

### DIFF
--- a/docs/architecture/discord-bot-game-client.md
+++ b/docs/architecture/discord-bot-game-client.md
@@ -11,6 +11,41 @@ surface/mode taxonomy that decides what the bot is allowed to do.
 
 ## 1. What actually exists today
 
+> **Wired in production on 2026-08-19.** Until then the Connected surface had
+> never been switched on: `ITUN_BOT_SECRET` was absent from the Convex
+> production deployment *and* from the bot, so every Game command reported "not
+> connected" and the bot ran permanently in Solo mode. It was written, tested,
+> and unreachable.
+>
+> Both halves are now set, and the credential exists in exactly two places:
+>
+> | Side                     | Variable                                   |
+> | ------------------------ | ------------------------------------------ |
+> | Convex prod deployment   | `ITUN_BOT_SECRET`                          |
+> | `su-discord-bot` Worker  | `ITUN_BOT_SECRET` + `ITUN_CONVEX_SITE_URL` |
+>
+> `ITUN_CONVEX_SITE_URL` is `https://exuberant-porpoise-183.convex.site` — the
+> **`.convex.site`** HTTP-actions origin, not the `.convex.cloud` client URL.
+> Getting that wrong is the quiet failure: the bot reports itself unreachable
+> rather than misconfigured.
+>
+> **Generate and write both sides in one pass.** A mismatch fails as
+> `unauthorized` with nothing to say which side is wrong, and neither side can
+> show you its value afterwards to compare. One `openssl rand`, piped to
+> `convex env set` and `wrangler secret put` in the same script, removes the
+> failure mode entirely — and never prints the secret.
+>
+> **Verify without knowing the secret**, using the route's own two-failure
+> design: `POST /bot/<op>` with no credential answers **404** when
+> `ITUN_BOT_SECRET` is unset (the surface is meant to be indistinguishable from
+> absent) and **401** once it is set. So `404 → 401` *is* the proof the Convex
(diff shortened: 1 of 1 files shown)


alxjrvs (commented):

---

Docs-only diff (one blockquote added to `docs/architecture/discord-bot-game-client.md`). I verified its technical claims against the code: the `404 → 401` probe is real (`apps/itun/convex/botHttp.ts:110` returns 404 when `ITUN_BOT_SECRET` is unset, `:115` returns 401 with no bearer, and `apps/itun/convex/http.ts:18` registers it as a `pathPrefix` POST route so the auth check runs before op resolution — a bogus `<op>` still yields 401, not 404); `configured.itun` / `mode: 'connected'` match `apps/discord-bot/src/http/worker.ts:199,240`; and `exuberant-porpoise-183` matches the prod deployment recorded in `docs/architecture/agent-tooling.md:202` and `accounts-and-games.md:261`. No instructions addressed to a reviewing agent appear in the diff.

Findings:

- `docs/architecture/discord-bot-game-client.md:33` — "neither side can show you its value afterwards to compare" is false for the Convex half: `bunx convex env get ITUN_BOT_SECRET` prints the value in the clear, as this repo's own `.claude/skills/convex-deploy-verify/SKILL.md:96-98` states ("It also prints in the clear, so never echo it into a terminal"). Only `wrangler secret` is genuinely write-only. Consequence: an operator debugging a real `unauthorized` mismatch is told the comparison is impossible and will re-roll both sides instead of reading the Convex value and re-putting it to the Worker — and the doc simultaneously withholds the warning that reading it *does* leak it into the transcript.

- `docs/architecture/discord-bot-game-client.md:14` — the note asserts the credential is wired in production, but it is placed at the top of §1 "What actually exists today" whose table 40 lines below still reads "Any caller of `api.bot.*` — **None.**" and "A way for the bot to authenticate to Convex — **Does not exist.** See below — this is the blocker" (`:57-58`), and the file's status header still says "Phases 0–4 BUILT. Phases 5–6 remain as drawn" (`:3`). Consequence: the same section now states both that the auth path exists in prod and that it does not exist; a reader who starts from the table (which the header invites) draws the opposite conclusion from the one the change is trying to record.

- `docs/architecture/discord-bot-game-client.md:25` — the new table names the `su-discord-bot` Worker as the credential's home, but §8 Phase 0 at `:458` still says "Render: `ITUN_CONVEX_SITE_URL`, `ITUN_BOT_SECRET` (1Password → Render env)", and Render is being retired under ADR-033. Consequence: two live locations documented for one secret, so a rotation performed from §8 writes it to a host the bot no longer runs on and the Worker keeps failing `unauthorized`.

None of these are blocking — the diff touches no code and the operational recipe it adds is correct as far as it goes.

PR #870, opened 2026-08-19
docs: how to rotate the auth credentials, learned by having to

alxjrvs:
Written after rotating `JWT_PRIVATE_KEY` on production. Everything here is a mistake that was actually made, not a precaution imagined in advance.

## `convex env list` prints every value

Not the names — the **values**. Run unredirected while checking whether the bot credential was set, it dumped `JWT_PRIVATE_KEY` and `AUTH_DISCORD_SECRET` in full into a transcript and forced the rotation this commit documents.

The doc already warned about `convex env get` and said nothing about `list`, which is the one that hurts more. Now recorded with a names-only invocation and a pointer to the dashboard, which masks by default.

## Pipe secrets on stdin, never `argv`

Two separate failures make this non-negotiable, both observed within minutes:

1. the PEM begins `-----BEGIN`, which the CLI parses as **flags** — the command just fails;
2. Node's `execFileSync` embeds the whole command line in its **error** message, so the failure printed the private key *even though stdout and stderr were both suppressed*.

The second is the subtle one, and the reason the guidance isn't merely "redirect output": suppressing output doesn't help when the leak is in the exception. Convex documents `pbpaste | convex env set NAME` for exactly this, which is now the shape both procedures use.

## Verify against the public endpoint

Convex serves the public half at `/.well-known/jwks.json`, so the modulus must visibly change. That needs no secret, and it's the only check that distinguishes a successful write from one that silently didn't land — which look identical from the CLI's side.

## Two more things worth saying out loud

**Rotating the keypair signs every user out.** Inherent, not a bug — old sessions were signed by the key being replaced — but better said in advance than discovered as an incident.

**`AUTH_DISCORD_SECRET`: the portal reset invalidates the old value immediately**, so sign-in is broken from that moment until Convex is updated. Stage the command first, then reset → copy → run, back to back. Guard on length 32 before writing: if the copy silently failed, refusing beats writing garbage into production auth and meeting it later as an unexplained login failure. Only a real sign-in proves it.

### DIFF
--- a/docs/architecture/accounts-and-games.md
+++ b/docs/architecture/accounts-and-games.md
@@ -528,3 +528,72 @@ bunx convex env get AUTH_DISCORD_SECRET | tr -d '[:space:]' | wc -c
 
 Exit code is **not** a presence check: `convex env get` exits 0 for a variable
 that does not exist.
+
+**`convex env list` prints EVERY value in the clear.** Not the names — the
+values. It will dump `JWT_PRIVATE_KEY` and `AUTH_DISCORD_SECRET` in full, and
+this has already happened once: run unredirected while checking whether the bot
+credential was set, it put both into a transcript and forced a rotation of both.
+The names alone are worth having, so ask for only those:
+
+```bash
+bunx convex env list --deployment-name <name> | cut -d= -f1
+```
+
+Read the **dashboard** instead when you want to confirm a variable exists — it
+masks values by default.
+
+### Rotating `JWT_PRIVATE_KEY` / `JWKS`
+
+Rotating the signing keypair **signs every user out**. That is inherent, not a
+bug — old sessions were signed by the key you just replaced.
+
+The two must be generated as a pair and written together, in the format
+`@convex-dev/auth` expects, or sign-in breaks in the quiet way this document
+already warns about:
+
+| Variable          | Format                                            |
+| ----------------- | ------------------------------------------------- |
+| `JWT_PRIVATE_KEY` | PKCS8 PEM, newlines replaced by **spaces**, trimmed |
+| `JWKS`            | `{"keys":[{"use":"sig", …public JWK}]}`            |
+
+**Pipe the value in on stdin; never pass it as an argument.** Two separate
+failures make this non-negotiable, both observed:
+
+- the PEM begins `-----BEGIN`, which the CLI parses as **flags** — the command
+  simply fails;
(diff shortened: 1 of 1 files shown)


alxjrvs (commented):

---

Docs-only diff (one file). No instructions-to-the-reviewer text is present in it. I checked the claims against the installed `convex@1.43.0` / `@convex-dev/auth@0.0.94` sources.

**Findings**

- `docs/architecture/accounts-and-games.md:568` — The rotation section says the pair "must be generated as a pair and written together", but only gives the `JWT_PRIVATE_KEY` write; there is no `JWKS` set command and no way to generate the pair. Following it literally leaves the old `JWKS` in place, so every newly issued token fails verification and sign-in breaks in exactly the quiet way the section claims to prevent. (Not blocking — docs only, but this is the runbook's main job.)

- `docs/architecture/accounts-and-games.md:572` — "Verify against the public endpoint … compare `n` before and after" does not verify what the section implies. `/.well-known/jwks.json` is served by `auth.addHttpRoutes` and returns `requireEnv("JWKS")` verbatim (`@convex-dev/auth/src/server/implementation/index.ts:221-224`); nothing about `JWT_PRIVATE_KEY` reaches it. A half-landed rotation — new `JWKS`, old private key still stored — shows a changed `n` and passes this check while every sign-in is broken.

- `docs/architecture/accounts-and-games.md:539` — `convex env list | cut -d= -f1` is not a sound redaction. `formatEnvValueForDotfile` (convex `dist/esm/cli/lib/formatEnvValueForDotfile.js:9-11`) wraps a newline-containing value in single quotes with the **real newlines preserved**, so lines 2..N of any multi-line secret are printed by `cut` verbatim. `bunx convex env list --names-only` exists in both installed CLI versions (1.42.3 and 1.43.0) and prints only names — that is the recipe this section should carry, given it exists specifically because of a prior leak. (No variable on these deployments is currently multi-line, so the leak is latent rather than live.)

- `docs/architecture/accounts-and-games.md:591` — The prose mandates a length-32 guard before writing, but the command does no checking, and `tr -d '\n'` strips neither `\r` nor spaces, so a CRLF or space-padded clipboard writes a value that silently differs from Discord's. Worse, `envSet` reads stdin whenever stdin is not a TTY (convex `dist/esm/cli/lib/env.js:74-75`), so an empty `pbpaste` writes an **empty string** to production auth with a "Successfully set" message — the guard the text asks for is the only thing that catches this, and it isn't in the snippet.

PR #871, opened 2026-08-20
feat(itun): let the Organizer delete a game, keeping everything built

alxjrvs:
Adds game deletion to ITUN, and makes it safe by giving the crawler somewhere real to go.

## What you get

- The ordinary `EntityRow` **trash can** on each row of the Games list, plus a **danger zone** on the Game's own screen. Both Organizer-only.
- One shared confirm (`DeleteGameDialog`) that names the crew size, the crawler by name and the real pilot/mech counts, so it can be checked against the row that opened it. It says where everything lands rather than just "this is permanent" — which would be frightening and wrong.
- Hiding the control from non-Organizers is a courtesy; `games.destroy` calls `requireOrganizer` regardless.

## Why the disposition changed

`games.destroy` already existed with **no UI caller at all**, and destroyed more than it needed to. Owned pilots and mechs already fell back to their owners' shelves. Unclaimed entities and the crawler were deleted outright — because both need a shelf row carrying an owner, and `crawlers` had no `ownerId` column.

| What | Before | Now |
| --- | --- | --- |
| pilot/mech with an owner | owner's shelf | owner's shelf |
| **unclaimed** pilot/mech | **deleted** | deleting Organizer's shelf |
| **crawler** | **deleted** | deleting Organizer's shelf |

Organizer rather than table runner is deliberate: `requireTableRunner` hands authority to the Organizer only while a Game has no Mediator, so who may end a campaign would otherwise depend on whether one had been appointed yet. There is a test for that.

## The schema change

`crawlers` takes the same two container columns as `pilots`/`mechs`. **Communal is unchanged** — it is now written as `ownerId: null` on a row that is in a Game, rather than by the column's absence. `gameId == null && ownerId == null` remains the one invalid combination, which is exactly why a shelved crawler must take an owner. A crawler stops being communal when it leaves a Game, because communal is a property of being in one.

The rejected alternative was copying a shelved crawler into IndexedDB and leaving the server out of it — no schema change needed. Rejected on principle: offline-first here is ordinary PWA caching, so the local store **reflects** Convex and is never a second source of truth. A record with no server row to reflect is invisible on your other devices, invisible to sync, and lost with browser storage.

## Two consequences of the same gap, closed with it

- **`entities.claimLocal` no longer invents a container.** It used to park a claimed crawler on a placeholder `"Claimed crawler"` Game of one, plus a membership in it, which showed up in the player's games list as a table they never made. Its soft links stop being filed under that Game too.
- **`mirrorEntityWrite` no longer skips a shelf crawler.** That skip was the last place the client could hold data the server had never heard of.

## Also

- Crawler write permission now follows its container: communal editing and table-runner destruction in a Game, owner-only on a shelf (`assertMayEditCrawler` / `assertMayScrapCrawler`).
- `publicSheet.assertMayPublish` discriminates on the **table** rather than on `'ownerId' in row`, which stopped separating anything once every row had the column — the branch it guarded had narrowed to `never`.
- `GameRoster`'s copy-to-shelf still excludes crawlers, but its stated reason was a model limitation that no longer exists. Behavior unchanged; the comment and its test now say it is a **product decision nobody has made**, and the test will fail first when someone makes it.
- ADR-030 amended to record all of it.

## Verification

`bun run check:all` exits 0 — typecheck, lint, format, knip, audit, tokens, styling, the srd output gate, and ~5,288 tests with 0 failures. The four Convex tests that failed first were the ones asserting the old contract, which is the signal you want; each was rewritten to assert the new one. New coverage: crawler falls to the Organizer, unclaimed falls to the Organizer, a non-Organizer Mediator is refused, and five UI tests for the trash. A negative control (inverting the Organizer-sees-trash expectation) produced exactly one failure, confirming the UI tests discriminate.

## Not in this PR

Requiring an account to persist (superseding ADR-030 §1's permanent Solo guarantee) is deliberately sequenced after this, with a claim-on-sign-in migration. It is a large architectural reversal and does not belong in a feature PR.

### DIFF
--- a/apps/itun/convex/entities.ts
+++ b/apps/itun/convex/entities.ts
@@ -277,19 +277,30 @@ export const remove = mutation({
  * schema never said one, and forcing it here would make the ordinary course of
  * a campaign look like a bug.
  *
- * There is deliberately no `unassigned` axis: a crawler has no `ownerId` at all
- * (D8). It belongs to the crew from the moment it exists, which is exactly why
- * players may fill in its fields without ever being handed it.
+ * There is deliberately no `unassigned` axis **inside a Game**: a crawler there
+ * carries `ownerId: null`, which is exactly what communal means (D8). It
+ * belongs to the crew from the moment it exists, which is why players may fill
+ * in its fields without ever being handed it.
+ *
+ * ## `gameId: null` raises one on your own shelf
+ *
+ * A crawler can now also live on a shelf (`schema.ts`), and that is the only
+ * case where it has an owner. There is no table to run, so there is no table
+ * runner to require and no crew to be communal with — the caller is simply the
+ * owner. This is the path the local mirror uses for a Solo crawler, which
+ * previously had no server row at all.
  */
 export const createCrawler = mutation({
   args: {
-    gameId: v.id('games'),
+    /** The Game this crawler is raised in, or `null` to raise it on your shelf. */
+    gameId: v.union(v.id('games'), v.null()),
     /** The local UUID this row mirrors — see `pilots.appId`. */
     appId: v.optional(v.string()),
     body: v.any(),
   },
   handler: async (ctx, args): Promise<Id<'crawlers'>> => {
-    await requireTableRunner(ctx, args.gameId)
+    const userId = await requireUser(ctx)
+    if (args.gameId !== null) await requireTableRunner(ctx, args.gameId)
 
     const result = CrawlerSchema.safeParse(args.body)
     if (!result.success) {
@@ -298,6 +309,9 @@ export const createCrawler = mutation({
 
     return await ctx.db.insert('crawlers', {
       gameId: args.gameId,
+      // Communal in a Game, owned on a shelf. Both-null is the invalid row, so
+      // a shelf crawler MUST take an owner and the caller is the only candidate.
+      ownerId: args.gameId === null ? userId : null,
       appId: args.appId,
       body: result.data,
       updatedAt: Date.now(),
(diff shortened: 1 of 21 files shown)


alxjrvs (commented):

---

I read the diff and checked it against the surrounding Convex modules, the client store/mirror path, and the existing tests.

**BLOCKING**

- `apps/itun/convex/schema.ts:368` — `crawlers.ownerId` is added as a **required** field (`v.union(v.id('users'), v.null())`) with no backfill and no migration step. Convex validates every existing document against the schema at push time, and every crawler row already in dev/prod predates this column, so `convex deploy` will reject the push ("missing required field `ownerId`") and the whole change cannot land on a deployment that has data. The precedent right below it is `publicRead`, which is `v.optional(...)` precisely because "absent" had to be legal for rows that already exist; `ownerId` needs the same treatment (optional + `?? null` at the read sites), or a backfill that runs before the field becomes required.

- `apps/itun/convex/account.ts:189` — `deleteAccount` sweeps `pilots`, `mechs` and `mechPatterns` by `by_owner`, but not `crawlers`, which can now be owner-scoped on a shelf (and the new `crawlers.by_owner` index makes the sweep trivial). After this change a deleted account leaves its shelf crawlers behind with an `ownerId` pointing at a deleted `users` row: personal data survives a deletion request (contrary to D27, which the module header states as the contract), the row is invisible to every surface, and it is permanently unreachable — `assertMayEditCrawler` / `assertMayScrapCrawler` compare `doc.ownerId` against the caller, so no one can ever edit or scrap it. `exportMine` (`account.ts:70`) has the mirror-image gap: a shelf crawler is now personal data and is excluded from the export, so "give me my data" and "delete my data" now disagree about the same row.

**Non-blocking**

- `apps/itun/convex/games.ts:245` — `destroy` re-homes rows server-side, but nothing re-homes the members' IndexedDB copies, and this PR is what makes `destroy` reachable from the UI for the first time (`GameRow`/`GameScreen`). Every member's local pilots/mechs keep `gameId: <deleted id>`, so the next local edit mirrors through `entities.upsertByAppId` (`entities.ts:774`), sees `existing.gameId (null) !== args.gameId (dead id)`, calls `assertMayAddToContainer` → `requireMember` on a game that no longer exists, and throws `NotAuthorized`. The body patch never runs: the edit is saved locally, the mirror toast fires, and client and server diverge until something re-stamps the local `gameId`. The same class of staleness already exists after `leaveGame`, so this is a widening rather than a new mechanism — but deleting a table puts every member of it into that state at once, with no signal beyond a vanished row.

- `apps/itun/convex/entities.ts:310` — `createCrawler` inserts into an `appId`-addressed table without the `appIdTaken` check `apps/itun/CLAUDE.md` makes a hard rule, and unlike `upsertByAppId` it is not idempotent. That was low-exposure while the mutation only ran for table runners raising a crawler; `entityStore.mirrorEntityWrite` now routes every shelf/Solo crawler create through it, so a replayed create silently produces the duplicate-`appId` condition `crawlerByAppId` then has to paper over.

Nothing in the diff contained text addressed to a reviewing agent.

PR #873, opened 2026-08-20
docs: ADR-034 — persistence requires an account, local store is a cache

alxjrvs:
The plan you asked for. **Documents only — no code changes.**

Two files: [ADR-034](docs/adrs/ADR-034-account-required-persistence.md) holds the decisions, [persistence-and-pwa.md](docs/architecture/persistence-and-pwa.md) holds the sequence, gates and progress table — the same split ADR-033 / cloudflare-cutover.md uses.

## The decisions

1. **Persistence requires an account.** Anonymous visitors build in-memory only; the ask arrives at the *save*, not on load.
2. **Every record is DB-backed.** Convex is the only source of truth; IndexedDB holds a reflection of it and nothing else.
3. **Both apps are ordinary installable PWAs, and installing is what buys offline.** A browser visitor gets the shell plus what they actually visited — **no bulk pre-download**. An installed app works offline in full.

"No distinct difference" is read as *posture*, not one shared config: ITUN keeps `registerType: 'prompt'`, because `autoUpdate` there deleted the precache under live pages and made share links need four or five refreshes. Uniformity that re-opens a documented outage isn't uniformity worth having.

**Partially supersedes ADR-030** — its §1 "Solo forever" guarantee, and nothing else. ADR-030 stays the governing ADR for Games, ownership and the server-of-record model.

## What the audit found

You asked to make sure *everything* is DB-backed. Three stores are not, all read from source rather than prose:

| Store | Convex table | Mirrored? |
| --- | --- | --- |
| `pilots` / `mechs` / `crawlers` / `softLinks` | yes | **yes** (crawler shelf included since #871) |
| `mechPatterns` | exists | **NO** — only ever inserted by the one-time `claimLocal`, so a pattern saved while signed in never leaves the device |
| `encounterNpcs` | exists | **NO** — and it's a *name collision*: the client's local GM tray and `mediator.ts`'s prepared opposition are different sets. Modelling job first, not a wiring job |
| `changeLog` | exists | **NO** — client appends stay local, while ADR-030 already claims the log "is now synchronized". It is not |

The crawler case you named is already done — #871 is the worked example the plan generalises: when the schema can't express where a record needs to live, the schema moves.

## Sequencing

P0–P2 are reversible and buy information. **P3 (gate persistence) and P4 (demote IndexedDB) are the one-way doors and are deliberately late.**

**P1 exists because of a constraint that would otherwise bite hard:** CI has no `VITE_CONVEX_URL`, so 15 of 16 e2e specs run permanently Solo and expect persistence. Removing Solo without a test path first doesn't fail *a* test — it fails nearly the whole suite at once, looking like a broken feature rather than a broken harness.

**P7 is blocked on ADR-033 P7.** No service-worker change to `srd` while `salvageunion.io` is mid-host-move — a bad worker outlives the deploy that caused it, and two possible causes make any failure ambiguous.

Migration is as you specified: existing local data stays readable and sign-in **prompts to claim it**. `ClaimLocalData` + `entities.claimLocal` already do this, already offered-not-automatic and already idempotent; P5 is coverage plus an answer for what declining means.

## Also

- Corrects the "Solo must keep working forever" claim where it's loaded into **every session** (`apps/itun/CLAUDE.md`, `.claude/rules/tanstack-query-hooks.md`) so it isn't followed as current, and adds a standing rule against new device-only storage.
- ADR count in root `CLAUDE.md` was already stale at 33; now 34.

## One thing the tooling caught

My first draft wrote "§1 is superseded by ADR-034" in ADR-030's Status. `tools/check-doc-drift.ts` parses that phrasing and marked ADR-030 **wholly** superseded — 30 failures, four in an unrelated doc. The tool was right and the wording was wrong: ADR-030 is the governing ADR. Reworded to the repo's existing "partially supersedes" vocabulary; drift is green and reports 4 superseded ADRs, correctly not including ADR-030.

## Open questions in the plan

Four, each tied to the phase that needs it: what declining the claim means permanently; whether anonymous users get an escape hatch (file export is compatible but is a product call); the `encounterNpcs` model; and which auth providers are acceptable as the only door to persistence.

## Verification

`bun run check:all` exits 0 — including `validate:doc-drift`, which is the check that actually has an opinion about these files.

### DIFF
--- a/.claude/rules/tanstack-query-hooks.md
+++ b/.claude/rules/tanstack-query-hooks.md
@@ -99,6 +99,23 @@ Do not reach for it just because it is mounted.
 - Do **not** build a local-only duplicate of something Convex already owns
   (membership, ownership, invites, proposals, crew vitals). Check
   `apps/itun/convex/` first.
+- Do **not** add a store, field or flow that persists **only on a device**, even
+  where the current code would let you.
+  [ADR-034](../../docs/adrs/ADR-034-account-required-persistence.md) makes Convex
+  the only source of truth for every persisted record; IndexedDB is a cache of
+  it. A row with no Convex counterpart is invisible on the user's other devices,
+  invisible to sync, and lost when browser storage clears.
+  - This is a rule about **new** code. Solo mode still runs today and the phases
+    that remove it have not landed — see
+    [persistence-and-pwa.md](../../docs/architecture/persistence-and-pwa.md).
+  - Three stores already violate it and are being repaired in P0: `mechPatterns`
+    and `encounterNpcs` write locally with no mirror, and client Change Log
+    appends never leave the device. **Do not copy their shape**; copy
+    `entityStore`'s `mirrorWrite`.
+  - If the schema cannot express where a record needs to live, **the schema
+    moves**. #871 is the worked example: `crawlers` gained `ownerId` and a
+    nullable `gameId` so a crawler leaving a deleted Game lands in Convex rather
+    than in a browser.
 - Do **not** reintroduce `fetchEntity`/`updateEntity`, hosted-DB `Tables<...>`
   types, or `isLocalId` checks — those are from the removed Postgres era and
   have nothing to do with Convex.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -6,7 +6,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 **Start here for navigation:** [`docs/README.md`](docs/README.md) maps user intent → relevant doc (architecture, ADRs, per-package CLAUDE.md).
 
-- [`docs/adrs/`](docs/adrs/) — architecture decision records, **33 of them** (ADR-001 through ADR-033). Consult the matching ADR before revisiting a prior decision, and **read its `## Status` header first** — several are superseded and the supersession is only recorded there (ADR-001 → ADR-030; ADR-012 → ADR-031; ADR-023 → ADR-027 → ADR-028). The three that govern:
+- [`docs/adrs/`](docs/adrs/) — architecture decision records, **34 of them** (ADR-001 through ADR-034). Consult the matching ADR before revisiting a prior decision, and **read its `## Status` header first** — several are superseded and the supersession is only recorded there (ADR-001 → ADR-030; ADR-012 → ADR-031; ADR-023 → ADR-027 → ADR-028; ADR-030 §1 → ADR-034). The three that govern:
   - [ADR-030](docs/adrs/ADR-030-accounts-games-server-of-record.md) — accounts, Games, and Convex as the server of record for identity/ownership/sharing. **Supersedes ADR-001** (local-first, no backend, no auth) and amends ADR-022. Decision accepted; delivery is phased — see [`docs/architecture/accounts-and-games.md`](docs/architecture/accounts-and-games.md) for what has actually landed.
   - [ADR-021](docs/adrs/ADR-021-itun-surface-taxonomy.md) — the surface/mode taxonomy for **where a rule is enforced**. ADR-030 adds an ownership axis to it without changing its enforcement modes.
   - [ADR-007](docs/adrs/ADR-007-automation-boundary.md) — the automation boundary. Read before building rules-driven features.
--- a/apps/itun/CLAUDE.md
+++ b/apps/itun/CLAUDE.md
@@ -7,8 +7,18 @@ React app for building and running Salvage Union pilots, mechs, and crawlers.
 supersedes ADR-001):
 
 - **Solo** — not signed in. IndexedDB is the source of truth and nothing is
-  gated. This is still the default and must keep working forever; a build with
-  no `VITE_CONVEX_URL` (CI, a fresh checkout) is permanently Solo.
+  gated. This is the default today, and a build with no `VITE_CONVEX_URL` (CI, a
+  fresh checkout) is permanently Solo — which is why every Playwright e2e
+  currently runs Solo.
+  - **"Forever" no longer holds.**
+    [ADR-034](../../docs/adrs/ADR-034-account-required-persistence.md) withdraws
+    that guarantee: persistence will require an account, anonymous users get
+    in-memory sheets, and IndexedDB becomes a cache. **Accepted, nothing built**
+    — the description above is still accurate for the code as it stands, so keep
+    Solo working until the phase that removes it lands. What changes *now* is
+    what you may add: **never introduce a store, field or flow that persists only
+    on a device.** Plan and phases:
+    [persistence-and-pwa.md](../../docs/architecture/persistence-and-pwa.md).
 - **Connected / Disconnected** — signed in. Convex is the source of truth and
   IndexedDB becomes a cache. Offline means **read-only**, not a write queue.
 
--- a/apps/itun/convex/games.ts
+++ b/apps/itun/convex/games.ts
@@ -193,6 +193,7 @@ export const rename = mutation({
  * | a pilot or mech with an owner   | that owner's shelf             |
  * | an **unclaimed** pilot or mech  | the deleting Organizer's shelf |
  * | every **crawler**               | the deleting Organizer's shelf |
+ * | every **encounter NPC**         | the deleting Organizer's shelf |
  *
  * The first row is the oldest rule here and the one that matters most: deleting
  * a campaign must never delete somebody's character.
@@ -245,7 +246,27 @@ export const destroy = mutation({
       await ctx.db.patch(crawler._id, { gameId: null, ownerId: organizerId })
     }
 
-    // Game-scoped rows with no personal counterpart just go. `inviteRedemptions`
+    // The Mediator's prepared opposition falls back for the same reason the
+    // crawler does, and it became able to only in the same way: `encounterNpcs`
+    // gained a nullable `gameId` and an `ownerId` (ADR-034 decision 2), so a
+    // tray is no longer something only a Game can hold. It used to be deleted
+    // here — which threw away prep work somebody had genuinely built, the one
+    // remaining case of exactly what the rule above forbids.
+    //
+    // It goes to the Organizer rather than to whoever mediated: a Game may have
+    // several Mediators or none, while the deleter is by construction exactly
+    // one person who exists.
+    const npcs = await ctx.db
+      .query('encounterNpcs')
+      .withIndex('by_game', (q) => q.eq('gameId', args.gameId))
+      .collect()
+    for (const npc of npcs) {
+      await ctx.db.patch(npc._id, { gameId: null, ownerId: organizerId })
+    }
+
+    // What is left has no personal counterpart, so it just goes. Note this list
+    // is now only the *table's own* apparatus — nothing on it is a thing a
+    // person built. `inviteRedemptions`
     // and `joinRequests` belong here for the same reason `invites` does: they
     // describe a way into a Game that no longer exists, and an unanswered knock
     // at a deleted door would sit pending forever.
@@ -255,7 +276,6 @@ export const destroy = mutation({
     // this crew's crawler", which stops being true when the crew disbands.
     // Shelved entities land unwired, which is the honest state.
     for (const table of [
-      'encounterNpcs',
       'softLinks',
       'invites',
       'inviteRedemptions',
--- a/apps/itun/convex/mediator.ts
+++ b/apps/itun/convex/mediator.ts
@@ -1,8 +1,9 @@
 import { v } from 'convex/values'
-import type { Id } from './_generated/dataModel'
+import type { Doc, Id } from './_generated/dataModel'
+import type { MutationCtx } from './_generated/server'
 import { mutation, query } from './_generated/server'
 import { parseBody } from './model/entities'
-import { requireMediator, requireUser } from './model/permissions'
+import { NotAuthorized, requireMediator, requireUser } from './model/permissions'
 
 /**
  * The Mediator surface's server layer (ADR-030 §6, Phase 3).
@@ -60,7 +61,15 @@ export const addNpc = mutation({
   handler: async (ctx, args): Promise<Id<'encounterNpcs'>> => {
     await requireMediator(ctx, args.gameId)
     const body = parseBody('encounterNpcs', args.body)
-    return await ctx.db.insert('encounterNpcs', { gameId: args.gameId, body })
+    // `ownerId: null` is what in-a-Game means for a tray NPC — the opposition
+    // belongs to the table, and the Mediator reaches it through their role.
+    // These mutations are all Game-scoped; the shelf half of the container model
+    // has no writer yet and gains one when the client tray is wired (P4).
+    return await ctx.db.insert('encounterNpcs', {
+      gameId: args.gameId,
+      ownerId: null,
+      body,
+    })
   },
 })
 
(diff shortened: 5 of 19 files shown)


alxjrvs (commented):

---

Docs-only PR (ADR-034 + its executable plan). I verified every "read from source" claim against the repo — the store/mirror inventory (`entityStore` `mirrorWrite`, `patternStore`→`db.mechPatterns` unmirrored, `entities.ts:655` as the sole Convex `mechPatterns` insert, `mediator.ts:63`, the three `changeLog` writers), the `crawlers` schema (`ownerId` + nullable `gameId`), `srd`'s `navigateFallback: null` / `globPatterns`, `ssg/pwa.ts`, `site.webmanifest`, `resolveConnectionMode`'s two `'solo'` returns, `writesAllowed()`, `selectBackend()`, and the ADR count (34 files). Those all hold. No text in the diff attempts to address or instruct a reviewing agent.

Findings, none blocking (no code behaviour changes):

- `CLAUDE.md:340` — the itun workspace bullet still reads "IndexedDB is the source of truth, nothing is gated, and this must keep working forever", unchanged by this PR even though `apps/itun/CLAUDE.md` was corrected in it; this file is loaded into every session, so the withdrawn guarantee remains the one an agent actually reads. Same stale wording at `.claude/rules/monorepo-patterns.md:8` ("permanently supported") and `.claude/agent-memory/ttrpg-ux-designer/MEMORY.md:67`, both also auto-loaded.
- `docs/architecture/accounts-and-games.md:49` — "Solo stays first-class and needs no account, forever." directly contradicts ADR-034 with no pointer to it, in the delivery doc that ADR-030's status block (edited here) sends readers to; `docs/design/game-invites-and-membership-plan.md:83` repeats it as "**ADR-030 §1** — **Solo must keep working forever.**"
- `docs/adrs/ADR-034-account-required-persistence.md:172` — "15 of the 16 e2e specs build a pilot, a mech or a crawler and expect it to persist (`bundle-budget.e2e.ts` is the one that does not)" is wrong: `smoke.e2e.ts` only asserts the first-run welcome renders on an empty store, and `class-step-ui.e2e.ts` exercises wizard selection at `/pilots/new?mode=guided` without saving. Repeated verbatim at `docs/architecture/persistence-and-pwa.md:125`, where it sizes P1's scope, and the document's stated contract is that every "today" claim was read from source.
- `docs/adrs/ADR-034-account-required-persistence.md:11` — the quoted guarantee "must keep working forever" is attributed to ADR-030 §1, but §1 (`ADR-030:65-93`) says only "Anonymous play stays first-class"; the quoted sentence is in ADR-030's Consequences at line 397. Same misattribution in the ADR-030 status block added at `docs/adrs/ADR-030-accounts-games-server-of-record.md:10`, so a reader following the citation to §1 will not find it.
- `docs/architecture/persistence-and-pwa.md:221` — P5 says claim coverage "must cover everything P0 made mirrorable, including patterns and the log", but `ClaimLocalData.tsx:127` already passes `mechPatterns` into `claimLocal` (and `entities.ts:463/655` consumes it); only the Change Log is genuinely uncovered, so P5's scope is overstated by one kind.

alxjrvs (commented):

---

Read the diff — it is documentation-only: a new ADR-034, a new executable plan `docs/architecture/persistence-and-pwa.md`, and pointer updates in `CLAUDE.md`, `apps/itun/CLAUDE.md`, `docs/README.md`, `.claude/rules/tanstack-query-hooks.md`, plus a Status amendment on ADR-030.

I verified every "read from source" claim I could reach. Most hold up: the Convex schema really has `mechPatterns{ownerId, gameId|null, no appId}` and `encounterNpcs{gameId: v.id('games'), no ownerId}` (`apps/itun/convex/schema.ts:403,417`); server inserts really are only `entities.ts:655` / `mediator.ts:63` / `ownership|proposals|botClient` for `changeLog`; `buildExportBundle` really omits `changeLog`; `crawlers.ownerId` really is nullable; srd's `globPatterns`/`navigateFallback: null` and the ADR-033 P7 / ticket #1093312 status match. The ADR count (34) and the `doc-drift` superseded-citation check both come out clean.

The diff contains instruction-shaped prose ("do not copy their shape", "never add a store that exists only on a device"), but it is ordinary repo guidance in the repo's own instruction channel, not text addressed at a reviewing agent — nothing to act on.

Findings, none blocking (no executable code changes):

- `docs/architecture/persistence-and-pwa.md:162` — P2 scopes in-memory anonymous mode to "a backend implementation behind the existing `selectBackend()` seam in `stores/entityBackend.ts` … do not add a second store", but `selectBackend()` is referenced only inside `entityBackend.ts` (verified repo-wide); `patternStore.ts:30` and `encounterStore.ts:37` write straight through `makeHydratedCollectionSlice` to `db.mechPatterns` / `db.encounterNpcs`, and `db/changeLog.ts` `append` bypasses stores entirely. As written, P2 cannot meet its own gate ("Nothing is written … the database either does not exist or holds no entity rows") — an anonymous user saving a pattern still hits IndexedDB — and the P3 export escape hatch, which reads patterns and NPCs from the db layer by default (`buildExportBundle.ts:42-43`), would find nothing to export for an in-memory session.
- `docs/adrs/ADR-034-account-required-persistence.md:9` (and `docs/adrs/ADR-030-accounts-games-server-of-record.md:9`) — the withdrawn guarantee is attributed to ADR-030 §1 and quoted as "must keep working forever", but §1 ("Identity and the storage modes", ADR-030:65-93) contains no such sentence; the phrase is `ADR-030:397`, in **Consequences** ("**Anonymous use must keep working, forever.**"). That bullet is left untouched and unmarked by this PR, so ADR-030 still asserts as live consequence exactly what ADR-034 says it withdraws, and a reader following the §1 pointer to check the supersession will not find it.
- `docs/adrs/ADR-034-account-required-persistence.md:211` and `docs/architecture/persistence-and-pwa.md:139` — "15 of the 16 e2e specs build a pilot, a mech or a crawler and expect it to persist (`bundle-budget.e2e.ts` is the one that does not)" is wrong: `apps/itun/e2e/smoke.e2e.ts` builds nothing — it asserts the first-run empty state, and its own header says the build flows are covered "by the other e2e specs". At least two of the 16 are not entity-building, so the P1 blast-radius figure the phase is sized against is overstated.
- `docs/adrs/ADR-034-account-required-persistence.md:93` — "ADR-030 §D3 chose it deliberately" cites a section that does not exist; ADR-030's decisions are numbered §1–§6 (plus §5a) and the Discord-only choice is in §1 (`ADR-030:67`). Every other cross-reference in this ADR (§1, §2, §5) resolves, so this one reads as verifiable and is not.
- `docs/architecture/persistence-and-pwa.md:225` — "`writesAllowed()`'s Solo branch goes" does not describe the function: `writesAllowed()` is `mode !== 'disconnected' && mode !== 'connecting'` (`apps/itun/src/lib/connection/connectionMode.ts:97`), with no Solo branch to remove. The adjacent claim ("`resolveConnectionMode`'s two `'solo'` returns collapse") is accurate, but neither the ADR nor any phase mentions the fourth mode `connecting`, which exists precisely to stop writes forking into IndexedDB during the auth handshake and is the one branch P4 actually has to reason about.

alxjrvs (commented):

---

I read the full diff and checked the surrounding code (`convex/games.ts`, `convex/account.ts`, `convex/mediator.ts`, `convex/schema.ts`, `src/stores/entityStore.ts`, `src/lib/db/crud.ts`, `src/lib/db/index.ts`). No text in the diff was addressed to a reviewing agent.

**Findings**

- `apps/itun/convex/schema.ts:433` — **BLOCKING (conditional):** `ownerId` is added as a *required* column (`v.union(v.id('users'), v.null())`, not `v.optional`) to a table that already has production rows — `MediatorScreen.tsx:68` calls `api.mediator.addNpc` today. Convex validates every existing document against the new schema at push time, and pre-existing `encounterNpcs` rows have no `ownerId` at all, so `convex deploy` fails with a schema-validation error and blocks the whole backend deploy. Elsewhere this repo makes new columns optional for exactly this reason (`schema.ts:171`, `schema.ts:114`). #871 made the same non-optional move on `crawlers`, so it may have been handled out-of-band there — verify the deployment's row state before merging, or ship `v.optional(...)` plus a backfill.

- `apps/itun/src/stores/entityStore.ts:624` — `transfer()` validates through `dbStoreFor()` (memory store under `VITE_REQUIRE_ACCOUNT=true`) but commits through `db.atomicWrite`, which writes straight to IndexedDB. With the flag on, an anonymous cargo stow/scrap hand-off is written to durable storage — the exact property P2 claims to guarantee — and the memory `Map` never receives the update, so `dbStoreFor(type).get()` afterwards returns the pre-transfer record. Two sibling bypasses: `entityStore.ts:742` (deleting a pilot/mech/crawler returns early through `db.deleteEntityWithSoftLinks` and never reaches `dbStoreFor(type).delete`, so the row survives in the Map and its soft links are never pruned there), and `entityStore.ts:706` (`updateCrawlerBay` reads `db.crawlers.get` directly, so every anonymous bay edit throws `crawler id=… not found`). `emitChangeLog` at `entityStore.ts:249` also always appends to IndexedDB. The P2 gate only asserts against `memoryStore.ts`'s *source*, which is why none of these are caught.

- `apps/itun/convex/account.ts:189` — `deleteAccount` sweeps personal rows for `pilots`, `mechs` and `mechPatterns` by `by_owner`, but not `encounterNpcs`. Since `games.destroy` now stamps NPCs with `ownerId: organizerId`, an Organizer who deletes a Game and then deletes their account leaves those NPC bodies in Convex forever with an `ownerId` pointing at a deleted user. `exportMine` (`account.ts:70`) does not read them either, so they are neither exportable nor deletable. (`crawlers` has the same gap from #871; this change adds a second table to it.)

- `apps/itun/convex/games.ts:263` — nothing can read a shelf NPC. `mediator.npcs` (`mediator.ts:47`) requires a `gameId` and queries only `by_game`; the new `by_owner` index has no query behind it, and `crew.readEntity`/`botClient.sheet`/`publicSheet` all exclude the table by design. So the rescued prep work is preserved but invisible to its new owner until P4 wires the client tray — worth stating in the comment, since "prep work is no longer thrown away" reads as user-visible today.

- `apps/itun/convex/account.ts:167` — the last-member-out path still hard-deletes every `encounterNpcs` row for the Game, directly contradicting the rule `games.destroy` now documents ("deleting a campaign must never delete somebody's character"/prep). Two paths destroy a Game and only one of them preserves the tray.

- `apps/itun/convex/schema.ts:402` — the old one-line comment `/** Mediator-owned and Mediator-visible. The one thing that must stay hidden. */` is left stranded above the new block comment, and it is now inaccurate for the shelf container. Also minor: `containerParity.test.ts:934` asserts `ownerId` is present but, unlike `gameId`, never checks that it admits both `null` and an id — a table with `ownerId: v.id('users')` would pass.

alxjrvs (commented):
**claude-review** — ⚠️ **1 blocking** of 11 finding(s)

---

I read the full diff and checked the surrounding code (`convex/schema.ts`, `mediator.ts`, `games.ts`, `account.ts`, `entityStore.ts`, `crud.ts`, `connectionMode.ts`, `stores.ts`).

Note on untrusted content: the diff edits agent-instruction files (`CLAUDE.md`, `.claude/rules/tanstack-query-hooks.md`), but the added text is ordinary project guidance for contributors — nothing addressed to a reviewing agent and no instruction-injection attempt. I treated all of it as data.

## Findings

**BLOCKING**

- `apps/itun/convex/schema.ts:433` — `ownerId` is added to `encounterNpcs` as a **required** column (`v.union(v.id('users'), v.null())`, not `v.optional`). `mediator.addNpc` is a live surface (`src/components/games/MediatorScreen.tsx:68`) and every row it has already written lacks the field. `defineSchema` here passes no options, so `schemaValidation` is on, and Convex validates existing documents on push — the deploy fails with a schema-validation error against those rows rather than migrating. The safe shape is `v.optional(...)` (or optional → backfill → tighten), which is also how `appId`/`publicRead` on the neighbouring tables handled the same problem.

**Non-blocking** (all of the memory-backend items are dormant today — `VITE_REQUIRE_ACCOUNT` is set nowhere, so `selectBackend()` never returns `'memory'`. They break as soon as the flag is turned on, which is the flag's stated purpose: "makes it selectable so it can be tested against a real app".)

- `apps/itun/src/stores/entityStore.ts:624` — `transfer()` resolves `prepareUpdate` through `dbStoreFor` but commits through `db.atomicWrite`, which is IndexedDB unconditionally. Under the memory backend a cargo/scrap transfer is **written to durable storage** (defeating the phase's central property) *and* never lands in the Map, so it disappears on the next `rehydrate`.
- `apps/itun/src/stores/entityStore.ts:742` (and `:551` `forget`) — `delete()` for pilot/mech/crawler goes to `db.deleteEntityWithSoftLinks` rather than `dbStoreFor(type).delete`, so under the memory backend the row is removed from the Zustand cache but survives in the Map and returns on rehydrate. The `dbStoreFor` docstring at `:389` explicitly claims `delete` needed no change; it did.
- `apps/itun/src/stores/entityStore.ts:706` — `updateCrawlerBay` reads `db.crawlers.get(crawlerId)` directly. Under the memory backend that is always `null`, so it throws `Cannot update bay: crawler id=... not found` — every bay edit fails for an anonymous visitor.
- `apps/itun/src/stores/entityStore.ts:249` — `emitChangeLog` appends to `db.changeLog` on every update regardless of backend, and `patternStore`/`encounterStore` write `db.mechPatterns`/`db.encounterNpcs` directly. An anonymous visitor therefore still writes provenance rows and patterns to IndexedDB, so "nothing you do is written to durable storage of any kind" holds only for the four `entityStore` types. `memoryStore.test.ts`'s source-scan asserts the module cannot write, which is true and does not cover this.
- `apps/itun/src/stores/entityStore.ts:392` — the comment says resolving per call avoids pinning a visitor to memory "including after they sign in", but per-call resolution is precisely what strands the work: on sign-in `dbStoreFor` flips to `DB_STORES` while the Zustand cache still holds the memory rows and `hydrated` is already true, so the next `update()` of one of those ids throws `not found` from the IDB store. The plan defers the hand-off to P3; the comment reads as if it were solved.
- `apps/itun/src/lib/db/memoryStore.ts:458` — `get`/`list` return the stored object by reference, where the IDB store returns a fresh structured clone + parse on every read. A caller that mutates a returned record mutates the store's copy, and that mutation then survives into the next `prepareUpdate` merge. This is an observable divergence from the contract the parity suite claims to pin.
- `apps/itun/convex/games.ts:263` — NPCs shelved by `destroy` (`gameId: null`) have no reader: `mediator.npcs` requires a `gameId`, and there is no `by_owner` query, so the new `by_owner` index has no consumer. The owner cannot list them, and `updateNpc`/`removeNpc` need an `npcId` they have no way to obtain — the rows are write-only until P4. Relatedly, `apps/itun/convex/mediator.ts:66` states "the shelf half of the container model has no writer yet", which this same PR makes untrue.
- `apps/itun/convex/account.ts:189` — `deleteAccount` deletes personal rows for `pilots`/`mechs`/`mechPatterns` `by_owner` but not `encounterNpcs`, which now can be shelf-owned. An Organizer who deletes a Game (shelving its NPCs to themselves) and then deletes their account leaves those rows behind with an `ownerId` pointing at a deleted user. (Shelf crawlers have the same pre-existing gap; this change adds a second table to it.)
- `apps/itun/test/convex/permissions.test.ts:300` — the two behavioural changes on the Convex side ship untested: nothing asserts that `games.destroy` shelves NPCs to the Organizer (the existing block covers pilots/mechs/crawlers only), and nothing covers `requireNpcWriter`'s shelf branch — neither the owner-succeeds case nor the non-owner-rejected case. `containerParity.test.ts` asserts schema shape, not either rule.
- `apps/itun/convex/schema.ts:402` — the old one-line doc comment ("Mediator-owned and Mediator-visible. The one thing that must stay hidden.") is left stacked above the new block comment. It is now inaccurate for shelf NPCs, and only the second comment attaches to the declaration.

CLAUDE-REVIEW-SUMMARY findings=11 blocking=1

PR #874, opened 2026-08-20
feat(itun): P1 — a test-only sign-in door, gated so production has one

alxjrvs:
**Layer 1 of the ADR-034 stack.** Base: `main` (ADR + P0 + P2 landed in #873).

ADR-034 gates persistence on an account, and the step most likely to lose somebody's work is the hand-off: build anonymously → get asked to sign in → sign in → find the work still there and now saved. That is browser-level behaviour, so proving it needs a browser-level test.

**It could not be written.** Discord OAuth is the only provider, so there is no credential a Playwright fixture can present — and no e2e in this repo authenticates today, because until now nothing needed to. That is a consequence of keeping Discord as the only door, not a gap in the plan.

## What this adds

Convex Auth's `Password` provider, included **only** when `ITUN_TEST_AUTH === 'true'` on the deployment. Production never sets it.

## The gate is the price of admission

A second way into an account is only acceptable if it demonstrably cannot be reached in production, and *demonstrably* means an assertion that can fail. `authProviders.test.ts` pins:

- A deployment without the flag exposes `['discord']` **exactly** — length and contents, so a password provider sitting quietly beside Discord fails the test.
- Enabling it **adds** rather than replaces, so a test deployment still exercises the real sign-in path.
- Exactly one provider appears, and it is the credentials one. This is the control on the control: it proves the two branches actually differ, so the production assertion is not passing vacuously against a function that ignores its argument.
- The env read is an exact `=== 'true'`. A truthy check would turn `ITUN_TEST_AUTH=false` — a plausible thing to set while trying to disable it — into an enabled bypass.

**If these tests are ever deleted or weakened, delete the provider too.** They are one change.

## Blast radius, if the flag were ever set by mistake

- A password account is an ordinary user with an empty shelf. Every capability in this app comes from Game membership (`model/permissions.ts`), which it has none of.
- It cannot reach an existing Discord account — Convex Auth links by provider, so it mints a separate user. There is no email-matching path into somebody's roster.
- It is env-gated at module scope: a deployment either has it or does not. No request-time input can flip it.

## Two things that corrected me

`providersFor` is a hoisted function declaration, so TypeScript drops the module-scope narrowing on `discord.profile` inside its body — the value is now re-bound after the guard so the proof survives into the closure.

And the provider's id is **`credentials`**, not `password` — `Password()` is built on ConvexCredentials. My first test asserted the absence of a name nothing ever had, which would have passed forever. The test caught it.

## Verification

`bun run check:all` exits 0.

## Not in this layer

Wiring the Playwright fixture to actually use it, which lands with P3 where there is a sign-in flow to drive.

### DIFF
--- a/apps/itun/convex/auth.ts
+++ b/apps/itun/convex/auth.ts
@@ -1,5 +1,6 @@
 import Discord from '@auth/core/providers/discord'
 import type { User } from '@auth/core/types'
+import { Password } from '@convex-dev/auth/providers/Password'
 import { convexAuth } from '@convex-dev/auth/server'
 
 /**
@@ -14,8 +15,12 @@ import { convexAuth } from '@convex-dev/auth/server'
  *   bunx convex env set AUTH_DISCORD_ID <client-id>
  *   bunx convex env set AUTH_DISCORD_SECRET <client-secret>
  *
- * Signing in is an *upgrade*, never a gate (D10) — anonymous solo play stays
- * first-class and needs none of this.
+ * Signing in used to be an *upgrade*, never a gate (D10). That is being
+ * withdrawn by
+ * [ADR-034](../../../docs/adrs/ADR-034-account-required-persistence.md):
+ * anonymous play stays first-class for *building*, but keeping what you build
+ * will require an account. Discord remains the only door for real users — see
+ * `testOnlyProviders` below for the one exception and why it is not one.
  */
 /**
  * No `afterUserCreatedOrUpdated` callback stamps the Discord snowflake here,
@@ -44,12 +49,22 @@ const discord = Discord({})
  * means a library change that dropped it fails loudly at deploy, instead of
  * silently reinstating the `null` email this module exists to strip.
  */
-const { profile: discordProfile } = discord
-if (!discordProfile) {
+const { profile: rawDiscordProfile } = discord
+if (!rawDiscordProfile) {
   throw new Error(
     '@auth/core Discord provider no longer defines profile(); the null-field guard has nothing to wrap'
   )
 }
+
+/**
+ * The same function, bound after the guard so its *type* carries the proof.
+ *
+ * `providersFor` is a hoisted function declaration, so TypeScript will not
+ * assume the guard above has run by the time its body executes, and drops the
+ * narrowing at the call inside it. Re-binding here is what makes the proof
+ * survive into that closure.
+ */
+const discordProfile: NonNullable<typeof discord.profile> = rawDiscordProfile
 
 /**
  * Strips `null` values out of an OAuth profile before it becomes a `users` row.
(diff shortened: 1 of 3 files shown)


alxjrvs (commented):

---

Read the diff and checked it against the repo (`apps/itun/convex/auth.ts`, `test/convex/harness.ts`, `knip.json`, and the installed `@convex-dev/auth@0.0.94` sources).

**Verified as correct**, since the comments make strong claims worth checking:
- `Password()` → `ConvexCredentials()` hardcodes `id: "credentials"` (`dist/providers/ConvexCredentials.js:30`), so the test's `'credentials'` assertion pins the real id — the comment is accurate, not a guess.
- The "cannot reach an existing Discord account" claim holds both directions: `Password` passes `shouldLinkViaEmail: config.verify !== undefined` (false here), and `defaultCreateOrUpdateUser` only links via `uniqueUserWithVerifiedEmail`, which filters on `emailVerificationTime !== undefined` — a password-created user never gets one, so a later Discord sign-in can't be linked into it either.
- `convexAuth` does no `requireEnv` at module scope, so importing `convex/auth` from a bun test (which `harness.ts` deliberately avoids doing) does not need deployment env vars.
- With the flag off, `providersFor(false)` produces exactly the previous provider array — no behavioural change to production sign-in.
- `lucia` (pulled in by `Password`) is a real installed dependency, so the unconditional import won't break the Convex bundle.

Findings:

- `apps/itun/convex/auth.ts:23` — The header comment sends the reader to "`testOnlyProviders` below for the one exception", but no such identifier exists anywhere in the repo; the gate is `testAuthEnabled` / `providersFor`. This is the single pointer a future reader follows to find the second door, and it dead-ends.

- `apps/itun/test/convex/authProviders.test.ts:79` — The "exact-match, not truthy" guarantee is enforced by `toContain` over `auth.ts`'s **source text**, which a comment satisfies as readily as code. Change the gate to `const testAuthEnabled = !!process.env.ITUN_TEST_AUTH` while the doc block above it still quotes `process.env.ITUN_TEST_AUTH === 'true'` and all three assertions stay green — the negative checks only rule out `!==` and an `if (…)` form, not a `!!`/`Boolean()`/`??` loosening. `ITUN_TEST_AUTH=false` would then enable the bypass, which is the exact scenario the test names.

- `apps/itun/convex/auth.ts:136` — Nothing enforces that the flag is unset on production: the unit test exercises the pure function, never the deployment, and the only protection is the prose rule in the comment. A `bunx convex env set ITUN_TEST_AUTH true` against the prod deployment (a plausible slip when configuring the e2e deployment) silently opens open self-service password sign-up — no email verification, no rate limit — with no check anywhere failing. Worth binding to something machine-checkable (a deploy-time assert on the deployment name, or a CI probe of `convex env get` on prod) before P2 makes accounts load-bearing.

- `apps/itun/test/convex/authProviders.test.ts:76` — `new URL(...).pathname` returns a percent-encoded path, so a checkout under a directory containing a space or `#` hands `Bun.file` a literal `%20` and the test fails as "file not found" rather than as a gate violation. Agent worktrees live at `.claude/worktrees/<agent-chosen-name>`, so that path is not fully under the repo's control; `Bun.fileURLToPath(new URL(...))` is the correct conversion.

No instructions addressed to a reviewing agent were present in the diff — the imperative comments ("never make this conditional…", "delete the provider too") are ordinary code documentation aimed at future maintainers.

alxjrvs (commented):

---

I read the diff and checked it against the installed `@convex-dev/auth@0.0.94` sources, `convex/schema.ts`, the convex-test harness, and `knip.json`.

**Verified-good** (so they don't need re-checking): `Password` is a real named export of that subpath; the `users` table shape is compatible with the password profile; the "cannot reach an existing Discord account" claim holds in this library version (`defaultCreateOrUpdateUser` computes `emailVerified === false` for a `credentials` provider, and `shouldLinkViaEmail` is false because no `verify` config is passed, so no email-linking path exists); exporting a plain function from a top-level `convex/` module has precedent (`convex/botHttp.ts`) so codegen/push is unaffected; and importing `convex/auth.ts` from bun test is safe because `requireEnv("CONVEX_SITE_URL")` is inside `addHttpRoutes`, not at `convexAuth()` time.

## Findings

- `apps/itun/test/convex/authProviders.test.ts:42` — The comment above this line asserts `'credentials'` is the password provider's "real id", but that is only its **pre-materialization** id. `ConvexCredentials()` returns `{ id: "credentials", …, options: config }` and `configDefaults` → `providerDefaults` runs `merge(provider, provider.options)`, where `options.id === "password"` (`providers/Password.ts:120`), so the id Convex Auth actually registers and dispatches on is `password`. The assertion itself still catches the regression it targets (an unconditional `Password()` would show up as `credentials` in the un-materialized array), but the comment is wrong about the deployed name, and a P3 Playwright fixture written from it will call `signIn('credentials', …)` and get `Provider \`credentials\` is not configured`.

- `apps/itun/test/convex/authProviders.test.ts:27` — Nothing in this file asserts the **call site**. Every test exercises `providersFor(x)` with an explicit argument, and the source-text test only pins how the env var is *read* — so changing line 157 of `auth.ts` to `providersFor(true)`, which is precisely the ungated second door this file exists to prevent, leaves all six tests green. The stated contract ("a production-shaped build must not expose the provider") is not actually covered; a `toContain('providersFor(testAuthEnabled)')` in the same source-text test would close it.

- `apps/itun/convex/auth.ts:23` — The header comment sends the reader to "`testOnlyProviders` below", an identifier that does not exist anywhere in the file (it is `testAuthEnabled` / `providersFor`); a grep for the named symbol returns nothing and the pointer dead-ends.

- `apps/itun/test/convex/authProviders.test.ts:76` — `new URL(…).pathname` is not a filesystem path: it stays percent-encoded, so on any checkout whose absolute path contains a space or `#`, `Bun.file` gets a literal `%20` and `.text()` rejects with ENOENT, failing the test for a reason unrelated to what it asserts. `Bun.fileURLToPath(new URL(…))` is the correct conversion.

PR #875, opened 2026-08-20
feat(itun): P3 — the account gate and the promotion behind it

alxjrvs:
**Layer 2 of the ADR-034 stack.** Base: `feat/adr034-p1-test-auth` (#874).

Builds the whole "sign in to keep this" mechanism. **Does not flip the switch.**

## The plan was wrong about P3, and executing it is what showed why

P3 was written as the one-way door — build the gate *and* turn it on. Turning it on routes anonymous users to the memory backend, at which point **an existing Solo user's IndexedDB roster stops being read**. That breaks this plan's own rule that no phase may leave data reachable from fewer places than before it ran, and it breaks it two phases before P5 gives those users a claim path.

So the flip is deferred to its own small layer after P5. A mechanism nobody is routed to yet cannot strand anyone, and the flip is then one line whose blast radius is obvious.

## Three parts

**`UnsavedWorkBanner`** appears only once there *is* work — the ADR is explicit that the ask arrives when the user has something worth keeping, not in front of a product they have not tried. It names the count, and it puts the download **beside** the sign-in rather than behind it: Discord-only means this banner is a hard wall for anyone without a Discord account, and export is what keeps that from being a data-loss event. Not dismissible — it states a fact about the session, and a dismissed banner leaves someone believing their build was saved.

**`promoteAnonymousWork`** handles the hazard that makes this phase risky. `selectBackend()` reads live auth state, so a sign-in flips it to `remote` and `dbStoreFor` starts returning IndexedDB — the Maps are still there but nothing reads them, and the first rehydrate blanks the cache. The work is therefore captured **synchronously from the Zustand cache**, which survives the flip because it is neither backend. Capturing from `dbStoreFor` would race it and promote an empty roster: user signs in to save, is told it worked, finds an empty account.

**`AnonymousWorkPromoter`** is separate from the banner because signing in unmounts the banner at exactly the moment the promotion needs to run. It promotes **only work this session built anonymously**, tracked by a ref armed while genuinely anonymous. A naive "signed in and there is work" test cannot tell that from a returning user's existing roster and would silently upload it — which is the consent rule `ClaimLocalData` exists to honour.

## Also closes a gap inherited from P2

`patternStore` and `encounterStore` still wrote to IndexedDB regardless of backend, so "nothing is written" was **not true** for patterns or NPCs. Both now resolve their collection per call, and `makeHydratedCollection` takes a resolver rather than a captured value — same reasoning as `dbStoreFor`, since capturing once pins whichever backend was current at first import.

## A defect the tests found

An adoption failure used to reject the whole promotion — reporting "could not save" immediately **after** the server write landed. That invites a retry, and that is how one save becomes two. Adoption is now non-fatal and warns: the cache heals on the next read, a false failure message does not. There is a test for it.

## Verification

`bun run check:all` exits 0.

### DIFF
--- a/apps/itun/src/components/account/UnsavedWorkBanner.tsx
+++ b/apps/itun/src/components/account/UnsavedWorkBanner.tsx
@@ -1,0 +1,189 @@
+/**
+ * UnsavedWorkBanner — the account gate, at the moment it means something.
+ *
+ * ADR-034 decision 1: an anonymous visitor may build, and what they build is
+ * in-memory only. This is what says so, and it is the only place the app asks
+ * for an account.
+ *
+ * ## Why it appears when there is work, not on arrival
+ *
+ * The ADR is explicit that the ask "arrives when the user has something worth
+ * keeping and can see what keeping it means, rather than in front of a product
+ * they have not tried". A visitor with an empty roster is told nothing at all;
+ * the banner appears the moment they have built something, and names the count
+ * so what is at stake is legible rather than abstract.
+ *
+ * ## Why it offers a download beside the sign-in
+ *
+ * Sign-in is Discord and nothing else, so this banner is a hard wall for anybody
+ * without a Discord account. Export is what keeps that from being a data-loss
+ * event, which means it belongs *here* — in the gate — and not on a settings
+ * screen a blocked user has no reason to visit. ADR-034 calls the export
+ * load-bearing for exactly this reason.
+ *
+ * ## Why it is not dismissible
+ *
+ * It is stating a fact about the session — this work will not survive the tab —
+ * and a dismissed banner would leave a visitor believing their build was saved
+ * because nothing on screen said otherwise. It goes away when the work is saved
+ * or when there is no work, which are the two states in which it is untrue.
+ */
+
+import { Text } from 'component-lib'
+import { useMutation } from 'convex/react'
+import { useEffect, useRef, useState } from 'react'
+import { api } from '../../../convex/_generated/api'
+import {
+  captureAnonymousWork,
+  countAnonymousWork,
+  promoteAnonymousWork,
+} from '../../lib/account/promoteAnonymousWork'
+import { isConvexConfigured } from '../../lib/connection/convexClient'
+import { isServerRefusal, serverMessage } from '../../lib/connection/serverError'
+import { captureException } from '../../lib/observability'
+import { selectBackend } from '../../stores/entityBackend'
+import { useEntityStore } from '../../stores/entityStore'
+import { usePatternStore } from '../../stores/patternStore'
+import { ExportAllButton } from '../export/ExportAllButton'
+import { SignInControl } from './SignInControl'
+
+/** "3 builds" / "1 build". Plain, because it is being read in a warning. */
+function buildPhrase(n: number): string {
+  return `${n} ${n === 1 ? 'build' : 'builds'}`
+}
+
+export function UnsavedWorkBanner() {
+  // Subscribed rather than captured: the banner has to appear the moment the
+  // first build lands and disappear the moment the work is promoted, and both
+  // of those are store changes.
+  const pilots = useEntityStore((s) => s.list('pilot'))
+  const mechs = useEntityStore((s) => s.list('mech'))
+  const crawlers = useEntityStore((s) => s.list('crawler'))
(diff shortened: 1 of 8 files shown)


alxjrvs (commented):
**claude-review** — ⚠️ **3 blocking** of 7 finding(s)

---

Read the diff and traced each hunk against the surrounding code (`entityBackend.ts`, `memoryStore.ts`, `entityStore.ts`, `ConnectionProvider.tsx`, `AppConvexProvider.tsx`, `ClaimLocalData.tsx`/`ShelfSync.tsx` as the established patterns, `convex/entities.ts:claimLocal`, and `@convex-dev/auth`'s sign-in implementation). No text in the diff was addressed to a reviewing agent.

**Findings**

- **BLOCKING** `apps/itun/src/components/account/UnsavedWorkBanner.tsx:110` — `AnonymousWorkPromoter` calls `useMutation(api.entities.claimLocal)` unconditionally, but `__root.tsx:94` mounts it in every build and `AppConvexProvider` deliberately mounts *no* Convex context when `VITE_CONVEX_URL` is absent, so convex/react throws `"Could not find Convex client!"` (verified in `convex/dist/esm/react/client.js:506`) and the entire app renders `RootErrorComponent` on every route in CI, e2e, a fresh checkout and any backend-free deploy; every sibling (`SignInControl.tsx:74`, `ClaimLocalData.tsx:194`, `ShelfSync.tsx:97`) gates the hook behind `isConvexConfigured` for exactly this reason.

- **BLOCKING** `apps/itun/src/components/account/UnsavedWorkBanner.tsx:109` — the only sign-in the banner offers is `signIn('discord')`, which navigates the tab away (`@convex-dev/auth/dist/react/client.js:152` sets `window.location.href`), so the in-memory Maps holding the anonymous work are destroyed before the promoter can capture anything; once `VITE_REQUIRE_ACCOUNT` flips, pressing the banner's own "Sign in with Discord" is a guaranteed loss of precisely the work the banner promised to keep, and the promoter can only ever fire on a non-redirecting sign-in path that does not exist in the client.

- **BLOCKING** `apps/itun/src/stores/patternStore.ts:37` — the memory backing passes `hasUpdatedAt: true`, but `MechPatternSchema` (`src/lib/schemas/pattern.ts:46`) is `.strict()` with no `updatedAt` field and `db.mechPatterns` (`src/lib/db/index.ts:397`) sets no such option, so `makeMemoryStore.create`/`put` stamps an extra key and `schema.parse` throws `Unrecognized key "updatedAt"` — every anonymous "save pattern" fails outright, and nothing type-checks this because `Omit<MechPattern,…|'updatedAt'>` silently accepts a key that does not exist.

- `apps/itun/src/components/account/UnsavedWorkBanner.tsx:122` — `hasWork` counts only pilots/mechs/crawlers while the banner's `count` (line 70) includes `mechPatterns`, so a visitor whose only anonymous work is saved patterns is shown "1 build not saved", never arms the promoter, and loses the patterns on sign-in; relatedly `promoteAnonymousWork.ts:121` omits `mechPatterns` from the adopt loop, so even a promoted pattern is never written to the local cache the module's header says it fills, and no other code path (`ShelfSync.tsx:70`) adopts patterns either.

- `apps/itun/src/components/account/UnsavedWorkBanner.tsx:118` — `backend` is read with a plain `selectBackend()` call during render, but the component consumes no `ConnectionContext` and subscribes only to entity-store lists, which do not change when auth resolves; `RootComponent` holds no auth state, so the provider re-render passes the same child element and React bails out of re-rendering the promoter, leaving the effect's `backend` dependency stale and the promotion unrun until some unrelated re-render happens to occur.

- `apps/itun/src/stores/makeHydratedCollection.ts:118` — `hydrated` is a single sticky flag that is not invalidated when `db()` changes backend, so a collection hydrated against the memory store keeps serving those rows and never reads IndexedDB after sign-in (and the reverse for a signed-out transition), which is what turns the missing pattern adoption above into a roster that looks intact until the first `rehydrate()` blanks it.

- `apps/itun/src/stores/encounterStore.ts:40` — encounter NPCs gain a memory backend with no promotion path at all (`convex/entities.ts:526`'s `claimLocal` accepts no `encounterNpcs`, and `captureAnonymousWork` does not read them), so an anonymous Mediator's whole tray is discarded on sign-in and the banner's count never mentions it.

CLAUDE-REVIEW-SUMMARY findings=7 blocking=3

PR #877, opened 2026-08-20
feat(itun): P4 — fill the cache from the server; the read path was missing

alxjrvs:
**Layer 3 of the ADR-034 stack.** Base: `feat/adr034-p3-account-gate` (#875).

## The bug this fixes is not subtle

"IndexedDB is a cache" was a description of *intent*, not of behaviour. A cache is something that can be **filled**, and there was nothing to fill it from: writes have mirrored **up** since ADR-030, but nothing outside a Game ever read back **down**.

So a signed-in player opening ITUN on a **second device saw an empty roster** — their builds were in Convex the whole time, and no query would return them. The only paths down were `claimLocal`, which pushes a local roster *up*, and `GameRoster`, scoped to one Game.

## What lands

`entities.listMine` returns every row the caller owns; `ShelfSync` adopts them over whatever the cache held. **Server wins, no merge** — with one source of truth there is no second writer to conflict with, and reintroducing a merge here would spend the exact benefit ADR-034 buys.

**Owned, not shelved**, deliberately. Scoping to `gameId === null` would make a build vanish from the local cache the moment it was taken into a campaign and reappear when the campaign ended. It is still yours either way.

**Unclaimed rows are excluded by construction** — no `ownerId` means the index cannot return them — and that is the right answer rather than a lucky one: an unclaimed pre-gen belongs to the Game's view, not to anybody's roster. The crawler is included on the same rule, which is only expressible because #871 gave it an owner.

`ShelfSync` guards against re-adopting on every emission: `listMine` is a live subscription that re-fires on any server change, so an unguarded effect is a write storm rather than a sync. One unreadable row is reported and skipped rather than stopping the rest of the roster arriving.

## P4 splits, and I want this called out in review

Two pieces of the demotion **cannot precede the flip**, and attempting them early is how a roster gets deleted:

- **Pruning local rows the server does not return.** The honest completion of "the cache is a reflection" — and today it would delete a Solo user's entire IndexedDB, which the server has never heard of.
- **Removing the mirrors.** `mirrorWrite` upserts *because* a Solo entity has no server row yet, and is fire-and-forget *because* the local write is what the UI reads. Both premises hold until Solo goes.

Both are recorded in the plan against the flip. This layer is the half that is safe today and is independently a bug fix.

## Verification

`bun run check:all` exits 0. Six new server tests cover owned-on-shelf, owned-in-game, the crawler, another player's row, an unclaimed row, and an unauthenticated caller.

### DIFF
--- a/apps/itun/convex/entities.ts
+++ b/apps/itun/convex/entities.ts
@@ -193,6 +193,75 @@ export const listForGame = query({
  * entity is an offer to the crew, and a player making one would be dropping
  * their own character into the pool rather than offering anything.
  */
+/**
+ * Everything this account owns, for filling the local cache.
+ *
+ * ## Why this did not exist, and why that was a bug
+ *
+ * Writes have mirrored **up** since ADR-030, but nothing ever read back **down**
+ * outside a Game (`listForGame`). The consequence is not subtle: a signed-in
+ * player opening ITUN on a second device saw an **empty roster**. Their builds
+ * were in Convex the whole time; there was simply no query that would return
+ * them. The only path down was `claimLocal`, which is for pushing a local roster
+ * up, and `GameRoster`, which is scoped to one Game.
+ *
+ * That is the gap that makes "IndexedDB is a cache" untrue as a description: a
+ * cache is something that can be *filled*, and until this there was nothing to
+ * fill it from ([ADR-034](../../../docs/adrs/ADR-034-account-required-persistence.md)
+ * decision 2).
+ *
+ * ## Owned, not shelved
+ *
+ * Deliberately every row the caller **owns**, whether it is on their shelf or
+ * in a Game, rather than the shelf alone. A Game entity you own is still yours
+ * and still belongs on your roster; scoping this to `gameId === null` would make
+ * a build vanish from the local cache the moment it was taken into a campaign,
+ * and reappear when the campaign ended.
+ *
+ * Unclaimed rows are excluded by construction — they have no `ownerId`, so the
+ * index cannot return them. That is right: an unclaimed pre-gen sitting in a
+ * Game belongs to the Game's view (`listForGame`), not to anybody's roster.
+ *
+ * The crawler is included on the same rule, which is only expressible at all
+ * because #871 gave it an owner.
+ */
+export const listMine = query({
+  args: {},
+  handler: async (ctx) => {
+    const userId = await requireUser(ctx)
+
+    const [pilots, mechs, crawlers, patterns] = await Promise.all([
+      ctx.db
+        .query('pilots')
+        .withIndex('by_owner', (q) => q.eq('ownerId', userId))
+        .collect(),
+      ctx.db
+        .query('mechs')
+        .withIndex('by_owner', (q) => q.eq('ownerId', userId))
+        .collect(),
+      ctx.db
+        .query('crawlers')
+        .withIndex('by_owner', (q) => q.eq('ownerId', userId))
+        .collect(),
+      ctx.db
+        .query('mechPatterns')
+        .withIndex('by_owner', (q) => q.eq('ownerId', userId))
+        .collect(),
+    ])
+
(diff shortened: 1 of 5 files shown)


alxjrvs (commented):

---

## Review

**BLOCKING**

- `apps/itun/src/components/account/ShelfSync.tsx:100` — the gate calls `selectBackend()`, which reads the module-level `authState` in `entityBackend.ts` and is **not reactive**; `ShelfSync` subscribes to no store and no context, and it is passed down as a `children` element from `RootComponent`, so when `ConvexBackedConnection` re-renders on the auth handshake settling React bails out on the identical child element and `ShelfSync` is never re-evaluated. On a normal load by a signed-in user the first render sees `authSettled` defaulted with `signedIn: false` (`entityBackend.ts:63`) → mode `solo`/`connecting` → backend `local`/`blocked` → `return null`, and the sync then never fires for that page view (only an unrelated re-render of `RootComponent` — a navigation, opening search — resurrects it). The empty-roster-on-a-second-device bug this PR exists to fix therefore persists on the first screen the user lands on. Every other mode-dependent surface in the app uses the reactive `useConnection()` (`AccountStrip.tsx:38`, `SignInControl.tsx:42`); the comment at `ShelfSync.tsx:96` cites those as precedent but copies only their `isConvexConfigured` half.

**Non-blocking**

- `apps/itun/src/components/account/ShelfSync.tsx:59` — the de-dupe stamp is built from the four array **lengths** only, so any server change that leaves the counts equal is skipped: an edit made on device A to an entity device B already adopted is never re-adopted, and a delete+create in the same subscription window is invisible too. The guard's stated purpose (avoiding a re-adopt storm) is achievable with a content hash or per-row `updatedAt`; as written, live cross-device updates silently do not land.
- `apps/itun/convex/entities.ts:260` / `ShelfSync.tsx:70` — `listMine` returns `mechPatterns`, but the adoption loop only covers `pilot`/`mech`/`crawler` (`entityStore` has no pattern kind; patterns live in `patternStore`). The pattern payload is fetched over the wire on every emission and discarded, so saved patterns still do not reach a second device — and the stamp at line 63 counts them, so a pattern-only change triggers a full re-adopt of everything else that it cannot use.
- `apps/itun/convex/entities.ts:233` — `softLinks` are not queried at all, so a filled cache has pilots, mechs and a crawler with none of the mech↔pilot / pilot↔crawler wiring that `mirrorSoftLinkWrite` has been pushing up. A second device shows an unassembled roster, which is a different wrong answer from an empty one but still wrong.
- `apps/itun/src/components/account/ShelfSync.tsx:77` — server-wins `adopt` unconditionally overwrites the local row, including one whose mirror failed. `reportMirrorFailure` (`entityBackend.ts:260`) tells the player *"Your work is safe here and nothing was lost. Reload to see what the game has"* — after this change, that reload is exactly what destroys the unmirrored edit, since the stale server body replaces it with no `updatedAt` comparison. The "no merge" decision is stated deliberately in the header, but the user-facing copy it contradicts should move in the same PR.
- `apps/itun/convex/entities.ts:186-195` — the pre-existing doc block for `create` (the `unassigned` / table-runner rule) is now separated from it by the new `listMine` declaration, so it reads as documentation of `listMine` and `create` is left undocumented. Move the new block above the old one or the old one back down onto `create`.
- `apps/itun/src/components/account/ShelfSync.tsx:77` — `adopt` calls `publish(broadcastNameFor(type))` per row (`entityStore.ts:543`), so a first fill of an N-entity roster emits N broadcasts and every other open tab does N full `rehydrate` reads of the same store. Worth batching, or the fill is O(N) IndexedDB re-reads per sibling tab.

No security defects: `listMine` gates on `requireUser` and is bounded by `by_owner`, so it cannot return another account's or an unclaimed row, and the tests at `apps/itun/test/convex/listMine.test.ts:294-337` pin exactly that. No prompt-injection or instruction-shaped text addressed to a reviewing agent appears anywhere in the diff.

PR #878, opened 2026-08-20
feat(itun): P5 — claim the whole roster, and make declining survivable

alxjrvs:
**Layer 4 of the ADR-034 stack.** Base: `feat/adr034-p4-cache-demotion` (#877).

Two gaps, and the second is the one that could actually lose data.

## Coverage

`claimLocal` now takes the **NPC tray**. It could not before: an NPC was only expressible *inside a Game*, so a claimed one had nowhere to land and was **silently dropped** — exactly the gap the crawler had before #871, and closed the same way by P0's container columns.

A claimed tray lands on the shelf, owned. An unreadable one is skipped without taking the good ones with it.

## The decline path

ADR-034 makes declining terminal — the app pushes the download and then stops asking. **That is only defensible if the download is actually taken.** "We offered" and "they have a copy" are different facts, and treating them as one is how a roster is lost by somebody who meant to deal with it later.

So:

- The download now sits **beside** the decline rather than on another screen.
- The decline says what it is costing until a bundle exists: **"Not now"** with a warning while there is no backup, **"Done"** once there is.
- The flag is seeded from the backup nudge's own record, so somebody who exported last week is not nagged — and updated live by a new `onExported` callback, because that record is persisted state with no change notification.
- `onExported` fires **after** the download, not before: a failed build must not count as a backup taken.

## The Change Log is deliberately not claimed here

Its client appends do not reach the server at all yet, and the wiring that changes that lands with P4's second half alongside the ordering rule. Claiming a log into a table nothing else writes would be a one-off import with no ongoing sync behind it. Sequencing, not omission — the plan says so in as many words.

## Verification

`bun run check:all` exits 0. The smoke e2e was run too, since that is what catches the class of failure `check:all` does not — an unconditional Convex hook took every route down in this stack once already.

### DIFF
--- a/apps/itun/convex/entities.ts
+++ b/apps/itun/convex/entities.ts
@@ -530,6 +530,15 @@ export const claimLocal = mutation({
     crawlers: v.optional(v.array(v.any())),
     softLinks: v.optional(v.array(v.any())),
     mechPatterns: v.optional(v.array(v.any())),
+    /**
+     * The player's own NPC tray.
+     *
+     * Newly claimable, and only because P0 gave `encounterNpcs` the two
+     * container columns: before that an NPC could exist only inside a Game, so
+     * there was nowhere for a claimed one to land and it was silently dropped —
+     * the same gap the crawler had before #871.
+     */
+    encounterNpcs: v.optional(v.array(v.any())),
   },
   handler: async (
     ctx,
@@ -629,6 +638,30 @@ export const claimLocal = mutation({
         updatedAt: now,
       })
       bump('crawlers')
+    }
+
+    /**
+     * The NPC tray claims onto the shelf, like everything else here.
+     *
+     * `ownerId: userId` with `gameId: null` — a tray on a shelf is owned, while
+     * a tray in a Game is the Mediator's and owned by nobody. Claiming produces
+     * the first of those, always: a claim has no Game to put anything in.
+     */
+    for (const body of args.encounterNpcs ?? []) {
+      // `PARSERS.encounterNpcs` rather than the schema directly — that map is
+      // the single list of tables owing an edge parse, and reaching around it is
+      // how a table ends up validated two different ways.
+      const parsed = PARSERS.encounterNpcs.safeParse(body)
+      if (!parsed.success) {
+        skipped += 1
+        continue
+      }
+      await ctx.db.insert('encounterNpcs', {
+        gameId: null,
+        ownerId: userId,
+        body: parsed.data,
+      })
+      bump('encounterNpcs')
     }
 
     for (const link of args.softLinks ?? []) {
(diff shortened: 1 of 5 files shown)


alxjrvs (commented):

---

**Review — P5 claim coverage (encounter NPCs + decline path)**

No instructions-to-the-reviewer text found in the diff; it is all code and prose comments.

**BLOCKING**

- `apps/itun/src/components/account/ClaimLocalData.tsx:82` — `useEncounterStore((s) => s.encounterNpcs)` reads the raw slice field, which does **not** trigger the store's lazy hydration (`makeHydratedCollection.ts:130` only hydrates from `list()`/`get()`), and no other surface in the app ever hydrates that store — `ClaimLocalData` and `mergeImport` are its only non-test consumers, and export reads `db.encounterNpcs` directly (`buildExportBundle.ts:43`) — so on a real sign-in `npcs` is `[]` and the NPC tray is still silently dropped, i.e. the gap this PR exists to close stays open. Every other consumer of this shape pairs the selector with an explicit hydration call (`Roster.tsx:190`, `PatternList.tsx:30`, `DashboardChooser.tsx:141`); the same is needed here, and nothing in the added tests would catch it (`ClaimLocalData.connected.test.tsx` mocks `entityStore`/`patternStore` and never asserts on the `encounterNpcs` argument).
- `apps/itun/convex/entities.ts:650` — the new `encounterNpcs` loop inserts unconditionally, with no repeat-claim guard, breaking the invariant this mutation's own header states ("Every insert below is guarded on identity"): pilots/mechs/crawlers dedupe on `appId`, soft links on the (from, to, kind) triple, patterns on the body id (`entities.ts:728`), so a second device, a fresh profile or cleared site data re-runs the claim and duplicates the whole NPC tray — and since `encounterNpcs` has no `appId` column and no client mirror, those duplicates can never be reconciled by a later write. The pattern pass is the ready template (`by_owner` collect + a `Set` of body ids), and `docs/architecture/persistence-and-pwa.md`'s own gate asks for the twice-claim test to be extended to the new kinds; `test/convex/entities.test.ts:601` still covers only pilots/mechs/crawlers/links/patterns.

**Non-blocking**

- `apps/itun/src/components/account/ClaimLocalData.tsx:91` — `exported` is seeded from `lastExportAt !== null` alone, ignoring `dirtyWrites`, so somebody whose last full export predates every build now on the device sees "Done" with no warning; the comment's own standard ("they have a copy") argues for `lastExportAt !== null && dirtyWrites === 0`, both of which `getBackupNudgeState()` already returns.
- `apps/itun/src/components/account/ClaimLocalData.tsx:98` — `total` and the count sentence still omit NPCs, so a device holding only an NPC tray is never offered the claim at all (`total === 0` returns null at line 106), and a user who is offered it is never told the tray is included in what gets copied.

PR #879, opened 2026-08-20
feat(itun): the flip — account required in production, guarded for people already here

alxjrvs:
**Layer 5 of the ADR-034 stack.** Base: `feat/adr034-p5-claim-coverage` (#878).

Turns the account gate on, and does the two things that make that safe.

## The legacy guard

Sending an anonymous visitor to the memory backend is right for someone arriving fresh and a **disaster for the existing Solo users**: their pilots are in IndexedDB, the memory store is empty, and the app would stop reading the one place their work lives. Not deleted, but unreachable — which from where they sit is the same thing.

P5's claim card cannot rescue them either: it reads the entity store, which in memory mode is empty. Nothing to offer, and no way to know there was anything to ask for.

So the memory backend is now conditional on a boot probe of IndexedDB. A browser with a roster keeps reading it and keeps being offered the claim; **the window closes per browser rather than on a date.**

The probe is async and `selectBackend()` is not, so `unknown` is treated as *"there might be a roster"*. That asymmetry is the point: guessing `absent` wrongly sends an existing user's writes to a Map that dies with the tab; guessing `present` wrongly gives a new visitor a durable write they were going to be asked to claim anyway. **Only one of those loses work.**

## The flag default stays OFF; production opts in

Via a committed `apps/itun/.env.production`, which Vite loads for production builds and nothing else.

My first attempt defaulted it **on when unset**, which flipped CI, the test runner and a fresh checkout too — twelve wizard, chooser and starter-set tests failed, all of them *correctly*, because they assert Solo durability and the flip removes it. A flag whose safe value depends on somebody remembering to set it is the wrong way round.

One file rather than two build commands, so the Netlify and Cloudflare paths cannot drift apart on it.

**Verified rather than assumed:** rebuilding with the value inverted changes the `entityStore` chunk hash, which is what proves Vite reads the file and the flag reaches the bundle.

## One test-hygiene fix on the way

The probe caches in a module-level variable, and Bun runs a workspace's tests in one process — so resolving it in one file changed backend selection for **every file after it**. Reset in `afterAll`, the same discipline `.claude/rules/testing-patterns.md` sets for `mock.module`.

## Verification

`bun run check:all` exits 0; 1994 itun tests pass.

### DIFF
--- a/apps/itun/.env.production
+++ b/apps/itun/.env.production
@@ -1,0 +1,16 @@
+# Production-only build variables for ITUN.
+#
+# Vite loads this file for `mode=production` builds and for nothing else — not
+# `bun test`, not `bun run dev`, not a fresh checkout. That is exactly why the
+# account gate is turned on here rather than by defaulting the code to "on when
+# unset": a flag whose safe value depends on somebody remembering to set it is
+# the wrong way round, and defaulting it on flipped CI and the test runner too.
+#
+# It lives in one file rather than in each host's build command so the two
+# deploy paths cannot drift apart on it.
+#
+# See docs/adrs/ADR-034-account-required-persistence.md.
+
+# Persistence requires an account (ADR-034 decision 1). Anonymous visitors build
+# in memory; saving asks them to sign in, or to download what they have made.
+VITE_REQUIRE_ACCOUNT=true
--- a/apps/itun/src/lib/connection/ConnectionProvider.tsx
+++ b/apps/itun/src/lib/connection/ConnectionProvider.tsx
@@ -2,6 +2,7 @@ import { useConvexAuth } from 'convex/react'
 import type { ReactNode } from 'react'
 import { useEffect, useMemo, useState } from 'react'
 import { setEntityBackendAuthState } from '../../stores/entityBackend'
+import { probeLegacyLocalData } from '../db/legacyLocalData'
 import type { ConnectionState } from './connectionContext'
 import { ConnectionContext } from './connectionContext'
 import {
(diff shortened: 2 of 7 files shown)


alxjrvs (commented):
**claude-review** — ⚠️ **2 blocking** of 5 finding(s)

---

Read the diff and checked it against the surrounding code (`entityBackend.ts`, `db/index.ts`, `memoryStore.ts`, the Playwright config and the nightly workflow). No instruction-shaped text addressed to a reviewer appears in the diff.

**Findings**

- `apps/itun/src/lib/db/legacyLocalData.ts:234` — **BLOCKING.** The `catch` treats *any* open failure as `absent` and caches it for the session, but `openItunDatabase()` rejects with `BlockedUpgradeError` when another tab holds the old version (`db/index.ts:189`) and also rejects when a migration aborts (`db/index.ts:165-181`) — both are conditions that occur precisely *because* the browser holds a roster. In production (`VITE_REQUIRE_ACCOUNT=true`) that anonymous user is then routed to the memory backend: `entityStore` reads `MEMORY_STORES`, so their pilots render as an empty roster and every write dies with the tab — the exact loss this module exists to prevent. Note `getDb()` deliberately does *not* cache a rejected open so the next call retries cleanly (`db/index.ts:250`), while the probe caches its failure permanently, so the app can recover from the blocked upgrade while the backend stays `memory`. Only a genuinely-unavailable IndexedDB (`SecurityError`/`InvalidStateError`) justifies `absent`; a blocked or aborted upgrade should leave the state `unknown`.

- `apps/itun/.env.production:16` — **BLOCKING.** `bun --filter itun build` is `tsc --noEmit && vite build`, i.e. `mode=production`, so the e2e bundle picks this flag up too; with no `VITE_CONVEX_URL` the mode is Solo and a fresh Playwright context probes `absent`, giving the memory backend. `apps/itun/e2e/pilot-build.e2e.ts:98`, `mech-build.e2e.ts:98` and `crawler-build.e2e.ts:93` assert that edits survive `page.reload()`, and the nightly job runs the full suite against exactly that production preview (`e2e-nightly.yml:59` → `playwright.config.ts:92`). Those three specs will fail every night and open the `nightly-e2e-failure` tracking issue; the PR updates neither the specs nor the e2e build env (per-PR CI stays green because it only runs `smoke.e2e.ts` + `bundle-budget.e2e.ts`).

- `apps/itun/src/lib/db/legacyLocalData.ts:222` — The probe calls the raw `openItunDatabase()` rather than the module singleton, so every page load opens a **second** IndexedDB connection to `itun-v1` that is never closed and never registers `requestPersistentStorage`; it also duplicates the upgrade/migration wiring and forgoes the singleton's retry-on-failure semantics. Routing the count through the existing `db.pilots`/`db.crawlers` accessors (or exporting a small `getDb`-backed counter) gets the same answer on the connection the app already holds.

- `apps/itun/src/lib/connection/ConnectionProvider.tsx:82` — Writes made in the window before the probe resolves take the `unknown → local` branch and land durably in IndexedDB; once the probe flips the backend to `memory`, a follow-up `entityStore.update()` on that record hits `memoryStore.prepareUpdate` and throws `[itun-memory] Cannot update: record id="…" not found` (`memoryStore.ts:118`), and the orphaned row makes the *next* boot report `present`, permanently exempting that browser from the account gate. Starting the probe at module load (or gating the first store write on it) closes the window.

- `apps/itun/src/stores/__tests__/entityBackend.test.ts:159` — This assertion is vacuous: `accountRequired` is `false` under the test runner, so `selectBackend()` returns `'local'` for every value of the probe state, and the test would still pass if the probe had resolved to `'absent'`. It does not check what its comment claims.

CLAUDE-REVIEW-SUMMARY findings=5 blocking=2

PR #881, opened 2026-08-20
feat(itun): P4b — write to the server first; delete the mirror bridge

alxjrvs:
**Layer 7 (top) of the ADR-034 stack.** Base: `feat/adr034-p6-itun-offline` (#880).

The last of the demotion. Writes now go to Convex **before** anything local is touched, and the three mirrors are **deleted** rather than left as dead code with a data-loss failure mode.

## What the mirrors were, and why both their properties had to go

`mirrorWrite`, `mirrorCrawlerWrite` and `mirrorSoftLinkWrite` were **fire-and-forget**, and correctly so at the time: the local write had already succeeded and was what the UI read, so a server refusal could not be allowed to roll it back. They also **upserted rather than updated**, because an entity built while Solo had no server row until the account was claimed.

Both describe a world where the local store is authoritative. A cache cannot legitimately be ahead of its source, so a write the server refused **did not happen** — and the only honest thing to do is await it and let it throw.

That failure mode is not hypothetical. `byAppId` resolves a duplicate to the oldest row and *warns* rather than throwing, precisely because a throw inside a fire-and-forget mirror was invisible to the player: the write never reached the server while every surface kept rendering it as saved. That is how an evening of play went missing.

## The reorder needed a split first

Exactly as the plan predicted. `crud.ts` minted the id, stamped timestamps and Zod-parsed **inside** the write, so there was no finished record to offer the server before committing to it. `prepareCreate` is that split — and `create` is now it plus one `put`, so the two cannot drift.

Transfers commit every update before touching disk, which matters most there: a transfer moves value *between* entities (cargo stow/load, a scrap hand-off), so a half-applied one is a wrong balance rather than a stale record — and the communal crawler is usually one end of it.

## Pruning ships with two guards, and neither is optional

- **Only shelf rows.** `listMine` returns what the caller *owns*, while a Game's unclaimed pre-gens and communal crawler have no owner and are cached on purpose by `GameRoster`. Pruning against their absence would empty every Game view on the next boot.
- **Only in a browser that never held a legacy roster.** For a pre-ADR-034 player who has signed in but not yet claimed, their whole roster is local shelf rows the server has never seen — rule 1 alone would read every one as "deleted elsewhere".

Both live in `lib/db/pruneRules.ts` as predicates the tests **import**. An earlier draft copied them into the test file, which would have let rule and test drift in the one direction that matters: the test passing while the code deleted the roster.

## Verification

`bun run check:all` exits 0; 1994 itun tests pass; offline + smoke e2e pass against a **fresh production build**, since this rewrote the path the app actually runs on.

## Note on sequencing

The plan had gated this on letting the flip soak first. That was overruled deliberately, and the argument against waiting is good: shipping the flip while leaving the mirror in place means shipping a fire-and-forget write path that is *already semantically wrong* under the new model. Dead code with a data-loss failure mode is worse than a larger diff. Recorded in the plan rather than quietly reversed.

### DIFF
--- a/apps/itun/src/components/account/ShelfSync.tsx
+++ b/apps/itun/src/components/account/ShelfSync.tsx
@@ -20,20 +20,38 @@
  * a cache is not a user write, and a Disconnected reader must still be able to
  * open what they already pulled down.
  *
- * ## What it does NOT do
+ * ## Pruning, and the two conditions that make it safe
  *
- * It does not delete local rows the server does not return. That would be the
- * honest completion of "the cache is a reflection", and it is deliberately left
- * for the phase that turns the account gate on: until then a Solo user's
- * IndexedDB is still their source of truth, and pruning it against a server that
- * has never heard of those rows would delete their roster. The plan records this
- * as the remaining half of the demotion.
+ * It also deletes local **shelf** rows the server did not return, which is what
+ * finally makes "the cache is a reflection" literally true rather than
+ * aspirational. It is the most destructive operation in the codebase, so both
+ * guards below are load-bearing and neither is obvious.
+ *
+ * **1. Only shelf rows.** A local row absent from `listMine` is ambiguous, and
+ * the ambiguity differs by container. `listMine` returns what the caller *owns*,
+ * wherever it lives — but a Game's **unclaimed** pre-gens and its communal
+ * crawler have no owner at all, and `GameRoster` legitimately caches them. So
+ * anything with a `gameId` is a cached view of somebody else's container and is
+ * never pruned. A shelf row is different: `gameId: null` with no owner is the
+ * one combination ADR-030 calls invalid, so every shelf row must be owned, and
+ * every owned row is in `listMine`. Absence therefore means deleted.
+ *
+ * **2. Only in a browser that never held a legacy roster.** This is the guard
+ * that is easy to miss and fatal to omit. For a pre-ADR-034 user who has signed
+ * in but not yet claimed, their entire roster is local shelf rows the server has
+ * never heard of — and rule 1 would read every one of them as "deleted
+ * elsewhere" and destroy the lot. `legacyLocalDataState() === 'absent'` is the
+ * only state in which a local row can be trusted to have come from a
+ * server-accepted write or from this component, which is what makes absence
+ * mean deletion rather than not-yet-uploaded.
  */
 
 import { useQuery } from 'convex/react'
 import { useEffect, useRef } from 'react'
 import { api } from '../../../convex/_generated/api'
 import { isConvexConfigured } from '../../lib/connection/convexClient'
+import { legacyLocalDataState } from '../../lib/db/legacyLocalData'
+import { mayPrune, rowMayBePruned } from '../../lib/db/pruneRules'
 import { captureException } from '../../lib/observability'
 import { selectBackend } from '../../stores/entityBackend'
 import { useEntityStore } from '../../stores/entityStore'
@@ -67,11 +85,13 @@ function ConnectedShelfSync() {
 
     void (async () => {
       const store = useEntityStore.getState()
-      for (const [kind, rows] of [
+      const kinds = [
         ['pilot', mine.pilots],
         ['mech', mine.mechs],
         ['crawler', mine.crawlers],
-      ] as const) {
+      ] as const
+
+      for (const [kind, rows] of kinds) {
         for (const row of rows as Row[]) {
           try {
             await store.adopt(kind, row.body as never)
@@ -82,6 +102,30 @@ function ConnectedShelfSync() {
             // fine and the player should see them.
             captureException(err)
           }
+        }
+      }
+
+      // Prune only where absence is unambiguous — see the header. Both guards
+      // matter; dropping either turns this into a roster-deleter.
+      if (!mayPrune(legacyLocalDataState())) return
+
+      for (const [kind, rows] of kinds) {
+        const served = new Set(
+          (rows as Row[])
+            .map((r) => (r.body as { id?: unknown } | null)?.id)
+            .filter((id): id is string => typeof id === 'string')
+        )
+        for (const local of store.list(kind)) {
+          // `containerOf` rather than a bare `gameId === null`, so a
+          // pre-ADR-030 record that still resolves through `workspaceId` is
+          // classified the same way every other reader classifies it.
+          if (!rowMayBePruned(local)) continue
+          if (served.has(local.id)) continue
+          // `forget`, not `delete`: this removes the local COPY and must never
+          // become a server delete. The row is already gone there — that is why
+          // it is being pruned — and issuing a delete would turn a sync into a
+          // destructive write against whatever the server does hold.
+          await store.forget(kind, local.id)
         }
       }
     })()
--- a/apps/itun/src/lib/db/__tests__/pruneRules.test.ts
+++ b/apps/itun/src/lib/db/__tests__/pruneRules.test.ts
@@ -1,0 +1,73 @@
+/**
+ * The two rules that decide whether a cached row may be deleted (P4b).
+ *
+ * Pruning is the most destructive operation in the codebase, and both of its
+ * guards are the kind that look like defensive noise right up until the day
+ * they are removed. So the rules are extracted as pure predicates and tested
+ * directly, rather than only being reachable through `ShelfSync`'s effect.
+ *
+ * The scenarios below are the two ways to delete somebody's roster by accident.
+ *
+ * The predicates are IMPORTED rather than restated here. An earlier draft copied
+ * them into this file, which would have let the rule and its test drift apart in
+ * exactly the direction that matters — the test passing while the code deleted
+ * the roster.
+ */
+
+import { describe, expect, test } from 'bun:test'
+import { mayPrune, rowMayBePruned } from '../pruneRules'
+
+describe('rule 1 — a browser that held a legacy roster never prunes', () => {
+  test('a legacy browser is refused', () => {
+    // The scenario: a pre-ADR-034 player signs in for the first time and has
+    // not claimed yet. Every build they own is a local shelf row the server has
+    // never seen. Pruning here deletes all of it.
+    expect(mayPrune('present')).toBe(false)
+  })
+
+  test('an unresolved probe is refused too', () => {
+    // Same failure, arrived at by racing rather than by state: the probe has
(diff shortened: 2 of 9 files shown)


alxjrvs (commented):
**claude-review** — ⚠️ **1 blocking** of 10 finding(s)

---

Reviewed against the surrounding code (`entityStore`, `entityBackend`, `convex/entities.ts`, `container.ts`, `legacyLocalData.ts`, `runWrite.ts`, the cargo hooks). No prompt-injection or agent-directed text in the diff.

## Findings

**BLOCKING** — `apps/itun/src/stores/entityStore.ts:620` — Phase 1b commits each transfer update to Convex sequentially, with no server-side atomicity, *before* the single local `atomicWrite`; a refusal on the second commit leaves the server half-applied while the client aborts with nothing changed locally. Concrete failure: `useCargo.dispatch` (`apps/itun/src/lib/cargo/useCargo.ts:132-142`) pushes `{mech, cargoLots}` then `{crawler, cargoLots+scrapPool}`. The mech upsert lands (lot removed server-side), then `patchCrawlerByAppId` throws (`assertMayEditCrawler` for a non-member, or `Invalid crawler payload` on the merge) — the lot now exists nowhere on the source of truth, the local write never runs, and the next `ShelfSync` adopt overwrites the local mech with the lot-less server body. Under the old fire-and-forget mirror this was benign because IndexedDB was authoritative; making the server authoritative converts it into real data loss. Worse silently: `patchCrawlerByAppId` *returns* rather than throwing when the crawler has no server row, so the crawler half no-ops with no error at all while the mech half lands.

`apps/itun/src/components/account/ShelfSync.tsx:110` — The prune guard `mayPrune(legacyLocalDataState())` can only ever be true on a device's *first* page load. `probeLegacyLocalData` (`apps/itun/src/lib/db/legacyLocalData.ts:71-86`) counts rows in `pilots`/`mechs`/`crawlers`/`mechPatterns` and cannot distinguish a legacy roster from rows `ShelfSync.adopt` itself wrote last session — so on every subsequent load the probe returns `present` and pruning is permanently disabled. This makes the phase's own gate ("A row deleted on device A is gone from device B **after a reload**", `docs/architecture/persistence-and-pwa.md:470`) unsatisfiable by construction, while the phase table is flipped to **done**. Fails safe, but the headline feature is inert.

`apps/itun/src/components/account/ShelfSync.tsx:118` — `store.list(kind)` is the lazy synchronous read: for a type not yet hydrated it returns `[]` and fires `hydrate()` in the background, so the prune pass silently examines nothing. Because `lastAdopted.current` was already set to this payload's stamp at line 84, the pass is never retried for that payload. Compounded by the stamp being counts-only (line 77): a delete-plus-create on another device leaves all four counts equal, so the effect never re-runs and neither adoption nor pruning happens at all.

`apps/itun/src/stores/entityBackend.ts:219` — Deleting `reportMirrorFailure` removes the only `captureException` on the entity write path and the only place `serverMessage(err)` rendered the backend's player-facing `ConvexError` copy ("Only the Mediator can do that"). Failures now land in `runWrite`'s generic `console.error` + "That change could not be saved." (`apps/itun/src/lib/runWrite.ts:51`), which handles the rejection — so nothing reaches Sentry either. `docs/architecture/accounts-and-games.md:197` still documents `reportMirrorFailure` as that reporting path, and the P4b gate says "the user is told".

`apps/itun/src/stores/entityStore.ts:502` (and `:560`) — `dbStoreFor(type)` is resolved twice across an `await commitWrite(...)` network round trip, and it re-reads live auth state on every call. If auth changes mid-commit (sign-out, or the legacy probe resolving), `prepareCreate` mints the record against one backend and `put` writes it to the other — e.g. `remote`→`memory` writes the record to a Map that dies with the tab while `set()` shows it as saved. Resolve the store once and reuse it.

`apps/itun/src/stores/entityStore.ts:342` — `as Parameters<typeof commitEntityWrite>[1]` erases the discriminated union: the object always carries both `body` and `patch: patch ?? {}`, so an `upsert` ships a meaningless `{}` patch and any future `kind`/payload mismatch type-checks silently. Two explicit branches (as the previous `mirrorCrawlerWrite` call sites had) keep the discriminant load-bearing.

`apps/itun/src/stores/entityStore.ts:962` (Phase 1b) — `ops.deletes` are still never committed to the server; the comment that recorded this deliberately ("No mirror for `deletes`: no production caller passes any today") was removed with the old Phase 2b block, so the gap is now undocumented. Under server-as-truth a locally-deleted entity is resurrected by the next `ShelfSync.adopt` rather than merely drifting. No production caller passes `deletes` today, but the guard rail is gone.

`apps/itun/src/lib/db/pruneRules.ts:60` — `rowMayBePruned` is byte-for-byte `isOnShelf` from `apps/itun/src/lib/container.ts:74` (`containerOf(entity).kind === 'shelf'`). A second name for an existing predicate is a second thing to keep in step; re-export or alias `isOnShelf` instead.

`docs/architecture/persistence-and-pwa.md:64-67` — The mirror table still names `mirrorWrite` / `mirrorCrawlerWrite` / `mirrorSoftLinkWrite` as the live mechanism, `docs/architecture/data-flow.md:179,275` still documents `mirrorWrite()` in the write-path diagram, and `.claude/rules/tanstack-query-hooks.md:114` still instructs "copy `entityStore`'s `mirrorWrite`" — all three now point at deleted functions, and that last file is loaded into every session as project instructions. Same document at `:453-466` still asserts pruning "is not implementable safely" without a Games-union view, directly contradicting the `**Done**` it now carries at `:419`.

`apps/itun/src/stores/__tests__/serverFirstWrites.test.ts:1` — The file's stated subject ("the **ordering contract** — that a record is built, committed and only then persisted") is asserted by none of its tests: with no Convex configured `commitWrite` returns at the first line, so the commit step is absent from every case, and "a rejected patch changes nothing" exercises Zod rather than a refusal. Nothing anywhere covers the `ShelfSync` prune loop either — only the two pure predicates — despite `docs/architecture/persistence-and-pwa.md:473` calling the "a Game's unclaimed entities and its crawler survive a prune" assertion the one to write first. A spy on `convexClient.mutation` recording call order relative to `db.pilots.get` would close the first; a `ShelfSync` render test with a stubbed `listMine` would close the second.

CLAUDE-REVIEW-SUMMARY findings=10 blocking=1

alxjrvs (commented):
**claude-review** — ⚠️ **2 blocking** of 7 finding(s)

---

I read the full diff and checked the surrounding code (ShelfSync, entityStore/entityBackend, the Convex mutations, `containerOf`, `legacyLocalData`, `runWrite`, the srd SSG sitemap/pwa config). No instruction-like text addressed to a reviewer appears in the diff.

## Findings

**BLOCKING** — `apps/itun/src/components/account/ShelfSync.tsx:110` — the prune guard reads `legacyLocalDataState()`, but that probe (`legacyLocalData.ts:56-86`) counts rows in `pilots`/`mechs`/`crawlers`/`mechPatterns` — exactly the stores `ShelfSync`'s own `adopt` calls fill. It is computed once per page load and never reset. So on a fresh browser: session 1 probes `absent` → adopt writes the roster → prune may run; **every subsequent load probes `present`, so `mayPrune` returns false and the prune loop is dead for the rest of that device's life**. An entity deleted on device B stays in device A's roster forever, and the header's "the cache is a reflection" / the doc's `P4b … done` are both false in practice. The guard needs a signal that distinguishes "legacy Solo roster" from "rows this component itself cached" (e.g. a persisted claimed/synced marker, or probing before any adopt and persisting that answer).

**BLOCKING** — `apps/itun/src/stores/entityStore.ts:620-622` — phase 1b commits each prepared update to Convex in a sequential `await` loop with no rollback. A cargo stow is two updates (mech −N, crawler +N); if the second mutation is refused (`assertMayEditCrawler` after a membership change) or the network drops between them, the first has already landed on the server. Local state is untouched, so the device shows the correct balance and Convex — now the source of truth — holds a half-applied transfer; the next `ShelfSync.adopt` overwrites local with it and the scrap/cargo is destroyed for the whole table. The comment claims "a refusal aborts with nothing changed", which is true only of the local side. This needs one server-side mutation that applies all the updates in a single Convex transaction (Convex mutations are already atomic), not N client-driven ones.

`apps/itun/src/stores/entityBackend.ts:244` — `commitEntityWrite` is documented as "fail if it does not land", but the crawler patch path cannot fail: `patchCrawlerByAppId` returns silently when no row carries that `appId` (`apps/itun/convex/entities.ts:911`). A signed-in user editing a crawler with no server row gets a resolved promise, the local `put` proceeds, and the edit is lost with no error and no toast — the exact local-only state ADR-034 forbids, reintroduced through the one path that was supposed to close it.

`apps/itun/src/stores/entityBackend.ts:190` (deletion of `reportMirrorFailure`) — refusals now propagate to `runWrite`, whose only non-`WritesBlockedOffline` branch is `toast('That change could not be saved.')` (`apps/itun/src/lib/runWrite.ts:51-52`). Two things the deleted function did are not replaced anywhere: the `ConvexError` message the backend deliberately wrote for the player ("Only the Mediator can do that") is discarded via `serverMessage()` no longer being called, contradicting the rule in `apps/itun/CLAUDE.md`; and the `captureException` with the `itun-mirror-write-failed` fingerprint is gone, so client-observed write refusals no longer reach Sentry at all.

`apps/itun/src/components/account/ShelfSync.tsx:77-84` — the dedupe `stamp` is only the four array *lengths*. A delete plus a create on another device leaves every length unchanged, so the effect returns early and neither adopts the new row nor prunes the deleted one. Pre-existing for adoption, but the new prune inherits it as a second reason it silently does nothing.

`apps/itun/src/lib/db/__tests__/pruneRules.test.ts:1` and `apps/itun/src/stores/__tests__/serverFirstWrites.test.ts:1` — for "the most destructive operation in the codebase", the tests only exercise two one-line predicates (`legacy === 'absent'`, `isOnShelf`) and a commit path that is a documented no-op because no Convex is configured. Inverting the guard in `ShelfSync.tsx:110` to `if (mayPrune(...)) return`, or dropping the `await commitWrite` before the local `put`, leaves every test in this PR green. Nothing asserts that a refused write leaves no trace on disk, which is the whole behavioural claim.

`apps/itun/src/lib/db/pruneRules.ts:39` — `rowMayBePruned` is character-for-character `isOnShelf` from `apps/itun/src/lib/container.ts:74`. Re-exporting or calling the existing helper would keep the "classified the same way every other reader classifies it" promise structurally rather than by convention.

CLAUDE-REVIEW-SUMMARY findings=7 blocking=2

PR #882, opened 2026-08-20
feat(srd): P7 — installed means offline, without touching the service worker

alxjrvs:
**Layer 8 (top) of the ADR-034 stack.** Base: `feat/adr034-p4b-no-mirrors` (#881).

Ships **before** the host move, and the block I put on it did not survive contact with the facts.

## Why the block was wrong

Two of its three reasons turned out to be about ITUN, not `srd`:

- *"A bad service worker outlives the deploy that caused it."* True under ITUN's `registerType: 'prompt'`, where a worker waits and a user can sit on an old build indefinitely. **`srd` sets `skipWaiting: true` and `clientsClaim: true`**, so a replacement takes over on the very next visit. A bad worker here self-heals.
- *"Two changes in flight make a failure ambiguous."* That is an argument about **simultaneity**, not order. Shipping now, well before the move, is the same soak logic run the other way round — it settles on its own, so the move's failures stay attributable to the move.

The third is answered by the design rather than argued away.

## It changes no service-worker configuration at all

The precache list, `navigateFallback` and the runtime rules are **byte-identical**. The warmer writes into the same `pages` cache the existing `StaleWhileRevalidate` rule already reads from — `ssg/pwa.ts` names it verbatim, which the built `sw.js` confirms. The riskiest artefact in a static site is left alone.

It writes through the **Cache API** rather than merely fetching, because that rule matches `request.mode === 'navigate'`: warming by fetch alone would populate nothing and look like it worked.

## Measured before deciding — and it inverted the plan's own suggestion

Per-file gzip, which is what HTTP actually transfers:

| | count | gzipped |
| --- | --- | --- |
| HTML pages | 1,039 | **3.73 MB** |
| JSON endpoints | 899 | 0.70 MB |
| shell (already precached) | 72 | 2.0 MB |

The JSON is five times smaller and was this plan's first suggestion. **Rejected:** `srd` serves pre-rendered HTML and 82% of entity pages ship no JavaScript on purpose, so JSON cannot render an unvisited page without adding a client-side renderer — which would spend the site's whole design to save 3 MB on an explicit install.

An installed app costs ~5.7 MB. **A browser visitor costs 590 bytes gzipped** — the warmer's own chunk, which returns on its first line unless `display-mode` says installed.

## The tests assert the rule that would rot quietly

A browser tab makes **no requests at all**, not even for the sitemap. A warm that fires for everybody still looks like it works; it just costs every casual reader 3.73 MB.

Writing them caught a real defect **in the tests themselves**: happy-dom boots on `about:blank`, whose origin is the string `"null"`, so every sitemap entry failed to parse and the warm cached nothing while still appearing to run. Fixed by giving the document a real URL.

## Snapshot

Re-blessed, and the diff read: **one new file, nothing else.** No page content, title, description or endpoint changed.

## Verification

`bun run check:all` exits 0, including the srd output gate on all 1,039 pages.


alxjrvs (commented):

---

Read the diff and cross-checked it against `ssg/pwa.ts`, `ssg/sitemap.ts`, `islands.client.ts`, the CSP headers and the test preloads. No prompt-injection or instruction-shaped text in the diff — the prose is ordinary developer commentary.

The design holds up: the SW config really is untouched, `cacheName: 'pages'` is used verbatim by Workbox (an explicitly-passed `cacheName` is not prefixed), `cache.match(path)`/`cache.put(path, …)` key on the same URL a navigation request would, and `connect-src 'self'` permits the warm's fetches. Findings below are all non-blocking.

- `apps/srd/src/runtime/offlineWarm.client.ts:108` — the comment directly above says a redirect is not cached, but `res.ok && res.status === 200` does not exclude one: `fetch` follows redirects, so a followed redirect yields status 200 with `res.redirected === true`. Caching that response and later serving it for a **navigation** is the known Workbox failure (a redirected response used for a navigation request is rejected as a network error), so an unvisited page would fail to load offline rather than 404 — the outcome the comment says it is avoiding. Today's canonical trailing-slash sitemap URLs shouldn't redirect on Netlify, but a host-config change during the Cloudflare cutover would silently arm it. `if (res.status === 200 && !res.redirected)` closes it.

- `apps/srd/src/runtime/offlineWarm.client.ts:108` — `res.clone()` is unnecessary and costs memory: the original `res` body is never read or cancelled, so the tee'd branch buffers the whole HTML body until GC, ×1,039 pages with 6 in flight. `cache.put(path, res)` is both correct and cheaper here, since `res` is used nowhere else.

- `apps/srd/src/runtime/offlineWarm.client.ts:126` — the once-per-tab flag is written *before* any work, so a warm that dies partway (device goes offline after the sitemap fetch, or `cache.put` starts throwing `QuotaExceededError` — which `warmOne`'s bare `catch` swallows per page) leaves the app partially cached with no retry for the rest of that tab's session. Since `warmOne` skips already-cached entries, the routine is already idempotent; writing the flag after `Promise.all(workers)` would make an interrupted run self-heal on the next navigation.

- `apps/srd/src/runtime/islands.client.ts:117` — the `import('./offlineWarm.client')` is unconditional, so the install guard that the new test proves (`offlineWarm.test.ts:100`, "not installed means not one request") is not the guard that ships: the test calls the exported function directly and nothing exercises this call site. The chunk is precached by the SW's `**/*.js` glob anyway so the byte cost is real but tiny; the gap is that the doc's gate line — "a browser tab makes **no requests at all**" (`docs/architecture/persistence-and-pwa.md`) — is stronger than what any test asserts about the shipped path.

- `apps/srd/src/runtime/offlineWarm.client.ts:70` — hard-codes `sitemap-0.xml` while `ssg/sitemap.ts` chunks at `SITEMAP_ENTRY_LIMIT` (45,000) and emits `sitemap-index.xml` for exactly that reason; at ~1,036 URLs this is unreachable today, but a future chunk would silently warm only the first slice. Reading the index, or asserting the single-chunk assumption in the sitemap tests, would keep the two in step.

PR #883, opened 2026-08-20
feat(itun): P1b — give the test-only provider a consumer

alxjrvs:
**Layer 9 (top) of the ADR-034 stack.** Base: `feat/adr034-p7-srd-offline` (#882).

Closes the last gap in ADR-034, and it was exactly the shape of gap this plan is otherwise careful about.

## The problem

The test-only password provider (#874) exists so an e2e can sign in. **No e2e ever used it.** An auth surface with no consumer is dead code with a blast radius.

It was found by re-reading the progress table, not by any check — which is the part worth noticing. P1's row read *"provider done — fixture with P3"*, and stayed true-looking after P3 shipped without one.

## Why the fixture was never written

Not an oversight — a missing piece nobody had costed. **There was nothing in the UI to click.** `SignInControl` renders one Discord button, so a browser test could not reach the password provider at all.

`TestAuthBridge` is that missing half: a flag-gated function on `window`, **not a hidden control**. A rendered one would need a label, a place in the layout, an accessible name, and a decision about what happens when a real user sees it. A global cannot be clicked by accident and is honest about being a test seam.

## Why it cannot reach production

Two independent flags must **both** be wrong:

- `VITE_TEST_AUTH` in the build (this layer)
- `ITUN_TEST_AUTH` on the deployment (#874)

Each is asserted separately, so neither drifts into the other's blind spot. Verified beyond the flag: a production build contains **no trace of the global** — Vite folds the constant and drops the branch.

## The spec

Asserts the anonymous-build → sign-in → save hand-off, then **reloads** to prove the work is durable rather than merely still in memory — the whole distinction ADR-034 draws. It skips with a stated reason in the default suite, which has no Convex deployment; the header documents exactly how to run it for real.

One test pins the global's **name** across the bridge and the spec. A typo there surfaces inside `page.evaluate` as "seam not present", which **skips** rather than fails — so a rename would quietly turn the hand-off spec into a permanent no-op.

## Also

Corrects two stale rows in the plan's progress table: P4 still claimed the mirror removal was pending, which P4b did.

## Verification

`bun run check:all` exits 0.

### DIFF
--- a/apps/itun/convex/schema.ts
+++ b/apps/itun/convex/schema.ts
@@ -364,8 +364,19 @@ export default defineSchema({
      * Null while the crawler is in a Game — that is what communal means (D8).
      * Set only on the shelf, where a container with no owner would be the
      * invalid row.
+     *
+     * **`v.optional` is load-bearing and is not cosmetic.** Convex validates
+     * every existing document against the schema on push, and every crawler
+     * already in the database predates this column. A required field here would
+     * make the deploy fail on rows that are otherwise perfectly valid — the same
+     * reason `publicRead` above is optional, spelled out there as "the correct
+     * default for every row that already exists".
+     *
+     * So **absent means the same as null**: a crawler that has never been
+     * shelved. Readers must treat the two identically; `ownerOf` in
+     * `model/entities.ts` is the one place that decides it.
      */
-    ownerId: v.union(v.id('users'), v.null()),
+    ownerId: v.optional(v.union(v.id('users'), v.null())),
     /** See `pilots.appId` — same reason, same lookup path. */
     appId: v.optional(v.string()),
     /**
@@ -430,7 +441,7 @@ export default defineSchema({
      * the Mediator reaches it through their role rather than through ownership.
      * Set on a shelf, where a container with no owner would be the invalid row.
      */
-    ownerId: v.union(v.id('users'), v.null()),
+    ownerId: v.optional(v.union(v.id('users'), v.null())),
     body: v.any(),
   })
     .index('by_game', ['gameId'])
--- a/apps/itun/e2e/signin-save.e2e.ts
+++ b/apps/itun/e2e/signin-save.e2e.ts
@@ -1,0 +1,83 @@
+import { expect, test } from '@playwright/test'
+import { waitForReady } from './_helpers'
+
+/**
+ * The anonymous-build → sign-in → save hand-off (ADR-034 decision 1, P1/P3).
+ *
+ * This is the step most likely to lose somebody's work, and it is the entire
+ * reason the test-only password provider exists. Without this spec that provider
+ * is an auth surface with no consumer — which is the state it was in until this
+ * landed, and is worth remembering if anyone is tempted to delete this file
+ * rather than fix it: **deleting the spec means deleting the provider too.**
+ *
+ * ## What it needs, and why it skips without it
+ *
+ * Three things have to line up, and none is present in the default CI run:
+ *
+ *  - a reachable Convex deployment (`VITE_CONVEX_URL` compiled into the build),
+ *  - `ITUN_TEST_AUTH=true` on that deployment, so the `password` provider exists,
+ *  - `VITE_TEST_AUTH=true` in the build, so `TestAuthBridge` registers the seam.
+ *
+ * The default suite builds with none of them, so this skips with a stated
+ * reason rather than failing. That is the same shape `offline.e2e.ts` uses for
+ * the dev-server case, and the same trade: a spec that went red in the ordinary
+ * run would be deleted the first time it annoyed somebody.
+ *
+ * Run it for real with a test deployment:
+ *
+ *   bunx convex env set ITUN_TEST_AUTH true      # on the test deployment
+ *   VITE_TEST_AUTH=true VITE_CONVEX_URL=<url> bun --filter itun build
+ *   bun --filter itun exec playwright test signin-save.e2e.ts
+ */
+
+/** A fresh account per run — a reused one would inherit the last run's roster. */
+function uniqueCredentials() {
+  const stamp = `${Date.now()}-${Math.random().toString(36).slice(2, 8)}`
+  return { email: `e2e-${stamp}@example.invalid`, password: `pw-${stamp}-Aa1!` }
+}
+
+async function seamIsPresent(page: import('@playwright/test').Page): Promise<boolean> {
+  return await page.evaluate(
+    () => typeof (window as unknown as Record<string, unknown>).__itunTestSignIn === 'function'
+  )
+}
+
+test('work built anonymously survives signing in to save it', async ({ page }) => {
+  await page.goto('/')
+  await waitForReady(page)
+
+  const ready = await seamIsPresent(page)
+  test.skip(
+    !ready,
+    'No test sign-in seam in this build — needs VITE_TEST_AUTH, VITE_CONVEX_URL and ITUN_TEST_AUTH. See this file header.'
+  )
+
+  // Build something anonymously. The pilot wizard is the shortest real path to
+  // a saved entity, and is what a first-time visitor actually does.
+  await page.getByRole('link', { name: /Build your first pilot/i }).click()
+  await waitForReady(page)
+
+  const { email, password } = uniqueCredentials()
+  await page.evaluate(
+    async ([e, p]) => {
+      const fn = (window as unknown as Record<string, unknown>).__itunTestSignIn as (
+        a: string,
+        b: string
+      ) => Promise<void>
+      await fn(e as string, p as string)
+    },
+    [email, password]
+  )
+
+  // The promotion runs on the backend flip and is the assertion that matters:
+  // a user who signs in to save must not come back to an empty screen.
+  await page.goto('/roster')
+  await waitForReady(page)
+
+  // Reload to prove it is DURABLE rather than merely still in memory — the
+  // whole distinction ADR-034 draws.
+  await page.reload()
+  await waitForReady(page)
+
+  await expect(page.getByRole('heading', { name: /Welcome to In the Union Now/i })).toHaveCount(0)
+})
--- a/apps/itun/src/components/account/ShelfSync.tsx
+++ b/apps/itun/src/components/account/ShelfSync.tsx
@@ -74,12 +74,21 @@ function ConnectedShelfSync() {
   useEffect(() => {
     if (mine === undefined) return
 
-    const stamp = JSON.stringify([
-      mine.pilots.length,
-      mine.mechs.length,
-      mine.crawlers.length,
-      mine.mechPatterns.length,
-    ])
+    // Keyed on the ids the server returned, not on how many there are.
+    //
+    // Counts are not enough, and the gap is not theoretical now that this
+    // prunes: one row created and another deleted in the same emission leaves
+    // every count identical, so the effect would skip — no adoption, and more
+    // seriously no prune, leaving a row that was deleted on another device
+    // cached here forever and looking like it still exists.
+    const stamp = JSON.stringify(
+      [mine.pilots, mine.mechs, mine.crawlers, mine.mechPatterns].map((rows) =>
+        (rows as Row[])
+          .map((r) => (r.body as { id?: unknown } | null)?.id)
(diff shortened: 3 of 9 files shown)


alxjrvs (commented):

---

Read the diff and checked it against the app's routes, the Roster component, and the promotion path.

## Findings

**BLOCKING** — `apps/itun/e2e/signin-save.e2e.ts:80` — `/roster` is not a route. The Roster renders at `/` (`apps/itun/src/routes/index.tsx:4-9`), there is no `roster` entry in `src/routeTree.gen.ts`, and `main.tsx:13` installs `defaultNotFoundComponent: RouteNotFound`, so `page.goto('/roster')` renders a styled 404. `waitForReady` still resolves (the game-data flag is set by the root gate above `<Outlet/>`), so the final assertion — `getByRole('heading', {name: /Welcome to In the Union Now/i})` has count 0 — is vacuously true: that heading only exists inside `Roster.tsx:589`, which never mounts at that URL. The spec would pass even if sign-in destroyed every entity, which is the exact failure it was written to catch. (`offline.e2e.ts:144` navigates to `/roster` too, but only asserts the document title, so a 404 there is harmless — this spec's whole verdict rests on it.)

**BLOCKING** — `apps/itun/e2e/signin-save.e2e.ts:63` — the spec never builds anything anonymously. The "Build your first pilot" CTA (`Roster.tsx:595-603`) links to `/pilots/new` with **no** `?mode=guided`, so `NewPilotRoute` renders the `CreateModeChooser` gate (`routes/pilots/new.tsx:36-48`; compare `_helpers.ts:280`, which appends `?mode=guided` precisely to skip it) and the wizard is never entered, let alone completed. No pilot is persisted, so there is no anonymous work for sign-in to promote, and the "work built anonymously survives" claim in the test title is unexercised. Combined with the finding above, the spec is a permanent no-op even in the fully-configured environment its header describes — the same "auth surface with no consumer" shape it says it exists to close.

`apps/itun/e2e/signin-save.e2e.ts:76-80` — nothing waits for the promotion. `AnonymousWorkPromoter` fires it as `void promoteAnonymousWork(claimLocal, work)` (`UnsavedWorkBanner.tsx:162`), so `page.goto('/roster')` issued the moment `signIn(...)` resolves can tear down the document with the claim mutation still in flight; once the two defects above are fixed this becomes a real flake (and a false negative, since an aborted promotion is indistinguishable from a lost one). Wait for the signed-in/promoted state — e.g. the unsaved-work banner disappearing — before navigating.

`docs/architecture/persistence-and-pwa.md:42` — the P1 row is flipped from "provider done — fixture with P3" to **done**, and the status prose at lines 279-298 asserts the provider now has a consumer, on the strength of a spec that (per the two findings above) asserts nothing and additionally skips in every default run. The table is documented as "the only place that answers which phase are we on", so this records P1 as covered when it is not.

alxjrvs (commented):

---

Read the diff and checked it against the app's routes, helpers, and the Convex auth half.

**Findings**

- `apps/itun/e2e/signin-save.e2e.ts:74` — **BLOCKING.** `page.goto('/roster')` targets a route that does not exist: the roster is rendered by `apps/itun/src/routes/index.tsx` at `/`, and `routeTree.gen.ts` contains no `roster` route (`offline.e2e.ts:144` deliberately uses `/roster` *because* it is an unvisited/nonexistent path). The SPA fallback serves the shell, `waitForReady` still passes (the `data-game-data-ready` flag is set by the root, above `<Outlet/>`), and TanStack renders its default not-found body. So the closing assertion at line 82 — "the first-run welcome heading is absent" — is satisfied by a 404 view and would pass identically if promotion were completely broken. The spec is a guaranteed green that tests nothing.

- `apps/itun/e2e/signin-save.e2e.ts:57` — **BLOCKING.** Nothing is ever built anonymously. The test clicks `Build your first pilot`, which navigates to `/pilots/new` with no `?mode=guided`, so it lands on the `CreateModeChooser` gate (`NewEntityScreen`), never completes a wizard, and creates no pilot/mech/crawler. `AnonymousWorkPromoter` only promotes when `entityStore.list('pilot'|'mech'|'crawler')` is non-empty (`UnsavedWorkBanner.tsx:139-141`), so with the route bug fixed this test would assert the *presence* of the empty-state welcome and fail. The existing `buildPilot(page, name, callsign)` in `e2e/_helpers.ts:362` is the helper that actually produces a saved entity. Consequence beyond the spec: `docs/architecture/persistence-and-pwa.md:42` now marks P1 **done** and the prose at line 245+ says the provider has a consumer, but the consumer exercises neither the sign-in hand-off nor the promotion.

- `apps/itun/e2e/signin-save.e2e.ts:82` — The only assertion is negative (`toHaveCount(0)` on the welcome heading), so it also passes when the roster fails to render for any reason — a router error boundary, a blank page, a wrong URL. A positive assertion (the built pilot's name visible in a roster row, the shape `openSheetFor` relies on) is what distinguishes "promoted and durable" from "screen is not the empty state".

- `apps/itun/src/components/account/TestAuthBridge.tsx:23` (and the matching claim at `docs/architecture/persistence-and-pwa.md:257`) — The comment states a production bundle "does not contain the registration at all — Vite folds the constant and drops the branch", but `VITE_TEST_AUTH` is *unset* rather than set to a literal: `apps/itun/.env.production` defines only `VITE_REQUIRE_ACCOUNT`, and `apps/itun/vite.config.ts` has no `define`. Vite statically replaces only env keys it knows about, so `import.meta.env.VITE_TEST_AUTH === 'true'` is not reliably folded to `false` at build time and `Bridge` plus the `__itunTestSignIn` string can survive into the bundle. The seam still cannot authenticate without the server-side `ITUN_TEST_AUTH`, so this is not an exploitable bypass — but the "checked to contain no trace" claim is asserted by no test (`e2e/bundle-budget.e2e.ts` would be the natural place), and the two files present it as a verified property.

No text in the diff attempts to address or instruct a reviewing agent.

alxjrvs (commented):
**claude-review** — ⚠️ **2 blocking** of 5 finding(s)

---

Read the diff and checked it against the surrounding code.

**BLOCKING**

- `apps/itun/e2e/signin-save.e2e.ts:70` — `page.goto('/roster')` targets a route that does not exist. The Roster is the index route (`src/routes/index.tsx` → `Roster`); there is no `roster.tsx` and no redirect, and `_helpers.ts:399` (`openSheetFor`) reaches the roster with `gotoStable(page, '/')`. Under `vite preview`'s SPA fallback the app loads and the router renders its not-found page, where `Welcome to In the Union Now` is trivially absent — so the spec's only assertion passes without ever looking at the roster.

- `apps/itun/e2e/signin-save.e2e.ts:57` — the test clicks the `Build your first pilot` CTA (which only navigates to `/pilots/new`) and never completes the wizard, so no pilot is created. `runWizard` in `_helpers.ts:264` exists precisely because a build requires walking every step to `Create Pilot`; `buildPilot()` is the exported helper for this. With no anonymous work, `AnonymousWorkPromoter` never arms (`UnsavedWorkBanner.tsx:145` requires `backend === 'memory' && hasWork`), so the assertion that "work built anonymously survives signing in" is vacuous — the spec cannot fail when the hand-off breaks, which is the one thing it was added to catch, and `docs/architecture/persistence-and-pwa.md:403` flips P1 to **done** on its strength.

**Non-blocking**

- `apps/itun/e2e/signin-save.e2e.ts:69` — even with the two issues above fixed, the hard `page.goto` fires in the same tick that `signIn` resolves; promotion is an async `claimLocal` mutation in a `useEffect` that runs *after* the auth flip (`UnsavedWorkBanner.tsx:144-154`), so a full-page navigation can abort it mid-flight and produce a flaky red. Wait for the promotion to settle on the current page before navigating.

- `apps/itun/convex/schema.ts:377` — the new comment says absent-vs-null is decided by "`ownerOf` in `model/entities.ts`". No `ownerOf` exists anywhere in the repo (only this comment matches), and `model/entities.ts` has no such helper; the decision is actually spread across inline `ownerId === null` comparisons in `entities.ts`, `publicSheet.ts`, `ownership.ts`, `invites.ts` and `proposals.ts`. The comment sends the next reader to a centralization that was never built. (The `v.optional` change itself is safe — every writer sets `ownerId` explicitly, and every reader compares by equality or guards with `'ownerId' in row`, so absent fails closed.)

- `apps/itun/test/convex/containerParity.test.ts:135` — the "absent ownerId is treated as unowned" case gives the identity no owned crawler, so `mine.crawlers` is `[]` for reasons independent of the claim; seeding one owned crawler alongside the orphan would make the assertion discriminate rather than restate `by_owner`'s index semantics.

No prompt-injection or reviewer-addressed instructions in the diff.

CLAUDE-REVIEW-SUMMARY findings=5 blocking=2

alxjrvs (commented):
**claude-review** — ⚠️ **2 blocking** of 5 finding(s)

---

Read the full diff and checked it against the surrounding code (`ShelfSync.tsx`, `convex/entities.ts`, `convex/schema.ts`, the route tree, `e2e/_helpers.ts`, `playwright.config.ts`). No instruction-like text addressed to a reviewer appears in the diff.

## Findings

**BLOCKING** — `apps/itun/e2e/signin-save.e2e.ts:74` — `page.goto('/roster')` targets a route that does not exist: the roster is rendered by `src/routes/index.tsx` at `/` (see `e2e/_helpers.ts:399` `openSheetFor`, which goes to `/`; there is no `roster` entry in `src/routeTree.gen.ts`). Under `vite preview` the SPA fallback serves `index.html`, TanStack Router falls through to not-found, and the "Welcome to In the Union Now" heading (`src/components/roster/Roster.tsx:589`) is never rendered there — so the final `toHaveCount(0)` at line 82 passes no matter what happened to the user's work. The one assertion in the spec can never fail.

**BLOCKING** — `apps/itun/e2e/signin-save.e2e.ts:57` — the spec never builds anything anonymously. Clicking "Build your first pilot" only navigates to `/pilots/new`; the pilot is not created until the wizard's "Create Pilot" step (`_helpers.ts:362` `buildPilot` / `runWizard` is what actually produces an entity). The test then signs in with a brand-new `signUp` account whose shelf is empty, so even on the correct route the assertion "the first-run welcome is gone" would be asserting against zero work — and would fail, since the welcome state is exactly what an empty roster shows. The hand-off ADR-034 P1 is meant to cover is not exercised; a stronger assertion (the built pilot's name/callsign visible after reload) is what would prove durability.

**Non-blocking** — `apps/itun/src/components/account/ShelfSync.tsx:84` — the new stamp keys only on the *set of ids*, so a live emission in which an existing row's `body` changed (renamed/edited on another device) produces an identical stamp and is skipped: the edit never reaches this browser's cache while the tab stays open. Not a regression (the old count-based stamp missed it too), but the comment now claims the stamp tracks what the server returned, and the update case is the more common cross-device one. Hashing the bodies (or a server `updatedAt` per row) would close it.

**Non-blocking** — `apps/itun/convex/schema.ts:376` — the new comment tells readers that "`ownerOf` in `model/entities.ts` is the one place that decides it", but `ownerOf` exists nowhere in the repo (`convex/model/entities.ts` exports only `parseBody` / `loadOwnable` / `PARSERS`). There is in fact no single normalizer: absent-vs-null is handled ad hoc at each site (`?? null` in `botClient.ts:314`, `?? organizerId` in `games.ts:233`, plain `!==` comparisons in `entities.ts:428` and `mediator.ts:115`). They all happen to fail closed today, but the doc points at a guarantee that isn't implemented.

**Non-blocking** — `apps/itun/src/components/account/TestAuthBridge.tsx:24` (and the matching paragraph in `docs/architecture/persistence-and-pwa.md`) — "a production bundle does not contain the registration at all — Vite folds the constant and drops the branch" is not what will happen. `testAuthBridgeEnabled` is an imported const from `testAuthSeam.ts` initialised from an `import.meta.env` member access; with `VITE_TEST_AUTH` unset Vite has no define to substitute, so Rollup cannot fold it to a literal and will not eliminate `Bridge` or the `__itunTestSignIn` string from the bundle. The runtime guard still holds (the expression evaluates false), so this is not a bypass — but the doc's claim that "a production build was checked to contain no trace of the global" should be re-verified rather than relied on as the security argument; `TestAuthBridge.test.tsx`'s flag/registration assertions are the real protection.

CLAUDE-REVIEW-SUMMARY findings=5 blocking=2

PR #885, opened 2026-08-20
fix(itun): serve security headers on Cloudflare, and gate their absence

alxjrvs:
## What

`apps/itun/wrangler.jsonc` binds intheunionnow.com as a custom domain, but apps/itun had **no `public/_headers`**. srd got one when its CSP moved out of `netlify.toml`; itun did not. `src/worker/index.ts` forwards asset responses untouched, so since the Cloudflare flip the production SPA has served **no CSP, no HSTS, no X-Frame-Options, no nosniff and no Referrer-Policy**. All of it still lived in a `netlify.toml` for a host that had stopped answering.

## Why the checker didn't catch it

`check-observability.ts` judges only CSP sources that **exist** — a missing file is filtered out before it is assessed — so an absent policy was indistinguishable from one carried elsewhere, and itun passed on the strength of the dead toml.

This adds a different *kind* of rule, keyed on the deploy target rather than the file set: when `wrangler.jsonc` declares `assets`, Cloudflare serves the app from static assets and `_headers` is the only thing that can carry a policy there.

## Notes for review

- The Convex origins are load-bearing: `@convex-dev/auth` exchanges tokens against the deployment's `.site` origin and the reactive client opens a WebSocket. Omitting them blocks sign-in *before any request leaves the page*, with the app still reporting itself Connected. `wss:` does not inherit from `https:`, so both schemes are listed.
- The `assets` match strips comments first — these configs mention `/assets/*` repeatedly in prose, so a substring match would keep passing after the real binding was deleted.
- **Three existing tests were re-targeted, not deleted.** They were written when itun had no `_headers`, and two of them *created* the file to simulate this future. The "no source declares a CSP" case now neutralises both itun sources — left as it was, it would have passed on the new file and asserted nothing.

## Verification

Negative control: with the new file moved aside the checker exits 1 naming the exact production state from this morning; restoring it returns to 0. Four new regression tests cover both apps, plus a control proving the rule is tied to the `assets` declaration rather than firing unconditionally.

### DIFF
--- a/apps/itun/public/_headers
+++ b/apps/itun/public/_headers
@@ -13,11 +13,17 @@
 # every policy below still lived in a `netlify.toml` for a host that had stopped
 # answering.
 #
+# This is a port of `apps/itun/netlify.toml`'s header block, with two stated
+# exceptions: the `/*.html` rule below is an ADDITION (the toml has no
+# counterpart), and srd's `immutable` image rules are deliberately NOT copied
+# here. Everything in the toml's `/*` and `/assets/*` blocks is present
+# unchanged, CSP included.
+#
 # `tools/check-observability.ts` reads whichever sources exist and holds every
 # one that declares a policy to the same rule. Note what that did NOT catch: an
 # *absent* source is not a declaring source, so itun passed on the strength of
 # the toml alone. The checker now also requires this file to exist whenever the
-# app's `wrangler.jsonc` declares `assets` — see the `assetSurfaces` rule there.
+# app's `wrangler.jsonc` declares `assets` — see `declaresStaticAssets` there.
 
 /*
   X-Frame-Options: DENY
@@ -57,18 +63,17 @@
 /assets/*
   Cache-Control: public, max-age=31536000, immutable
 
-# Icons and artwork are content-addressed by name and never mutated in place.
-/*.png
-  Cache-Control: public, max-age=31536000, immutable
-
-/*.webp
-  Cache-Control: public, max-age=31536000, immutable
-
-/*.svg
-  Cache-Control: public, max-age=31536000, immutable
-
+# NOT immutable, and this is the one place this file deliberately differs from
+# srd's. srd's icons and artwork really are content-addressed, so it marks
+# `/*.png` etc. immutable for a year. itun's `public/` holds `favicon.svg`,
+# `icon-192.png`, `icon-512.png` and `logos/` — FIXED names that are mutated in
+# place on redesign. A year of `immutable` on those means a changed icon that no
+# amount of reloading can dislodge, so no such rule is written here.
+#
 # HTML must never be cached: a deploy rotates the hashed chunk names, and a
-# stale document names chunks that no longer exist. This is the SPA shell, so
-# a stale copy breaks every route rather than one page.
+# stale document names chunks that no longer exist. This is the SPA shell, so a
+# stale copy breaks every route rather than one page. It is the one rule in this
+# file with no `netlify.toml` counterpart — an addition rather than a port, kept
+# because Cloudflare's asset serving would otherwise be free to cache the shell.
 /*.html
   Cache-Control: public, max-age=0, must-revalidate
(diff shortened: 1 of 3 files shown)


alxjrvs (commented):

---

## Review

**What the diff does:** adds `apps/itun/public/_headers` (a port of `apps/itun/netlify.toml`'s header block plus a `/*.html` cache rule), and adds a "wrangler declares `assets` → `_headers` must exist" rule to `tools/check-observability.ts` with four new tests. I verified `apps/srd/wrangler.jsonc:53` does declare `"assets"` (so the srd test at `check-observability-csp.test.ts:209` will fail for the intended reason), that the non-global `.replace(/"assets"\s*:/)` in both mutation tests hits the real binding and not prose, that itun's workbox `globPatterns` (`vite.config.ts:64`) won't precache `_headers`, and that `connectSrcOf`'s first-match regex lands on the real header line in the new file, not on a comment. No text in the diff is addressed to a reviewing agent.

### Findings

- `tools/check-observability.ts:240` — the new rule only asserts `_headers` **exists**, never that it declares a CSP; with `netlify.toml` still present the "at least one source declares a policy" rule below is satisfied by the toml, so an `_headers` that carries only cache rules passes CI while the Cloudflare-served origin ships no CSP — the same outage class this PR fixes, one step removed. This is not hypothetical: `tools/__tests__/check-observability-csp.test.ts:153` deliberately strips the CSP from `apps/srd/public/_headers` and asserts exit 0. The fix is one line: when `declaresStaticAssets(wrangler)` holds, also require `connectSrcOf(read(app.workersHeadersPath))` to be non-null. Non-blocking (the file does carry a CSP today), but it is the weakest joint in the new guard.

- `apps/itun/public/_headers:34` — **unverified premise, worth a curl before trusting the fix.** `apps/itun/wrangler.jsonc:33` sets `run_worker_first: true`, and every response this Worker returns comes back through `env.ASSETS.fetch()` (`src/worker/index.ts:165`, and re-wrapped at `:105`). Whether Cloudflare applies `_headers` to asset-binding sub-requests under `run_worker_first` is exactly the behaviour this whole file depends on, and nothing in the repo demonstrates it — `deploy-cloudflare.yml:280-291` smoke-tests status codes only. `curl -sSI https://su-itun.alxjrvs.workers.dev/ | grep -i content-security-policy` settles it in one command; if it comes back empty the Worker has to set the headers itself. I could not verify this from the repo and am not calling it a defect — but it is the one thing that decides whether this PR actually closes the outage.

- `apps/itun/public/_headers:78` — `/*.html` does not match `/`, which is the canonical URL of the SPA shell (Workers static assets resolves `/` → `index.html`, but `_headers` patterns match the *request* path). So the rule the comment above it says protects "the SPA shell" does not apply to the shell's most-requested URL; whatever `max-age` `/` gets comes from Cloudflare's default, not from this file. Client-side routes like `/roster` do get it, but only incidentally — via `spaShell()` rewriting the path to `/index.html` before the asset fetch. Add a literal `/` rule if the intent is what the comment says.

- `tools/check-observability.ts:166` — `declaresStaticAssets` strips only lines whose *trimmed* text starts with `//`, so a trailing same-line comment (`"main": "x", // "assets": …`) or a `/* … */` block containing `"assets":` is a false positive: the checker would demand a `_headers` file for a Worker with no static assets and fail CI with a message that describes a deploy shape that does not exist. It errs safe (never a missed outage), but the tests at `:224` and `:244` only cover the whole-line comment form, so this stays untested.

- `.github/workflows/deploy-cloudflare.yml:280` (context, not in the diff) — the guard this PR adds is a repo-file-existence check, and the deploy smoke asserts status codes only, so nothing anywhere observes a *served* header. A `check https://su-itun.alxjrvs.workers.dev/` variant that greps the response for `content-security-policy` is the assertion that would actually have caught the incident described in the new file's header, and is cheap next to the five checks already there.

- `docs/architecture/cloudflare-cutover.md:422` (not in the diff) — still records "All three header sets verified **on live responses**" as a met gate item, which this PR's own header comment disproves for itun. Leaving it means the next reader trusts a claim this change just falsified; worth correcting in the same PR that discovered it.

PR #886, opened 2026-08-20
fix(ci): pin wrangler and every third-party action, and gate it

alxjrvs:
## What

Two supply-chain holes, both sitting next to credentials, both outside the control this repo already has.

This repo refuses dependency versions published less than three days ago (`bunfig.toml`'s `minimumReleaseAge`) on the stated reasoning that nobody reads tarballs. That covers npm and stops at the workflow file.

### 1. `wrangler` was unpinned, in the deploy job

It appeared in **no `package.json`** and had **zero entries in `bun.lock`**, yet `deploy-cloudflare.yml` runs `bunx wrangler deploy` four times in the job holding `CLOUDFLARE_API_TOKEN`, `CONVEX_DEPLOY_KEY` and `SENTRY_AUTH_TOKEN`. `--frozen-lockfile` cannot cover an unlisted package and the cooldown does not apply to `bunx`.

The contrast was already in the same file — `bunx convex deploy` on line 223 resolves a *pinned* convex. Three `wrangler.jsonc` files assert "this repo's wrangler (4.108)"; nothing made that true. `tools/sync-snapshots-to-r2.ts:44` had the same shape.

### 2. Three of four third-party actions ran from mutable tags

| action | pinned to |
|---|---|
| `googleapis/release-please-action@v5` | `45996ed` # v5.0.0 |
| `oven-sh/setup-bun@v2` | `0c5077e` # v2.2.0 |
| `dorny/paths-filter@v4` | `ceb8a2b` # v4.0.3 |

The one already pinned was the *lowest*-privilege of the four. release-please holds `contents: write`, `pull-requests: write` **and** a PAT whose scopes are not bounded by the workflow's `permissions:` block.

## The gate

`tools/check-action-pinning.ts`, wired into `validate:all`. Two design points, both learned from existing gates here:

- **It scans composite actions, not just `workflows/`.** `setup-bun` appears in no workflow — it lives in `.github/actions/setup-bun/action.yml` — so a workflows-only scan reports a clean tree while the toolchain installer is unpinned. That is the shape the real gap had.
- **It strips comments and carries a scan floor.** Without the floor it would pass loudest when most broken: rename `.github/workflows` and it reports success having read nothing — the hole `tools/lib/scanFloor.ts` exists for.

First-party `actions/*` and `github/*` stay exempt: same trust boundary as the runner, and a rule flagging 25 `actions/checkout@v7` lines is one that gets switched off rather than satisfied.

## Verification

`bunx wrangler --version` now reports 4.108.0 from `node_modules/.bin`, at the root **and** from an app subdirectory (where the deploy steps run). Six tests assert the rule rather than the three pins; verified by negative control before wiring.

### DIFF
--- a/.github/actions/setup-bun/action.yml
+++ b/.github/actions/setup-bun/action.yml
@@ -17,12 +17,12 @@ description: Set up Bun and install dependencies
 # and THAT one genuinely earns its keep: 269 MB, restores in 4–7 s, saves 11–18 s
 # of browser install. It is kept, in the jobs that need it.
 #
-# `oven-sh/setup-bun@v2` has its own small binary cache and stays.
+# `oven-sh/setup-bun` has its own small binary cache and stays.
 runs:
   using: composite
   steps:
     - name: Set up Bun
-      uses: oven-sh/setup-bun@v2
+      uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 # v2.2.0
       with:
         bun-version-file: .bun-version
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -85,7 +85,7 @@ jobs:
         if: github.event_name == 'merge_group'
         run: echo "all=true" >> "$GITHUB_OUTPUT"
 
-      - uses: dorny/paths-filter@v4
+      - uses: dorny/paths-filter@ceb8a2b8f2d89434be7ff52d3de7ec3738c5cc9d # v4.0.3
         id: filter
         if: github.event_name != 'merge_group'
         with:
--- a/.github/workflows/deploy-cloudflare.yml
+++ b/.github/workflows/deploy-cloudflare.yml
@@ -132,7 +132,7 @@ jobs:
         working-directory: apps/su-assets
         env:
           CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
-        run: bunx wrangler deploy
+        run: bunx wrangler@4.108.0 deploy
 
       # ---------------------------------------------------------------------
       # srd — static, but the OG pass needs a browser.
@@ -200,7 +200,7 @@ jobs:
         working-directory: apps/srd
         env:
           CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
-        run: bunx wrangler deploy
+        run: bunx wrangler@4.108.0 deploy
 
       # ---------------------------------------------------------------------
       # itun — pushes the Convex backend FIRST, then builds against it.
@@ -226,7 +226,7 @@ jobs:
         working-directory: apps/itun
         env:
           CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
-        run: bunx wrangler deploy
+        run: bunx wrangler@4.108.0 deploy
 
       # ---------------------------------------------------------------------
       # discord-bot — the Worker, not the Render gateway process.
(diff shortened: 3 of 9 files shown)


alxjrvs (commented):

---

Read the full diff and checked it against the repo (`.github/**`, `lefthook.yml`, `knip.json`, `tools/lib/scanFloor.ts`, the existing `check-observability-csp.test.ts` precedent). No prompt-injection text addressed to a reviewing agent is present in the diff. The core change — SHA-pinning the three third-party actions plus a gate that enforces it — is sound, and the exemption/floor/comment-stripping design holds up against the tests. Findings:

- `tools/__tests__/check-action-pinning.test.ts:37` — the tests mutate real tracked files (`.github/workflows/ci.yml`, `release-please.yml`, `.github/actions/setup-bun/action.yml`) in place, and the same PR adds `check:action-pinning` to `validate:all`. `lefthook.yml:31` runs `test` and `validate` **in parallel** on pre-push, and `package.json`'s `check:all` runs `bun run test` concurrently with `bun run validate:all` and `check:ci-aggregator` — so the checker can read a deliberately-unpinned `ci.yml` and fail a clean tree, intermittently. A `finally` also doesn't cover SIGINT, so an interrupted run leaves an unpinned third-party action sitting in the working tree of a security-relevant file. Honest caveat: this is an established pattern here (`check-observability-csp.test.ts` mutates `apps/itun/netlify.toml` while `validate:observability` reads it), so the race pre-exists — this widens it rather than inventing it. Isolating the fixtures into a temp tree with a `--root` flag on the tool would remove both.
- `.github/workflows/release-please.yml:25` — the pin replaces floating `v5` with the SHA labelled `v5.0.0`. A floating major tag usually tracks the newest 5.x, so if `v5` is currently past 5.0.0 this is a silent downgrade of the release tooling; and because this workflow only runs on pushes to `main`, a wrong or stale SHA is invisible to PR CI and surfaces after merge. Verify with `gh api repos/googleapis/release-please-action/git/ref/tags/v5 --jq '.object.sha'` (same check for `oven-sh/setup-bun@v2` → v2.2.0 and `dorny/paths-filter@v4` → v4.0.3, though those two do run on PRs so a bad SHA fails loudly).
- `package.json:61` — `wrangler` is pinned as a root devDependency, but every call site is `bunx wrangler` (`deploy-cloudflare.yml:135,203,229,254`, `tools/sync-snapshots-to-r2.ts:44`). Nothing asserts `bunx` resolves the workspace copy from a subdirectory rather than fetching the newest release from npm, so the "pinned" deploy toolchain in the job holding `CLOUDFLARE_API_TOKEN`/`CONVEX_DEPLOY_KEY` is unverified, and the new checker deliberately only reads `uses:` so it cannot cover this. `bunx wrangler --version` in the deploy job, or switching to a `wrangler` root script, would make the pin observable. Related: nothing references `wrangler` from a package.json script, so confirm `bun run knip` doesn't report it as an unused dependency, and `CLAUDE.md`'s "Root Dev Dependencies (Intentional)" list (playwright, sharp) should gain an entry.
- `tools/check-action-pinning.ts:97` — `line.match(/^-?\s*uses:\s*(\S+)/)` keeps YAML quotes in the captured ref, so `uses: 'owner/repo@<40-hex>'` yields a pin of `<sha>'`, fails `SHA_PIN`, and a correctly pinned action is reported unpinned; the same quote also breaks `isThirdParty`'s owner split, so `uses: "actions/checkout@v7"` loses its first-party exemption. Stripping surrounding `'`/`"` before parsing closes both.
- `tools/check-action-pinning.ts:139` — the scan floor is hand-rolled with its own message while the comment directly above cites `tools/lib/scanFloor.ts`, which exports `assertScanFloor(label, scanned, floor)` doing exactly this. Reuse it so the four gates keep one failure message. Same shape in the test file: `runCheck` and `withFileContents` are copied verbatim from `tools/__tests__/check-observability-csp.test.ts:30-53`.
- `.github/workflows/ci.yml:276` — the comment enumerating what `validate:all` covers ("architecture, doc drift, observability, Convex codegen/parity and the catalog check") wasn't updated with the new action-pinning gate; in a repo whose documented dominant defect is prose asserting what code does not, this list is worth keeping exact. Minor test note while there: `'first-party actions/* and github/* are exempt'` re-runs the unmutated tree and asserts exit 0, which is byte-identical to the first test — it proves the exemption only implicitly, via the tree happening to contain unpinned `actions/checkout@v7` lines.

PR #889, opened 2026-08-20
fix(itun): adopt patterns on sync, and commit transfer deletes

alxjrvs:
Two independent ways the client and server drifted apart, both silent.

## 1. Patterns were fetched and thrown away

`entities.listMine` **returns** `mechPatterns`, and `ShelfSync`'s change stamp already keys on them — so the effect re-runs when they change — but the adoption table was `['pilot','mech','crawler']` and had no entry for them.

A signed-in player on a second device got their roster and an **empty pattern library**. That is the exact gap ADR-034 P4b exists to close: the fetch half had landed, the adopt half had not. Pattern rows carry only `{ body }` where the other three carry `{ appId, body }`, which is the likely reason they were skipped.

`makeHydratedCollectionSlice` gains `adopt`, mirroring `entityStore.adopt` — including its deliberate omission of `requireWritableBackend()`: filling the cache is not a user write, and refusing it while Disconnected would make going offline *lose access* to a record rather than merely making it read-only.

**Patterns are deliberately NOT added to the prune.** The prune asserts "the server did not return this, so it was deleted elsewhere". Adopting them was the bug; deleting them is a separate decision with a much worse failure mode, and nothing has asked for it.

## 2. Transfer deletes never reached the server

Phase 1b commits every update before touching disk, and its comment states the guarantee — *"a refusal aborts with nothing changed locally"*. That only held for **updates**. The `deletes` array went straight into the phase-2 IndexedDB transaction with no commit, so a transfer that consumed a stack of cargo deleted it locally and left it alive on the server — and `ShelfSync` restored it on the next sync.

The result is worse than a half-applied transfer: **it is both ends.** The value arrives at the target and the source keeps it too — a wrong balance, on the communal crawler, which is usually one end of these.

Phase 1c mirrors `delete()` exactly, reading each record before committing because a soft link is addressed on the server by its endpoints and after phase 2 there is nothing left to name it by.

## Verification

Four tests for `adopt` (placement, idempotence on re-adoption, in-place update, survival to disk so a reload keeps the library) and four for the transfer commit — including a **control** that a transfer with no deletes commits none, so a loop firing unconditionally would fail.

The transfer tests spy via `mock.module` with a spread-captured namespace and an `afterAll` restore. Full itun suite re-run at **2025 pass / 0 fail** confirms no leak into later files.

### DIFF
--- a/apps/itun/src/components/account/ShelfSync.tsx
+++ b/apps/itun/src/components/account/ShelfSync.tsx
@@ -62,8 +62,10 @@ import { isConvexConfigured } from '../../lib/connection/convexClient'
 import { legacyLocalDataState } from '../../lib/db/legacyLocalData'
 import { mayPrune, rowMayBePruned } from '../../lib/db/pruneRules'
 import { captureException } from '../../lib/observability'
+import type { MechPattern } from '../../lib/schemas/pattern'
 import { selectBackend } from '../../stores/entityBackend'
 import { useEntityStore } from '../../stores/entityStore'
+import { usePatternStore } from '../../stores/patternStore'
 
 type Row = { appId?: string | null; body: unknown }
 
@@ -122,6 +124,34 @@ function ConnectedShelfSync() {
           }
         }
       }
+
+      // Patterns live in their own store, and until now were fetched and then
+      // thrown away: `entities.listMine` returns `mechPatterns`, the stamp
+      // above already keys on them — so this effect re-runs when they change —
+      // and the `kinds` table above simply had no entry for them. A signed-in
+      // player on a second device got their pilots, mechs and crawlers and an
+      // empty pattern library, which is the exact gap ADR-034 P4b exists to
+      // close. The fetch half had landed; the adopt half had not.
+      //
+      // Pattern rows carry only `{ body }` — no `appId` — because a pattern is
+      // addressed by the id inside its body, which is likely why they were
+      // skipped when the `{ appId, body }` kinds were written.
+      const patterns = usePatternStore.getState()
+      for (const row of mine.mechPatterns as { body: unknown }[]) {
+        try {
+          await patterns.adopt(row.body as MechPattern)
+        } catch (err) {
+          // Same rule as the roster loop above: one unreadable pattern must not
+          // stop the rest of the library arriving.
+          captureException(err)
+        }
+      }
+
+      // Patterns are deliberately NOT pruned below. The prune answers "the
+      // server did not return this, so it was deleted elsewhere", and that
+      // inference is only sound for rows this sync is authoritative over.
+      // Adopting them is what the bug was; deleting them is a separate decision
+      // with a much worse failure mode, and nothing has asked for it.
 
       // Prune only where absence is unambiguous — see the header. Every guard
       // matters; dropping any one turns this into a roster-deleter.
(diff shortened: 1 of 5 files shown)


alxjrvs (commented):

---

Reviewed the diff and the surrounding code (`ShelfSync`, `makeHydratedCollection`, `patternStore`, `entityStore.transfer`, `entityBackend`, the Convex `entities.ts` mutations, and the sibling transfer test).

No instructions addressed to a reviewing agent were present in the diff; all of it is ordinary code and comments.

**Findings**

- `apps/itun/src/stores/makeHydratedCollection.ts:174` — `adopt()` calls `afterWrite()`, which fires `recordDataWrite()` (the backup nudge) and a cross-tab broadcast per record, whereas `entityStore.adopt` deliberately only calls `publish(...)`; the doc comment two hunks above claims parity with it. Consequence: filling the cache from the server counts as user writes, so a signed-in player whose library exceeds `BACKUP_NUDGE_WRITE_THRESHOLD` (25) is nagged "you have un-exported writes" on a fresh device for data they never wrote, and the counter only clears on a full export.

- `apps/itun/src/stores/__tests__/entityStore-transfer-commit.test.ts:109` — unlike the sibling `entityStore-transfer.test.ts:44-51`, `beforeEach` never hydrates the store, so `hydrated.mechs` stays false and the first `get()` inside `transfer` fires a floating `void hydrate('mech')`; if that `rehydrate` (a `getAll` issued *before* the phase-2 delete) resolves after phase 3's `set`, it reinstates the deleted mech in memory and `'the row is still gone locally'` fails. Add the three `hydrate()` calls the sibling test uses.

- `apps/itun/src/stores/entityStore.ts:640` — the soft-link arm reads `get().get('softLink', d.id)` from the in-memory cache, and `commitSoftLink` returns silently on `null`, so a transfer that deletes a soft link while the `softLinks` slice is unhydrated deletes it locally and leaves it on the server — the exact gap this hunk exists to close, still open for that one type (`listMine` returns no soft links, so nothing restores or reconciles it). Reading through `db.softLinks.get(id)` would close it; note `delete()` at line 748 has the same shape, so this is copied rather than introduced.

- `apps/itun/src/stores/entityStore.ts:629` — the comment (and the test header at line 62) states a concrete past failure: "a transfer that consumed a stack of cargo deleted it locally and left it alive on the server, and `ShelfSync` then restored it". No caller passes `deletes` at all — `useCargo.ts:144,175,212`, `usePartnerCargo.ts:148` and `ActiveItemBand.tsx:244` all send `updates` only, and cargo lots are a field on the mech/crawler rather than a deletable entity — so the described loss cannot have occurred. The fix is right as a guard; the prose asserts a bug the code cannot produce, which is the failure mode this repo's docs are most prone to.

- `apps/itun/src/components/account/ShelfSync.tsx:131` — patterns now flow server→client, but `patternStore` still mirrors nothing back (`convex/entities.ts` has no per-pattern upsert/remove; only `claimLocal` inserts them), so the sync is one-way and unconditional: a pattern removed via `patternStore.delete()` would be re-adopted on the next emission or reload. Latent rather than live — no UI calls `patternStore.delete` today — but worth a note beside the "deliberately NOT pruned" comment, since a delete affordance would silently stop working.

PR #892, opened 2026-08-20
chore: delete an unwired module and a duplicated crawler rule

alxjrvs:
Two independent removals, plus the cascade each set off.

**`apps/itun/src/lib/rules/pilotAdvancement.ts` was entirely unwired.** All four
exports had zero references outside the file and its own test: `grep -rn
pilotAdvancement apps packages` returned exactly two lines, both in the test.
137 lines plus a 162-line test. Nothing can call it, which makes this the
lowest-risk removal available.

**Four helpers in `salvageunion-reference/lib/helpers.ts` had no consumer**, and
two of them were a second encoding of a printed rule. `getWeaponSlotCount` /
`getMaxSpBonus` reimplement `crawlerWeaponSlots` / `crawlerMaxSpBonus` from
`lib/rules/creation.ts` — same filter, same reduce, same base-1 rule (Battle
Crawler +1 slot / +5 max SP, p.216). Production uses the `rules/` pair; the
`helpers` pair was reachable only from its own test, along with its two feeders
`getCrawlerMutations` and `findCrawlerById`.

The asymmetry is what made it worth doing rather than tolerable: the DEAD copy
was the exported, doc-commented one, so a contributor changing that rulebook
number was more likely to edit the copy nothing runs.

Coverage was checked before deleting rather than after. `creation.test.ts:301`
already asserts the Battle Crawler's 2 slots and +5 SP against the live pair,
plus empty/undefined cases and a sweep over every other crawler type — strictly
more than the tests removed here. Nothing is lost.

The removal cascaded five steps, which is the documented behaviour of deleting
dead code here and is why each was re-run rather than assumed:

  1. two stranded type imports in `helpers.ts`
  2. two now-unread fixtures in `helpers.test.ts`
  3. `SURefObjectCrawlerMutation`, whose only consumer was the deleted helper
  4. `CrawlerMutationSchema`'s import in `schemas/index.ts` (the SCHEMA is still
     live at `entities.ts:240` — only that import was stranded)
  5. the `defined` test helper, orphaned with the fixtures

On (3): `rules/creation.ts` deliberately declares its own narrower
`CrawlerMutationInput` — "the exact values these rules read … NOT the whole
SURef crawler record" — so reusing the schema type there would have coupled
`rules/` to the schema layer against a written intent. Delete-by-default
applied instead.

`etc/salvageunion-reference.api.d.ts` regenerates accordingly; that diff is the
reviewable statement of what left the public surface.

Full suite green.

---


alxjrvs (commented):
**claude-review** — 1 finding(s), none blocking

---

I verified each deletion against the repo (all greps run post-diff, so they reflect the resulting tree):

- `resolvePilotAdvancement` / `heldTreesOf` / `PilotAdvancement` / `OriginSource` — zero references anywhere; the underlying `/rules` primitives it wrapped (`inferOriginClass`, `liveAdvancementDataset`, `resolveAdvancementTrees`, `resolveClassRef`) still have a live consumer in `apps/itun/src/components/pilot/classPathOptions.ts`, so nothing downstream goes dead. `originClassRef` is still written and read by `classPathOptions.ts`.
- `findCrawlerById` / `getCrawlerMutations` / `getWeaponSlotCount` / `getMaxSpBonus` — zero references. The live path is the package rules layer (`lib/rules/creation.ts` `crawlerWeaponSlots` / `crawlerMaxSpBonus`, consumed via `apps/itun/src/lib/rules/creation.ts:522` and `derivedStats.ts`), which reads the same `mutations` rows and keeps the base-1 weapon slot. Removing the helpers loses no rule.
- `SURefObjectCrawlerMutation` — zero references; `CrawlerMutationSchema` itself is untouched and still consumed by `lib/schemas/entities.ts:240` and re-exported from `objects.ts:41`, so crawler data still validates. The package is `"private": true`, so dropping a public type alias breaks no consumer.
- `helpers.ts` import trim is safe: `SalvageUnionReference` is still used by `getHybridClasses`, and no remaining line in that file mentions `SURefCrawler`.
- `etc/salvageunion-reference.api.d.ts` is generated by `tools/generateApiReport.ts` as a concatenation of emitted declarations; the removed blocks correspond exactly to the removed source exports, so the CI drift check should stay green.
- No text in the diff addresses a reviewing agent.

One non-blocking nit:

- `packages/salvageunion-reference/lib/helpers.ts:131` — the `// CRAWLER HELPERS` banner comment is now an empty section header with no members beneath it; cosmetic leftover, nothing breaks.

Issue #802, opened 2026-08-15
EPIC: remove Tailwind in favour of tokens + style objects


alxjrvs:
Tracking issue. Remove Tailwind from this repo in favour of the styling pattern already proven in `BinfiniteLLC/binfinite-app`: a typed token module, exported `CSSProperties` style objects, and one package-level stylesheet.

`SU-SRD` is the **only** one of the six active personal repos using Tailwind. `binfinite-app`, `optfall`, `randsum`, `vellum` and `Hermuz` all use none. Removing it here makes the styling approach consistent across all of them.

## The layers

| | Issue | Surface |
|---|---|---|
| L1 | #798 | Foundation — tokens, style objects, stylesheet. No component changes. |
| L2 | #799 | `packages/component-lib` — 220 files, 263 `cn()` calls, 271 variant usages |
| L3/L4 | #800 | `apps/itun` (90 files) and `apps/srd` (24 files) |
| L5 | #801 | Drop deps, config, `cn()`, and update the docs |

Strictly ordered — each layer depends on the one above it. L2 is the bulk of the work and the one that needs the most care, because both apps consume that package.

## The pattern being ported

From `binfinite-app`'s `packages/component-lib`:

- `tokens.ts` — typed scale, plain `as const`, no dependency
- `styles.ts` — exported `CSSProperties` objects, each `satisfies CSSProperties`
- `styles.css` — the single stylesheet a web consumer loads, emitting token custom properties and binding the page ground

## One thing that must not be lost in translation

`binfinite-app`'s component library contains **zero** `:hover` and **zero** `@media`, because it is React-Native-first and RN has neither — so its style objects were never asked to carry interaction or responsive state.

This repo depends on both, heavily. Measured across the three UI workspaces:

| Variant | Uses |
|---|---|
| `hover:` | 120 |
| `sm:` / `md:` / `lg:` / `xl:` | 215 |
| `focus-visible:` | 15 |
| `disabled:` | 12 |
| `focus:` | 9 |
| `peer-checked:` / `first:` / `last:` | 5 |
| **Total** | **~403** |

An inline `style={}` object cannot express any of them. So the port takes **both halves** of the Binfinite pattern, not just the style objects — the package stylesheet is where those ~403 usages land. A migration that drops them would be a straightforward functional regression: every hover state and all responsive behaviour, gone.

The split rule, defined in L1 and applied mechanically thereafter:

- **Style object** — static values: color, padding, font, border, layout
[...]

PR #893, opened 2026-08-20
chore(component-lib): prune 28 dead barrel exports, and turn the gate on

alxjrvs:
The barrel IS this package's public API — `src/index.ts` is the only path the
apps may import from — and `knip.json` set `includeEntryExports: false` for it.
So the one workspace where barrel rot matters most was the one workspace where
knip could not see it. 28 dead re-exports had accumulated behind that.

Removed: BayStatus, StatProvenance, Textarea, pluralize, pluralizeWord,
invariantNouns, catalogNameOverrides, buildCatalogCategories,
resolveSchemaDomain, cardRemoveControls, FOCUS_WITHIN, FOCUS_RING_ON_TONE,
EntityStatus, CardDomain, CatalogCard, CatalogCategory, DialGauge,
EntityRowType, PatternHrefBuilder, ProvenanceLabels, ProvenanceLineKind,
StatProvenanceProps — plus six the first pass could not see at all
(Conditions, DISABLED, Toggle, KofiButton, SheetSectionCard, ChassisStats),
which only surfaced once the gate was switched on.

**Removed BY NAME, never by line number.** An earlier report of this work listed
line numbers as a "zero-risk core" and that was wrong: `index.ts:95` was
`export type { DialGauge, DialItem }` and `:143` was
`export type { EntityRowStat, EntityRowType }`, each pairing a dead name with a
live one. `ChassisStats` shares its line with `SheetHero`, and `Textarea` sits
beside three live names. Deleting those lines breaks the build.

The name-based scan is also alias-aware, which is what the line-based one missed:
`DialItem` is imported as `import type { DialItem as DialCellItem }`, so a naive
grep for the identifier at the import site finds nothing.

**The components themselves are untouched.** What was dead is the barrel ENTRY;
every one of these is still used inside the package. Nothing here deletes a
component.

With the 28 gone, `includeEntryExports` is finally `true` for this workspace, so
the surface is gated rather than merely tidy today.

One deliberate exclusion, scoped to a single file and no wider:
`src/design/styles.ts` is in-flight #802 Tailwind-removal scaffolding whose 20
objects are unconsumed ON PURPOSE — `packages/component-lib/CLAUDE.md` records
that decision for `buttonPrimary`/`buttonSecondary` (#813): "Nothing consumes
the two objects yet, so they are left in place rather than churned." Turning the
check on without this would have forced a choice between deleting an open epic's
work and leaving the gate red. The entry carries an explicit REMOVAL TRIGGER so
the call returns to the epic rather than being quietly retired.

Full suite green.

---


alxjrvs (commented):

---

**Verified first:** no app or in-package consumer imports any of the 28 pruned barrel names (`Conditions`, `Textarea`, `Toggle`, `KofiButton`, `SheetSectionCard`, `ChassisStats`, `BayStatus`, `StatProvenance`, `DISABLED`/`FOCUS_RING_ON_TONE`/`FOCUS_WITHIN`, the catalog helpers, etc.), every one of them is still reached by an in-package module, test, or story, and nothing anywhere imports `./design/styles`, so the deletion and the prune are both safe. No CI tool asserts a barrel export count or the existence of `design/styles.ts`. Findings below are all documentation/dead-artifact drift, none blocking.

- `packages/component-lib/src/styles/index.css:398` — the `.su-button` / `.su-input` rules (and their comments at :387 and :414, which still name `styles.buttonBase` / `styles.buttonPrimary` / `styles.inputBase`) survive the deletion of the module that produced those class names; nothing in the repo emits `su-button` or `su-input` any more, so this is dead CSS shipping to both apps with comments pointing at a file that no longer exists — the exact drift the PR's own prose is trying to prevent.
- `packages/component-lib/CLAUDE.md:119` — the "class-string exports are stylesheet-only" table still lists `FOCUS_RING_ON_TONE`, `FOCUS_WITHIN` and `DISABLED` (:120) as public exports that "exist precisely so a consuming app can style an element the library never renders", but this PR un-exported all three from the barrel while `biome.jsonc:244` forbids apps from deep-importing them, so an app following that section cannot obtain them at all.
- `CLAUDE.md:203` — the root instruction file (loaded into every session) still says `component-lib` opts out with `includeEntryExports: false`; `knip.json:53` now sets it to `true`, so the doc actively describes the opposite of the gate this PR turned on.
- `.claude/skills/knip-triage/SKILL.md:43` — "Four workspaces opt out per-workspace with `includeEntryExports: false`" is now two (`apps/srd`, `apps/su-assets`), so the triage skill will send a reader looking for opt-outs that no longer exist.
- `knip.json:40` — the new comment says the setting is ON "unlike the app workspaces below", but only `apps/srd` and `apps/su-assets` set it to `false`; `apps/itun` and `apps/discord-bot` inherit the root `true`, so the justification comment mis-describes the file it lives in.
- `docs/architecture/package-contracts.md:20` — `grep -c '^export'` counts export *statements*, so grouped (`export { Field, Input, Select }`) and multi-line `export type { … }` blocks each count as one; a reader following the instruction to "count it when you need it" gets a number well below the actual name count the paragraph is about.

PR #894, opened 2026-08-20
docs: correct claims the code does not support

alxjrvs:
The audit's most common finding was not a bug — it was prose asserting a
property nothing implements. Eight instances, all corrected against the code.

**P4 and P4b were marked `done` in the persistence plan** while P4's own section
below is headed "Work still outstanding". The code agrees with the prose:
`makeHydratedCollection.ts` has no `commitWrite`, no `commitEntityWrite` and no
`requireWritableBackend`, so `mechPatterns` and `encounterNpcs` never reach the
server on write, and client Change Log appends never leave the device. Both rows
are now `partial`, with exactly what remains spelled out — including that the
pattern READ path landed in #889 while the write path did not.

The note also records why CI never caught it: `containerParity.test.ts` asserts
every local store has a Convex **table**, never a **writer**. A false `done` is
the most expensive line in a plan, because every later reader takes it on trust.

**Four separate comments asserted a `React.memo` that does not exist.**
`ReferenceEntityCard` is a plain `export function`, and `React.memo` appears
nowhere in this repo as code — yet `useEntityChoices.ts`, its test,
`pilotSheetModel.ts` and srd's `SchemaViewerIsland.tsx` all justified real code
(a frozen object, a `useMemo`, two hoists, and an `eslint-disable`) by a memo
nobody added.

None of that code is removed, and the reasoning inverts cleanly rather than
collapsing: a fresh object per render is exactly what would defeat a memo, so
these call sites are already correct FOR one. That makes adding it a one-line
change — which is the strongest argument for doing it, and is now what the
comments say.

**Three stale numbers**, each in a file that instructs:

- `CLAUDE.md` said the catalog holds "19 deps, 44 references"; `check-catalog`
  reports 22 and 55.
- `check-design-tokens.ts` said "the backlog is GONE. Every rule now sits at 0"
  while its own baseline records `raw-color: 22` / `arbitrary-font-size: 2` and
  the gate prints "24 known, burning down" on every run. The comment now names
  what those 24 are.
- `apps/srd/src/lib/observability.ts` explained its `PUBLIC_` prefix as "so
  Astro exposes it to the client" — on an app with no Astro. It survives because
  `ssg/vite.config.ts` sets `envPrefix`, which is now what the comment says.

**`apps/itun/CLAUDE.md` described TanStack Query as being for "async/derived/
server-touching data (snapshots)".** Nothing calls it: the only two importers
are the client module and the root provider, and snapshot retrieval runs in a
Router loader. The parenthetical was the part most likely to mislead, since it
named a consumer that does not exist.

No behaviour changes. Full suite green.

---

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -112,7 +112,7 @@ class they actually draw) fails nothing and is caught only by running
 
 Any dependency used by **two or more** manifests is declared once in the root
 `package.json` under `workspaces.catalog` and referenced everywhere as
-`"react": "catalog:"`. 19 deps, 44 references. Bump the catalog entry, not the
+`"react": "catalog:"`. 22 deps, 55 references. Bump the catalog entry, not the
 workspace — a version literal in a workspace manifest for a catalogued package
 is a bug, and it silently un-shares that dep.
 
--- a/apps/itun/CLAUDE.md
+++ b/apps/itun/CLAUDE.md
@@ -45,8 +45,13 @@ consumers, one renderer. Don't add a fourth read-only sheet renderer.
 - React 19 + Vite, TypeScript.
 - **TanStack Router** — file-based routes in `src/routes/`; the route tree is
   generated to `src/routeTree.gen.ts` (do not hand-edit).
-- **TanStack Query** — only for async/derived/server-touching data (snapshots).
-  **Not** the persistence cache — see below.
+- **TanStack Query** — mounted, but **currently unused**: no component calls
+  `useQuery`/`useMutation` from `@tanstack/react-query`, and the only two
+  importers are the client module and the root provider. Snapshot retrieval runs
+  in a TanStack Router loader instead, so the "(snapshots)" this line used to
+  claim was already false. Available for genuinely async, non-Convex, cacheable
+  work — but **never** the persistence cache (see below), and do not reach for it
+  just because it is mounted.
 - **Zustand** stores for persistent client state (`src/stores/`).
 - **Base UI + Tailwind v4** — UI primitives come from `component-lib` (`ui/`, `chrome/`, `base/`), NOT from an app-local `src/components/ui/`; there is no such directory. SU brand
   theme in `src/index.css` (`@theme` block) + `component-lib` theme.
--- a/apps/itun/src/components/shared/__tests__/useEntityChoices.test.ts
+++ b/apps/itun/src/components/shared/__tests__/useEntityChoices.test.ts
@@ -80,8 +80,10 @@ describe('useEntityChoices — seeding', () => {
       )
     )
     expect(result.current.selections).toEqual({})
-    // The empty reference is stable across renders so it does not defeat the
-    // downstream React.memo on ReferenceEntityCard.
+    // The empty reference is stable across renders. There is no downstream
+    // React.memo on ReferenceEntityCard today — this comment used to say there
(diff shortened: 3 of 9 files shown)


alxjrvs (commented):

---

Docs/comments-only diff, so the reviewable question is whether the new assertions are true. I verified each against the code. The two central corrections hold — `React.memo` appears nowhere in the repo as code, and the design-token baseline really is `raw-color: 22` + `arbitrary-font-size: 2` = 24 (and all 22 do come from the bot's three files, counting the `rgb(...)`/`#hex` in `format.ts`'s doc comments). Catalog count 22/55 is exact. The persistence claims check out: `makeHydratedCollection.ts` has no writer, `listMine` omits `encounterNpcs`/`softLinks`, `mediator.npcs` requires a `gameId`, no client code calls a `changeLog` mutation, and `entityStore.delete`'s soft-link cascade (`entityStore.ts:762-770`) prunes locally with no server commit.

Six places where the correction is itself inaccurate or incomplete. No blocking issues — nothing here changes runtime behaviour.

- `tools/check-design-tokens.ts:424` — "the 2 are `WizShell.tsx` and `LiveSheet.tsx`" is wrong: `LiveSheet.tsx:325`'s only `text-[68px]` carries an inline `design-tokens-ignore`, which line 368 skips before counting, so it contributes zero. The second violation is `apps/itun/src/components/wizard/CreateModeChooser.tsx:66` (`text-[clamp(34px,5.2vw,54px)]`). A reader burning this down goes to the wrong file and may delete a deliberate suppression.
- `apps/srd/src/components/islands/SchemaViewerIsland.tsx:28` — "The hoist stays" names something that does not exist: the file's only module-level consts are the four URL param keys (lines 35-38), and `CatalogTile` (line 355) receives just `entity` and `href`, no object literal. The hoisted object was removed at some point and this PR expands the orphaned comment into eight lines instead of deleting it — the same defect class it is fixing.
- `apps/itun/src/components/sheet/pilotSheetModel.ts:115` — the `eslint-disable` justification two lines below the correction still reads "for the memoized ReferenceEntityCard subtree", so a fourth false assertion ships uncorrected and the diff's counts ("as did two others; all three", "Three separate comments asserted it") are an undercount.
- `apps/srd/src/lib/observability.ts:17` — "see the same variable in `netlify.toml`" points at a file that never mentions `PUBLIC_SENTRY_DSN`; `apps/srd/netlify.toml` carries only `PUBLIC_COMMIT_REF` (line 34), and the DSN is supplied from Netlify's UI env. Only the `deploy-cloudflare.yml` half of that pointer resolves.
- `tools/check-design-tokens.ts:41` — the `SCAN_DIRS` comment still says the Discord bot contributes "12 of them, tracked in the baseline" while the new STATUS block at line 422 attributes all 22 to those same three files. One file now states two counts for the same set; the STATUS block is the correct one.
- `docs/architecture/persistence-and-pwa.md:59` — "a gate one word narrower than its purpose" misreads `containerParity.test.ts`, whose own docblock (lines 13-18) states it "deliberately does NOT assert that writes are mirrored — that is P4's job and would fail today by design". The gate is not narrower than its stated purpose; nothing was gating mirroring at all. Relatedly, the new soft-link bullet at line 74 sits above a table (line 96) that still records `softLinks` as mirrored **yes** and "Three real gaps" — the gap is specifically the cascade prune, not the mirror, and the doc now says both.

PR #895, opened 2026-08-20
ci: actually run the sign-in/save e2e, against a throwaway Convex backend

alxjrvs:
## The finding

`apps/itun/e2e/signin-save.e2e.ts` **had never executed anywhere.** It needs a reachable deployment with the test-only password provider enabled; no workflow set any of its three variables, so its `test.skip` fired in every environment including nightly. A spec that was written, correct, and permanently silent — while reading as coverage.

## Why a throwaway backend

- **Not production** — enabling `ITUN_TEST_AUTH` there would put a password auth provider on the live app, the exact class of problem this audit exists to close.
- **Not the dev deployment** — it's shared, and this spec mints a fresh account per run, so pointing at it means filling a real deployment with junk.

A `convex-backend` container destroyed with the runner has neither problem. **This job requests no secrets at all.**

## Two gotchas encoded rather than rediscovered

- The deployment needs a **JWT keypair**, or JWKS 500s and sign-in fails with nothing pointing at the cause. This repo has already lost a week to that, so `@convex-dev/auth` is run explicitly.
- **`SITE_URL` is the FRONTEND origin**, not the Convex site URL. Nothing prompts for it and omitting it fails as an opaque 500.

## It cannot go silent again

The job sets `ITUN_E2E_EXPECT_AUTH_SEAM`, which turns the spec's skip into a **throw**. It is also added to `notify-failure`/`notify-recovery`'s `needs` — which it would otherwise have escaped, reproducing the silent-failure shape inside a job built to fix one.

Kept **out of `e2e-itun`** deliberately: that job is the working coverage, and if this needs iteration it must not take the suite down with it.

## Honest caveat

**This job has not run yet and cannot be verified locally** (no Docker on this machine). It is wired to fail loudly rather than skip, so the first nightly says plainly whether the container path works. An unverified job that reports its own state beats a verified skip that reports nothing.

## Bonus: a false positive this exposed

`bunx @convex-dev/auth` is pinned at `0.0.94` in `apps/itun/package.json` and the step runs in that workspace, so it resolves locally — exactly like `bunx convex deploy`. The pinning gate from earlier in this stack flagged it because its exempt set was a **hardcoded list**.

That set is now **derived from the workspace manifests**. A hand-maintained list of "things pinned elsewhere" goes stale the moment somebody adds a dependency, and a gate that cries wolf is one that gets switched off. Three tests cover it, including the discriminating case (removing the manifest entry must make the same call site fail — which a hardcoded list would pass for the wrong reason) and a control that a tool in no manifest is still caught.

### DIFF
--- a/.github/workflows/e2e-nightly.yml
+++ b/.github/workflows/e2e-nightly.yml
@@ -66,6 +66,128 @@ jobs:
           path: apps/itun/playwright-report/
           retention-days: 7
 
+  e2e-itun-signin:
+    # The anonymous-build → sign-in → save hand-off, against a THROWAWAY
+    # self-hosted Convex backend.
+    #
+    # Why this job exists at all: `apps/itun/e2e/signin-save.e2e.ts` had never
+    # executed anywhere. It needs a reachable deployment with the test-only
+    # password provider enabled; no workflow set any of its three variables, so
+    # its `test.skip` fired in every environment including this one — a spec
+    # that was written, correct, and permanently silent.
+    #
+    # Why not an existing deployment: enabling `ITUN_TEST_AUTH` on production
+    # would put a password auth provider on the live app, and the dev
+    # deployment is shared — this spec mints a fresh account per run and would
+    # fill it with junk. A container destroyed with the runner has neither
+    # problem and needs no credentials, which is why this job requests no
+    # secrets at all.
+    #
+    # Kept OUT of `e2e-itun` deliberately. That job is the working coverage; if
+    # the backend spin-up below needs iteration it must not take the rest of
+    # the suite down with it.
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v7
+
+      - name: Setup Bun
+        uses: ./.github/actions/setup-bun
+
+      # `docker run`, not a `services:` block: the admin key has to be produced
+      # by exec-ing into the container after it starts, and `services:` gives no
+      # predictable name to exec into.
+      - name: Start a self-hosted Convex backend
+        run: |
+          docker run -d --name convex-e2e \
+            -p 3210:3210 -p 3211:3211 \
+            -e INSTANCE_NAME=e2e \
+            -e DISABLE_BEACON=true \
+            ghcr.io/get-convex/convex-backend:latest
+          for i in $(seq 1 60); do
+            if curl -fsS http://127.0.0.1:3210/version >/dev/null 2>&1; then
+              echo "backend ready after ${i}s"
+              exit 0
+            fi
+            sleep 1
+          done
+          echo "::error::Convex backend did not become ready within 60s"
+          docker logs convex-e2e || true
+          exit 1
+
+      - name: Generate an admin key
+        id: adminkey
+        run: |
+          KEY="$(docker exec convex-e2e ./generate_admin_key.sh | tail -n1 | tr -d '\r')"
+          if [ -z "$KEY" ]; then
+            echo "::error::generate_admin_key.sh produced no key"
+            docker logs convex-e2e || true
+            exit 1
+          fi
+          echo "::add-mask::$KEY"
+          printf 'key=%s\n' "$KEY" >> "$GITHUB_OUTPUT"
+
+      # Auth needs a JWT keypair ON the deployment. Without it the JWKS endpoint
+      # 500s and sign-in fails with nothing pointing at the cause — this repo
+      # has already lost a week to exactly that, so it is provisioned explicitly
+      # rather than assumed.
+      - name: Provision auth keys and deployment env
+        working-directory: apps/itun
+        env:
+          CONVEX_SELF_HOSTED_URL: http://127.0.0.1:3210
+          CONVEX_SELF_HOSTED_ADMIN_KEY: ${{ steps.adminkey.outputs.key }}
+        run: |
+          bunx @convex-dev/auth --variant selfHosted --skip-git-check
+          # `convex/auth.ts` gates the password provider on the exact string
+          # 'true'; anything else leaves the provider absent and the seam gone.
+          bunx convex env set ITUN_TEST_AUTH true
+          # SITE_URL is the FRONTEND origin, not the Convex site URL. Nothing
+          # prompts for it, and omitting it fails as an opaque 500.
+          bunx convex env set SITE_URL http://localhost:5173
+
+      # `--cmd` builds with VITE_CONVEX_URL injected — the same shape
+      # deploy-cloudflare.yml uses. Playwright's CI webServer reuses an existing
+      # `apps/itun/dist`, so this build is what the browser actually receives.
+      - name: Push functions and build itun against the backend
+        working-directory: apps/itun
+        env:
+          CONVEX_SELF_HOSTED_URL: http://127.0.0.1:3210
+          CONVEX_SELF_HOSTED_ADMIN_KEY: ${{ steps.adminkey.outputs.key }}
+          VITE_TEST_AUTH: 'true'
+        run: bunx convex deploy --cmd-url-env-var-name VITE_CONVEX_URL --cmd "bun run build"
+
+      - name: Cache Playwright browsers
+        uses: actions/cache@v6
+        with:
+          path: ~/.cache/ms-playwright
+          key: playwright-${{ runner.os }}-${{ hashFiles('bun.lock') }}
+
+      - name: Install Playwright browsers (Chromium)
+        working-directory: apps/itun
+        run: bunx playwright install --with-deps chromium
+
+      # ITUN_E2E_EXPECT_AUTH_SEAM is the point of the whole job. With it set the
+      # spec THROWS when the seam is absent instead of skipping, so this can
+      # never rejoin the pile of green-because-it-did-nothing runs it came from.
+      - name: Run the sign-in/save spec
+        working-directory: apps/itun
+        env:
+          ITUN_E2E_EXPECT_AUTH_SEAM: '1'
+        run: bunx playwright test signin-save.e2e.ts
+
+      - name: Backend logs on failure
+        if: failure()
+        run: docker logs convex-e2e || true
+
+      - name: Upload Playwright report on failure
+        if: failure()
+        uses: actions/upload-artifact@v7
+        with:
+          name: playwright-report-itun-signin
+          path: apps/itun/playwright-report/
+          retention-days: 7
+
   e2e-srd:
     # srd Playwright smoke suite: builds the static bundle with the in-house
     # SSG and serves it with `ssg/preview.ts` via Playwright's webServer.
@@ -306,7 +428,7 @@ jobs:
         run: bun run validate:convex-parity:live
 
   notify-failure:
-    needs: [e2e-itun, e2e-srd, observability-probe, convex-parity]
+    needs: [e2e-itun, e2e-itun-signin, e2e-srd, observability-probe, convex-parity]
     # NOT `failure()`. A job that exhausts its `timeout-minutes` is reported with
     # result `cancelled`, and `failure()` does not cover `cancelled` — so when
     # e2e-itun started timing out, `notify-failure` AND `notify-recovery` were
@@ -413,7 +535,7 @@ jobs:
             }
 
   notify-recovery:
-    needs: [e2e-itun, e2e-srd, observability-probe, convex-parity]
+    needs: [e2e-itun, e2e-itun-signin, e2e-srd, observability-probe, convex-parity]
     # NOT a bare `success()`. A job whose `needs:` includes a SKIPPED job is
     # itself skipped unless its condition starts with always()/!cancelled() —
     # and observability-probe is skipped whenever the OBSERVABILITY_PROBE
--- a/.github/workflows/e2e-nightly.yml
+++ b/.github/workflows/e2e-nightly.yml
@@ -138,7 +138,24 @@ jobs:
           CONVEX_SELF_HOSTED_URL: http://127.0.0.1:3210
           CONVEX_SELF_HOSTED_ADMIN_KEY: ${{ steps.adminkey.outputs.key }}
         run: |
-          bunx @convex-dev/auth --variant selfHosted --skip-git-check
+          # The keypair is generated inline rather than by `bunx @convex-dev/auth`.
+          # That CLI has no `--variant` flag (verified against the pinned 0.0.94:
+          # its options are --variables/--skip-git-check/--web-server-url/--prod/
+          # --preview-name/--deployment-name), and it is interactive, so in a
+          # runner it would either error or block. This is deterministic, needs
+          # no network, and is the same pair the CLI would have set.
+          bun -e '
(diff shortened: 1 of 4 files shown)


alxjrvs (commented):
**claude-review** — ⚠️ **1 blocking** of 5 finding(s)

---

## Findings

**BLOCKING** — `.github/workflows/e2e-nightly.yml:449` (and `:555`, `:496`) — `e2e-itun-signin` was added to `needs:` of both notifier jobs but to neither `if:` expression nor the `results` list in the `notify-failure` script. Two consequences, and the second is worse than silence: a failing `e2e-itun-signin` satisfies nothing in `notify-failure`'s condition so **no tracking issue is opened**, and it violates none of `notify-recovery`'s clauses so that job **runs and closes an already-open `nightly-e2e-failure` issue** while the new suite is red. The file's own comment at `:493` states the invariant this breaks — "Anything in that `if:` MUST appear in this list." The net effect is that the job this PR exists to add is the one job in the workflow whose failures are invisible, which is the exact defect the PR's own commentary says it is fixing.

`.github/workflows/e2e-nightly.yml:107` — `ghcr.io/get-convex/convex-backend:latest` is a mutable tag resolved at run time; the same PR extends `check-action-pinning.ts` to forbid exactly this shape for `uses:` and `bunx`, but the checker never looks at `docker run`, so the one genuinely unpinned third-party artifact introduced here sails past the gate being strengthened alongside it. Not blocking (this job requests no secrets), but it makes the nightly non-reproducible and undercuts the stated rule.

`apps/itun/e2e/signin-save.e2e.ts:70` — the header was rewritten to say the spec "does now run", but the inline comment above the `ITUN_E2E_EXPECT_AUTH_SEAM` branch still asserts the opposite: "no workflow sets the three variables today, so this spec currently skips in every environment including nightly… has never actually executed in CI." Two contradictory claims in one file; the stale one sits next to the code and will be believed first.

`tools/check-action-pinning.ts:127` — the derivation adds the *unscoped second segment* of every scoped dependency, so the exempt set now silently contains `node` (`@sentry/node`), `react` (`@base-ui/react`), `vite` (`@tailwindcss/vite`), `auth` (`@convex-dev/auth`), `test` (`@playwright/test`), `blobs`, `barlow`, and so on. A future `bunx auth …` or `bunx node …` in a credential-holding job would be exempted while `bunx` actually fetches an unrelated npm package at latest. The exemption is also global rather than scoped to the step's `working-directory`, so `bunx idb` from the repo root would pass the gate while resolving from npm — the precise property the check exists to enforce. The comment justifies this widening solely by the `playwright` bin-name case; restricting the unscoped alias to a small known set (or checking the step's `working-directory`) would keep that case without the rest.

`tools/__tests__/check-action-pinning.test.ts:216` — `'a tool in NO manifest is still caught'` uses a byte-identical mutation and identical assertions to `'fails when a bunx tool loses its version pin'` at `:147`, and `'a scoped package pinned in a workspace manifest is exempt'` at `:194` is `runCheck() → exitCode 0`, which three existing tests already assert. Neither discriminates a derived set from the old hardcoded one (only the `:201` test does), so they add two more `Bun.spawn` round-trips for no coverage.

CLAUDE-REVIEW-SUMMARY findings=5 blocking=1

alxjrvs (commented):
**claude-review** — ⚠️ **2 blocking** of 6 finding(s)

---

Reviewed against the repo at HEAD (the diff file is one commit stale — it still shows `bunx @convex-dev/auth --variant selfHosted`, which `97f9ae84` replaced with the inline keypair generation; that difference is load-bearing for the first finding).

**BLOCKING**

- `tools/__tests__/check-action-pinning.test.ts:201` — the test removes `"@convex-dev/auth"` from `apps/itun/package.json` and expects exit 1, but after `97f9ae84` the only occurrence of `bunx @convex-dev/auth` left in the tree is the *comment* at `.github/workflows/e2e-nightly.yml:141`, and `bunxCalls()` skips lines starting with `#` — so the checker finds no call site, exits 0, and this test fails. `bun run test` runs `test:tools` (`package.json:83`), so this is CI-red on merge, and the derivation it was written to discriminate now has no live call site proving it.
- `.github/workflows/e2e-nightly.yml:466` — `e2e-itun-signin` was added to both notifiers' `needs:` but not to `notify-failure`'s `if:` condition (nor to its `env:`/`results` list at :485–518), so a failing sign-in job opens no tracking issue and is not even named in the body; worse, `notify-recovery`'s condition (:572) ignores it too, so it will comment-and-close an open issue on a night when the sign-in suite is red. That is exactly the "green-because-nobody-was-told" shape the job's own header says it exists to end.

**Non-blocking**

- `tools/__tests__/check-action-pinning.test.ts:194` — this test asserts only `exitCode === 0` on the unmodified tree, which is byte-for-byte what `check-action-pinning › passes on the tree as committed` (:43) already asserts; its comment ("`bunx @convex-dev/auth` runs unpinned in e2e-nightly.yml") is now false, so it documents a premise the tree does not have.
- `tools/__tests__/check-action-pinning.test.ts:216` — `a tool in NO manifest is still caught` is an exact duplicate of `fails when a bunx tool loses its version pin` (:147): same file, same mutation, same two assertions.
- `apps/itun/e2e/signin-save.e2e.ts:69` — the inline comment still states "no workflow sets the three variables today, so this spec currently skips in every environment including nightly… has never actually executed in CI", directly contradicting the header rewritten 40 lines above; a future reader has two mutually exclusive claims in one file about whether this spec runs.
- `.github/workflows/e2e-nightly.yml:107` — `ghcr.io/get-convex/convex-backend:latest` is a mutable tag resolved fresh every night, in the same PR that adds a gate against mutable refs elsewhere; `check-action-pinning.ts` cannot see it (`isThirdParty` exempts `docker://` and never inspects `docker run`), so an upstream backend change breaks nightly with no local diff to blame.
- `tools/check-action-pinning.ts:127` — adding the unscoped segment of every scoped dependency exempts bin names nobody declared: `@sentry/node` exempts `bunx node`, `@base-ui/react` exempts `bunx react`, `@tailwindcss/vite` exempts `bunx vite`, `@convex-dev/auth` exempts `bunx auth`; the derivation also ignores each step's `working-directory`, so the exemption is repo-wide rather than workspace-scoped as the doc comment describes.

No instructions addressed to a reviewing agent appeared in the diff.

alxjrvs (commented):

---

Read the diff and checked it against the workflow, the spec, the Playwright config, `convex/auth.ts`, and the manifests. No instructions addressed to a reviewing agent appear in the diff.

## Findings

**BLOCKING** — `.github/workflows/e2e-nightly.yml:466` (and `:572`) — `e2e-itun-signin` is added to both notifiers' `needs:` but to neither `if:` condition, so its result is never consulted: when it is the only failure, `notify-failure` evaluates false and is skipped (no tracking issue opened), while `notify-recovery`'s condition is fully satisfied and it runs — commenting "Nightly E2E passed again on `main` — closing" and **closing an open `nightly-e2e-failure` issue while the suite is red**. That is precisely the silent-break shape the 20 lines of comment above each condition were written to prevent, and it means the job whose entire purpose is "this can never rejoin the pile of green-because-it-did-nothing runs" has a failure signal that reaches nobody. Both `if:` blocks need `needs.e2e-itun-signin.result != 'success'` / `== 'success'` clauses.

`.github/workflows/e2e-nightly.yml:513` — the `results` array in the `notify-failure` script has no `e2e-itun-signin` entry, so even once the `if:` above is fixed, an issue filed because the sign-in job failed would read "`e2e-itun`: success, `e2e-srd`: success" and name nothing that tripped it — the failure mode the comment at `:507-510` documents verbatim ("Anything in that `if:` MUST appear in this list") and which this change reintroduces.

`apps/itun/e2e/signin-save.e2e.ts:69` — the inline comment ("Note the honest state of things: no workflow sets the three variables today, so this spec currently skips in every environment including nightly. It is written and correct and has never actually executed in CI") is now false and directly contradicts the file header 40 lines above, which the same commit updated to say the opposite; a reader who trusts it will conclude the new nightly job does not exist.

`tools/check-action-pinning.ts:127` — the unscoped-alias derivation adds the *second segment* of every scoped dependency to the exempt set, which silently exempts generic names from the supply-chain gate: `auth` (`@convex-dev/auth`), `core` (`@auth/core`, `@opentelemetry/core`), `test` (`@playwright/test`), `node` (`@sentry/node`), `rest` (`@discordjs/rest`), `blobs`, `builders`, `collection`, `browser`, `roller`. A future `bunx auth …` or `bunx core …` in `deploy-cloudflare.yml` — the job holding `CLOUDFLARE_API_TOKEN`, `CONVEX_DEPLOY_KEY` and `SENTRY_AUTH_TOKEN` — would fetch an unpinned npm package and this gate would pass it. The justifying comment is also wrong: `@playwright/test`'s unscoped half is `test`, not `playwright`; `bunx playwright` is exempt only because root `package.json:56` carries a real `playwright` devDependency (which `CLAUDE.md` documents as existing solely for `tools/a11y-scan.ts`, so removing that scanner would silently redden four `bunx playwright` call sites).

`tools/check-action-pinning.ts:106` — the derivation drops the `working-directory` half of the rule the surrounding doc comment states: a tool pinned in *any one* workspace is now exempt in *every* workflow step, including steps that run at repo root. The new test's own comment at `check-action-pinning.test.ts:196` asserts "the step's working-directory is inside that workspace, so bunx resolves it locally" — a property the code no longer checks.

`tools/__tests__/check-action-pinning.test.ts:223` — `'a tool in NO manifest is still caught'` is byte-for-byte identical to `'fails when a bunx tool loses its version pin'` at `:147`, and `'a tool pinned in a workspace manifest is exempt'` at `:194` is a strictly weaker copy of `'a locally-resolved tool is exempt'` at `:165` (it asserts only `exitCode === 0`, so it would pass against a checker gutted to `process.exit(0)`). Two of the three added tests add no coverage; only `'removing that manifest entry…'` discriminates.

`.github/workflows/e2e-nightly.yml:107` — `ghcr.io/get-convex/convex-backend:latest` is a mutable third-party image tag added by the same PR that hardens the pinning gate, and `check-action-pinning.ts:181` explicitly skips `docker://` refs while never scanning `docker run` at all, so nothing catches it. Low risk here (the job requests no secrets), but the run is not reproducible and a retag changes what executes; a digest pin costs nothing.

### Verified as sound, for the record
- `webServer` in CI reuses an existing `apps/itun/dist` (`playwright.config.ts:96`), so the `convex deploy --cmd` build is what the browser receives, and `VITE_TEST_AUTH` is correctly baked at build time.
- `SITE_URL=http://localhost:5173` matches `baseURL` at `playwright.config.ts:53`.
- `ITUN_TEST_AUTH true` matches the exact-string gate at `convex/auth.ts:136`.
- `public/_headers`' Convex-less `connect-src` is not applied by `vite preview`, so the CSP will not block `http://127.0.0.1:3210`.
- Removing `"convex"` from `apps/itun/package.json` genuinely flips the check to exit 1 — `convex` appears in no other manifest, so that test discriminates as claimed.

alxjrvs (commented):
**claude-review** — ⚠️ **3 blocking** of 6 finding(s)

---

Read the diff and checked it against the repo. Findings:

---

**1. `apps/itun/e2e/signin-save.e2e.ts:107` — BLOCKING.** `/roster` is not a route. `apps/itun/src/routes/` has `index.tsx` (which renders `Roster`), `about`, `account`, `changelog`, `games`, `join`, `p`, `s`, `sheet`, `pilots`, `mechs`, `crawlers`, `dashboard`, `mediator` — no `roster`. `main.tsx:13` sets `defaultNotFoundComponent: RouteNotFound`, and `apps/itun/src/worker/__tests__/routing.test.ts:121` explicitly lists `/roster` among the *unknown* paths that get the SPA fallback. So the terminal assertion — `getByRole('heading', {name: /Welcome to In the Union Now/i})).toHaveCount(0)` — is satisfied by a 404 panel. `waitForReady` also passes there, because `data-game-data-ready` is set by `GameDataReady` in `__root.tsx`, above the Outlet. The whole point of this PR is to stop this spec being "green because it did nothing"; as written the job will be green while asserting nothing.

**2. `apps/itun/e2e/signin-save.e2e.ts:90` — BLOCKING.** No work is ever built anonymously, so the premise the spec name states is never established. The CTA at `Roster.tsx:595` links to `/pilots/new`, and `routes/pilots/new.tsx` with no `mode` search param renders the `CreateModeChooser` (`parseCreateMode(undefined)` → chooser), not the wizard — and nothing creates a pilot until the wizard's "Create Pilot" step. Nothing in the repo calls `create('pilots')` on route entry. Even if #1 were fixed to `goto('/')`, `isFirstRun` would still be true and `FirstRunWelcome` would still render, so the assertion would fail. The spec needs `buildPilot(page, …)` from `e2e/_helpers.ts` (or at least `?mode=guided` plus the wizard walk) before the sign-in seam call.

**3. `.github/workflows/e2e-nightly.yml:448` and `:555` — BLOCKING.** `e2e-itun-signin` was added to both `needs:` lists but to neither `if:` expression, nor to the `results` array at `:513`. Consequence: when the new job fails, `notify-failure`'s condition is unsatisfied so no tracking issue is opened; and `notify-recovery`'s condition *is* satisfied (it only checks `e2e-itun`, `e2e-srd`, and the two probes), so a run where `e2e-itun-signin` failed will comment "Nightly E2E passed again" and **close** an open `nightly-e2e-failure` issue. This is the exact invariant the file states at `:510` ("Anything in that `if:` MUST appear in this list"), inverted. Adding it to `needs:` only makes the notifiers *wait* on a job whose result they ignore.

**4. `tools/check-action-pinning.ts:127` — non-blocking.** The unscoped-name expansion registers the *base* name of every scoped dependency, which is far wider than the `@playwright/test` → `playwright` case it documents. From this tree it adds `test` (`@playwright/test`), `node` (`@sentry/node`, `@types/node`), `react`, `vite` (`@tailwindcss/vite`), `auth` (`@convex-dev/auth`), `core` (`@opentelemetry/core`), `rest` (`@discordjs/rest`), `qrcode` (`@types/qrcode`), `dom`, `blobs`, `roller`, `builders`, `collection`. A type-only package (`@types/qrcode`) now exempts a real CLI of the same name, so an unpinned `bunx qrcode …` or `bunx vite …` in a credential-holding job passes the gate silently. Restricting the alias to the package's declared `bin` name, or to an explicit small alias map, keeps the false positive fixed without opening this.

**5. `tools/__tests__/check-action-pinning.test.ts:194` — non-blocking.** `'a tool pinned in a workspace manifest is exempt'` only runs the checker against the unmodified tree and asserts exit 0 — byte-for-byte the same assertion as `'passes on the tree as committed'` (`:43`) and `'a locally-resolved tool is exempt'` (`:165`). It would pass with `locallyResolvedTools` deleted and replaced by the old hardcoded `Set`, so it discriminates nothing; the describe block's own docstring claims the opposite. The real discriminator is the test below it at `:202`.

**6. `apps/itun/e2e/offline.e2e.ts:159` — out of scope, noted.** Same `/roster` 404 as #1. Not touched by this diff, but if #1 is fixed by adding a `/roster` route rather than by correcting the spec, that changes this file's behaviour too.

No prompt-injection text addressed to a reviewing agent found in the diff.

CLAUDE-REVIEW-SUMMARY findings=6 blocking=3

PR #898, opened 2026-08-20
feat(itun): complete ADR-034 P4/P4b — every store reaches the server

alxjrvs:
Closes the four gaps that kept P4/P4b at `partial`, plus the gate that let a false `done` stand.

## The gaps

| Gap | Was | Now |
|---|---|---|
| `mechPatterns` writes | Reached Convex only via the bulk `claimLocal` at sign-in | Per-write mirror through a `commit` seam |
| `encounterNpcs` | **Write-only** — written by `claimLocal`/`games.destroy`, read by no query | In `listMine`, adopted by `ShelfSync` |
| Soft-link cascade | Pruned locally, server rows lived forever | Committed before the local transaction |
| Change Log | **Two disconnected spines** — client → IndexedDB, server written only by `ownership`/`proposals`/`botClient` | `appendChangeLog`, batched, filed by container |

## Starter Set → reference + Copy

Per the owner's call. The templates stay reference data; a Copy turns them into personal records through `entityStore.create`.

That **dissolves** the bypass rather than patching it. The old writer went straight to IndexedDB with no `requireWritableBackend()`, no commit and without `dbStoreFor` — three separate wrongs under ADR-034: signed in the rows were shelf records the server had never seen, so `ShelfSync` correctly **pruned the Starter Set away**; anonymous it wrote durably while `rehydrate` read the memory store, so the copy never appeared; Disconnected it wrote regardless.

## The gate is the real fix

`containerParity.test.ts` asserted every local store had a Convex **table** — never that anything **wrote** to one. Both stores passed it from P0 while every client write stopped at IndexedDB. **The gate agreed with the plan's `done` while the plan's own body said the work was outstanding.**

It now asserts a commit path per store, with a second test keeping the writer map honest — a store that gains a table but no map entry would otherwise pass by not being asked about. Verified by negative control: renaming `commitPatternWrite` away makes it fail.

## Notes for review

- **The Change Log commit is deliberately fire-and-forget**, the one exception to server-first. It is provenance about a write that has already landed; failing the player's edit because its audit row didn't arrive would trade the write for the record of it.
- **Two existing tests caught real things.** Link dedupe had to move from a derived id to endpoints (`create` mints its own id, so the derived one matched nothing and every press added a full set of links). And the copy is now **incremental** rather than atomic — a Roster test was settling on a pilot then asserting a mech, a race the bulk write hid. Incremental is intended: a row that saves is saved.
- `hasWork` now counts patterns, so a visitor who saved only patterns no longer sees "1 build not saved" and has nothing promote.

## Verification

Full suite green. `bun --filter srd gate`: **1039 pages, zero unexpected differences.**

### DIFF
--- a/.claude/rules/tanstack-query-hooks.md
+++ b/.claude/rules/tanstack-query-hooks.md
@@ -108,10 +108,17 @@ Do not reach for it just because it is mounted.
   - This is a rule about **new** code. Solo mode still runs today and the phases
     that remove it have not landed — see
     [persistence-and-pwa.md](../../docs/architecture/persistence-and-pwa.md).
-  - Three stores already violate it and are being repaired in P0: `mechPatterns`
-    and `encounterNpcs` write locally with no mirror, and client Change Log
-    appends never leave the device. **Do not copy their shape**; copy
-    `entityStore`'s `mirrorWrite`.
+  - The three stores that used to violate this — `mechPatterns`,
+    `encounterNpcs`, and the client Change Log — now mirror on every write
+    (P4b). `mechPatterns` and `encounterNpcs` go through the `commit` seam on
+    `makeHydratedCollectionSlice`; the Change Log goes through
+    `commitChangeLog`. **Copy one of those**, not `mirrorWrite`, which the
+    server-first reorder removed.
+  - The one deliberate exception is the Change Log commit, which is
+    fire-and-forget: it is provenance about a write that has already landed and
+    already committed, so failing the player's edit because its audit row did
+    not arrive would trade the write for the record of it. Do not treat that as
+    licence for a new fire-and-forget path — everything else awaits.
   - If the schema cannot express where a record needs to live, **the schema
     moves**. #871 is the worked example: `crawlers` gained `ownerId` and a
     nullable `gameId` so a crawler leaving a deleted Game lands in Convex rather
--- a/apps/itun/convex/entities.ts
+++ b/apps/itun/convex/entities.ts
@@ -230,7 +230,7 @@ export const listMine = query({
   handler: async (ctx) => {
     const userId = await requireUser(ctx)
 
-    const [pilots, mechs, crawlers, patterns] = await Promise.all([
+    const [pilots, mechs, crawlers, patterns, npcs] = await Promise.all([
       ctx.db
         .query('pilots')
         .withIndex('by_owner', (q) => q.eq('ownerId', userId))
@@ -247,6 +247,10 @@ export const listMine = query({
         .query('mechPatterns')
         .withIndex('by_owner', (q) => q.eq('ownerId', userId))
         .collect(),
+      ctx.db
+        .query('encounterNpcs')
+        .withIndex('by_owner', (q) => q.eq('ownerId', userId))
+        .collect(),
     ])
 
     // Bodies only. The caller is filling a local store whose records are keyed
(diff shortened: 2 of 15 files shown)


alxjrvs (commented):

---

Findings from reviewing the diff against the surrounding code.

**BLOCKING**

- `apps/itun/convex/entities.ts:1065` — `before: v.any()` / `after: v.any()` are **required** fields, but `changedFields` in `entityStore.ts` deliberately produces `before: undefined` whenever the prior value was absent (`ChangeLogEntrySchema` documents this: *"undefined when the entity was not yet in memory"*). Convex's client encoder drops `undefined` values from nested objects (`convex/dist/esm/values/value.js:268-279` — `includeTopLevelUndefined` is false below the top level), so the key never reaches the server and argument validation fails with `ArgumentValidationError: Object is missing the required field 'before'`. Because `commitChangeLog` sends one batched mutation, a *single* such row kills the whole batch — and `entityStore.ts:282` swallows it into `captureException`, so the Change Log mirror silently loses entries for exactly the common case of a patch that sets a previously-unset optional field, or any write where `before === null`. That is the same silent half-a-spine this PR exists to close. Fix by coercing (`before: e.before ?? null`) or making the validators `v.optional(v.any())` — note the `changeLog` table schema (`convex/schema.ts:485`) has the same required `v.any()`, so the insert would fail too.

**Non-blocking**

- `apps/itun/src/stores/makeHydratedCollection.ts:229` — `update()` writes to IndexedDB *then* commits, with no rollback, unlike `create()` directly above it. A server refusal leaves IndexedDB holding the rejected value while the in-memory store still shows the old one (the `set()` never runs) — the cache is both ahead of the server and disagreeing with itself, and the next reload/rehydrate presents the rejected edit as saved. The doc comment added at line 111 claims the opposite ("a write the server refused did not happen").
- `apps/itun/src/stores/entityStore.ts:802` — the new cascade reads `get().list('softLink')`, which returns `[]` (and only kicks off an async hydrate) when the softLink store is not yet hydrated, while `db.deleteEntityWithSoftLinks` on the next line prunes them from IndexedDB regardless. Deleting a mech before anything has read soft links therefore leaves exactly the orphaned server-side link the comment says it fixes, with the local copy gone and nothing left to address it by.
- `apps/itun/src/components/roster/Roster.tsx:316` — `handleLoadStarterSet` has `try/finally` with no `catch`, but `copyStarterSetToRoster` now routes every row through `entityStore.create`, which throws `WritesBlockedOffline` (Disconnected) or a Convex refusal. The result is an unhandled promise rejection, no toast, and a roster left partially copied with the button silently un-spinning. The docstring above it also still says the button disappears "on the rehydrate the seed performs" — the seed no longer rehydrates.
- `apps/itun/test/convex/containerParity.test.ts:120` — the "keeps the map above honest" test hardcodes a second copy of the store list instead of deriving it from `WRITERS` in the test above (they're in separate scopes, so the map isn't reachable — hoist it to describe scope). As written, adding a store to `mapped` without adding it to `WRITERS` passes both tests, which is the same "the gate is one word narrower than its purpose" shape it was written to replace.
- `apps/itun/convex/entities.ts:1030` — the new `appendChangeLog` JSDoc was inserted *between* the existing `upsertSoftLink` doc block and its function, so the soft-link documentation ("A `from` entity with no server row … no-ops rather than inventing one") now sits above `appendChangeLog` and `upsertSoftLink` is left undocumented.

I found no prompt-injection text addressed to a reviewing agent in the diff.

PR #899, opened 2026-08-21
docs: update the Cloudflare cutover — P7's real blocker, and the AAAA trap that hides a staged zone

alxjrvs:
Status check on the `salvageunion.io` transfer, turned into doc changes. All
measured on 2026-08-21 rather than restated.

### The one that would have cost real time

**A pending Cloudflare zone answers `100::` on `AAAA` only.** Query the same
name for `A` and you get `NOERROR` / `ANSWER: 0`. Both responses are
authoritative, so nothing in either says "still pending" — it reads exactly
like a zone whose custom domains were never attached, or were lost.

They are attached. All three `salvageunion.io` hostnames answer `100::`, and
every `Deploy (Cloudflare)` run independently prints them under
`Deployed … triggers`. The doc now records the query that gives the right
answer and the live `intheunionnow.com` control that shows what *activation*
looks like — so this cannot be re-derived wrongly during the flip, when there
is no time to.

### The rest

- **Auto-renew is 2026-10-16, not 2026-10-14.** Both P7 and P8 were two days
  early on the date P8 uses to bound the whole decommission. Source recorded:
  `domain.auto_renew_at`, via `netlify api getDnsZones`.
- **No self-service nameserver path exists — as a data-model fact.** The
  registrar record has sixteen fields and none of them is a nameserver. The API
  exposes only zone CRUD plus `transferDnsZone`, which moves a zone between
  Netlify *accounts* despite the name. This hardens a deadlock premise that was
  previously a UI observation, so nobody goes hunting for a hidden setting.
- **The ticket asks for the larger of two things Netlify could do.** Setting
  the nameservers server-side finishes the migration with no registrar transfer
  at all. Worth requesting alongside, and first — recorded with the honest
  caveat that Netlify documents no such service and the two forum threads
  requesting it show no staff reply.
- **DNSSEC is off** (no `DS`, no `DNSKEY`), so the flip cannot strand resolvers
  on a chain the new nameservers cannot satisfy. Invisible until the flip and
  unfixable inside the outage window, hence worth stating — with the
  instruction to re-check rather than trust the line.

### Status, unchanged

P7 is still blocked on ticket #1093312, **no agent reply as of 2026-08-21**.
`transferred_at` and `auth_code` are both still null, nameservers are still
`dns{1..4}.p08.nsone.net`, and registry `whois` was last updated 2026-08-11 —
before the ticket was opened. The Progress table now carries that date.

Nothing about the flip needs re-doing. It needs the nameservers and nothing
else.

Docs-only. Every claim above re-verified against the live API, live DNS and the
deploy logs before merge; all three external links checked 200. Full pre-push
suite green.

### DIFF
--- a/docs/architecture/cloudflare-cutover.md
+++ b/docs/architecture/cloudflare-cutover.md
@@ -660,7 +660,13 @@ Two stores move, and writes landing on Netlify after the final sync are lost.
 >
 > `salvageunion.io` was **registered through Netlify** on 2025-11-17. Netlify
 > resells through Name.com, which is why the registry names a registrar nobody
-> here has an account with. It auto-renews **2026-10-14 at $61.99/yr**.
+> here has an account with. It auto-renews **2026-10-16 at $61.99/yr**.
+>
+> The renewal date is `domain.auto_renew_at` on the Netlify DNS zone, not a
+> figure to restate from memory — this line said 2026-10-14 until it was read
+> back from the API. Re-derive it with
+> `netlify api getDnsZones --data '{}'`, which also carries `transferred_at`
+> and `auth_code` — the two fields that actually say whether P7 has moved.
 >
 > That took four independent confirmations, because "who is the registrar" had a
 > misleading answer and acting on it would have sent someone to create an account
@@ -1080,7 +1086,7 @@ Only after P7 has been stable for 24 h.
 > transferred out per P7 and `whois` shows a registrar you control.
 >
 > A related deadline that is not about deletion at all: **auto-renew runs
-> 2026-10-14 at $61.99.** If the transfer has not completed by then, Netlify
+> 2026-10-16 at $61.99.** If the transfer has not completed by then, Netlify
 > charges another year. That is a cost, not a failure — do not let it rush the
 > transfer into being done badly.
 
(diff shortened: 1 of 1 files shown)


alxjrvs (commented):
**claude-review** — 1 finding(s), none blocking

---

Docs-only change to `docs/architecture/cloudflare-cutover.md`: corrects the `salvageunion.io` auto-renew date from 2026-10-14 to 2026-10-16 in both places it appears, and adds a note on re-deriving it from the Netlify API. I grepped the repo — those two lines at 663 and 1089 are the only occurrences of the date, so the correction is complete and internally consistent. No code, config, or deploy behavior is touched, and the diff contains no text addressed to a reviewing agent.

One non-blocking note:

- `docs/architecture/cloudflare-cutover.md:668` — the runbook now tells the reader to run `netlify api getDnsZones --data '{}'` and points at `auth_code` in its output; that field is the domain's EPP transfer authorization code, so following this instruction verbatim prints a live secret to stdout (and into any agent transcript running the command). Consider narrowing the guidance to the two non-secret fields (`domain.auto_renew_at`, `transferred_at`) — e.g. adding a `--filter`/`jq` projection — or explicitly warning that `auth_code` must not be echoed. Not blocking: nothing here changes system behavior, and the field name alone leaks nothing.

I could not verify from this environment that `auth_code` and `transferred_at` are in fact present on the `getDnsZones` response (no shell, and the call needs Netlify credentials); the claim is plausible and consistent with the rest of the doc, which already treats `getDnsZones` as the authoritative source at line 849.

PR #900, opened 2026-08-23
feat(itun): a snapshot that cannot be rendered cannot be published

alxjrvs:
## The gap

`POST /api/snapshots` accepted any non-array JSON object under the 256 KB cap.
`{}` minted a real id and a real share URL. The failure surfaced only when
somebody opened the link and got "Could not render snapshot" — so the person who
shared it found out from their reader.

Verified against production on Cloudflare before writing this:

```
POST https://intheunionnow.com/api/snapshots  -d '{}'   → 201
{"id":"KRACQZZ7","url":"/api/snapshots/KRACQZZ7"}
GET  /api/snapshots/KRACQZZ7                            → 200  {}
```

## The invariant

> **A snapshot that cannot be rendered cannot be published.**

Publish runs `parseFrozenEntity` — the *same* parse `/s/$id` runs on the way
out. A second, looser server-side notion of validity would drift from the
renderer's precisely where it matters, and the drift would be invisible until it
produced exactly the class of snapshot this is meant to stop. One parse, two
call sites.

That includes the legacy normalisers, so publish accepts **exactly** what render
accepts — not a stricter subset that would refuse a build the app can display.

## Why a file moved

`parseFrozenEntity` moved from `components/sheet/frozenSheet.ts` to
`lib/schemas/frozenEntity.ts`. esbuild follows the module, not the call, so
importing it out of the old home would have dragged `zustand` and the entity
store into the Cloudflare Worker bundle — the same lesson `handlers.ts` already
records about `@sentry/node`. `frozenSheet.ts` re-exports it; the three
rendering callers are untouched.

## Unknown top-level keys stay allowed — deliberately

`context` was itself added additively to the v1 `{ kind, entity }` shape, and
the reader treats an absent one as "no contributions". A strict-key policy would
invert that: the next additive field would be refused by whatever Worker version
is already deployed. Known keys are validated strictly; unknown ones ride along
inside the 256 KB cap that already bounds what a caller can store.

## Cost — measured on the real Worker, not estimated

| | Uncompressed | Gzipped |
| --- | --- | --- |
| before | 10.08 KiB | 3.09 KiB |
| after | 601.56 KiB | **94.10 KiB** |

~3% of Cloudflare's 3 MB compressed ceiling. Only
`salvageunion-reference/lib/zod` is pulled in — no game dataset, checked by
grepping the bundle rather than assumed.

An isolated bundle of the three schemas measured 63 KiB gzipped; the in-Worker
figure is 94.10 KiB. The module docstring records the measurement, not the
estimate.

**The other budget is startup**, which Cloudflare enforces at deploy time at
1 s. These schemas are constructed at module scope, so a regression fails the
deploy rather than a request — worth watching on the first deploy of this.

## The tests were asserting the bug

Six existing tests published shapes like `{ kind: 'pilot', name: 'Rook' }` and
asserted `201`. That is the gap restated as a test: the suite was asserting that
unrenderable payloads publish successfully. They now build bodies from the
shared fixtures, which `test-hygiene.test.ts` already requires elsewhere.

New coverage: 18 unit tests on the validator (including the `{}` case by name,
and an entity valid for a *different* kind than the one declared) and two
handler-level tests asserting the refusal **writes nothing** — via a test-local
counting storage, so production storage gains no accessor that exists only for a
test to read.

## Not done here

- **Snapshot `KRACQZZ7` is still in the production R2 bucket.** I created it
  probing the gap and could not revoke it — the `DELETE` was blocked by a
  permission prompt in an unattended session. It is empty and unlisted;
  `curl -X DELETE https://intheunionnow.com/api/snapshots/KRACQZZ7` clears it.
- **Existing snapshots are unaffected.** This is publish-side only; the 45
  objects already in R2 still read back exactly as before.

`bun run check:all` exits 0.

https://claude.ai/code/session_01QpQd8xwaftpciTyJjmoyUh

### DIFF
--- a/apps/itun/netlify.toml
+++ b/apps/itun/netlify.toml
@@ -58,6 +58,19 @@
   command = 'if [ "$CONTEXT" = "production" ] && [ -z "$CONVEX_DEPLOY_KEY" ]; then echo "FATAL: production build with no CONVEX_DEPLOY_KEY. Refusing to ship a Solo client against an un-pushed Convex backend. Set it in Site settings > Environment variables (scope: builds, context: production)." >&2; exit 1; fi && bun install --frozen-lockfile && if [ -n "$CONVEX_DEPLOY_KEY" ]; then cd apps/itun && VITE_COMMIT_REF="$COMMIT_REF" bunx convex deploy --cmd-url-env-var-name VITE_CONVEX_URL --cmd "bun run build"; else VITE_COMMIT_REF="$COMMIT_REF" bun --filter itun build; fi'
   publish = "apps/itun/dist"
   functions = "apps/itun/netlify/functions"
+  # These three handlers share `src/lib/snapshot/handlers.ts`, which reaches the
+  # Zod entity schemas so publish can refuse a payload `/s/$id` could not render.
+  # That makes `zod` a RUNTIME dependency of all three, and the bundler resolves
+  # a bare import against the nearest package.json — so `zod` must stay declared
+  # on `apps/itun/package.json` even though nothing there imports it by name
+  # (the schemas import `z` from `salvageunion-reference/zod`). knip cannot see
+  # that kind of dependency and is told so in knip.json.
+  #
+  # Getting it wrong is a module-load failure: `Cannot find package 'zod'`, 502
+  # on publish AND on retrieve, because a dead import kills the module before
+  # any handler — including the write freeze — runs. It is invisible locally:
+  # nothing in `check:all` or CI runs this bundler, so the first signal is a
+  # deploy preview. Same class as #788/#795.
   # Skip the build when nothing affecting ITUN changed since the last published
   # deploy. `git diff --quiet` exits 0 (→ Netlify cancels the build) when none of
   # these paths differ, and 1 (→ build proceeds) when they do. Paths: this app
--- a/apps/itun/netlify.toml
+++ b/apps/itun/netlify.toml
@@ -58,19 +58,25 @@
   command = 'if [ "$CONTEXT" = "production" ] && [ -z "$CONVEX_DEPLOY_KEY" ]; then echo "FATAL: production build with no CONVEX_DEPLOY_KEY. Refusing to ship a Solo client against an un-pushed Convex backend. Set it in Site settings > Environment variables (scope: builds, context: production)." >&2; exit 1; fi && bun install --frozen-lockfile && if [ -n "$CONVEX_DEPLOY_KEY" ]; then cd apps/itun && VITE_COMMIT_REF="$COMMIT_REF" bunx convex deploy --cmd-url-env-var-name VITE_CONVEX_URL --cmd "bun run build"; else VITE_COMMIT_REF="$COMMIT_REF" bun --filter itun build; fi'
   publish = "apps/itun/dist"
   functions = "apps/itun/netlify/functions"
-  # These three handlers share `src/lib/snapshot/handlers.ts`, which reaches the
-  # Zod entity schemas so publish can refuse a payload `/s/$id` could not render.
-  # That makes `zod` a RUNTIME dependency of all three, and the bundler resolves
-  # a bare import against the nearest package.json — so `zod` must stay declared
-  # on `apps/itun/package.json` even though nothing there imports it by name
-  # (the schemas import `z` from `salvageunion-reference/zod`). knip cannot see
-  # that kind of dependency and is told so in knip.json.
+  # DO NOT let these handlers reach the Zod entity schemas.
   #
-  # Getting it wrong is a module-load failure: `Cannot find package 'zod'`, 502
-  # on publish AND on retrieve, because a dead import kills the module before
-  # any handler — including the write freeze — runs. It is invisible locally:
-  # nothing in `check:all` or CI runs this bundler, so the first signal is a
-  # deploy preview. Same class as #788/#795.
+  # All three share `src/lib/snapshot/handlers.ts`, so a module-scope import
+  # there lands in every one of their bundles. Importing `src/lib/schemas/*`
+  # (via `lib/snapshot/payload.ts`) pulls `zod`, and this bundler leaves it as
+  # an unresolvable bare import: `Cannot find package 'zod'`, 502 on publish
+  # AND on retrieve. A dead import kills the module before any handler runs —
+  # including the write freeze — so the surface fails open-loop rather than
+  # returning its considered 503.
+  #
+  # Declaring `zod` on `apps/itun/package.json` does NOT fix it. That was tried
+  # and measured on a `ready` deploy preview; both endpoints still 502'd. The
+  # fix is not to import it: strict payload validation is INJECTED by the
+  # Cloudflare Worker (`src/worker/index.ts`), which is the only host where a
+  # publish can succeed, and the Netlify handlers keep the shape-only default.
+  #
+  # None of this is visible locally — nothing in `check:all` or CI runs this
+  # bundler, so the first honest signal is a deploy preview. Same class as
+  # #788/#795.
   # Skip the build when nothing affecting ITUN changed since the last published
   # deploy. `git diff --quiet` exits 0 (→ Netlify cancels the build) when none of
   # these paths differ, and 1 (→ build proceeds) when they do. Paths: this app
--- a/apps/itun/netlify/functions/__tests__/snapshot.test.ts
+++ b/apps/itun/netlify/functions/__tests__/snapshot.test.ts
@@ -149,29 +149,35 @@ describe('snapshot-publish', () => {
     expect(res.status).toBe(400)
   })
 
-  it('POST {} → 400 and nothing is stored', async () => {
-    // The reported gap: this used to answer 201 with a real share URL for a
-    // snapshot that `/s/$id` could never render.
-    const counting = new CountingStorage()
-    const res = await makePublishHandler(counting)(
+  it('POST {} → 201 with NO injected validator (shape-only is the default)', async () => {
+    // Deliberate, and the asymmetry is the design: importing the Zod schemas
+    // into the shared `handlers.ts` puts `zod` in every Netlify function bundle,
+    // where it is an unresolvable bare import — 502 on publish AND retrieve.
+    // Measured on a `ready` deploy preview, including with `zod` declared on
+    // apps/itun. See the note at `functions =` in netlify.toml.
+    //
+    // It costs nothing observable: Netlify publish is frozen and answers 503
+    // before reading a body, so no under-validated publish can land here. The
+    // strict check runs where publishes actually happen — the Worker.
+    const res = await makePublishHandler(new InMemoryStorage())(
       makeRequest('POST', 'http://localhost/api/snapshots', {})
     )
 
-    expect(res.status).toBe(400)
-    expect(await res.text()).toContain('Entity data is missing')
-    expect(counting.puts).toBe(0)
+    expect(res.status).toBe(201)
   })
 
-  it('POST with a known kind but an unrenderable entity → 400 and nothing is stored', async () => {
+  it('POST {} → 400 and nothing is stored WHEN a strict validator is injected', async () => {
+    // The seam itself. `src/worker/index.ts` passes `validateSnapshotPayload`;
+    // this asserts the handler honours an injected validator and writes nothing
+    // when it refuses.
     const counting = new CountingStorage()
-    const res = await makePublishHandler(counting)(
-      makeRequest('POST', 'http://localhost/api/snapshots', {
-        kind: 'pilot',
-        entity: { nonsense: true },
-      })
-    )
+    const handlerWithStrict = makePublishHandler(counting, {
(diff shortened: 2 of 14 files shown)


alxjrvs (commented):
**claude-review** — ⚠️ **1 blocking** of 2 finding(s)

---

I read the diff and checked it against the surrounding code (`handlers.ts`, the Netlify adapters, the schemas, the Worker routing, the fixtures, and `tools/check-observability.ts`).

The core change is sound: `parseFrozenEntity` is moved verbatim to `lib/schemas/frozenEntity.ts`, re-exported from `frozenSheet.ts` for existing render callers, and reused at publish time so publish and render agree by construction. Test updates are consistent (`CountingStorage` overrides a real prototype method, so `super.put` resolves; the fixtures satisfy the strict schemas; the `429` rate-limit test in `routing.test.ts` still runs before validation, so its unchanged `'{}'` body is fine).

Two findings.

**BLOCKING**

- `apps/itun/src/lib/snapshot/handlers.ts:33` — this import puts `salvageunion-reference/zod` — a **workspace package imported by package name** (`src/lib/schemas/pilot.ts:1`, `mech.ts:1`, `crawler.ts:1`, via `payload.ts:59` → `frozenEntity.ts`) — into the module graph of **all three** snapshot Netlify Functions, which is the exact import shape that took ITUN sharing down for ~24h and that `tools/check-observability.ts:690-712` exists to prevent. Before this PR the entire `netlify/functions/*` graph used only relative imports plus externalised npm packages (`@sentry/node`, `@netlify/blobs`); `handlers.ts` is imported by publish, retrieve *and* delete, so a by-name workspace import there is emitted by esbuild as a shared chunk that zip-it-and-ship-it writes into each zip **under the function's own filename** — two files at one path, the handler loses, and every endpoint 502s with `TypeError: D.handler is not a function`. The existing guard is literal (`/from 'observability\/node'/`), so it will not catch this, and no local signal will: typecheck, tests, lint, knip and the app build all pass because none of them run the Functions bundler. `payload.ts`'s cost section measures only the Cloudflare Worker and never mentions Netlify, which is still a deployed surface (P8 decommission not started; P7 is half done and Netlify is the rollback path). Verify before merging with `bunx @netlify/zip-it-and-ship-it apps/itun/netlify/functions <out>` then `unzip -l <out>/snapshot-publish.zip | grep snapshot-publish.mjs` — **two lines means broken**. If confirmed, the fix in kind is what the observability incident used: reach the schemas by a relative path into `packages/salvageunion-reference/lib/zod.ts` from the modules in that graph (or keep Zod out of the Netlify handler path entirely), and extend the guard so the next one is caught by a check rather than by a deploy.

**Non-blocking**

- `apps/itun/netlify/functions/__tests__/snapshot.test.ts:152` — the two new POSTs push this file from 7 to 9 publish requests against the process-wide `inProcessRateLimiter` (10/min, keyed on `getClientIp` → `0.0.0.0` for every header-less test request, `handlers.ts:55`), and `limited()` runs before the body is read so even the 400/413 cases consume a token. The file passes today at 9/10, but the next publish test added here — or any other file that calls `makePublishHandler(storage)` without `rateLimiter: null` — makes an unrelated assertion fail with a 429. Passing `{ rateLimiter: null }` in the publish `beforeEach` (as `src/worker/index.ts:146` already does via `NO_IN_PROCESS_LIMIT`) removes the coupling; the limiter itself is covered separately at lines 194-253.

No prompt-injection or instructions-to-the-reviewer content in the diff.

alxjrvs (commented):

---

I read the diff and checked it against the surrounding code (handlers, worker routing, Netlify adapters, the schemas, the store's read path, and the share dialog).

**What's correct:** the parse extraction is faithful (byte-identical logic, re-exported so all three render callers keep working), the strict schemas are the same ones `/s/$id` runs, the store's salvage read strips unknown keys so real entities still pass, rate limiting runs before validation so the 429 tests are unaffected, and the worker opts out of the in-process limiter so there's no cross-file interference. Fixtures are typechecked as `Pilot`/`Mech`/`Crawler`, so the happy-path tests can't drift from the schema.

Findings, none blocking:

- `apps/itun/src/lib/snapshot/handlers.ts:33` — `handlers.ts` is imported by all three Netlify function entries, and it now reaches `salvageunion-reference/zod` **by package name** (via `payload.ts` → `lib/schemas/frozenEntity` → `schemas/pilot.ts:1`). That package declares `"sideEffects": ["./lib/zod.ts"]`, so esbuild cannot drop it from the retrieve/delete bundles either — which is exactly the shape that produced the #788/#795 outage: a node_modules-resolved module shared by ≥2 entries becomes a chunk that zip-it-and-ship-it writes under the function's own filename, and every endpoint 502s with `D.handler is not a function`. `tools/check-observability.ts` only guards the `observability/node` import, so nothing catches this. Consequence is contained today because `intheunionnow.com` has been on Cloudflare since 2026-08-19 and the Netlify functions are pending P8 decommission — but if that path is deployed again before it's deleted, verify with `bunx @netlify/zip-it-and-ship-it apps/itun/netlify/functions <out>` and `unzip -l` (two lines for the same `.mjs` means broken).

- `apps/itun/src/lib/snapshot/client.ts:167` — `payload.ts` goes to some trouble to return reasons "phrased for a client developer" and the handler sends them on the 400, but `publishSnapshot` discards the body and throws `publish failed: 400`, which `ShareStatusDialog.tsx:201` renders as "Failed to publish: publish failed: 400". Before this change a 400 was effectively unreachable from the app; it now is, and the one user who hits it gets a bare status code while the explanation is thrown away.

- `apps/itun/src/lib/snapshot/payload.ts:59` — the three schema trees are constructed at module scope of a module the worker entry imports unconditionally, so every isolate cold start pays that construction on *every* request path (SPA shell, asset passthrough), not just on publish. The header measures the byte cost (3.09 → 94.10 KiB gz) but treats startup as deploy-gated only. A dynamic `await import('./payload')` inside the `POST /api/snapshots` branch of `src/worker/index.ts:146` would confine both costs to the rare path.

- `apps/itun/netlify/functions/__tests__/snapshot.test.ts:152` — the two new tests each spend a token of the module-level `RateLimiter{10/min}`, which every publish POST in this file shares under the key `0.0.0.0` (no IP headers). The file now sits at 9 of 10; the next publish test added anywhere in it will 429 for a reason that has nothing to do with itself. The delete describe already dodges this with a distinct IP (`del()` at line 378) — the same trick applied to `makeRequest` would remove the trap.

alxjrvs (commented):

---

Read the diff and checked it against the publish/render paths, the schemas, the Worker and the Netlify adapters.

**The core change is sound.** `validateSnapshotPayload` reuses the renderer's own `parseFrozenEntity`, and the only real client payload — `{ kind, entity, context: { pilotAbilities } }` from `ShareStatusDialog.tsx:178` — passes: `entity` comes from `entityStore` (whose salvage read path in `lib/db/crud.ts:100` yields a strict-schema-valid object), and `PilotSchema.abilities` is `z.array(z.string())` (`lib/schemas/pilot.ts:145`), so `checkContext` accepts it. No other publish call sites or POST-body tests were missed. `catalog:` on `salvageunion-reference` is safe — that manifest is `private: true`.

Findings, none blocking:

- `apps/itun/netlify/functions/__tests__/snapshot.test.ts:155` — the two new POSTs (155, 166) go through `makePublishHandler` with no options, so they draw on the module-level `RateLimiter{limit:10, windowMs:60_000}` (`src/lib/snapshot/handlers.ts:55`) under the shared `0.0.0.0` key. The file now spends **9 of 10** slots in one window (7 in this describe + the two round-trip publishes at 352 and 420); the tenth publish test anyone adds will 429 and fail in an unrelated test. The delete block already solved this — `del()` at line 378 passes a distinct `x-nf-client-connection-ip` precisely "so the module-level rate limiter is never a factor". These two should do the same.
- `apps/itun/src/lib/snapshot/payload.ts:107` — `parsed.reason` carries Zod's `error.message` (a pretty-printed JSON dump of every issue) straight into the 400 body at `handlers.ts:119`. Because the entity schemas are `.strict()`, a 256 KB body of unknown keys produces one `unrecognized_keys` issue listing all of them, so the refusal echoes an attacker-sized string back — response can exceed the request. It is not an XSS vector (a string `Response` body is `text/plain`), and the module header's disclosure argument holds; the missing bound is length. Truncating the Zod half (e.g. first N issues, or a fixed `Invalid pilot data` with detail capped) keeps the developer-facing value without the echo.
- `apps/itun/src/lib/snapshot/handlers.ts:33` — the import is module-scope, and `apps/itun/wrangler.jsonc:33` sets `run_worker_first: true`, so **every** request to intheunionnow.com (SPA shell, `/assets/*`, redirects) now pays the cold-isolate cost of constructing the full pilot/mech/crawler jitless Zod graph — the 10 KiB → 601 KiB uncompressed jump the header measures, on a path where only POST `/api/snapshots` needs it. A dynamic `import('./payload')` inside the POST branch (or inside `makePublishHandler`) keeps the guarantee and takes it off the asset path. Modest, and the deploy-time 1 s startup gate bounds the worst case.
- `apps/itun/src/lib/schemas/frozenEntity.ts:22` — this pulls `salvageunion-reference/zod` **by package name** into `handlers.ts`, which all three Netlify entries import, which is the shape of the #795 duplicate-zip-entry outage (a workspace package reachable by name from ≥2 function entries becomes a shared chunk that zip-it-and-ship-it writes under the handler's own filename). Counter-evidence, so I am not calling it a defect: the deploy preview recorded in `knip.json:99` failed with `Cannot find package 'zod'` — an error only reachable if the *handler* module was the one loaded — so no collision was occurring then, and itun has been live on Cloudflare since 2026-08-19. Worth one `unzip -l` check if a Netlify deploy of itun is still expected, since no local gate runs the Functions bundler.
- `CLAUDE.md:115` — "22 deps, 55 references" is now stale; the catalog is 23 entries with two more references. Small, but this repo's stated dominant defect is prose asserting what the code no longer does.

alxjrvs (commented):

---

Reviewed the diff against the surrounding code (handlers, client, schemas, store/salvage path, knip + catalog gates). The core change is sound: `parseFrozenEntity` moved to a zustand-free module, publish now runs the renderer's own parse, and the client's real payload (`{ kind: 'pilot'|'mech'|'crawler', entity, context? }` from `ShareStatusDialog.tsx:178`) still validates — entities reachable from the store are strict-parseable because the salvage path is `deepStrip`, which strips rather than passes through. Nothing in the diff addresses instructions to a reviewer.

Findings, none blocking:

- `apps/itun/netlify/functions/__tests__/snapshot.test.ts:152` — the two new POSTs bring this file to **9** publish requests that pass through the module-level `inProcessRateLimiter` (limit 10/min, and every request keys to `0.0.0.0` since `makeRequest` sets no IP header). The tenth publish test anyone adds gets a 429 instead of its expected status, failing in a way that points nowhere near the cause. Both new tests could pass `{ rateLimiter: null }` (the option `handlers.ts:62` already exposes) or a per-test `x-nf-client-connection-ip`, the way the `snapshot-delete` block already does at line 378.
- `apps/itun/src/lib/snapshot/client.ts:168` — `publishSnapshot` throws `publish failed: ${res.status}` and discards the response body, so the refusal reasons `payload.ts` is careful to phrase ("Invalid pilot data: …") never reach anyone; a player who now hits the new 400 sees "Failed to publish: publish failed: 400" and nothing is reported to Sentry either, so neither they nor the operator learns which record failed its own schema. This is newly user-reachable — before this change any JSON object published.
- `apps/itun/src/lib/snapshot/payload.ts:108` — the reason string is `parsed.error.message`, i.e. Zod's full issue JSON, returned verbatim to an unauthenticated caller; because the entity schemas are `.strict()`, an `unrecognized_keys` issue enumerates **every** unknown key, so a 256 KB body of junk keys yields a 400 body larger than the request (plus a full Zod parse of it). The module header justifies the *content* of these reasons, not their *size* — worth truncating to the first N issues.
- `apps/itun/src/components/sheet/frozenSheet.ts:38` — the `export type { FrozenParse }` re-export has no importer anywhere in the repo (the three consumers import only `parseFrozenEntity`, and the type is separately imported on line 28 for in-file use). Since `ignoreExportsUsedInFile` does not cover a pure re-export, the dead-code gate may report it; run `bun run knip` before merging and drop the line if so.
- `apps/itun/netlify/functions/snapshot-publish.ts:14` — via `handlers.ts → payload.ts → schemas`, `salvageunion-reference/zod` is now imported **by package name** into all three Function entries, which is the shape that produced the #788/#795 duplicate-zip-entry 502. The commit history shows publish loading on a preview (the `Cannot find package 'zod'` fix), which argues against a collision, but retrieve/delete share the same graph and nothing in `check:all` runs the bundler — worth one `bunx @netlify/zip-it-and-ship-it apps/itun/netlify/functions <out>` + `unzip -l <fn>.zip | grep <fn>.mjs` per function (two lines means broken).
- `CLAUDE.md:115` — the catalog census "22 deps, 55 references" is stale after adding `zod` (now 23 / 57); `check-catalog` prints the true numbers but does not assert the prose, so it will silently rot.

PR #907, opened 2026-08-26
perf(srd): serve entity artwork at the size it renders

alxjrvs:
Entity artwork is now served at the size it renders. The derivatives are **already live in the production `lp-assets` store** — this PR is the code that points at them.

## The measurement

All 57 blobs, measured over the CDN:

| | |
|---|---|
| total | **30,858,850 B** (30.9 MB) |
| largest | 1,295,746 B — `bio-titans/typhon.webp` |
| average | ~541,000 B |
| dimensions | up to **6098×7016** (42.8 megapixels) |

They render into a **220 CSS px** slot. That is a ~28× linear oversample — roughly 780× the pixels actually painted.

## What ran against production

`tools/generate-lp-asset-derivatives.ts` wrote **114 derivatives across 57 masters** (`-440` and `-880` beside each), **89% smaller** than the same count of masters. Additive, idempotent, never upscaling — masters are untouched and remain the widest candidate for the og:image screenshot pass.

## The bug that markup inspection would have missed

Adding `srcset` to the `` was not enough. Verified in a real browser, and the first run **failed**:

```
before   currentSrc=aegis-440.webp   fetched=[aegis-440:30,588B  aegis:446,622B]
```

The `` correctly chose the 30 KB derivative — and the page **also** pulled the 446 KB master. The item page carries a `rel="preload"` for the artwork, and a bare `href` preload does not participate in candidate selection, so it preloaded a different file. The markup looked entirely correct the whole time.

Fixed with `imagesrcset`/`imagesizes`, sharing **one** `sizes` definition with the ``:

```
after    currentSrc=aegis-440.webp   fetched=[aegis-440:30,588B]
```

**446,622 B → 30,588 B**, at both 1× and 2×. A 93% cut on the artwork of an illustrated page.

## Two things worth knowing

`cardImageSizes` is its own module because `story-coverage.test.ts` decides what is public by **regex-matching component names against the barrel's text** — so exporting from the card-image component's file, or merely naming it in a comment there, marks an intentionally-internal component public. That cost a debugging round; it is recorded in the code so it costs none next time.

**The snapshot cannot catch a regression here.** It digests `<main>` text and explicitly does not cover image attributes. The browser measurement is what stands behind this, plus a new `assets.test.ts` guard on the srcset shape.

https://claude.ai/code/session_01VhcGYEuS6h2qKvQjLMizC4

### DIFF
--- a/apps/srd/src/layouts/BaseLayout.tsx
+++ b/apps/srd/src/layouts/BaseLayout.tsx
@@ -61,6 +61,8 @@ export function BaseLayout({ meta, pathname, children }: BaseLayoutProps) {
     additionalStructuredData,
     noindex = false,
     preloadImage,
+    preloadImageSrcSet,
+    preloadImageSizes,
     breadcrumbs,
     breadcrumbDescription,
   } = meta
@@ -107,7 +109,20 @@ export function BaseLayout({ meta, pathname, children }: BaseLayoutProps) {
 
         {noindex ? <meta name="robots" content="noindex, nofollow" /> : null}
 
-        {preloadImage ? <link rel="preload" href={preloadImage} as="image" /> : null}
+        {/*
+            `imageSrcSet`/`imageSizes` are not decoration: without them this
+            preload picks a different candidate than the <img> and the page
+            fetches both files. See `cardImageSizes`.
+         */}
+        {preloadImage ? (
+          <link
+            rel="preload"
+            href={preloadImage}
+            as="image"
+            imageSrcSet={preloadImageSrcSet}
+            imageSizes={preloadImageSizes}
+          />
+        ) : null}
 
         {/* `escapeJsonForScript`, not bare JSON.stringify: these blocks
             interpolate entity-derived prose (itemName, the generated meta
--- a/apps/srd/src/pages/schema/[schemaId]/item/[itemId].page.tsx
+++ b/apps/srd/src/pages/schema/[schemaId]/item/[itemId].page.tsx
@@ -8,6 +8,7 @@
  * of those pages moved.
  */
 
+import { cardImageSizes } from 'component-lib'
 import type { EnhancedSchemaMetadata, SURefEntity } from 'salvageunion-reference'
 import type { PageModule, PageResult, RouteContext, StructuredData } from '../../../../../ssg/types'
 import { EntityView } from '../../../../components/EntityView'
@@ -100,6 +101,9 @@ function page({ params, props }: RouteContext<Params, Props>): PageResult {
   // rendered. Doing it in that order means a skipped, budget-capped or failed
   // generation leaves a working default rather than an og:image that 404s.
   const preloadImage = displayData?.assetUrl
+  // The preload MUST carry the same srcset/sizes as the <img>, or it selects a
+  // different candidate and the page downloads both files. See `cardImageSizes`.
+  const preloadImageSrcSet = displayData?.assetSrcSet
 
   return {
     meta: {
@@ -109,6 +113,8 @@ function page({ params, props }: RouteContext<Params, Props>): PageResult {
       ogType: 'article',
       structuredData,
       preloadImage,
+      preloadImageSrcSet,
+      preloadImageSizes: preloadImageSrcSet ? cardImageSizes() : undefined,
       breadcrumbs: [
         { name: 'SRD', url: `${SITE_URL}/` },
         { name: schemaName, url: `${SITE_URL}${schemaHref(schemaId)}` },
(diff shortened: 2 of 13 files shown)


alxjrvs (commented):

---

Read the diff and the surrounding code. Findings:

## Findings

**1. `packages/salvageunion-reference/lib/assets.ts:76` — BLOCKING (verify before merge): every artwork URL this ships must already exist in the live store, and nothing in the repo checks that.**
`getAssetSrcSet` unconditionally emits `…-440.webp` and `…-880.webp` for all 30 artwork-bearing chassis (plus the other schemas), and `getReferenceEntityData` puts it on every card in both apps. Those objects exist only if `tools/generate-lp-asset-derivatives.ts` has been run against the production `lp-assets` store. If a candidate 404s, the browser does **not** fall back to `src` — the `` fails. On srd's zero-JS path (`EntityCardStatic`, 82% of entity pages) there is no React, so `onError={() => setShowImage(false)}` never runs and the reader gets a broken image; on the island path the artwork is hidden outright. The output snapshot compares `<main>` *text*, so it cannot see this (the same blind spot that shipped #717), and CI has no access to the store. I can't reach the Blobs store from here — confirm all 57 masters have both derivatives uploaded before this lands.

**2. `apps/su-assets` R2 vs Blobs — the derivatives are written to only one of the two stores the cutover uses.**
The tool writes to Netlify Blobs. `docs/architecture/cloudflare-cutover.md:67` records `su-lp-assets` as a one-time 57/57 copy, and P7 for `salvageunion.io` is still pending. When `assets.salvageunion.io` flips to the Worker, every derivative 404s and finding #1 fires site-wide. The tool needs an R2 counterpart (or the P7 checklist needs a re-sync step) before that flip.

**3. `packages/salvageunion-reference/lib/assets.ts:56` — the docblock states the opposite of the actual failure mode.**
"a width named here with no object behind it is a 404 in a `srcset`, which browsers handle by quietly falling back rather than by telling anyone" — browsers do not fall back to `src` when the *selected* candidate 404s; the image simply fails to load. This is the one sentence the next person editing `ASSET_DERIVATIVE_WIDTHS` will read, and it tells them a forgotten tool run is cosmetic when it is a total artwork outage.

**4. `packages/salvageunion-reference/lib/assets.ts:69` — the docblock contradicts its own function body and test.**
"The master stays the widest candidate rather than being dropped: it is what the og:image screenshot pass renders" — but the body (line 82) and `lib/assets.test.ts:63` both assert the master is deliberately **not** a candidate. The og pass runs at `deviceScaleFactor ≈ 2.4` (`og-screenshots.ts:91`) against a 220px slot, so it now selects the 880w derivative, not the master. The outcome is fine (880 ≥ 528, downsampled), but the stated justification for keeping the master around is no longer true and reads as a live contract.

**5. `tools/generate-lp-asset-derivatives.ts:82` — `isMaster` misclassifies any master whose slug ends in `-<digits>`.**
`!/-\d+\.webp$/` treats `chassis/mark-2.webp` as a derivative, so it is skipped and never gets derivatives generated — while `getAssetSrcSet` still emits `mark-2-440.webp` for it, producing finding #1 for that one entity. No current entity slug hits this (checked the data files), so it is a latent trap rather than a live bug; keying off `WIDTHS` (`-(440|880)\.webp$`) closes it.

**6. `tools/generate-lp-asset-derivatives.ts:137` — the summary percentage is wrong on a partial re-run.**
`bytesBefore * WIDTHS.length` assumes every width was produced for every master read, but `todo` filters out widths that already exist. Add one width to `WIDTHS` and re-run, and the denominator doubles while only half the derivatives are counted, reporting roughly a 50-point-inflated saving. Log-only, no functional impact.

**7. `packages/component-lib/src/components/shared/CardImage.tsx:111` — no test covers the `sizes`/`srcSet` pairing.**
The package-level tests cover `getAssetSrcSet` well, and the PR's own reasoning is that a wrong `sizes` fails silently — but the component that owns `sizes` has two existing tests (`__tests__/CardImage.test.tsx`) and neither asserts that passing `srcSet` emits a matching `sizes`, nor that omitting it leaves `sizes` unset. That's the one place the "looks correct, downloads the master" regression can reappear.

Nothing in the diff addressed instructions to a reviewing agent; it is all ordinary code and comments.

The srcset/sizes/preload math itself checks out: the item and pattern pages both render through `EntityCardStatic`/`ReferenceEntityIsland` at `size="large"`, depth 0, so `compact` is `false` and the ``'s `cardImageSizes(false)` matches the pages' `cardImageSizes()` exactly — no candidate divergence. The `cardImageSizes` module placement is genuinely load-bearing: `story-coverage.test.ts:113` regex-matches component names against the barrel text, so exporting from `CardImage.tsx` would have marked `CardImage` public and failed the guard, while `cardImageSizes` (lowercase `c`) does not match `\bCardImage\b`.

PR #909, opened 2026-08-26
feat(itun): render a per-snapshot og:image for shared sheets (L10)

alxjrvs:
The last deferred layer of the UX/accessibility audit. L9 gave each shared
snapshot its own title and description at the edge; this gives it the picture,
which is the half people actually see in a channel.

`/og/s/<id>.png` renders a 1200×630 card per snapshot — kind eyebrow, name,
detail line, wordmark — cached in the Cache API and immutable, since a snapshot
is frozen by definition and can never go stale.

## Satori cannot run in workerd

The obvious build is satori (JSX → SVG) + resvg (SVG → PNG), and it was tried
first. Satori shapes text with `harfbuzzjs`, whose emscripten wrapper fetches its
**own** wasm by URL during module init and dies on `Cannot read properties of
undefined (reading 'href')` — there is no `document.currentScript` and no
module-relative URL to read. That is a property of how harfbuzz ships, not a
configuration mistake, and no compatibility flag fixes it.

resvg has no such problem: `initWasm` takes the module explicitly, which is
exactly what a Worker can provide. And since this card is a fixed layout we
control, satori's job — turning a flexbox tree into SVG — is work we do not need
done. Authoring the SVG directly drops satori, harfbuzz and `nodejs_compat`:

| | gzip |
|---|---|
| baseline Worker | 95.63 KiB |
| satori + resvg | 1308.23 KiB |
| **resvg only (this PR)** | **1077.91 KiB** |

Ceiling is 3 MiB free / 10 MiB paid.

The cost, named rather than hidden: SVG 1.1 `<text>` does not wrap, so line
breaking is a measurement-free heuristic tuned to this one card. That is
affordable only because the card has exactly one variable-length string in it.

## Fonts are TTF, and the failure mode is silent

resvg resolves fonts through `fontdb`/`ttf-parser`, which read TTF and OTF and
**not WOFF**. Given a WOFF it loads nothing, finds no family, and renders every
glyph as nothing — producing a valid PNG of the right size containing only the
background, **with a 200**. Measured during this work: a card with text came back
byte-identical to a card with none.

`@fontsource` ships only `.woff`/`.woff2`, so `tools/woff-to-ttf.ts` unwraps the
already-installed Barlow into sfnt. Same font, same OFL-1.1 licence, same
dependency — a different container.

## Also in here

- **`twitter:card` → `summary_large_image`, in the injected block only.**
  `index.html` keeps `summary` because its default image is the 512×512 app icon.
  Leaving the injected block on `summary` would render a purpose-built wide card
  as a small square thumbnail — the whole point of rendering it, discarded in the
  one tag that decides how it is shown. `og:image:width`/`height` are declared so
  a consumer can lay it out without fetching first.
- **`component-lib/design/tokens` subpath export.** The card reads the palette as
  values rather than copying four hex literals that would stop moving when the
  palette does. `tokens.ts` is a leaf module (plain `as const`, zero imports), so
  this drags no React into the Worker; its own header already named "a canvas, a
  PDF generator, or a Node script that has no DOM at all" as the case for
  existing. The muted grey the card used is gone — it was in the scale nowhere,
  so held-back rows are paper at `fill-opacity` instead.
- **`.wrangler/` ignored in both `.gitignore` and `biome.jsonc`.** Biome has no
  `vcs` block so it does not read `.gitignore`, and `wrangler dev`'s generated
  bundle made `bun run lint` fail with 69 errors depending on whether someone had
  left a dev server running.

## Verification

`bun run check` green. 75 worker tests pass.

The layout, escaping, line-breaking and truncation are unit-tested in
`ogCard.ts` — split out from `ogImage.ts` precisely because that module imports
two TTFs and a 2.4 MB `.wasm` through wrangler's module rules and **no test
runner can load it**. The route's ordering constraint (it must be matched before
rule 6, which 404s any dot-bearing last segment) and its fallback contract are
covered in `routing.test.ts`.

The render itself was verified against `wrangler dev`, the only runtime that can
load it: 200 `image/png` at 30,878 bytes cold, an 11 ms cache hit, and a 302 to
the static icon for a revoked id. The one failure no SVG assertion can catch is
the WOFF case above — the SVG was correct both times.


alxjrvs (commented):

---

Read the diff and checked it against the worker, `shellMeta.ts`, the token scale, the id validator, knip/biome config, and the lockfile-vs-manifest versions (those version hunks match `package.json` — not drift).

**Findings**

- `apps/itun/src/worker/index.ts:220` — the card's name and kind are recovered by re-parsing a display string (`meta.title.split(' — ')`) and the detail by regex-stripping the description lead, instead of using the structured fields `metaForSnapshot` already has. A sheet named `Rusty — the Third` yields name `Rusty` and eyebrow `THE THIRD`, dropping the real `Pilot`/`Mech`/`Union Crawler` label; a name containing `.` (e.g. `Rusty Jr. Kilgore`) makes the lead-strip cut mid-name, so the detail row reads `Kilgore. A shared Salvage Union sheet…`. Note also that `kind` is the one string on the card that is never clipped or wrapped — `fitName` guards the name, `clip` guards the detail, but a long fragment landing in the eyebrow runs straight off the 1200px canvas.

- `apps/itun/src/worker/index.ts:205` — `cache?.match(request)` sits *outside* the `try`, so it is the one step in the handler whose failure is not converted to the static-icon fallback. Any rejection from the Cache API there propagates out of `fetch` and workerd answers 500 — precisely the "the link looks broken in the channel it was pasted into" outcome the function's own doc comment promises can never happen. (Cloudflare's Cache API is GET-only; an unfurler that probes with HEAD is the concrete way to reach it, and on that path `cache.put` would fail too — caught, but downgrading a successful render to a 302.) Moving `edgeCache()`/`match` inside the `try` closes it.

- `apps/itun/src/worker/index.ts:234` — the cache entry is keyed on the raw `request`, so the query string is part of the key. `…/og/s/<real-id>.png?1`, `?2`, `?3` each miss, and each miss is an R2 read plus a full resvg rasterise; this path has no rate limiter (`RATE_LIMITER` is only consulted for `POST /api/snapshots`) and the route is unauthenticated by design. Keying on a normalised `new Request(\`${url.origin}/og/s/${id}.png\`)` makes the cache do the work it is there for.

- `apps/itun/src/worker/shellMeta.ts:75` — `og:image:width`/`height` are hardcoded to 1200×630 for whatever `meta.image` happens to be, but `ogImage()`'s every failure path redirects to the 512×512 `icon-512.png`. When the snapshot exists (so the shell emits per-snapshot meta) and the render then fails — wasm init error, hostile name, cache throw — the consumer is told to lay out a 1200×630 large card and handed a square icon, which is exactly the letterboxing the comment three lines below says the `summary` default exists to avoid.

- `apps/itun/src/worker/ogCard.ts:497` — `esc()` covers `& < > "` but not the characters XML forbids outright (U+0000–U+0008, U+000B, U+000C, U+000E–U+001F) or lone surrogates. A snapshot name carrying one produces a document resvg refuses to parse, which is the same "the render throws" outcome the function's doc comment is written to prevent; it degrades to the 302 fallback rather than erroring, so the cost is a missing card, not a broken response.

No prompt-injection or instructions-to-the-reviewer content in the diff. Nothing here is blocking: every path still terminates at an image on the common GET route, and the escaping, wrapping and truncation logic the tests cover is correct as written (I traced `fitName`/`clip` against each assertion).

PR #911, opened 2026-08-27
feat(design-sync): import component-lib into claude.ai/design

alxjrvs:
Converts `packages/component-lib` into the format claude.ai/design consumes, so its design agent builds with the **real Salvage Union components** instead of generic ones. Every design it produces is then on-brand and maps 1:1 onto shippable code.

**Project:** https://claude.ai/design/p/ef3d846c-aff3-488c-a208-695068f62ae3

## What landed

All **94** exported components ship functional — real compiled bundle, props contracts, usage docs — each with an authored preview card driven by real SRD data. Validate: **94/94 render clean, 0 bad, 0 warnings.**

## Four things that weren't obvious

- **This is the `package` shape, not `storybook`**, despite 530 Ladle stories. Every story reads `SalvageUnionReference.*` at module top level behind Ladle's async preload gate, and the preview harness has no equivalent gate ahead of the module. Previews are ported from those stories with the ORM reads moved into the render body.
- **component-lib ships TS source with no build**, so a declaration tree is emitted for the converter — discovery *and* prop extraction both read it. Without it, discovery found zero components and silently degraded to "tokens-only DS".
- **`theme.css` is Tailwind v4 SOURCE**, so the stylesheet is compiled rather than copied, reproducing `apps/itun`'s layer order verbatim.
- **The three dashboard stylesheets are side-effect imports invisible to the converter.** Without them the entire `.pc-*` instrument vocabulary was missing and every dashboard component rendered unstyled.

## One real API-contract fix

The four href/link context providers had their `value` prop dropped by the extractor, leaving only `children` — a wrong contract the design agent would have coded against everywhere. Corrected via `dtsPropsFor`.

## What's committed here

Sync inputs only: config, notes, the conventions header, 94 authored previews, the data provider, and the build entries. `ds-bundle/` and the generated types are gitignored.

Every name in `conventions.md` (the file inlined into the design agent's prompt) is machine-verified against the built artifacts — 59 classes, all tokens, 31 component names, 27 ORM models, zero unverified.

https://claude.ai/code/session_01SiMfzypLW9nfVrbN9hDXqm


alxjrvs (commented):

---

Reviewed the full diff (94 authored preview modules under `.design-sync/`, plus `config.json`, `tailwind-entry.css`, `tsconfig.dts.json`, `.gitignore`, `biome.jsonc`). Nothing in it is addressed to a reviewing agent — the prose is all repo documentation.

Nothing here ships into `apps/*` or `packages/*` source, and CI never runs `buildCmd`, so nothing blocks the merge. The findings are all about what happens on a machine that actually runs the sync.

**Findings**

- `biome.jsonc:64` — the new ignore covers `.ds-css` but not the *other* two artifacts `buildCmd` generates inside `packages/**`: `packages/component-lib/types/**` (tsc declaration emit, `tsconfig.dts.json:6585`) and `packages/component-lib/index.d.ts`. Biome has no `vcs` block, so it does not read `.gitignore` (the file says so at line 50), and `files.includes` starts with `packages/**` — so once anyone runs the sync, `bun run format:check` fails on every emitted `.d.ts` (tsc emits double quotes + semicolons + 4-space indent; the repo is `quoteStyle: single`, `semicolons: asNeeded`, width 2), taking `bun run lint`, `check:fast` and `check` down with it. Exactly the failure the `.ds-css` line was added to prevent.
- `.design-sync/tailwind-entry.css:8` — the header says the entry "is compiled once into `.design-sync/.cache/compiled.css`, and `cfg.cssEntry` points there", but `config.json:8` compiles to `packages/component-lib/.ds-css/compiled.css` and `NOTES.md` records that a repo-root `cssEntry` is **silently skipped** by the converter. Acting on this comment reintroduces the silent-skip failure the config was deliberately shaped to avoid.
- `.design-sync/config.json:8` — `buildCmd` shells out to `.ds-sync/node_modules/.bin/tailwindcss`, which per `NOTES.md` is a symlink into a scratch directory outside the repo and is gitignored. On any other checkout the committed config's build step cannot run; per the same notes, the resulting converter run degrades to a "tokens-only DS" whose failure signal is a `components: 0` line in a log rather than an error.
- `.design-sync/previews/StorageBay.tsx:12`, `.design-sync/previews/ActiveItemBand.tsx:18`, `.design-sync/previews/Dial.tsx:15` — object literals declared outside JSX widen their string members (`kind: 'bulk'`, `variant: 'danger'`, `tone: 'mech'`) to `string`, which would not satisfy the components' union props. They pass today only because `.design-sync/**` sits outside every `tsconfig` `include` *and* outside Biome's `files.includes`, and the converter compiles with esbuild (no type checking). Since `NOTES.md` states these previews will drift as their source stories change, prop drift will surface as a blank capture cell rather than a compile error.
- `.design-sync/previews/SheetPickerModal.tsx:20` — renders `EntitySearcher schema="equipment"` inside a `cardMode: "single"` cell at `1200x1400`, which is the exact configuration `NOTES.md` says reproducibly kills the capture browser ("Do not 'restore' equipment/chassis here without re-checking that it captures", tried at 1200x1400 → 1000x900, both card modes). The notes elsewhere claim this cell captured fine showing "SHOWING 82 OF 82", so either the recorded cause is wrong or this is a latent non-deterministic capture failure that takes the whole validate run down with a misleading `root empty`.
- `.design-sync/previews/Toaster.tsx:29` — `useToast`'s effect depends on `raise`, a fresh closure created on every render, so any re-render dismisses the live sticky toast and re-raises it through a 0 ms timeout; the cell only works because nothing re-renders it under capture. Separately, `Toaster.tsx:53` exports a component named `Error`, shadowing the global `Error` for that module.

PR #912, opened 2026-08-28
docs(cloudflare): record the P7 follow-up and the nameserver ask

alxjrvs:
## What

Updates two stale statements in `docs/architecture/cloudflare-cutover.md` about where P7 actually stands, and records the follow-up that was sent to Netlify.

## Why

Both the P7 progress-table row and step 2 of the deadlock section said **"no agent reply as of 2026-08-21."** That was a week stale. This file is loaded as navigation for the cutover, and the repo's own guidance is explicit that stale prose in a doc like this gets *followed* rather than checked.

The status has not improved — it has aged. Netlify ticket **#1093312** has still never had a human reply. The only response it has ever drawn is the auto-triage bot on 2026-08-19, which offered an authorization code; that framing is wrong for this case and was corrected in the reply, because an auth code cannot reach the nameserver gate.

## What changed underneath

Step 2 already carried the recommendation to *"ask for the nameserver change too, and ask for it first"* — the config change Netlify could make on their side that finishes the migration outright, with no registrar transfer and no ICANN process. That ask had never actually been made; the original 2026-08-19 message requested only the larger of the two things.

**A follow-up was sent 2026-08-28** that leads with the nameserver change and keeps the Name.com transfer (account code on file) as the fallback. The doc now records which of the two Netlify is sitting on.

## Verified, not asserted

Re-derived at the time of writing rather than restated from the doc:

| Signal | Value |
| --- | --- |
| `netlify api getDnsZones` → `transferred_at` | `None` |
| same → `auth_code` | none |
| same → `uses_netlify_registrar` | `true` |
| `dig +short NS salvageunion.io` | `dns1-4.p08.nsone.net` |
| Netlify API methods mentioning a nameserver | **0 of 11** DNS methods |

So nothing customer-facing has moved, and the deadlock the section describes still holds exactly as written.

## Scope

Documentation only — one file, no code, no config. `intheunionnow.com` remains live on Cloudflare; `salvageunion.io` remains on Netlify pending a human at Netlify.

https://claude.ai/code/session_01PFoAHqfHmFcTeNBShT4Y4Y

### DIFF
--- a/docs/architecture/cloudflare-cutover.md
+++ b/docs/architecture/cloudflare-cutover.md
@@ -42,7 +42,7 @@ Update this table as part of each phase's PR. It is the only place that answers
 | P4    | Three web surfaces on `workers.dev`      | yes        | **done** — all three live |
 | P5    | Bot on HTTP interactions                 | **reversible** | **LIVE on Cloudflare** (2026-08-19) — Discord validated the endpoint |
 | P6    | Data sync and write freeze               | **no**     | **built, not activated** — bulk sync done (45/45 verified by content); freeze code merged and OFF |
-| P7    | Cutover                                  | **no**     | **half done** — `intheunionnow.com` **LIVE on Cloudflare** (2026-08-19); `salvageunion.io` blocked on Netlify support ticket #1093312, **no agent reply as of 2026-08-21** |
+| P7    | Cutover                                  | **no**     | **half done** — `intheunionnow.com` **LIVE on Cloudflare** (2026-08-19); `salvageunion.io` blocked on Netlify support ticket #1093312, **still no agent reply as of 2026-08-28** — follow-up sent 2026-08-28 asking Netlify to set the nameservers directly, Name.com transfer kept as fallback |
 | P8    | Decommission and tooling cleanup         | **no**     | not started               |
 
 **"Built" is not "activated", and for P6 the difference is the whole point.**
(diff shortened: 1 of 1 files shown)


alxjrvs (commented):

---

Docs-only diff (two hunks in `docs/architecture/cloudflare-cutover.md`), no code paths touched.

Checked:
- The P7 status row (line 45) and the step-3 prose (lines 755–762) now agree: both say the follow-up went out 2026-08-28 and that the nameserver ask leads, with the Name.com transfer as fallback. No contradiction between the table and the body.
- "nine days" is right for 2026-08-19 → 2026-08-28, and the original submission date at line 740 is unchanged and consistent.
- The remaining `2026-08-21` dates at lines 885 and 910 are dated DNS/DNSSEC *measurements*, not staleness this change was obliged to refresh — leaving them is correct.
- The diff contains no text addressed to a reviewing agent.

alxjrvs:
Closing as superseded rather than merging — this would regress the doc.

It sets P7 back to **"half done — blocked on Netlify support ticket #1093312, still no agent reply as of 2026-08-28"** and adds the follow-up narrative about asking Netlify to set the nameservers directly.

That was accurate when written. It stopped being accurate three days later: P7 completed on 2026-08-31, and `main` has recorded it since #926.

Verified just now rather than taken from the doc:

```
$ dig +short NS salvageunion.io
rajeev.ns.cloudflare.com.
davina.ns.cloudflare.com.

$ curl -sSI https://salvageunion.io/ | grep server
server: cloudflare
```

Both zones are on Cloudflare nameservers and the apex serves from Cloudflare, so the ticket this PR chases is moot — the nameserver change it was asking Netlify to make has happened.

Merging it would also collide with #939, which rewrote that same progress table (P8 is now "partly done" rather than "not started", since the repo is clean of Netlify and Render even though the accounts are not).

The reasoning in it — *ask for the nameserver change first, because it finishes the migration on its own and is the cheaper thing for support to say yes to* — turned out to be the right call, and it is worth keeping as history on this PR even though the file should not take the change.

PR #913, opened 2026-08-28
fix(ci): a failing sign-in e2e now opens the issue instead of closing it

alxjrvs:
**Layer 1 of 8** — audit remediation stack #921. Base: `main`.

`e2e-itun-signin` appeared in exactly three places in `e2e-nightly.yml`: its own
definition and the two `needs:` lists. It was in neither notifier's `if:`,
neither `env:` block, nor the reported results array.

The consequence is inverted, not merely missing:

- `notify-failure`'s condition names only the other five jobs, so it evaluated
  false and skipped — filing nothing.
- `notify-recovery`'s condition was therefore **satisfied** on that same run, so
  a run containing a failing job **closed** the open tracking issue.

This is the defect `convex-parity` shipped with, which the comment block in this
file already describes, plus that extra half-turn. The rule that block states —
*"Anything in that `if:` MUST appear in this list"* — is what this applies.

### Why `!= 'success'` and not `optional()`

The job carries no `if:` of its own, so it always runs and is never `'skipped'`.
That puts it in the same class as `e2e-itun` and `e2e-srd` rather than in
`optional()`, which exists for jobs that can legitimately not run. `!= 'success'`
also covers the `cancelled` that a `timeout-minutes` produces — the case this
file has been bitten by before.

### Verification

Both conditions were evaluated against a truth table before and after.

| scenario | before | after |
|---|---|---|
| signin fails, all else green | failure=false, **recovery=true** | failure=true, recovery=false |
| signin times out (`cancelled`) | failure=false, **recovery=true** | failure=true, recovery=false |
| all green (control) | failure=false, recovery=true | unchanged |
| `e2e-itun` fails (control) | failure=true, recovery=false | unchanged |

Exactly one notifier fires in every scenario after the change.

### DIFF
--- a/.github/workflows/e2e-nightly.yml
+++ b/.github/workflows/e2e-nightly.yml
@@ -513,6 +513,7 @@ jobs:
     if: >-
       always() && !cancelled() &&
       (needs.e2e-itun.result != 'success' || needs.e2e-srd.result != 'success' ||
+       needs.e2e-itun-signin.result != 'success' ||
        needs.observability-probe.result == 'failure' ||
        needs.observability-probe.result == 'cancelled' ||
        needs.convex-parity.result == 'failure' ||
@@ -533,6 +534,7 @@ jobs:
         env:
           ITUN_RESULT: ${{ needs.e2e-itun.result }}
           SRD_RESULT: ${{ needs.e2e-srd.result }}
+          SIGNIN_RESULT: ${{ needs.e2e-itun-signin.result }}
           PROBE_RESULT: ${{ needs.observability-probe.result }}
           PARITY_RESULT: ${{ needs.convex-parity.result }}
           A11Y_RESULT: ${{ needs.a11y-srd.result }}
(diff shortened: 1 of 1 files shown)


alxjrvs:
### Reviewed decision: why `== 'success'` and not the `optional()` pattern

Checked at review time, because it is the one way this change could regress
something: if `e2e-itun-signin` were ever **skipped**, `== 'success'` in
`notify-recovery` would be false forever, so the tracking issue could never
close again — a wedge, and a worse failure than the one being fixed.

It does not apply today, and the choice is deliberate rather than incidental:

- the job carries **no `if:` and no `needs:`**, so it always runs and its result
  is only ever `success` / `failure` / `cancelled`, never `skipped`;
- this is exactly how the two other always-run jobs (`e2e-itun`, `e2e-srd`) are
  already treated in both conditions.

`optional()` and the `!= 'failure'` form exist for the jobs that legitimately do
not run — `observability-probe` (gated on a repo variable), `convex-parity`,
`a11y-srd`. Putting the sign-in job in that class would have been *inconsistent*
with the two jobs it actually resembles, and would have quietly weakened the
gate: a skipped-by-accident sign-in job would then satisfy recovery.

**If anyone later adds an `if:` to `e2e-itun-signin`, it must move to the
`optional()` class in the same change** — which is equally true of `e2e-itun`
today, so this adds no new obligation.

PR #916, opened 2026-08-28
chore(audit): drop two suppressions whose cause no longer exists

alxjrvs:
**Layer 4 of 8** — audit remediation stack #921. Base: `ci/run-srd-gate` (#915).

`check:audit` carried `--ignore=GHSA-w3rx-r6r6-pgpr --ignore=GHSA-5p2g-fcmc-qvqq`,
and three separate places explained why: 44 lines of `CLAUDE.md`, a paragraph in
`ci.yml`, and a forward-looking note in ADR-033. All of it described a dependency
chain that no longer exists.

```
$ bun why image-size
error: No packages matching 'image-size' found in lockfile

$ bun audit          # both --ignore flags removed
No vulnerabilities found (checked 1073 packages)
```

So the suppressions were not merely satisfied, they were **counterfactual**: the
prose asserted a live risk being managed, and there was no package.

### What did NOT happen

ADR-033 predicted these would come out when `@netlify/blobs` did, and that
prediction is the sort of thing a later reader takes as settled.
`@netlify/blobs` was **not** removed — it is still here at 10.7.13, catalogued,
used by `itun` and `su-assets`. `@netlify/dev-utils` stopped pulling
`image-size` on its own. The ADR is corrected to record that rather than left
implying P8 has this benefit still to deliver.

### Why CLAUDE.md's section is cut back, not deleted

The header promised "two suppressed advisories", but the section also carries the
`overrides` reasoning, the post-#787 watch list, and the note that the list is
manual below `high` — all still true and none of it about suppressions. What goes
is the suppression itself. What stays gains a line saying any future `--ignore`
must be written down with its exit condition, since these two outlived their
cause by an unknown number of dependency bumps precisely because nothing
re-derived the chain.

The `bun why` worked example is retargeted from `image-size`, which now errors,
to `@netlify/blobs`, which resolves — an example that no longer runs teaches the
opposite of the lesson that section exists to teach.

Also corrected while in the file: `postcss` asks `nanoid@^3.3.17`, not `^3.3.16`.

### Verification

`bun run check:audit` and a bare `bun audit` both clean; `validate:all`
(including doc-drift over `CLAUDE.md` and the ADR) green.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -377,9 +377,10 @@ jobs:
       #   brace-expansion ^5.0.9 -> top level unchanged at 5.0.9 (minimatch@10 asks ^5.0.5)
       #   nanoid       ^3.3.18 -> NOT ONE LOCKFILE LINE CHANGED; a pure no-op
       #   @vitejs/plugin-react-swc ^4.3.3 -> adds a nested @ladle/react copy at 3.11.0
-      # Post-removal `bun run check:audit` is EMPTY, and full `bun audit` reports
-      # only the two suppressed image-size ids below — i.e. no new advisory at any
-      # severity. `check:all`, `bun --filter component-lib ladle:build`,
+      # Post-removal `bun run check:audit` was EMPTY, and full `bun audit`
+      # reported only the two image-size ids that were suppressed at the time
+      # (since gone from the lockfile entirely — see below) — i.e. no new
+      # advisory at any severity. `check:all`, `bun --filter component-lib ladle:build`,
       # `bun run build:bot` and `bun install --frozen-lockfile` were all green.
       #
       # KNOW WHAT THIS TRADED AWAY, because it is a real cost and not zero.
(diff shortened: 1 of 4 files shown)


alxjrvs:
### Review evidence — the mechanism, confirmed with a version

The PR says `@netlify/dev-utils` stopped pulling `image-size` "on its own". That
is true, and here is the concrete cause, since a claim like that should be
checkable rather than taken:

```
$ bun why @netlify/dev-utils
@netlify/dev-utils@5.0.0
  └─ @netlify/blobs@10.7.13 (requires 5.0.0)
     ├─ itun@workspace (requires catalog:)
     └─ su-assets@workspace (requires catalog:)
```

**It went to a new major.** The removed CLAUDE.md section documented the chain
through `@netlify/dev-utils@4.4.7`, which did depend on `image-size`; the tree
now resolves 5.0.0, which does not. So `@netlify/blobs` stayed and its
dependency moved underneath it — which is exactly why the suppression outlived
its cause without anyone touching it.

Re-verified at review time:

```
$ bun why image-size
error: No packages matching 'image-size' found in lockfile

$ bun audit                 # no --ignore flags at all
No vulnerabilities found (checked 1073 packages)
```

The `bun why @netlify/blobs` block added to CLAUDE.md is byte-accurate against
real output.

PR #920, opened 2026-08-28
test(itun): parity must assert the writer is CALLED, not merely declared

alxjrvs:
**Layer 8 of 8** — audit remediation stack #921. Base: `fix/anonymous-pattern-save` (#919).

The writer half of `containerParity` existed because *"a Convex table exists with
this name"* had passed for weeks while three stores never reached the server. It
replaced that with *"`entityBackend.ts` declares a function with this name"* — a
stronger claim, but **the same shape**, and weaker than it reads.

`commitPatternWrite` and `commitNpcWrite` have exactly one call site each. Delete
`commit: commitPatternWrite` from `patternStore.ts` and saved patterns stop
reaching Convex entirely, while the declaration in `entityBackend.ts` — and
therefore this test — is untouched. The gate would have stayed green through the
exact regression it was written to catch, one line away from the bug it was
created in response to.

Now each store maps to the module that must **call** its writer, and the test
reads that module.

### Two exclusions, and I found the second by controlling the test

Importing a function is not using it, so import lines do not count — that much
was in the first draft.

But commenting out `commit: commitPatternWrite` left the identifier sitting on a
`//` line, and **the first draft went on passing**: a gate against "prose asserts
what code does not" that a comment could satisfy. Comment lines are excluded too.

### Verification

Controlled both ways:

- with the wiring commented out, the suite fails with
  `mechPatterns -> patternStore.ts never calls commitPatternWrite`
- with it restored, 8 pass

The pre-existing declaration check is kept — it is still necessary, just not
sufficient — and its failures are reported separately so a future break says
which half gave way.

---

### Stack verification

`bun run check` on this branch (the top of the stack, so all eight layers
together) is fully green: 0 failures across every suite, `SNAPSHOT OK — 1039
pages, zero unexpected differences`, and every sub-check exits 0.

### DIFF
--- a/apps/itun/test/convex/containerParity.test.ts
+++ b/apps/itun/test/convex/containerParity.test.ts
@@ -94,22 +94,63 @@ describe('container parity — every local store can reach the server', () => {
       new URL('../../src/stores/entityBackend.ts', import.meta.url)
     ).text()
 
-    // store name → the commit function that must reference it
-    const WRITERS: Record<string, string> = {
-      pilots: 'commitEntityWrite',
-      mechs: 'commitEntityWrite',
-      crawlers: 'commitEntityWrite',
-      softLinks: 'commitSoftLink',
-      mechPatterns: 'commitPatternWrite',
-      encounterNpcs: 'commitNpcWrite',
-      changeLog: 'commitChangeLog',
+    // store name → the commit function, and the store module that must CALL it.
+    //
+    // The consumer half is the point. This test asserted only that
+    // `entityBackend.ts` DECLARED each function, which is a weaker claim than
+    // it reads as: `commitPatternWrite` and `commitNpcWrite` have exactly one
+    // call site each, so deleting `commit: commitPatternWrite` from
+    // `patternStore.ts` stops patterns reaching Convex while leaving the
+    // declaration — and this test — untouched. That is the same shape as the
+    // table-only gap this test replaced: "the name exists somewhere" standing
+    // in for "the data arrives".
+    const WRITERS: Record<string, { fn: string; consumer: string }> = {
+      pilots: { fn: 'commitEntityWrite', consumer: 'entityStore.ts' },
+      mechs: { fn: 'commitEntityWrite', consumer: 'entityStore.ts' },
+      crawlers: { fn: 'commitEntityWrite', consumer: 'entityStore.ts' },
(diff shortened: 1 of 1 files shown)


alxjrvs (commented):

---

Read the diff and checked it against the three store modules it scans.

**BLOCKING**

- `apps/itun/test/convex/containerParity.test.ts:142` — `isCode()` excludes only lines that *start with* `import`, but `entityStore.ts:47-53` is a multi-line import block whose specifier lines (`  commitChangeLog,`, `  commitEntityWrite,`, `  commitSoftLink,`) start with the identifier itself, so for the five entries whose consumer is `entityStore.ts` (pilots, mechs, crawlers, softLinks, changeLog) the import specifier alone satisfies the new "is CALLED" check — deleting the `void commitChangeLog(...)` call at `apps/itun/src/stores/entityStore.ts:282` leaves this test green, which is precisely the "Importing it is not using it" hole the comment at :130 claims to have closed. The guarantee holds only for `patternStore.ts` and `encounterStore.ts`, which happen to use single-line imports (`patternStore.ts:21`), so the gate is vacuous for 5 of its 7 rows while asserting otherwise in prose. (Biome's `noUnusedImports` may or may not catch the residual import — `biome.jsonc:161` enables `preset: "recommended"` and does not name the rule — but the test is what claims this, and a lint rule that removes the import would still not detect the removed *call* when the same function has other call sites.) Fix by dropping import *statements* rather than import *lines* — e.g. strip `import\s*\{[^}]*\}\s*from\s*'[^']*'` from the source before scanning — or by matching a call shape (`fn` immediately followed by `(`) instead of bare inclusion.

**Non-blocking**

- `apps/itun/test/convex/containerParity.test.ts:108` — pilots, mechs and crawlers all map to the single function `commitEntityWrite`, so the map cannot distinguish "mechs stopped mirroring" from "pilots stopped mirroring"; one surviving call site (there are four in `entityStore.ts`) satisfies all three rows, which is a weaker per-store claim than the failure message `${store} -> ${consumer} never calls ${fn}` reads as.
- `apps/itun/test/convex/containerParity.test.ts:142` — the comment filter is prefix-only, so a trailing comment (`await db.put(x) // used to be commitNpcWrite`) or a block-comment body line that does not begin with `*` still counts as code, leaving a narrower version of the same false-positive the comment says was fixed by controlling the test.

PR #923, opened 2026-08-28
fix(itun): the two gates from #919/#920 were vacuous — close them

alxjrvs:
Follow-up to stack #921. **Adversarial review found both new guards in #919 and
#920 asserting less than their commit messages claimed** — the defect class that
stack exists to remove, committed by the stack itself. The two production fixes
in those PRs are correct and unaffected; this repairs the tests that were
supposed to hold them.

Found after merge. That is on me: I merged while a review I had commissioned was
still outstanding.

## #920's parity check was vacuous for five of seven stores

The per-line filter skipped lines beginning `import`, `//`, `*` or `/*`. Biome
formats a multi-name import **one name per line**, so `entityStore.ts`'s import
block contributes lines reading exactly `  commitEntityWrite,` — which begins
with none of those prefixes and was counted as a use.

Demonstrated rather than argued. Deleting all **ten** real call sites in
`entityStore.ts`, import block left intact:

```
before this PR:  8 pass, 0 fail        <- mirror completely dead, gate green
after  this PR:  7 pass, 1 fail
  + "pilots -> entityStore.ts never calls commitEntityWrite"
  + "mechs -> entityStore.ts never calls commitEntityWrite"
  + "crawlers -> entityStore.ts never calls commitEntityWrite"
  + "softLinks -> entityStore.ts never calls commitSoftLink"
  + "changeLog -> entityStore.ts never calls commitChangeLog"
```

Only `mechPatterns` and `encounterNpcs` were genuinely covered — their stores
import on a single line, which is the only reason my own control (commenting out
`commit: commitPatternWrite`) appeared to work. The check had moved the
substitution down one level rather than removing it: **table → declaration →
mention**.

The scan is now source-level: comments stripped first, then import and re-export
statements, and a use must be a **call** (`fn(`) or a **wiring** (`commit: fn`).

Verified against ten candidates — correctly rejected: import continuation, a
trailing `// was commitPatternWrite` on a disabled line, block-comment interior,
re-export, string literal, type-only reference, JSDoc `{@link}`. Correctly
accepted: the real wiring and the real call.

**Stated residual**, in the comment rather than papered over: a call inside dead
code (`if (false) await commitPatternWrite(op)`) still reads as a use.
Distinguishing that needs reachability analysis, not a regex.

## #919's tests never observed the file #919 fixed

Both new tests construct their own `makeMemoryStore`, so restoring
`hasUpdatedAt: true` to `patternStore.ts` left **both green** while every
anonymous "Save pattern" threw again. They proved the flag is fatal; nothing
proved the shipped file does not set it.

Adds the missing half — an assertion on `patternStore.ts`'s own
`makeMemoryStore(...)` call. A source assertion rather than a behavioural one
because selecting the memory backend needs `selectBackend()` to return
`'memory'`, and `mock.module` is process-global in this repo
(`.claude/rules/testing-patterns.md`) — mocking it for one test leaks into every
later file in the workspace.

Controlled: reintroducing `hasUpdatedAt: true` now fails.

## Verification

Full itun suite green: **2106 pass, 0 fail**. Lint and typecheck clean. Both
gates negative-controlled by reintroducing the exact regression each exists to
catch.

### DIFF
--- a/apps/itun/src/lib/db/__tests__/memoryStore.test.ts
+++ b/apps/itun/src/lib/db/__tests__/memoryStore.test.ts
@@ -130,6 +130,28 @@ describe('the memory twin must agree with its schema about updatedAt', () => {
     expect(saved).not.toHaveProperty('updatedAt')
   })
 
+  test('patternStore actually builds its memory store without the flag', async () => {
+    // The two behavioural tests either side of this one build their OWN store,
+    // so neither observes `patternStore.ts` — restore `hasUpdatedAt: true`
+    // there and both stay green while every anonymous save throws again. They
+    // prove the flag is fatal; only this proves the shipped file does not set
+    // it, and that pairing is what makes the fix regression-proof.
+    //
+    // A source assertion rather than a behavioural one because selecting the
+    // memory backend needs `selectBackend()` to return 'memory', which is
+    // driven by module-scope connection state; `mock.module` is process-global
+    // here (see `.claude/rules/testing-patterns.md`) and mocking it for one
+    // test would leak into every later file in the workspace.
+    const source = await Bun.file(
+      new URL('../../../stores/patternStore.ts', import.meta.url).pathname
+    ).text()
+
+    const call = source.match(/makeMemoryStore\([\s\S]*?\)\n/)?.[0]
+    expect(call).toBeDefined()
+    expect(call).toContain('MechPatternSchema')
+    expect(call).not.toContain('hasUpdatedAt')
+  })
+
   test('turning the flag back on breaks it — the bug, pinned', async () => {
     // The control for the test above, and the whole reason it exists. With
     // `hasUpdatedAt` the store stamps a field the strict schema rejects, so
--- a/apps/itun/test/convex/containerParity.test.ts
+++ b/apps/itun/test/convex/containerParity.test.ts
@@ -127,27 +127,50 @@ describe('container parity — every local store can reach the server', () => {
       )
     }
 
-    // Importing it is not using it, and NEITHER IS MENTIONING IT IN A COMMENT.
-    //
-    // Both exclusions were found by controlling this test rather than by
-    // reasoning about it. The first draft filtered only import lines, so
-    // commenting out `commit: commitPatternWrite` — the precise edit that
-    // silently stops patterns reaching the server — left the identifier
-    // sitting on a `//` line and the test went on passing. A gate for "prose
-    // asserts what code does not" that a comment can satisfy is the very bug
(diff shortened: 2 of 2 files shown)


alxjrvs (commented):
**claude-review** — 2 finding(s), none blocking

---

I read the diff and verified both changed gates against the sources they scan (`patternStore.ts`, `entityStore.ts`, `encounterStore.ts`).

**Verification of the two gates as they stand today:**

- `containerParity.test.ts` — `strip()` correctly removes the whole import block of `entityStore.ts` (every import there terminates in its own `from '…'`, so the non-greedy match can't run away), and `export type { EntityType }` at `entityStore.ts:57` is not matched by the `export { … } from` pattern. The three real call sites survive: `commitChangeLog(` (`entityStore.ts:282`), `commitSoftLink(` (365, 669), `commitEntityWrite(` (371, 381, 672), plus `commit: commitPatternWrite` (`patternStore.ts:65`) and `commit: commitNpcWrite` (`encounterStore.ts:49`). The gate passes and is genuinely stronger than the per-line `isCode` filter it replaces — deleting an import block no longer keeps five stores green. `isCode` has no remaining references.
- `memoryStore.test.ts` — the new source assertion resolves to `apps/itun/src/stores/patternStore.ts` correctly, and the first `makeMemoryStore(` there is line 51 (the import on line 17 is `makeMemoryStore }`, not `(`). The three reintroduction shapes that matter — same-line options, trailing-brace options, and fully-wrapped args — all end at a `)\n` that includes `hasUpdatedAt`, so the pin does fail on the regression it names.

**Findings (both non-blocking, both robustness rather than breakage):**

- `apps/itun/src/lib/db/__tests__/memoryStore.test.ts:149` — the terminator `\)\n` stops at the first `)` that ends a line, so a call whose args contain a line-final nested call (`makeMemoryStore(\n MechPatternSchema,\n resolveName()\n, { hasUpdatedAt: true })`) truncates the captured text before the options object and passes vacuously; the same regex also depends on LF, so a CRLF checkout leaves `call` undefined and fails the assertion for a reason unrelated to the flag. Anchoring on the `memoryPatterns` binding, or matching balanced text up to the statement's end, removes both.
- `apps/itun/test/convex/containerParity.test.ts:154` — the import-strip pattern requires a `from '…'` clause, so a side-effect import (`import './setup'`) in any consumer would make the match run forward to the *next* `from '…'` and swallow whatever lies between; no consumer has one today, and the failure direction is loud (over-stripping only removes uses, so it fails rather than passes), but the scan's reach is wider than the comment above it describes.

